PR #3, opened 2026-04-05
fix(BDHLNDR-1,BDHLNDR-2): compile hook script and purge stale ClaudeLander entries

William-Long-II:
## Summary

- **BDHLNDR-1**: Add `build:hooks` esbuild step so `dist/hooks/bodhilander-hook.js` is actually produced. The hook source was never compiled (excluded by `tsconfig.main.json`), leaving installed apps with an empty `dist/hooks/` and Claude Code failing with `Cannot find module .../bodhilander-hook.js` on every PreToolUse/PostToolUse/Stop event.
- **BDHLNDR-2**: Rewrite hook cleanup in `src/main/mcp-config.ts` to handle the rename from ClaudeLander to Bodhilander and several compounding latent bugs.

## The three bugs folded into BDHLNDR-2

1. **Wrong TypeScript shape.** `HookConfig.hooks` was declared `string[]` but Claude Code's real format is `{ type, command }[]`. Previous filter code called `h.includes(...)` on objects and would have thrown if it ever executed. Fixed by introducing a `HookCommand` interface and updating all reads/writes.
2. **Cleanup never ran.** `registerHooks()` bailed on the missing-file check before reaching cleanup, so stale entries from prior installs were never removed — on reinstall or rename, users were stuck with broken hook commands forever.
3. **Name match too narrow.** Cleanup matched the substring `bodhilander` only, so legacy `claudelander` / `ClaudeLander` entries from the pre-rename app were invisible to it.

## What changed

- New `build:hooks` script in `package.json` matching the existing `build:mcp` pattern; wired into the main `build` pipeline.
- New `HookCommand` type, `OUR_HOOK_IDENTIFIERS = ['bodhilander', 'claudelander']` (case-insensitive match), and `purgeOurHooks(settings, keepPath?)` helper.
- `registerHooks()` restructured: read → purge stale → persist if purge happened → check file exists → register current-path entries in correct `{type, command}` shape.
- `purgeOurHooks` walks **all** hook types (PreToolUse / PostToolUse / Stop / Notification), filters at the per-hook level (so mixed configs are handled correctly), and accepts an optional `keepPath` so repeated startups don't thrash `settings.json` when current entries are already correct.
- `areHooksConfigured` now matches on the exact current `hookScriptPath` so install-location changes trigger a re-register.
- `unregisterHooks` simplified to a one-line delegation to `purgeOurHooks`.

## Verification

- `bun run build:main` — tsc passes clean.
- `bun run build:hooks` — produces `dist/hooks/bodhilander-hook.js` (4.5 kB).
- `node -e "require('./dist/hooks/bodhilander-hook.js')" PreToolUse </dev/null` — loads and exits 0 with expected `{}` output.
- Inline sanity tests against the exact shape of a real user's broken `~/.claude/settings.json`:
  - Stale `ClaudeLander` entries in PreToolUse + PostToolUse + Stop → fully purged, empty `hooks` key removed
  - Mixed current-path + stale in the same config array → stale hook dropped, current hook kept
  - Third-party hooks present → untouched
  - Case-insensitive match (`ClaudeLander` with capitals)
  - `purgeOurHooks` without `keepPath` (unregister path) → removes everything ours
  - `purgeOurHooks` with current-path `keepPath` → no-op when only current entries exist (prevents startup thrash)

## Out of scope

- **BDHLNDR-3** (session-switch crash on Windows) — separate ticket; needs user logs/repro before a fix can be designed. May be ameliorated by fixing the hook infrastructure here if crashes are downstream of hook failures, but that's speculation until we see logs.

## Test plan

- [ ] Fresh install on a clean Windows machine → confirm `dist/hooks/bodhilander-hook.js` is present in `resources/app.asar.unpacked/dist/hooks/`.
- [ ] Seed `~/.claude/settings.json` with a legacy `ClaudeLander` hook entry → launch app → confirm stale entry is removed and a correct current-path entry is written.
- [ ] Run Claude Code in any project after app launch → confirm no "Cannot find module" errors on Stop/PreToolUse/PostToolUse.
- [ ] Launch app twice in a row → confirm `settings.json` is not unnecessarily rewritten on the second launch (no thrash).


brannon-bowden (approved):
Adds a dedicated esbuild step for the hooks bundle and integrates it into the main build pipeline. The core change in mcp-config.ts is well-structured: migrates hook format from plain strings to typed HookCommand objects, adds a purgeOurHooks() function that cleans up stale entries from both Bodhilander and legacy ClaudeLander installs, and makes cleanup idempotent via the keepPath parameter. Good defensive coding with the case-insensitive identifier matching.

PR #7, opened 2026-04-05
fix(BDHLNDR-6): prevent concurrent VectorSearch workers on session activity

William-Long-II:
## Summary

Root cause of [BDHLNDR-3](https://gobodhi.atlassian.net/browse/BDHLNDR-3) (Windows crash when leaving and returning to a session). Evidence from a user `main.log`:

```
10:30:50.251 [VectorSearch] Starting indexing for: C:\work\repos\bodhi-service-ai-engine
10:30:55.410 [VectorSearch] Starting indexing for: C:\work\repos\bodhi-service-ai-engine  ← same dir, 5s later
```

Two `[VectorSearch Worker] Starting worker thread...` blocks, each re-initializing `@huggingface/transformers` and tree-sitter against the same cache directory. The log ends mid-flow without a JS trace — a native crash during double-initialization.

### 1. `startIndexing` always cancelled-and-restarted

`useCodeSearch` auto-fires on `directoryPath` changes from multiple consumers (`IndexStatus.tsx:16`, `CodeSearchModal.tsx:35`). Each session switch → N simultaneous `startIndexing` IPC calls for the same path. The old top-of-function `this.cancelIndexing(index.id)` meant each call would tear down and respawn the worker, on top of whatever previous calls were already doing.

**Fix**: if a worker is already running for this index, log and return. This is the primary deduplication point — N concurrent IPC calls collapse to one worker.

### 2. `cancelIndexing` didn't await `worker.terminate()`

`Worker.terminate()` returns a Promise. It was called and immediately followed by `new Worker(...)` in the old startIndexing flow, so the dying worker was still unloading native modules (onnxruntime-node, tree-sitter, better-sqlite3) while the new worker was loading them.

**Fix**: `cancelIndexing` is now `async` and `await`s termination. The worker is removed from `this.workers` synchronously *before* `terminate()` is called, so any concurrent `startIndexing` correctly observes "no running worker".

## Caller updates

- `deleteIndex(directoryPath)` → `async`, awaits cancel (otherwise DB delete could race with in-flight worker writes).
- `vector-search:cancel-indexing` IPC handler → awaits cancel.
- `vector-search:delete-index` IPC handler → awaits delete.
- Worker-init timeout callback → uses `void this.cancelIndexing(...)` since it's inside a `setTimeout` and only updates DB/emits after (no concurrent-spawn risk).

## What this doesn't touch

- `useCodeSearch` renderer-side deduplication — left alone. Multiple consumer components still each fire their own auto-index effect, but the main-process guard catches them all. Renderer-level dedup would be cleaner code but is not required for correctness.

## Depends on / related

- **Depends on [BDHLNDR-4](https://gobodhi.atlassian.net/browse/BDHLNDR-4)** (tree-sitter/node-gyp-build packaging) — under the fallback parser the concurrent workers may be exercising additional unsafe paths. BDHLNDR-4 should ship in the same release.
- **Resolves [BDHLNDR-3](https://gobodhi.atlassian.net/browse/BDHLNDR-3)** pending post-release verification on Windows.

## Test plan

- [ ] Install a build containing this + BDHLNDR-4. Open multiple sessions with the same cwd. Switch between them repeatedly. Leave one session and return. Confirm the app no longer crashes and that `main.log` shows at most one `[VectorSearch] Starting indexing for:` event per directory per launch.
- [ ] Trigger a Re-index from `CodeSearchModal` (delete + start flow). Confirm the old worker is gone before the new one spawns (no overlap in worker log output).
- [ ] Force an init timeout (e.g. by interrupting network during first-run model download) and confirm the error path still reports the error and doesn't hang.
- [ ] `bun run build:main` — tsc passes clean.

[BDHLNDR-3]: https://gobodhi.atlassian.net/browse/BDHLNDR-3?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BDHLNDR-4]: https://gobodhi.atlassian.net/browse/BDHLNDR-4?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -969,13 +969,13 @@ safeHandle('vector-search:search-symbols', (directoryPath: string, name: string,
   return getVectorSearchManager().searchSymbols(directoryPath, name, symbolType as any, limit);
 });
 
-safeHandle('vector-search:cancel-indexing', (indexId: string) => {
-  getVectorSearchManager().cancelIndexing(indexId);
+safeHandle('vector-search:cancel-indexing', async (indexId: string) => {
+  await getVectorSearchManager().cancelIndexing(indexId);
   return { success: true };
 });
 
-safeHandle('vector-search:delete-index', (directoryPath: string) => {
-  getVectorSearchManager().deleteIndex(directoryPath);
+safeHandle('vector-search:delete-index', async (directoryPath: string) => {
+  await getVectorSearchManager().deleteIndex(directoryPath);
   return { success: true };
 });
 
--- a/src/main/vector-search/index.ts
+++ b/src/main/vector-search/index.ts
@@ -181,8 +181,16 @@ export class VectorSearchManager extends EventEmitter {
   async startIndexing(directoryPath: string): Promise<void> {
     const index = await this.getOrCreateIndex(directoryPath);
 
-    // Cancel any existing indexing for this directory
-    this.cancelIndexing(index.id);
+    // If a worker is already running for this index, don't spawn another one.
+    // Multiple renderer consumers (IndexStatus, CodeSearchModal) each mount
+    // their own useCodeSearch hook, and on session switch each fires its own
+    // auto-index effect — without this guard, concurrent workers race on
+    // native module initialization (tree-sitter / onnxruntime-node) and
+    // crash the process. (BDHLNDR-6 / root cause of BDHLNDR-3.)
+    if (this.workers.has(index.id)) {
+      console.log('[VectorSearch] Indexing already in progress for:', directoryPath);
+      return;
+    }
 
     // Clear the cancelled flag since we're starting fresh
     this.cancelledIndexes.delete(index.id);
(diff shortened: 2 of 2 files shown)


brannon-bowden (approved):
Fixes the root cause of BDHLNDR-3 (crash on session return) by adding a guard in startIndexing() that returns early if a worker already exists for the index, preventing concurrent workers from racing on native module init. Also makes cancelIndexing/deleteIndex properly async, awaiting worker.terminate() before proceeding — the synchronous workers.delete() before terminate is a smart touch to prevent races with concurrent startIndexing calls.

PR #9, opened 2026-04-05
fix(BDHLNDR-8): prevent WebGL addon disposal crash on session stop

William-Long-II:
## Summary

Fixes [BDHLNDR-8](https://gobodhi.atlassian.net/browse/BDHLNDR-8) — stopping a session was throwing `TypeError: Cannot read properties of undefined (reading 'onRequestRedraw')` from `xterm-addon-webgl` up to the React ErrorBoundary.

## Root cause

In `src/renderer/components/Terminal.tsx`, the `WebglAddon` was instantiated inside a `requestAnimationFrame` callback and never stored in a ref. When the effect tore down, `term.dispose()` cascaded through `AddonManager` into `webglAddon.dispose()` → `RenderService.setRenderer(...)`. If the render service was already torn down (or torn down in the same tick), `this._renderer` was undefined and the dispose crashed.

Known xterm.js lifecycle footgun when the addon is loaded async and disposal happens around the same time — triggered by stop, `restartKey` restart, or unmount.

## Fix

1. Track the addon in a new `webglAddonRef` so cleanup can reach it directly.
2. Declare a `mounted` flag at the top of the effect and guard the rAF body with `if (!mounted) return;` so the addon is never attached after teardown.
3. In the cleanup, dispose the WebGL addon explicitly **before** `term.dispose()`, inside a `try/catch`, so it cleans up while its internal state is still valid instead of via a late cascade.
4. Wrap `term.dispose()` in a `try/catch` as a last-resort safety net so no addon disposal error can escape to the React tree.

Canvas fallback path is unchanged (it's in the existing catch block that skips WebGL entirely).

## Diff

1 file, +29/−2 in `src/renderer/components/Terminal.tsx`.

## Test plan

- [ ] `bunx tsc --noEmit -p tsconfig.json` passes (verified locally, clean)
- [ ] Stop an active session — no ErrorBoundary, no console error
- [ ] Restart a session via the restart button (`restartKey` flow) — no crash
- [ ] Unmount a terminal (switch away) while it is still alive — no crash
- [ ] Force WebGL to fail (e.g. disable hardware acceleration) — canvas fallback still works

[BDHLNDR-8]: https://gobodhi.atlassian.net/browse/BDHLNDR-8?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/renderer/components/Terminal.tsx
+++ b/src/renderer/components/Terminal.tsx
@@ -29,6 +29,7 @@ const Terminal: React.FC<TerminalProps> = ({ sessionId, cwd, launchClaude = true
   const terminalRef = useRef<HTMLDivElement>(null);
   const xtermRef = useRef<XTerm | null>(null);
   const fitAddonRef = useRef<FitAddon | null>(null);
+  const webglAddonRef = useRef<WebglAddon | null>(null);
   const lastPasteTimeRef = useRef<number>(0);
   const [isRunning, setIsRunning] = useState(!isStopped);
   const [contextMenu, setContextMenu] = useState<ContextMenuState>({ visible: false, x: 0, y: 0, hasSelection: false });
@@ -181,6 +182,8 @@ const Terminal: React.FC<TerminalProps> = ({ sessionId, cwd, launchClaude = true
   useEffect(() => {
     if (!terminalRef.current || !isRunning) return;
 
+    let mounted = true;
+
     const term = new XTerm({
       theme: {
         background: '#1e1e1e',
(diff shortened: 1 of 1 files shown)


brannon-bowden (approved):
Fixes the WebGL addon disposal crash by tracking it in a ref, guarding the async rAF load with a mounted flag, and explicitly disposing the addon before term.dispose(). The try/catch safety net around both dispose calls ensures errors never escape to the React tree. Clean, defensive fix for a race condition between async addon loading and synchronous cleanup.

PR #10, opened 2026-04-05
feat(BDHLNDR-9): resume Claude sessions across restart/stop cycles

William-Long-II:
## Summary

Implements [BDHLNDR-9](https://gobodhi.atlassian.net/browse/BDHLNDR-9) — stopping and restarting a Claude session now preserves the conversation instead of spawning a fresh Claude with no context.

## Approach

Rather than capturing Claude's internally-generated session ID via the hook infrastructure (which would have a correlation problem — hooks are registered globally in `~/.claude/settings.json` so the hook can't trivially tell which Bodhilander session invoked it — and a capture race, since the ID isn't known until the first tool use or stop event), we generate a UUID on the Bodhilander side and hand it to Claude via `claude --session-id <uuid>` on first launch. On restart we pass the same UUID via `claude --resume <uuid>`.

**Verified via smoke test** that Claude persists the session under the caller-supplied UUID and `--resume` later restores conversation context.

## Changes

| File | Purpose |
|---|---|
| `src/shared/types.ts` | Add `claudeSessionId: string \| null` to `Session` |
| `src/main/database.ts` | Migration: nullable `claude_session_id` column on `sessions`, using the existing PRAGMA-based pattern |
| `src/main/repositories/sessions.ts` | Round-trip the column; dedicated `getClaudeSessionId` / `setClaudeSessionId` / `clearClaudeSessionId` helpers kept separate from the generic `updateSession` so arbitrary callers cannot stomp it |
| `src/main/claude-launcher.ts` | Extend `getClaudeCommand` with optional `claudeSession: { id, mode: 'new' \| 'resume' }`; emits `--session-id <uuid>` or `--resume <uuid>`; export `generateClaudeSessionId()` |
| `src/main/pty-manager.ts` | Look up stored UUID and resume, or generate + persist a new one; inject the session-id flag into all four shell spawn paths (WSL, PowerShell, cmd, bash/login); resume-failure fallback |
| `src/main/api/routes/sessions.ts` | Session literal: `claudeSessionId: null` |
| `src/renderer/store/sessions.ts` | Session literal: `claudeSessionId: null` |

**Diff:** 7 files, +173/-11.

## Resume-failure fallback

If a `--resume` spawn exits non-zero within a 5-second window, it's treated as a dead session:
1. The stored UUID is cleared from the DB.
2. The session is transparently respawned with a fresh `--session-id` UUID.
3. The renderer never sees the failed exit — it just sees new output from the replacement pty.

If the respawn itself fails, the original exit event is emitted so the UI recovers normally.

## Escaping / shell-safety

UUIDs are alphanumeric + hyphens, so the session-id flag can be inlined into each shell's command string without escaping. No shell-injection surface added.

## Acceptance criteria

- [x] Stop + restart restores prior conversation via `--resume <uuid>` _(static code path verified; needs runtime smoke test)_
- [x] First-time launch is visually unchanged (invisible `--session-id`) _(static code path verified; needs runtime smoke test)_
- [x] Resume failure falls back transparently; no crash, no error dialog _(static code path verified; needs runtime smoke test)_
- [x] Non-Claude shell sessions are unaffected (gated by `launchClaude`)
- [x] Typecheck passes (`bunx tsc --noEmit -p tsconfig.json` clean)

## Test plan

- [ ] Build and launch: `bun run dev`
- [ ] Create a Claude session, tell it \"the magic word is BANANA\"; verify `bodhilander.db` `sessions.claude_session_id` is populated
- [ ] Stop the session, start it again, ask \"what's the magic word?\" → should reply BANANA
- [ ] Edit the DB row to set `claude_session_id = 'deadbeef-0000-0000-0000-000000000000'`, restart the session → should transparently respawn (magic word gone, but no crash, no error)
- [ ] Create a plain bash/pwsh session → behaves identically to before

[BDHLNDR-9]: https://gobodhi.atlassian.net/browse/BDHLNDR-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/api/routes/sessions.ts
+++ b/src/main/api/routes/sessions.ts
@@ -79,6 +79,7 @@ export function createSessionsRouter(): Router {
           order: getNextOrder(),
           createdAt: now,
           lastActivityAt: now,
+          claudeSessionId: null,
         };
 
         sessionsRepo.createSession(session);
--- a/src/main/claude-launcher.ts
+++ b/src/main/claude-launcher.ts
@@ -1,12 +1,31 @@
 import * as path from 'path';
 import * as fs from 'fs';
 import * as os from 'os';
+import * as crypto from 'crypto';
 import { app } from 'electron';
+
+/**
+ * Which Claude CLI flag to use for the stored session UUID (BDHLNDR-9).
+ * - `new`    → `--session-id <uuid>` (first launch, set the session's ID)
+ * - `resume` → `--resume <uuid>` (subsequent launches, restore prior conversation)
+ */
+export type ClaudeSessionMode = 'new' | 'resume';
+
+export interface ClaudeSessionLaunch {
+  id: string;
+  mode: ClaudeSessionMode;
+}
 
 export interface ClaudeLaunchConfig {
   sessionId: string;
   projectDir: string;
   socketPath: string;
+  /**
+   * Claude session UUID + mode. When provided, the returned args include
+   * `--session-id <uuid>` or `--resume <uuid>`. Omit for the legacy fresh-launch
+   * behavior (no flag), which is only used if resume infrastructure is disabled.
+   */
+  claudeSession?: ClaudeSessionLaunch;
 }
 
 export function getClaudeCommand(config: ClaudeLaunchConfig): { command: string; args: string[]; env: NodeJS.ProcessEnv } {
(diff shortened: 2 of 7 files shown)


brannon-bowden (approved):
Implements Claude session resume (BDHLNDR-9) end-to-end: generates a UUID on first launch via --session-id, stores it in a new claude_session_id column, and passes --resume on subsequent spawns. The resume-failure fallback is well thought out — detects early non-zero exit within a 5s window and transparently respawns with a fresh session. DB migration, repository helpers, shared types, and all shell variants (WSL, PowerShell, cmd, bash) are covered.

PR #16, opened 2026-04-05
Release 3.1.1: BDHLNDR-12 (squished text) + BDHLNDR-13 (scrollback)

William-Long-II:
## Release summary

Merges `development` → `production` to cut **v3.1.1**, a patch release with two bug fixes that came out of user reports today.

### BDHLNDR-12 — Squished text on session switch (Bug, Medium)

When users switched away from and back to a session, the terminal content rendered hard-wrapped at the old narrow column count inside the now-wider container. Root cause: the `isActive` effect in `Terminal.tsx` called `fit()` on the xterm.js renderer but never propagated the new size to the PTY, so the shell kept emitting at stale cols. Fix hoists `handleResize` into a component-level `useCallback` and routes the `isActive` effect through it, so SIGWINCH fires on activation and the shell redraws at the correct width. Merged via #14.

### BDHLNDR-13 — Scrollback limited to 1000 lines (Bug, Medium)

`Terminal.tsx` and `RemoteTerminal.tsx` constructed `new XTerm({...})` without a `scrollback` option, inheriting xterm.js's default of 1000 lines — exhausted within a handful of Claude Code messages. Fix sets `scrollback: 10000` in both constructors (~2 MB per terminal, negligible). Merged via #15.

## Test plan

Both fixes have been merged to `development` and smoke-tested:
- [x] BDHLNDR-12: session switch no longer produces squished output
- [x] BDHLNDR-13: scroll history now shows ~10× more content

## Next steps after merge

1. Bump version on `production` with **patch** (`npm version patch`, 3.1.0 → 3.1.1 — both tickets are bug fixes, no new features)
2. `git push` **without** `--tags` so the release workflow creates the tag itself

### DIFF
--- a/src/renderer/components/RemoteTerminal.tsx
+++ b/src/renderer/components/RemoteTerminal.tsx
@@ -43,6 +43,8 @@ const RemoteTerminal: React.FC<RemoteTerminalProps> = ({ code, permission, isAct
       fontSize: 14,
       cursorBlink: permission === 'control',
       disableStdin: permission === 'read',
+      // BDHLNDR-13: match primary Terminal component — default 1000 is too small.
+      scrollback: 10000,
     });
 
     const fitAddon = new FitAddon();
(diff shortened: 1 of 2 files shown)


brannon-bowden (approved):
Release merge for v3.1.1 shipping BDHLNDR-12 (PTY resize on session switch) and BDHLNDR-13 (scrollback bump to 10k). Combined diff matches PRs #14 and #15 cleanly. Both are straightforward UI-quality fixes that should noticeably improve the day-to-day experience.

PR #25, opened 2026-04-07
fix(BDHLNDR-30): smart auto-scroll to prevent terminal viewport jumps

rockysco:
## Summary

- Fixes terminal viewport spontaneously jumping to random scroll positions during long Claude streaming output
- Adds smart auto-scroll: detects if user is near bottom before each `term.write()`, pins viewport after write completes
- Users who scroll up intentionally are not disturbed (5-line threshold)

## Details

Root cause: xterm.js viewport loses sync during rapid `term.write()` calls. The fix checks `buffer.active.baseY - buffer.active.viewportY` before each write, and calls `scrollToBottom()` in the write callback only if the user was within 5 lines of the bottom.

**Jira:** [BDHLNDR-30](https://gobodhi.atlassian.net/browse/BDHLNDR-30)

## Test plan

- [ ] Launch Bodhilander, start a Claude session
- [ ] Give Claude a long-running task that produces lots of streaming output
- [ ] Verify the terminal stays pinned to the bottom throughout
- [ ] Scroll up during streaming — verify it does NOT yank you back down
- [ ] Scroll back to bottom — verify auto-scroll re-engages

[BDHLNDR-30]: https://gobodhi.atlassian.net/browse/BDHLNDR-30?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/plans/2026-04-06-smart-auto-scroll-design.md
+++ b/docs/plans/2026-04-06-smart-auto-scroll-design.md
@@ -1,0 +1,40 @@
+# BDHLNDR-30: Smart Auto-Scroll for Terminal
+
+## Problem
+
+During long Claude Code streaming output, the xterm.js terminal viewport spontaneously jumps to random scroll positions. The jump distance varies. Root cause: xterm.js viewport loses sync during rapid `term.write()` calls.
+
+## Solution: Smart Auto-Scroll
+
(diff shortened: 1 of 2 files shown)


William-Long-II (approved):
looks good merging in thank you for the fix i know its aggrivating.

PR #34, opened 2026-04-09
fix(macos): default to /bin/zsh for Claude path resolution

brannon-bowden:
## Summary

- On macOS, default to `/bin/zsh` instead of `process.env.SHELL` in `detectUnixShell()`
- Fixes fresh installs not finding `claude` until the user manually sets the shell path in preferences
- `process.env.SHELL` can be stale or point to a shell (e.g. `/bin/bash`) whose init files don't have the claude binary on PATH when Electron launches from Finder/Dock
- zsh has been the default macOS shell since Catalina and is where users typically configure their PATH
- No behavior change on Linux (still uses `$SHELL` with `/bin/bash` fallback) or Windows

## Test plan

- [ ] Fresh macOS install: launch Bodhilander from Finder without setting custom shell path — claude should be found
- [ ] Existing installs with custom shell path set — should continue using the custom path (no change)
- [ ] Linux: verify `$SHELL` is still respected
- [ ] Windows/WSL: verify no regression

### DIFF
--- a/src/main/shell-detector.ts
+++ b/src/main/shell-detector.ts
@@ -83,11 +83,19 @@ function detectWindowsShell(): ShellInfo {
 }
 
 function detectUnixShell(): ShellInfo {
-  const shell = process.env.SHELL || '/bin/zsh';
+  let actualShell: string;
+
+  if (process.platform === 'darwin') {
+    // macOS: default to /bin/zsh (default shell since Catalina).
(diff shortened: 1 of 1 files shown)


William-Long-II (approved):
Clean, well-scoped fix. Only targets macOS, fallback chain still intact, no impact on Linux/Windows or users with custom shell path preference. LGTM.

PR #45, opened 2026-04-23
hotfix: fix Intel Mac splash hang + add Open Log Folder menu (BDHLNDR-38, BDHLNDR-39)

William-Long-II:
## Summary
- **BDHLNDR-38**: Intel Mac users on v3.2.9 were stuck at the splash screen. Log from affected user confirmed `dlopen ... incompatible architecture (have 'arm64', need 'x86_64h')` on `better_sqlite3.node` — postinstall's `electron-rebuild` ran only on the arm64 macos-latest runner, so the cross-packaged x64 DMG shipped arm64 native modules. Fix: split the mac matrix into `macos-14` (arm64) and `macos-13` (x64) jobs, each building natively. Drops the BDHLNDR-37 `npm pack` sharp workaround (no longer needed). Adds `scripts/merge-latest-mac.py` to combine per-arch `latest-mac.yml` into one so electron-updater resolves correctly for both archs.
- **BDHLNDR-39**: Adds an *Open Log Folder* item to the Help menu using `shell.showItemInFolder(log.transports.file.getFile().path)` so users can grab logs for support without hunting through `%APPDATA%` / `~/Library/Logs`.

## Test plan
- [ ] CI: both `mac-arm64` (macos-14) and `mac-x64` (macos-13) matrix jobs succeed and produce single-arch DMGs
- [ ] CI release job: `scripts/merge-latest-mac.py` runs cleanly; final `latest-mac.yml` lists both DMGs under `files:`
- [ ] Intel Mac user installs 3.2.10 x64 DMG → gets past splash to main UI
- [ ] Apple Silicon Mac user installs 3.2.10 arm64 DMG → still works (regression check)
- [ ] Help → Open Log Folder opens the logs dir with `main.log` highlighted on Windows, macOS, and Linux
- [ ] Auto-update from 3.2.9 → 3.2.10 works on both Intel and Apple Silicon Mac

## Release steps after merge
After merging to `production`, bump version and push (per CLAUDE.md, NO `--tags`):
```
git checkout production && git pull
npm version patch   # 3.2.9 → 3.2.10
git push
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,8 +46,18 @@ jobs:
             platform: linux
           - os: windows-latest
             platform: win
-          - os: macos-latest
-            platform: mac
+          # BDHLNDR-38: build each Mac arch on its native runner so postinstall
+          # (electron-rebuild) and sharp's optional-dep resolution produce
+          # host-native binaries. Cross-packaging from a single arm64 runner
+          # previously shipped arm64 .node files inside the x64 DMG, causing
+          # Intel Mac users to hang at the splash screen when better-sqlite3
+          # failed to load.
+          - os: macos-14
+            platform: mac-arm64
+            mac_arch: arm64
+          - os: macos-13
+            platform: mac-x64
+            mac_arch: x64
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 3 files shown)


brannon-bowden (approved):
Strong diagnosis — dlopen architecture mismatch on better_sqlite3 is exactly the footprint you'd expect from cross-arch packaging, and splitting the matrix into macos-14/macos-13 is the right native-rebuild fix rather than another postinstall workaround. Good call retiring the BDHLNDR-37 sharp hack in the same PR. The merge-latest-mac.py step is the piece that actually keeps auto-update working for both archs, worth calling out on the test plan. Help → Open Log Folder is a small quality-of-life win that'll pay for itself the next time someone is troubleshooting on an unfamiliar OS.

PR #47, opened 2026-04-23
fix(vector-search): stop pegging CPU during indexing (#46)

### LINKED ISSUE #46: Vector search indexing pegs CPU on all cores (100% multi-core)
## Summary

Starting a vector-search index over a large directory saturates every CPU core and makes the machine unresponsive. Observed on macOS (Apple Silicon), 3.3.0-beta.2, indexing `/Volumes/Secondary/DevelopmentWork/Temp` which contained 37,618 files. The worker ran ~11 minutes with no log output before the user had to force-quit the app.

Log excerpt:

```
[2026-04-23 12:48:04.327] [info]  [VectorSearch Worker] Embedding model initialized successfully
[2026-04-23 12:48:15.872] [info]  [VectorSearch Worker] [IndexingWorker] Discovered 37618 files in /Volumes/Secondary/DevelopmentWork/Temp
... (silent for ~11 minutes while all cores pinned at 100%) ...
[2026-04-23 12:59:27.003] [error] [Main] Child process gone: GPU killed 15
[2026-04-23 12:59:27.017] [error] [Main] Child process gone: Utility killed 15
[2026-04-23 12:59:27.019] [error] [Main] Render process gone: killed 15
```

## Root cause

Three compounding issues in `src/main/vector-search/`:

1. **ONNX Runtime thread count is unbounded.** `embedding-provider.ts` calls `pipeline('feature-extraction', modelName)` with no session options. `@huggingface/transformers` under `onnxruntime-node` defaults `intraOpNumThreads` to the number of physical cores. Every embedding call saturates all cores for the duration.

2. **`embed(texts)` is not actually batched.** `embedding-provider.ts:86-99` loops one text at a time:
   ```ts
   for (const text of texts) {
     const embedding = await this.embedWithRetry(text);
     results.push(embedding);
   }
   ```
   The `BATCH_SIZE = 32` in `indexing-worker.ts:181` is therefore 32 sequential single-text inferences, not a real batch. Proper batching would let ONNX amortize the forward pass.

3. **No yield between files in the indexing loop.** `indexing-worker.ts:166` iterates `for (const file of files)` with no cooperative yield, so the worker thread never checks for cancel messages and never lets progress IPC flush during a long run. The main log going silent for 11 minutes is a direct symptom.

## Contributing factor

`file-discovery.ts:51` defaults `maxFiles: 50000` with no pre-run confirmation. Pointing the indexer at a large junk directory silently grinds. Tracking this separately — it's a UX fix, not the CPU fix.

## Fix plan (this issue)

- Cap ONNX threads to a small fraction of cores (e.g. `max(1, floor(cpus/2))`) via `pipeline` session options and `env.backends.onnx.wasm.numThreads`.
- Pass the full text array to the pipeline in one call in `HuggingFaceEmbeddingProvider.embed`.
- `await new Promise(r => setImmediate(r))` between files in the worker loop so cancel and progress can be serviced.

## Out of scope (separate issues)

- Indexing scope UX: rather than capping `maxFiles`, add an **opt-in "Index this group/subgroup" toggle** when adding a group or subgroup, so users explicitly choose which directories get indexed instead of relying on a silent file-count limit.
- `HF_HOME`/`TRANSFORMERS_CACHE` not taking effect — model cache still writes inside the app bundle.
- Renderer `Cannot read properties of undefined (reading 'dimensions'/'handleResize')` race in xterm Viewport on PTY teardown.
- First-run UX when launched under AppTranslocation (read-only bundle).

brannon-bowden:
Closes #46.

## Summary

- Cap `onnxruntime` threads so a single embedding call no longer pins every physical core. `session_options.intraOpNumThreads = max(1, min(4, cores/2))` on the pipeline, plus `env.backends.onnx.wasm.numThreads` for the WASM fallback path.
- Make `HuggingFaceEmbeddingProvider.embed()` actually batch. Previously it looped one text at a time, so the worker's `BATCH_SIZE = 32` was 32 sequential inferences. Now the array is handed to the pipeline once and the output tensor is split per row. Per-text retry fallback remains on batch failure.
- Yield with `setImmediate` between files in the indexing worker so cancel messages and progress IPC get serviced during long runs (this is why the main log went silent for 11 minutes in the repro).

## Files touched

- `src/main/vector-search/embedding-provider.ts`
- `src/main/vector-search/indexing-worker.ts`

## Test plan

- [x] `tsc -p tsconfig.main.json --noEmit` clean
- [x] `bun run build:worker` clean (1.6mb bundle)
- [ ] Launch app from `/Applications/Bodhilander.app` (not translocated) and index a medium directory (~2-5k files). Confirm:
  - [ ] Other apps stay responsive during indexing (Activity Monitor should show the Bodhilander helper using ~half cores, not all).
  - [ ] Progress updates stream to the UI rather than going silent for minutes.
  - [ ] Cancel from the UI stops the worker within a second or two.
- [ ] Repeat the repro from the issue: point indexing at a very large directory and confirm it still completes (just not at the expense of every core).
- [ ] Query against the indexed corpus returns sensible results — batched output tensor slicing must line up with input order.

## Out of scope

Tracked in issue #46's "out of scope" section — separate tickets to follow.

### DIFF
--- a/src/main/vector-search/embedding-provider.ts
+++ b/src/main/vector-search/embedding-provider.ts
@@ -1,14 +1,28 @@
 console.log('[EmbeddingProvider] Loading @huggingface/transformers...');
 
+import os from 'os';
+
 // Import types for TypeScript, runtime value loaded dynamically
 import type { FeatureExtractionPipeline as FeatureExtractionPipelineType } from '@huggingface/transformers';
 
 let pipeline: typeof import('@huggingface/transformers').pipeline;
 
+// Cap ONNX threads so indexing doesn't saturate every CPU core.
+// Default onnxruntime behavior is intraOpNumThreads = physical cores, which
+// pins the machine during long indexing runs. Half the cores, max 4, min 1.
+const EMBEDDING_THREAD_COUNT = Math.max(1, Math.min(4, Math.floor(os.cpus().length / 2)));
+
 try {
   const transformers = require('@huggingface/transformers');
   pipeline = transformers.pipeline;
-  console.log('[EmbeddingProvider] @huggingface/transformers loaded successfully');
+  // WASM numThreads matters when transformers falls back to the WASM backend;
+  // harmless when onnxruntime-node is in use.
+  if (transformers.env?.backends?.onnx?.wasm) {
+    transformers.env.backends.onnx.wasm.numThreads = EMBEDDING_THREAD_COUNT;
+  }
+  console.log(
+    `[EmbeddingProvider] @huggingface/transformers loaded successfully (thread cap=${EMBEDDING_THREAD_COUNT})`
+  );
 } catch (e) {
   console.error('[EmbeddingProvider] Failed to load @huggingface/transformers:', e);
   throw e;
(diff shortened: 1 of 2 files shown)


William-Long-II (approved):
approved!

PR #48, opened 2026-04-23
chore: forward-merge production hotfixes into development

William-Long-II:
## Summary
Forward-merges the 3.2.x hotfix series from `production` into `development` so nothing gets lost when we promote beta → stable.

## What's coming in
- **BDHLNDR-37**: bundle darwin-x64 sharp prebuilt (3.2.8 hotfix)
- **BDHLNDR-38**: Intel Mac splash-hang fix
  - Split Mac matrix into native runners (`macos-14` arm64 + `macos-15-intel` x64)
  - Merge per-arch `latest-mac.yml` into one via `scripts/merge-latest-mac.py`
  - Removed `mac.target.arch` arrays from `electron-builder.yml` so each job builds only host arch (fixes the cross-pack duplicate-upload bug that broke auto-update for 3.2.10 Intel users)
- **BDHLNDR-39**: Open Log Folder item in Help menu

## Conflict resolutions
- **package.json**: kept dev's `3.3.0-beta.2` (we're still on the beta channel at this point)
- **.github/workflows/release.yml**: kept dev's beta channel-gating + `latest*.yml` → `beta*.yml` rename step, AND added prod's mac yml merge step. Ordered the merge BEFORE the rename so the merged `artifacts/latest-mac.yml` is what gets renamed to `beta-mac.yml` on prereleases.

## Test plan
- [ ] CI: `check-version` gates correctly — this branch is dev, version is `3.3.0-beta.2`, tag exists, so no release should fire (expected short-circuit).
- [ ] After this lands, next step is a `development` → `production` PR, where on production the version will be bumped `3.2.11` → `3.3.0` to promote beta to stable.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,8 +46,18 @@ jobs:
             platform: linux
           - os: windows-latest
             platform: win
-          - os: macos-latest
-            platform: mac
+          # BDHLNDR-38: build each Mac arch on its native runner so postinstall
+          # (electron-rebuild) and sharp's optional-dep resolution produce
+          # host-native binaries. Cross-packaging from a single arm64 runner
+          # previously shipped arm64 .node files inside the x64 DMG, causing
+          # Intel Mac users to hang at the splash screen when better-sqlite3
+          # failed to load.
+          - os: macos-14
+            platform: mac-arm64
+            mac_arch: arm64
+          - os: macos-13
+            platform: mac-x64
+            mac_arch: x64
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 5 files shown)


brannon-bowden (approved):
Forward-merge done right — carrying 3.2.x hotfixes into dev before the beta→stable promotion is exactly the sequencing that keeps those fixes from getting dropped on the floor. Conflict resolutions are the right calls: keeping dev's beta channel-gating and `latest*.yml` → `beta*.yml` rename while pulling in prod's mac-yml merge step, with the merge ordered before the rename so the beta channel still gets the merged manifest. That ordering detail is easy to get wrong and would've quietly broken beta auto-update — nice catch.

PR #93, opened 2026-07-13
Task: Add SonarQube analysis config and CI quality gate workflow

### LINKED ISSUE #92: Task: Add SonarQube analysis config and CI quality gate workflow
## What

- Add `sonar-project.properties` with the project key and analysis scoping/exclusions
- Add `.github/workflows/sonarqube.yml` running the SonarQube scan + quality-gate check on pushes to `development`/`production` and on pull requests

The required repository secrets are already configured.

## Why

Bring this repo under the same static-analysis and quality-gate coverage as our other repositories.

William-Long-II:
## Description
Wires the repo up for SonarQube analysis:

- `sonar-project.properties` — project key, analysis scoped to `src/`, excludes static PWA assets and `.d.ts` declaration files
- `.github/workflows/sonarqube.yml` — runs the SonarQube scan and quality-gate check on pushes to `development`/`production` and on every pull request

The required repository secrets are already configured, so this PR's own checks exercise the new workflow.

## Related issue
Closes #92

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Hotfix
- [x] Refactor / chore
- [ ] Docs

## How was this tested?
- Workflow YAML validated with js-yaml
- The SonarQube workflow runs on this PR itself — scan + quality gate must pass as a status check

## Checklist
- [ ] Tests pass locally (no test suite in this repo)

### DIFF
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -1,0 +1,39 @@
+name: SonarQube
+
+on:
+  # Analyze long-lived branches on push, and every pull request.
+  push:
+    branches:
+      - development
+      - production
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+# Cancel superseded analyses only on pull requests. Pushes to development/production
+# run to completion so quality-gate analysis history on protected branches is preserved.
+concurrency:
+  group: sonarqube-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  quality-gate:
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Disabling shallow clone is recommended for improving relevancy of reporting.
+          fetch-depth: 0
+
+      - name: SonarQube Scan
+        uses: sonarsource/sonarqube-scan-action@master
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
+
+      - name: SonarQube Quality Gate check
+        uses: sonarsource/sonarqube-quality-gate-action@master
+        timeout-minutes: 5
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -2,12 +2,17 @@ name: SonarQube
 
 on:
   # Analyze long-lived branches on push, and every pull request.
+  # Note: `pull_request` runs from forks receive no repo secrets, so the scan
+  # cannot gate external-fork PRs — acceptable because this repo is internal-only.
   push:
     branches:
       - development
       - production
   pull_request:
     types: [opened, synchronize, reopened]
+
+permissions:
+  contents: read
 
 # Cancel superseded analyses only on pull requests. Pushes to development/production
 # run to completion so quality-gate analysis history on protected branches is preserved.
@@ -19,20 +24,21 @@ jobs:
   quality-gate:
     if: github.actor != 'dependabot[bot]'
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           # Disabling shallow clone is recommended for improving relevancy of reporting.
           fetch-depth: 0
 
       - name: SonarQube Scan
-        uses: sonarsource/sonarqube-scan-action@master
+        uses: sonarsource/sonarqube-scan-action@713881670b6b3676cda39549040e2d88c70d582e # v8.2.0
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
           SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
 
       - name: SonarQube Quality Gate check
-        uses: sonarsource/sonarqube-quality-gate-action@master
+        uses: sonarsource/sonarqube-quality-gate-action@cf038b0e0cdecfa9e56c198bbb7d21d751d62c3b # v1.2.0
         timeout-minutes: 5
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,8 @@
+sonar.projectKey=Software-Development-LLC_Bodhilander_2d637ced-8612-41e6-a7d6-d6127023db10
+
+# Only application code is analyzed. Everything outside src/ is either
+# packaging assets (build/, resources/), docs, or repo tooling (scripts/).
+sonar.sources=src
+
+# Generated / vendored artifacts inside src that are not authored code.
+sonar.exclusions=src/pwa/public/**,**/*.d.ts


.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,39 @@
...
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Disabling shallow clone is recommended for improving relevancy of reporting.
+          fetch-depth: 0
+
+      - name: SonarQube Scan
+        uses: sonarsource/sonarqube-scan-action@master
```
brannon-bowden: Pin to a full-length commit SHA instead of the mutable `@master` branch — this action receives SONAR_TOKEN/SONAR_HOST_URL and a compromised or unexpectedly-changed master would run with access to those secrets.

.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,39 @@
...
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: ${{ secrets.SONAR_HOST_URL }}
+
+      - name: SonarQube Quality Gate check
+        uses: sonarsource/sonarqube-quality-gate-action@master
```
brannon-bowden: Same issue as the scan action above — pin `sonarqube-quality-gate-action` to a commit SHA rather than `@master`.

.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+# Cancel superseded analyses only on pull requests. Pushes to development/production
+# run to completion so quality-gate analysis history on protected branches is preserved.
+concurrency:
+  group: sonarqube-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
```
brannon-bowden, line 23: Consider adding an explicit `permissions: contents: read` at the workflow or job level rather than relying on the repo's default token permissions.

.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,39 @@
...
+  quality-gate:
+    if: github.actor != 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Disabling shallow clone is recommended for improving relevancy of reporting.
+          fetch-depth: 0
+
+      - name: SonarQube Scan
```
brannon-bowden, line 34: No timeout on this step (or the job); a hung scan could run for the default 6-hour job timeout.

brannon-bowden (changes requested):
I'll proceed with a direct thorough review since I already have a complete analysis of this workflow diff.




## Verdict

Request changes — the workflow is well-structured (sane concurrency handling, dependabot exclusion, shallow-clone disabled for better blame relevancy) but pins third-party actions to a mutable branch ref while injecting secrets into them, which is a supply-chain risk that should be fixed before merge.

## Blocking issues

- **`.github/workflows/sonarqube.yml:29` and `:35` — third-party actions pinned to `@master` instead of a version tag or commit SHA.** Both `sonarsource/sonarqube-scan-action@master` and `sonarsource/sonarqube-quality-gate-action@master` point at a mutable branch. `SONAR_TOKEN` and `SONAR_HOST_URL` are passed into these steps as env vars, so any future push to that branch (compromise, or an unreviewed breaking change) runs with access to those secrets and to the runner, on every PR and every push to `development`/`production`. This is exactly the class of attack that hit other widely-used Actions via mutable-ref compromise. Pin to a full-length commit SHA (or at minimum a released version tag) and update deliberately, e.g. `sonarsource/sonarqube-scan-action@<sha> # vX.Y.Z`.

## Non-blocking issues

- **Missing explicit `permissions:` block** (workflow or job level, e.g. around `.github/workflows/sonarqube.yml:18-21`). The job doesn't need more than `contents: read`. Declaring that explicitly is cheap defense-in-depth, especially paired with third-party actions running on `pull_request`.
- **No timeout on the "SonarQube Scan" step** (`.github/workflows/sonarqube.yml:28-29`) or on the job overall — only the quality-gate step has `timeout-minutes: 5`. A hung scan could run for the full default 6-hour job timeout, burning CI minutes silently.
- **Fork PR behavior worth confirming**: `pull_request` (not `pull_request_target`) triggers on PRs from forks, but GitHub withholds repo secrets from fork-triggered `pull_request` runs by default. If this repo accepts external contributions, `SONAR_TOKEN`/`SONAR_HOST_URL` will be empty for those PRs and the scan step will silently no-op or fail rather than gate anything — intentional if this is an internal-only repo, but worth a one-line comment confirming that assumption.

## Tests

Not applicable in the traditional sense — this PR adds CI/config only, not application logic, so there's no unit-testable behavior. That said, there's no evidence in the diff that the workflow was actually exercised against a real quality-gate failure to confirm the job fails the check correctly (as opposed to just running to completion); worth a manual dry run if not already done.

.github/workflows/sonarqube.yml
William-Long-II: Fixed in 7456076 — pinned to `713881670b6b3676cda39549040e2d88c70d582e` (v8.2.0, resolved from the official release tag via the GitHub API).

.github/workflows/sonarqube.yml
William-Long-II: Fixed in 7456076 — pinned to `cf038b0e0cdecfa9e56c198bbb7d21d751d62c3b` (v1.2.0). Also pinned `actions/checkout` to the v7.0.0 SHA for consistency, which clears the Node 20 deprecation annotation from the first run.

.github/workflows/sonarqube.yml
William-Long-II: Fixed in 7456076 — added workflow-level `permissions: contents: read`.

.github/workflows/sonarqube.yml
William-Long-II: Fixed in 7456076 — added job-level `timeout-minutes: 15` (covers checkout + scan + gate; first run completed in 44s). Also added a comment documenting that fork `pull_request` runs get no secrets, per the non-blocking note — this repo is internal-only.

brannon-bowden (approved):
Approve — this is a clean, security-conscious CI addition with no blocking issues.

**Blocking issues**

None.

**Non-blocking issues**

- `.github/workflows/sonarqube.yml:10-11` — The `pull_request` trigger isn't scoped to same-repo PRs. If this repo ever accepts a PR from a fork, the `quality-gate` job will still run but `SONAR_TOKEN`/`SONAR_HOST_URL` will resolve to empty strings, causing the scan step to fail rather than being skipped cleanly. The header comment acknowledges the secrets-are-unavailable tradeoff but doesn't gate the job on `github.event.pull_request.head.repo.fork == false`, so the failure mode is a red required check rather than a graceful skip. Low risk given the "internal-only" assumption, but worth an explicit guard if that assumption ever changes.
- `sonar-project.properties:5` — Only `sonar.sources` is set; there's no `sonar.tests` or coverage report path (e.g. `sonar.javascript.lcov.reportPaths` or language equivalent) configured. If coverage is meant to feed the quality gate, confirm this is intentional — otherwise the gate's coverage condition will likely always read 0%.
- `.github/workflows/sonarqube.yml:19` — `concurrency.group: sonarqube-${{ github.ref }}` is a repo-wide (not workflow-scoped) key. Unlikely to collide today, but worth namespacing with the workflow name (e.g. `sonarqube-scan-${{ github.ref }}`) if other workflows are added later.
- `sonar-project.properties:1` — No `sonar.projectName`; purely cosmetic for the SonarQube UI, not required.

Good practices worth calling out: actions are pinned to full commit SHAs with version comments (not floating tags), `permissions` is scoped to `contents: read` only, the trigger deliberately avoids `pull_request_target` so untrusted fork code can't access secrets, and `dependabot[bot]` is explicitly excluded.

**Tests**

This is CI/config-only (no application code), so unit tests aren't applicable. The listed CI status shows `SonarQube Code Analysis`, `quality-gate`, and `GitGuardian Security Checks` all passing, which is effectively the functional validation for a workflow-definition change like this.

Issue #98, opened 2026-07-13
Feature: New-session provider picker (default Claude)


William-Long-II:
### User story / problem

As a user creating a session I want to choose which provider it runs, defaulting to Claude, so provider choice is a one-click decision at session start.

### Proposed solution

Provider selector in the new-session flow (renderer session store `createSession` + creation UI). Defaults to Claude; undetected CLIs are disabled with a pointer to Settings → Providers.

### Acceptance criteria

- [ ] Creating a session with no explicit choice yields a Claude session (current behavior preserved)
- [ ] Picker lists detected providers; undetected ones are disabled with guidance
- [ ] Chosen provider is persisted on the session and used to spawn it

### Scope

Minor feature

### Priority

High

William-Long-II:
Heads-up from #96: the renderer mounts Terminal optimistically before `createDbSession` resolves, so `pty:create`'s DB-row lookup can race row persistence (pre-existing — groupId/memory injection has the same window). Once the picker introduces non-default providers, a lost race would silently launch Claude instead of the chosen provider. When implementing the picker, either await persistence before mounting the terminal or pass the provider explicitly through `electronAPI.createSession`.

Issue #95, opened 2026-07-13
Feature: Provider registry and generalized session launcher


William-Long-II:
### User story / problem

As a user I want sessions to run agentic CLIs other than Claude Code, so I can use the agent that fits my task and subscriptions.

### Proposed solution

Extract the hardcoded Claude launch path (`src/main/claude-launcher.ts`, literal `claude` command strings in `src/main/pty-manager.ts`) into a provider registry under `src/main/providers/`. Each provider definition declares: CLI command, arg builders (resume/session-id flag, system-prompt flag), env vars, terminal waiting-state detection patterns, and capability flags (hooks, resume, MCP). Ship definitions for Claude Code, OpenAI Codex, Gemini CLI, and Grok Build; providers without state patterns degrade to basic running/idle detection.

### Acceptance criteria

- [ ] `pty-manager` spawns sessions via provider lookup; no literal `claude` command strings remain outside the Claude provider definition
- [ ] Claude sessions behave exactly as before (resume, hooks, system prompt, state detection)
- [ ] Codex, Gemini, and Grok Build definitions launch their CLIs when installed
- [ ] Capability flags gate provider-specific features gracefully

### Scope

Major feature

### Priority

High

Issue #94, opened 2026-07-13
Epic: Multi-provider sessions and competitive arena mode


William-Long-II:
### Summary / goal

Let users create sessions backed by agentic CLIs other than Claude Code (OpenAI Codex, Gemini CLI, Grok Build) with Claude remaining the default, and add a competitive "arena" mode that sends one prompt to multiple models via their APIs and compares responses, latency, tokens, and cost side-by-side.

**Phase 1 — Multi-provider agentic sessions (PTY):** generalize the hardcoded Claude launch path into a provider registry; add a `provider` column to sessions; add Settings → Providers with installed-CLI detection and login guidance; add a provider picker to the new-session flow (default Claude).

**Phase 2 — Competitive arena mode (API):** encrypted API-key vault via Electron safeStorage; arena session kind that fans one prompt out to selected models (xAI, OpenAI, Gemini, Anthropic, Ollama) with streaming side-by-side results, per-model latency/token/cost metrics, and persisted run history.

### Why / business value

Users aren't locked to a single vendor; the app becomes a general agentic-session manager. Arena mode gives an objective way to pick the best/cheapest model for a given kind of task.

### Child issues / scope

(tracked as GitHub sub-issues)

### Definition of done

- A new session can be created against any detected provider CLI, defaulting to Claude; existing Claude sessions are unaffected.
- Settings shows provider availability and setup guidance; API keys are stored encrypted (OS keychain via safeStorage), never in plaintext.
- Arena mode runs one prompt across ≥2 configured models concurrently and displays responses with latency, token counts, and computed cost; runs are persisted.

### Priority

High

William-Long-II:
All six child issues are merged and closed — the epic is complete.

**Phase 1 — Multi-provider agentic sessions:**
- #95 Provider registry + generalized launcher (Claude Code, Codex, Gemini CLI, Grok Build)
- #96 Session provider persistence, resume-UUID invalidation on provider change
- #97 Settings → Providers with login-shell CLI detection + setup guidance
- #98 New-session provider picker (Claude default, first-launch race fixed)

**Phase 2 — Competitive arena + optional keys:**
- #100 Arena mode: one prompt fanned out to CLIs + Ollama, streamed side-by-side with TTFT/latency/token/cost metrics, persisted history — every contestant on its CLI's own login/subscription per the subscription-preservation decision
[...]

PR #101, opened 2026-07-13
Provider registry and generalized session launcher

William-Long-II:
## Description
First slice of the multi-provider epic (#94): replaces the hardcoded Claude launch path with a provider registry.

- **`src/main/providers/`** — `ProviderDefinition` (command, launch args/env, capability flags for resume/hooks/systemPrompt/accounts, TUI waiting patterns) with definitions for Claude Code, Codex (OpenAI), Gemini CLI, and Grok Build. Codex/Gemini/Grok share a `passthroughProvider` factory; the hook-env contract lives in one `baseSessionEnv` helper.
- **`pty-manager.ts`** — spawns sessions via provider lookup (default `claude`, so all existing call sites are untouched); waiting-state detection merges provider-specific patterns with the generic set (memoized per provider); shell wrapping (WSL/PowerShell/cmd/bash) extracted to a single `getShellLaunch` shared by regular and login ptys; the triplicated idle-timeout block deduplicated into `scheduleIdleTimeout`.
- `claude-launcher.ts` deleted; its logic moved verbatim into the Claude provider definition.

Capability-gated behavior: UUID resume machinery, account resolution, memory/system-prompt injection, and memory-file writes only run for providers that support them (Claude only, today).

Not in scope (tracked separately): sessions `provider` DB column + threading (#96), Settings CLI detection (#97), new-session picker (#98). Latent notes from self-review posted on #96 (UUID provider-namespacing, unknown-provider-id fallback).

## Related issue
Closes #95

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `tsc -p tsconfig.main.json` clean (two pre-existing unrelated errors in `mdns-advertiser.ts` excluded)
- Parity smoke test (node, mocked electron): Claude `buildCommand` produces byte-identical command/args/env/hook-script to the old launcher for resume + new modes; registry lookup/defaults/UUID format; passthrough providers launch bare with correct env
- Removed-behavior audit: all 17 waiting patterns accounted for; env construction, session flags, resume fallback, and memory gating verified equivalent for Claude
- SonarQube preflight on all changed files: providers clean; the two cognitive-complexity criticals raised on modified functions were refactored away
- Not yet run against a live Codex/Gemini/Grok CLI — worth a manual QA pass on a machine with those installed

## Checklist
- [ ] Tests pass locally (no test suite in this repo; smoke test above)
[...]

PR #102, opened 2026-07-13
Session provider persistence and launch-path threading

### LINKED ISSUE #96: Feature: Session provider persistence and state-detection generalization
### User story / problem

As a user I want each session to remember which provider it runs, so restarts, resume, and history work per provider.

### Proposed solution

Migration adding `provider` column to `sessions` (default `claude`; `shell_type` stays for bash). Thread provider through `src/shared/types.ts`, repositories, IPC, and renderer session store. Generalize state detection so non-Claude providers report sane states.

### Acceptance criteria

- [ ] Existing session rows migrate cleanly with `provider='claude'`
- [ ] Session create/resume round-trips the provider across restart
- [ ] Session list/UI shows the provider on non-Claude sessions

### Scope

Minor feature

### Priority

High

William-Long-II:
## Description
Second slice of the multi-provider epic (#94): sessions now remember which provider they run.

- **Migration**: `sessions.provider` column, `TEXT NOT NULL DEFAULT 'claude'` — existing rows migrate as Claude sessions with no behavior change.
- **Threading**: `Session` type → repo mapper/insert/update → `pty:create` (resolves provider from the stored row, so the preload/renderer signatures are unchanged) → `PtyManager.createSession`. Mobile API create/start and group import/export carry the field; PWA type gains it as optional.
- **Deferred findings from #95 closed out**: updating a session's provider clears its stored conversation UUID in the same SQL statement (one provider's resume id is never replayed against another CLI), and `pty:create` falls back to the default provider with a logged warning when a row references an unknown provider id instead of failing the launch.
- **UI**: sidebar session rows show a small provider badge when an agent session runs a non-Claude provider; Claude sessions look exactly as before.
- Renderer store `createSession` accepts an optional `provider` argument as groundwork for the #98 picker (a persistence/mount race relevant to that picker is documented on #98).

## Related issue
Closes #96

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `tsc` clean on main, base (renderer), and pwa tsconfigs
- SQL semantics verified against real SQLite: migration backfills `claude` on existing rows; same-provider updates preserve the stored UUID; provider changes clear it; explicit-provider inserts round-trip
- SonarQube preflight on the repo file with the new logic: clean
- Manual QA still worth doing: launch the app, confirm existing sessions open and resume exactly as before after the migration runs

## Checklist
- [ ] Tests pass locally (no test suite; SQL semantics + typecheck verification above)

### DIFF
--- a/src/main/api/routes/sessions.ts
+++ b/src/main/api/routes/sessions.ts
@@ -195,6 +195,7 @@ export function createSessionsRouter(): Router {
           endedAt: null,
           durationSeconds: 0,
           claudeAccountId: null,
+          provider: 'claude',
         };
 
         sessionsRepo.createSession(session);
@@ -208,7 +209,7 @@ export function createSessionsRouter(): Router {
 
         // Start PTY if requested
         if (launchClaude) {
-          ptyManager.createSession(id, session.workingDir, true, groupId);
+          ptyManager.createSession(id, session.workingDir, true, groupId, session.provider);
         }
 
         log.info(`[SessionsAPI] Created session: ${id}`);
@@ -305,7 +306,7 @@ export function createSessionsRouter(): Router {
           return;
         }
 
-        ptyManager.createSession(id, session.workingDir, launchClaude ?? false, session.groupId);
+        ptyManager.createSession(id, session.workingDir, launchClaude ?? false, session.groupId, session.provider);
 
         log.info(`[SessionsAPI] Started session: ${id}`);
         res.json({ success: true });
--- a/src/main/group-import-export.ts
+++ b/src/main/group-import-export.ts
@@ -40,6 +40,8 @@ interface PortableSession {
   order: number;
   createdAt: string;
   lastActivityAt: string;
+  /** Agent provider registry id (#96); absent in exports from older versions. */
+  provider?: string;
 }
 
 interface PortableData {
@@ -100,6 +102,7 @@ export async function exportGroupsAndSessions(): Promise<ExportResult> {
         order: s.order,
         createdAt: s.createdAt instanceof Date ? s.createdAt.toISOString() : String(s.createdAt),
         lastActivityAt: s.lastActivityAt instanceof Date ? s.lastActivityAt.toISOString() : String(s.lastActivityAt),
+        provider: s.provider ?? 'claude',
       })),
     };
 
@@ -225,6 +228,7 @@ export async function importGroupsAndSessions(): Promise<ImportResult> {
         endedAt: null,
         durationSeconds: 0,
         claudeAccountId: null,
+        provider: s.provider ?? 'claude',
       });
       sessionCount++;
     }
@@ -363,6 +367,7 @@ export async function importFromClaudeLander(): Promise<ImportResult> {
         endedAt: null,
         durationSeconds: 0,
         claudeAccountId: null,
+        provider: row.provider ?? 'claude',
       });
       sessionCount++;
     }
--- a/src/main/group-import-export.ts
+++ b/src/main/group-import-export.ts
@@ -13,7 +13,23 @@ import { randomUUID } from 'crypto';
 import Database from 'better-sqlite3';
 import * as groupsRepo from './repositories/groups';
 import * as sessionsRepo from './repositories/sessions';
+import { isKnownProvider, DEFAULT_PROVIDER_ID } from './providers';
 import log from 'electron-log';
+
+/**
+ * Validate an imported provider id at import time (#96). Unknown ids — e.g.
+ * an export written by a newer app version — are logged and defaulted here so
+ * they never land in the DB, rather than relying solely on the launch-path
+ * fallback. Exported for tests.
+ */
+export function sanitizeImportedProvider(provider: string | null | undefined): string {
+  const id = provider ?? DEFAULT_PROVIDER_ID;
+  if (!isKnownProvider(id)) {
+    log.warn(`[Import/Export] Unknown provider '${id}' in imported session; defaulting to '${DEFAULT_PROVIDER_ID}'`);
+    return DEFAULT_PROVIDER_ID;
+  }
+  return id;
+}
 
 // ---------------------------------------------------------------------------
 // Portable format types
@@ -228,7 +244,7 @@ export async function importGroupsAndSessions(): Promise<ImportResult> {
         endedAt: null,
         durationSeconds: 0,
         claudeAccountId: null,
-        provider: s.provider ?? 'claude',
+        provider: sanitizeImportedProvider(s.provider),
       });
       sessionCount++;
     }
@@ -367,7 +383,7 @@ export async function importFromClaudeLander(): Promise<ImportResult> {
         endedAt: null,
         durationSeconds: 0,
         claudeAccountId: null,
-        provider: row.provider ?? 'claude',
+        provider: sanitizeImportedProvider(row.provider),
       });
       sessionCount++;
     }
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import { execFile } from 'child_process';
 import { EventEmitter } from 'events';
 import {
-  getProvider,
+  resolveProvider,
   getSocketPath,
   generateAgentSessionId,
   claudeProvider,
@@ -230,8 +230,10 @@ export class PtyManager extends EventEmitter {
       throw new Error(errorMsg);
     }
 
-    // Track agent resume state for early-exit fallback (BDHLNDR-9)
-    const provider = launchClaude ? getProvider(providerId) : null;
+    // Track agent resume state for early-exit fallback (BDHLNDR-9).
+    // resolveProvider degrades unknown ids (e.g. rows written by a newer app
+    // version) to the default provider instead of failing the launch (#96).
+    const provider = launchClaude ? resolveProvider(providerId, `session ${id}`) : null;
     let resumeAttempted = false;
 
     if (provider) {
--- a/src/shared/types.ts
+++ b/src/shared/types.ts
@@ -25,7 +25,21 @@ export interface Session {
    * to the global default account, then to the legacy ~/.claude config).
    */
   claudeAccountId: string | null;
+  /**
+   * Agent provider this session runs (providers registry id: 'claude',
+   * 'codex', 'gemini', 'grok') (#96). Only meaningful when shellType is
+   * 'claude'; plain shell sessions keep the default.
+   */
+  provider: string;
 }
+
+/** Display labels for session providers (registry ids → short names). */
+export const PROVIDER_LABELS: Record<string, string> = {
+  claude: 'Claude',
+  codex: 'Codex',
+  gemini: 'Gemini',
+  grok: 'Grok',
+};
 
 export interface Group {
   id: string;
--- a/src/main/database.ts
+++ b/src/main/database.ts
@@ -143,6 +143,13 @@ function initializeTables(database: Database.Database): void {
     database.exec("ALTER TABLE sessions ADD COLUMN claude_account_id TEXT DEFAULT NULL");
   }
 
+  // Migration: Add provider column to sessions (#96, epic #94).
+  // Which agent provider the session runs (providers registry id). Existing
+  // rows are Claude sessions by definition, so they default to 'claude'.
+  if (!sessionColsBdhlndr17.includes('provider')) {
+    database.exec("ALTER TABLE sessions ADD COLUMN provider TEXT NOT NULL DEFAULT 'claude'");
+  }
+
   // Migration: Create session_events table (BDHLNDR-17)
   database.exec(`
     CREATE TABLE IF NOT EXISTS session_events (
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,7 +1,7 @@
 import { app, BrowserWindow, ipcMain, dialog, shell, crashReporter } from 'electron';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
-import { DEFAULT_PROVIDER_ID, isKnownProvider } from './providers';
+import { DEFAULT_PROVIDER_ID } from './providers';
 import { getDatabase, closeDatabase } from './database';
 import * as groupsRepo from './repositories/groups';
 import * as sessionsRepo from './repositories/sessions';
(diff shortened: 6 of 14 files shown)


src/main/api/routes/sessions.ts
```diff
@@ -305,7 +306,7 @@ export function createSessionsRouter(): Router {
           return;
         }
 
-        ptyManager.createSession(id, session.workingDir, launchClaude ?? false, session.groupId);
+        ptyManager.createSession(id, session.workingDir, launchClaude ?? false, session.groupId, session.provider);
```
brannon-bowden, line 309: Unlike src/main/index.ts's pty:create handler, this launch path doesn't validate session.provider against isKnownProvider before handing it to ptyManager.createSession. A session row with an unrecognized provider id (e.g. from an import written by a newer app version, or a downgrade) will fail here where the Electron IPC path falls back gracefully. Add the same isKnownProvider guard + fallback used in index.ts.

src/main/group-import-export.ts
```diff
@@ -225,6 +228,7 @@ export async function importGroupsAndSessions(): Promise<ImportResult> {
         endedAt: null,
         durationSeconds: 0,
         claudeAccountId: null,
+        provider: s.provider ?? 'claude',
```
brannon-bowden: Imported provider values are only defaulted when absent/null; an unrecognized-but-present provider id from a newer export is persisted without validation or logging. Consider validating against isKnownProvider at import time.

src/shared/types.ts
```diff
@@ -25,8 +25,22 @@ export interface Session {
    * to the global default account, then to the legacy ~/.claude config).
    */
   claudeAccountId: string | null;
+  /**
+   * Agent provider this session runs (providers registry id: 'claude',
+   * 'codex', 'gemini', 'grok') (#96). Only meaningful when shellType is
+   * 'claude'; plain shell sessions keep the default.
```
brannon-bowden, line 31: Session.provider is required here but optional in src/pwa/src/lib/types.ts. Confirm no other Session-literal construction site outside this diff is broken by making the field required.

brannon-bowden (changes requested):
Request changes. The core plumbing (migration, repo mapping, export/import, badge UI) is solid, but the safety net this PR itself introduces for unrecognized provider IDs is applied inconsistently across launch paths, and the one genuinely tricky piece of new logic (invalidating a stored resume token when the provider changes) ships with no tests.

## Blocking issues

- **`src/main/api/routes/sessions.ts:309`** — The "start session" route passes `session.provider` straight into `ptyManager.createSession` with no validation:
  ```js
  ptyManager.createSession(id, session.workingDir, launchClaude ?? false, session.groupId, session.provider);
  ```
  Compare this to `src/main/index.ts:664-673`, where the exact same input (a provider value read back off a stored session row) is explicitly guarded:
  ```js
  // Resolve the session's provider from its stored row (#96). Fall back to
  // the default when the row references a provider this app version doesn't
  // know (e.g. written by a newer version) instead of failing the launch.
  let providerId = session?.provider ?? DEFAULT_PROVIDER_ID;
  if (!isKnownProvider(providerId)) {
    log.warn(...);
    providerId = DEFAULT_PROVIDER_ID;
  }
  ```
  The comment shows the author is aware this scenario happens (imports from a newer app version, downgrades, hand-edited data). If `getProvider` throws for an unknown id (it does, per `src/main/providers/index.ts`), this route will fail to start a session that the desktop IPC path would have started fine. Same class of risk, same PR, inconsistent handling. Apply the same `isKnownProvider` fallback here (and in the create-session block, for defense in depth even though it's currently hardcoded to `'claude'`).

- **No tests for the non-trivial new logic.** Two behaviors introduced here are exactly the kind that need coverage, not just a docstring:
  - `src/main/repositories/sessions.ts` — the `updateSession` CASE expression that nulls out `claude_session_id` when `provider` changes. This is a correctness/safety mechanism (preventing a resume token from one provider's CLI being replayed against a different provider), relies on subtle SQL evaluation-order semantics, and has zero test coverage.
  - `src/main/index.ts` — the unknown-provider fallback path (`isKnownProvider` + warn + default) has no test exercising a session row with a bogus/foreign provider id.
  Given the PR touches sensitive, unchecked-out paths and changes DB write behavior, at least unit tests around `updateSession` (provider unchanged vs. changed) and the fallback resolution are warranted before merge.

## Non-blocking issues

- **`src/main/group-import-export.ts:231` (and the analogous `importFromClaudeLander` site ~line 371)** — imported sessions persist whatever string is in `s.provider`/`row.provider` verbatim (only `undefined`/`null` gets defaulted to `'claude'`). An import from a newer app version with an unrecognized provider id will silently land in the DB with no log at import time; the only enforcement is downstream at PTY launch, which (per the blocking issue above) isn't uniformly applied either. Consider validating with `isKnownProvider` at import time and logging/falling back there too.
- **`src/shared/types.ts:31`** — `Session.provider` is now a required `string`, while `src/pwa/src/lib/types.ts` keeps the equivalent field optional. That's plausibly intentional (renderer always gets a fully-populated row via IPC, PWA may talk to older backends), but worth double-checking no other `Session` literal construction elsewhere in the codebase (outside this diff) still omits `provider` — that would now be a compile error.
- `src/main/pty-manager.ts` isn't part of this diff. Worth confirming (if not already the case from earlier work) that its `createSession` actually consumes the new 5th `provider` argument to launch the right CLI — otherwise this PR only adds persistence/UI plumbing without a functional effect on what gets launched.

## Tests

No test files were changed. At minimum, add coverage for:
- `updateSession` provider-change → `claude_session_id` reset (and no-op when provider is unchanged or set to the same value).
- The unknown/stale provider fallback (`isKnownProvider` + default) in `src/main/index.ts`.
- Round-trip export/import preserving `provider`, including the "field absent" backward-compat default.

src/main/api/routes/sessions.ts
William-Long-II: Fixed in c6979f3, one level deeper than suggested: the fallback now lives inside `PtyManager.createSession` itself (via a new `resolveProvider()` in the registry), so this route, the Electron IPC path, and any future launch path share a single guard — the inline copy in `index.ts` was removed rather than duplicated here. Covered by tests in `src/main/providers/__tests__/resolve.test.ts`.

src/main/group-import-export.ts
William-Long-II: Fixed in c6979f3 — both import sites (portable JSON and direct ClaudeLander DB import) now run ids through `sanitizeImportedProvider()`, which validates against the registry and logs + defaults unknown ids at import time so they never land in the DB.

src/shared/types.ts
William-Long-II: Confirmed — `tsc --noEmit` is clean across the main, base (renderer), and pwa tsconfigs, so every `Session` literal site now supplies `provider` (store, mobile API, both import paths, test factory). The PWA field is intentionally optional since that client can talk to an older desktop backend over the pairing API.

William-Long-II:
All review items addressed in c6979f3:

- **Unknown-provider guard** — centralized in `PtyManager.createSession` via `resolveProvider()` so every launch path (IPC, mobile API create + start) is uniformly safe; the per-site copy in `index.ts` was removed. Import paths additionally validate at import time so bad ids never persist.
- **Tests** — added `bun test` suites (11 tests): provider resolution (known/unknown/null/throw contract) and the sessions repo (provider round-trip, pre-migration default, and the four UUID-invalidation cases: provider changed / set-to-same / provider-less update / switch-back-doesn't-resurrect). The repo suite runs against `bun:sqlite` so it needs no native build.
- **pty-manager 5th arg** — confirmed: `createSession` has consumed `providerId` since #101 (`.worktrees` PR); it selects the provider definition that builds the spawn command, so this PR's threading is functionally live, not just persisted.

Re-requesting review.

brannon-bowden (approved):
Approve — this is a well-scoped, carefully-tested change that threads the new `provider` field from storage through the launch path, with sensible defaulting for legacy data and imports.

**Blocking issues**

None. The core mechanics hold up under scrutiny:
- The migration (`src/main/database.ts:146-152`) adds `provider TEXT NOT NULL DEFAULT 'claude'`, which is a valid SQLite `ALTER TABLE ADD COLUMN` (constant default + NOT NULL), and the guard against re-running it is consistent with the existing pattern in this function.
- The resume-UUID invalidation in `updateSession` (`src/main/repositories/sessions.ts:121-130`) relies on SQL `UPDATE ... SET` evaluating all right-hand-side expressions against the pre-update row — that's correct SQLite/standard-SQL behavior, so `provider IS ?` in the `CASE` really does compare against the *old* value regardless of clause ordering. The accompanying tests exercise the interesting cases (change, no-op, unrelated update, round-trip) and all pass with the described semantics.
- `resolveProvider` vs. `getProvider` is a clean split: launch paths degrade unknown ids to Claude instead of failing a session start, while explicit validation call sites can still use the throwing `getProvider`. Import sanitization (`sanitizeImportedProvider`) mirrors the same "never let an unrecognized id reach the DB" principle at a different boundary.
- CI is green (SonarQube, quality-gate, GitGuardian), which is reassuring given that `Session.provider` in `src/shared/types.ts` was made a *required* field — a change that could have broken any other `Session` object-literal construction elsewhere in the codebase not touched by this diff; a clean build suggests all such sites were already covered.

**Non-blocking issues**

- `src/main/api/routes/sessions.ts:198` hardcodes `provider: 'claude'` on session creation via this HTTP route, while the renderer's local `createSession` (`src/renderer/store/sessions.ts:48`) now threads an arbitrary `provider` argument. If this route is reachable from a client that should be able to pick a provider (e.g. the PWA, given the new optional `provider?` field added to `src/pwa/src/lib/types.ts`), it currently can't — every session created through this path is silently pinned to Claude. Worth confirming this is an intentional scope boundary for this PR rather than an oversight.
- `src/main/providers/index.ts:41` — `resolveProvider`'s fallback returns the `claudeProvider` object literal directly rather than resolving `DEFAULT_PROVIDER_ID` through the registry. The log message on line 39 references `DEFAULT_PROVIDER_ID` while the return value is hardcoded; these will only ever diverge if someone changes `DEFAULT_PROVIDER_ID` without updating this function, but it's a needless coupling for a one-line fix (`REGISTRY.get(DEFAULT_PROVIDER_ID)!`).
- `src/main/group-import-export.ts:121` uses the literal `'claude'` as the export-time fallback rather than the already-imported `DEFAULT_PROVIDER_ID` constant — minor inconsistency, purely stylistic.

**Tests**

Solid coverage for the two riskiest pieces of logic (provider resolution and the resume-UUID invalidation on provider change), including edge cases like round-tripping and switching back to the original provider. One gap: `sanitizeImportedProvider` in `src/main/group-import-export.ts:19-27` is explicitly annotated "Exported for tests" but no test file exercises it (e.g. unknown provider id → defaults and logs, `null`/`undefined` → default, known id → passthrough). Given it's the last line of defense against bad data entering the DB via import, a small dedicated test would be worth adding.

src/main/pty-manager.ts
```diff
@@ -230,8 +230,10 @@ export class PtyManager extends EventEmitter {
       throw new Error(errorMsg);
     }
 
-    // Track agent resume state for early-exit fallback (BDHLNDR-9)
-    const provider = launchClaude ? getProvider(providerId) : null;
+    // Track agent resume state for early-exit fallback (BDHLNDR-9).
+    // resolveProvider degrades unknown ids (e.g. rows written by a newer app
+    // version) to the default provider instead of failing the launch (#96).
+    const provider = launchClaude ? resolveProvider(providerId, `session ${id}`) : null;
```
brannon-bowden, line 236: resolveProvider degrades an unknown provider id at launch time but never persists the corrected value back to the session row, so the DB (and the UI badge) can keep showing a stale/unknown provider while the session actually runs as claude. Consider writing the resolved id back via updateSession.

brannon-bowden (approved):
I'll proceed with a direct thorough manual review since the skill file isn't accessible in this environment.

Based on my analysis of the diff, here is the review:




## Verdict

Approve. The provider-persistence and launch-path threading is implemented soundly: the migration is backward-compatible, the `updateSession` UUID-invalidation logic correctly relies on SQLite's pre-update SET-clause semantics (and is explicitly tested for that), and unknown/imported provider ids are defensively sanitized before they ever reach the DB or launch path. CI (SonarQube, quality-gate, GitGuardian) is green.

## Blocking issues

None.

## Non-blocking issues

- **`src/main/pty-manager.ts:236`** (major) — When `resolveProvider` degrades an unknown/future provider id to `claude` at launch time, the corrected value is never written back to the session row. The session's persisted `provider` column keeps the stale/unknown value, so: (a) every future launch re-logs the same "unknown provider" warning instead of self-healing, and (b) the UI badge (`App.tsx`, `PROVIDER_LABELS[session.provider] ?? session.provider`) will keep showing the original unrecognized id even though the session is actually now running as a Claude session under the hood — a visible mismatch between displayed state and runtime behavior. Consider persisting the resolved id back via `updateSession` when a degrade occurs.
- **`src/main/api/routes/sessions.ts:198`** (minor) — The session-creation route hardcodes `provider: 'claude'` rather than accepting a provider from the request body, so sessions created through this API (e.g. remote/PWA access) can never be anything but Claude, even though the rest of the stack (registry, `resolveProvider`, `PROVIDER_LABELS`) supports `codex`/`gemini`/`grok`. If provider selection is meant to be desktop-only for now, that's fine, but worth confirming it's intentional rather than an oversight.
- **`src/renderer/styles/global.css:488`** (nit) — `.session-provider-badge` hardcodes colors (`#2d3a4d` / `#9fc3f0`) rather than referencing existing theme tokens/variables used elsewhere in the file; low priority but worth aligning if the codebase has a theming convention.

## Tests

- Good coverage: `resolveProvider`/`isKnownProvider` behavior (`providers/__tests__/resolve.test.ts`) and the provider-persistence + resume-UUID-invalidation logic in `sessions.repositories.__tests__/sessions.test.ts` are both well exercised, including the tricky "switch away and back" case.
- **Gap (minor):** `sanitizeImportedProvider` in `group-import-export.ts:26` is a new function explicitly annotated "Exported for tests," but no test file exercises it directly (no `group-import-export.test.ts` in this diff). It reuses already-tested `isKnownProvider`, so the residual risk is small, but it's the one piece of new import/export validation logic left unverified.

PR #103, opened 2026-07-13
Settings → Providers with CLI detection and setup guidance

### LINKED ISSUE #97: Feature: Settings → Providers with CLI detection and setup guidance
### User story / problem

As a user I want to see which provider CLIs are installed and how to set up the missing ones, so getting a new provider working is self-service.

### Proposed solution

New Providers section in `SettingsModal.tsx` backed by a main-process detector (PATH lookup + version probe per provider). Shows detected/not-detected per provider with install and login guidance links. Auth remains with each CLI in this phase.

### Acceptance criteria

- [ ] Providers section lists Claude, Codex, Gemini, Grok Build with live detected/missing status
- [ ] Missing providers show install/login instructions
- [ ] Detection re-runs on demand (refresh) without app restart

### Scope

Minor feature

### Priority

Medium

William-Long-II:
## Description
Third slice of the multi-provider epic (#94): a Providers tab in Settings showing which agent CLIs are installed, with self-service setup guidance for the missing ones.

- **Detection matches how sessions launch**: probes run through the user's login shell on macOS/Linux (GUI-launched Electron doesn't inherit shell PATH), via `wsl.exe bash -lc` when the configured shell is WSL, and via `where.exe` on native Windows. Command names come from the static provider registry only.
- **Per-provider status row**: green/grey dot, CLI command, detected version (best-effort first version-looking line of `--version`); missing providers expand with a copyable install command, sign-in hint, and a docs link (existing `shell:openExternal` bridge).
- **Refresh** re-runs detection on demand — no app restart (acceptance criterion).
- Setup metadata (`installHint`/`docsUrl`/`loginHint`) lives on each `ProviderDefinition`, so #98's picker can reuse it for disabled entries.
- New `providers:detect` IPC via the existing `safeHandle` wrapper + `detectProviders()` preload method.

## Related issue
Closes #97

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `bun test`: 12 pass, including a machine-independent contract test for `detectProviders` (one fully-populated status per registered provider)
- Real detection exercised on macOS: `claude` found through the login shell with version `2.1.207 (Claude Code)`; codex/gemini/grok correctly reported missing
- `tsc --noEmit` clean on main and renderer configs
- SonarQube preflight on the new detector: one minor (`??` vs `||`) found and fixed pre-push
- Manual QA worth doing: open Settings → Providers on Windows/WSL to confirm the platform branches

## Checklist
- [ ] Tests pass locally (bun test, 12 pass)

### DIFF
--- a/src/main/provider-detector.ts
+++ b/src/main/provider-detector.ts
@@ -1,0 +1,89 @@
+/**
+ * Provider CLI detection (#97).
+ *
+ * Probes each registered provider's CLI the same way sessions launch it:
+ * through the user's login shell on macOS/Linux (GUI-launched Electron does
+ * not inherit shell PATH additions), via `wsl.exe bash` when the configured
+ * shell is WSL, and via `where.exe` on native Windows. Command names come
+ * from the static provider registry, never from user input.
+ */
+import { execFile } from 'child_process';
+import { listProviders } from './providers';
+import { detectShell } from './shell-detector';
+import { getPreference } from './repositories/preferences';
+import { ProviderStatus } from '../shared/types';
+
+const PROBE_TIMEOUT_MS = 10_000;
+
+interface Probe {
+  /** argv to check the command exists (exit 0 = installed). */
+  lookup: [string, string[]];
+  /** argv to fetch the version string (best effort). */
+  version: [string, string[]];
+}
+
+function buildProbe(command: string): Probe {
+  const shellInfo = detectShell(getPreference('customShellPath') ?? '');
+
+  if (shellInfo.isWSL) {
+    const wrap = (cmd: string): [string, string[]] =>
+      ['wsl.exe', [...shellInfo.args, '--', 'bash', '-lc', cmd]];
+    return { lookup: wrap(`command -v ${command}`), version: wrap(`${command} --version`) };
+  }
+  if (process.platform === 'win32') {
+    return {
+      lookup: ['where.exe', [command]],
+      version: ['cmd.exe', ['/c', `${command} --version`]],
+    };
+  }
+  // Login (non-interactive) shell: picks up PATH from the user's profile
+  // without interactive-rc noise on stdout.
+  const wrap = (cmd: string): [string, string[]] => [shellInfo.shell, ['-l', '-c', cmd]];
+  return { lookup: wrap(`command -v ${command}`), version: wrap(`${command} --version`) };
+}
+
+function run([file, args]: [string, string[]]): Promise<{ ok: boolean; stdout: string }> {
+  return new Promise((resolve) => {
+    execFile(file, args, { timeout: PROBE_TIMEOUT_MS, windowsHide: true }, (err, stdout) => {
+      resolve({ ok: !err, stdout: stdout?.toString() ?? '' });
+    });
+  });
+}
+
+/** Pick the first output line that looks like a version (contains a digit). */
+function extractVersion(stdout: string): string | null {
+  const line = stdout
+    .split('\n')
+    .map((l) => l.trim())
+    .find((l) => l && /\d/.test(l));
+  return line ? line.slice(0, 100) : null;
+}
+
+async function probeCommand(command: string): Promise<{ installed: boolean; version: string | null }> {
+  const probe = buildProbe(command);
+  const lookup = await run(probe.lookup);
+  if (!lookup.ok) {
+    return { installed: false, version: null };
+  }
+  const version = await run(probe.version);
+  return { installed: true, version: version.ok ? extractVersion(version.stdout) : null };
+}
+
+/** Probe every registered provider concurrently. */
+export function detectProviders(): Promise<ProviderStatus[]> {
+  return Promise.all(
+    listProviders().map(async (p): Promise<ProviderStatus> => {
+      const { installed, version } = await probeCommand(p.command);
+      return {
+        id: p.id,
+        name: p.name,
+        command: p.command,
+        installed,
+        version,
+        installHint: p.setup.installHint,
+        docsUrl: p.setup.docsUrl,
+        loginHint: p.setup.loginHint,
+      };
+    })
+  );
+}
--- a/src/main/provider-detector.ts
+++ b/src/main/provider-detector.ts
@@ -22,7 +22,8 @@ interface Probe {
   version: [string, string[]];
 }
 
-function buildProbe(command: string): Probe {
+/** Exported for tests — platform branches can't all run on one machine. */
+export function buildProbe(command: string): Probe {
   const shellInfo = detectShell(getPreference('customShellPath') ?? '');
 
   if (shellInfo.isWSL) {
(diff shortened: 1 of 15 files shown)


src/main/providers/types.ts
```diff
@@ -75,6 +75,15 @@ export interface ProviderDefinition {
    * merged with the generic pattern set by the pty state detector.
    */
   waitingPatterns?: readonly RegExp[];
+  /** Setup guidance surfaced in Settings → Providers when the CLI is missing (#97). */
+  setup: {
```
brannon-bowden, line 79: `setup` is now a required field on ProviderDefinition. Confirm every provider returned by listProviders() (not just the four touched here) was updated, or this breaks the build.

src/main/provider-detector.ts
```diff
@@ -0,0 +1,89 @@
...
+const PROBE_TIMEOUT_MS = 10_000;
+
+interface Probe {
+  /** argv to check the command exists (exit 0 = installed). */
+  lookup: [string, string[]];
+  /** argv to fetch the version string (best effort). */
+  version: [string, string[]];
+}
+
+function buildProbe(command: string): Probe {
```
brannon-bowden: buildProbe has three platform-specific branches (WSL, native Windows, POSIX login shell) but only one is exercised by the new test — add unit coverage for the WSL and win32 paths.

src/main/providers/grok.ts
```diff
@@ -9,4 +9,9 @@ export const grokProvider = passthroughProvider({
   id: 'grok',
   name: 'Grok Build (xAI)',
   command: 'grok',
+  setup: {
+    installHint: 'curl -fsSL https://x.ai/cli/install.sh | bash',
+    docsUrl: 'https://docs.x.ai/build/overview',
+    loginHint: 'Run `grok-build login` (requires SuperGrok or X Premium+)',
```
brannon-bowden: loginHint references `grok-build login` but the detected/registered command is `grok` (line 11). Verify this is the correct command name — otherwise it will mislead users.

src/main/provider-detector.ts
```diff
@@ -0,0 +1,89 @@
...
+function run([file, args]: [string, string[]]): Promise<{ ok: boolean; stdout: string }> {
+  return new Promise((resolve) => {
+    execFile(file, args, { timeout: PROBE_TIMEOUT_MS, windowsHide: true }, (err, stdout) => {
+      resolve({ ok: !err, stdout: stdout?.toString() ?? '' });
+    });
+  });
+}
+
+/** Pick the first output line that looks like a version (contains a digit). */
+function extractVersion(stdout: string): string | null {
```
brannon-bowden: extractVersion picks the first line containing a digit; a pre-version banner line (update nag, runtime version, etc.) could be picked up instead of the actual version.

brannon-bowden (changes requested):
This PR adds a well-scoped "Providers" settings tab with CLI detection (install status, version, setup guidance) backed by a new `provider-detector` module, IPC handler, and preload/type plumbing. The core detection logic (login-shell vs. WSL vs. native Windows probing) is sound and the command names being probed come from a static, non-user-controlled registry, so there's no injection risk in the shell interpolation. However, CI is currently red and there are a couple of correctness/consistency issues worth addressing before merge.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate checks are both red on this PR. Please resolve before merge (GitGuardian is passing).
- `src/main/providers/types.ts:79` — `setup` was made a **required** field on `ProviderDefinition`. This diff only updates `claude.ts`, `codex.ts`, `gemini.ts`, and `grok.ts`. If any other registered provider definition exists outside this diff, it will fail to type-check against the new required field — this is a plausible cause of the failing quality gate. Worth double-checking that every entry returned by `listProviders()` was updated.
- `src/main/provider-detector.ts:25` — `buildProbe` has three distinct, non-trivial code paths (WSL via `wsl.exe`, native Windows via `where.exe`/`cmd.exe`, and POSIX login shell). The new test (`provider-detector.test.ts`) only exercises `detectProviders()` end-to-end on whatever platform CI runs on, so the WSL and Windows branches have no coverage at all. Given the platform-specific quoting/argv differences, at least a couple of unit tests around `buildProbe` (mocking `process.platform`/`detectShell`) would meaningfully de-risk this.
- `src/main/providers/grok.ts:15` — `loginHint` tells users to run `` `grok-build login` ``, but the registered/probed command for this provider is `grok` (`command: 'grok'`, line 11). If the login subcommand is really invoked via the `grok` binary, this hint will confuse users trying to authenticate; please confirm the actual CLI command name.
- `src/main/provider-detector.ts:54` — `extractVersion` returns the first line containing any digit. Some CLIs print an update-nag or environment banner (e.g., a Node.js version line, a "new version available" notice) before the actual version line on `--version` output, which could surface a misleading "version" string in the UI. Minor, since it's presented as best-effort.

## Suggestions (non-blocking)

- `installHint` for `grok` (`curl -fsSL https://x.ai/cli/install.sh | bash`) is a pipe-to-shell install pattern; since it's only ever displayed as copyable text (never executed by the app), this is fine as-is, but consider whether the docs page alone is a safer thing to point users to.

src/main/providers/types.ts
William-Long-II: Confirmed in 2905ae8's context — the registry contains exactly the four providers touched here (`listProviders()` maps over the static `[claudeProvider, codexProvider, geminiProvider, grokProvider]`), and `tsc --noEmit` is clean on both main and renderer configs. The failing gate was an unrelated S3776 on SettingsModal, now fixed by extracting the panel into `ProviderSettings.tsx`.

src/main/provider-detector.ts
William-Long-II: Added in 2905ae8 — `buildProbe` is exported for tests and all three branches are asserted directly (POSIX `-l -c`, native Windows `where.exe`/`cmd.exe /c`, WSL `wsl.exe <distro args> -- bash -lc`) with `process.platform` and `detectShell` stubbed. `extractVersion` got its own cases too.

src/main/providers/grok.ts
William-Long-II: Good catch — verified against current docs: the binary is `grok` and there is no separate login subcommand; auth is prompted interactively on first launch (X/xAI account). Hint corrected in 2905ae8.

src/main/provider-detector.ts
William-Long-II: Tightened in 2905ae8 — the heuristic now requires a semver-looking token (`v?\d+\.\d+(\.\d+)?`) rather than any digit, so plain-text banners are skipped. A banner that itself contains a version string can still win; it remains labeled best-effort and only affects the display string.

brannon-bowden (approved):
This is a clean, well-scoped feature addition: CLI detection for provider CLIs, surfaced in a new Settings → Providers panel, with solid platform-branch test coverage (POSIX/WSL/native Windows) and a sensible security posture (command names only ever come from the static provider registry, never from user input, so the shell-string interpolation in `buildProbe` isn't attacker-controllable). No blocking issues found. CI is green (SonarQube, quality-gate, GitGuardian).

## Issues

- `src/main/provider-detector.ts:48-49` — `run()` only reads `stdout`; if a CLI prints its version to stderr the version probe will succeed (`ok: true`) but `extractVersion` will find nothing and silently fall back to `null`. Worth capturing stderr too for the version probe, or documenting the limitation more explicitly (minor).
- `src/main/provider-detector.ts:48-49` — `execFile`'s `timeout` triggers a SIGTERM kill on expiry, which surfaces as `err` truthy and is treated identically to "command not found." A CLI that merely hangs (e.g. waiting on a prompt) would be reported as "not installed" rather than distinguished from a real lookup failure (minor).
- `src/main/provider-detector.ts:32,43` — `command -v <name>` also matches shell functions/aliases, not just real binaries on PATH. If a user's shell profile aliases the CLI name to something unrelated, detection would report "installed" even though the actual CLI isn't present (minor, edge case).

## Suggestions (non-blocking)

- `ProviderSettings.tsx` has no unmount guard around the async `refresh()` state updates; not currently harmful but could trigger a "set state on unmounted component" warning if the settings modal is closed mid-detection.
- Consider a lightweight test for `ProviderSettings.tsx`'s loading/error/refresh states, since the component owns non-trivial async state handling that isn't covered by the main-process tests.
- The 10s timeout per probe, run sequentially (lookup then version) per provider, means a single hung CLI can take ~20s before resolving — acceptable for a background settings refresh, but worth keeping in mind if more providers are added.

PR #104, opened 2026-07-13
New-session provider picker defaulting to Claude

### LINKED ISSUE #98: Feature: New-session provider picker (default Claude)

William-Long-II:
## Description
Final Phase 1 slice of the multi-provider epic (#94): choosing which agent a new session runs.

- **Picker in the New Session modal** (`NamePromptModal` gains an opt-in `providerPicker`): detection runs on modal open via the existing `providers:detect` IPC; detected CLIs are selectable, missing ones are disabled with " — not installed" and a hint pointing at Settings → Providers. Claude remains selectable even if undetected so creation never dead-ends, and it is always the preselected default — creating a session without touching the picker behaves exactly as today.
- **Race fix from the issue**: the Terminal now passes the session's provider explicitly through `pty:create`, which prefers the explicit value over the DB-row lookup. Previously a first launch could beat row persistence (terminals mount optimistically) and silently fall back to Claude. Restarts also use the explicit value from loaded state.
- `DEFAULT_SESSION_PROVIDER` shared constant keeps the renderer from importing the main-process registry (which pulls in electron).

## Related issue
Closes #98

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `tsc --noEmit` clean on main + renderer configs; full `bun test` suite passes (18 tests)
- SonarQube preflight on the modified modal: caught an unassociated `<label>` (S6853) and a `||`-vs-boolean flag (S6606), both fixed pre-push
- Acceptance walked through: default flow unchanged (picker preselects Claude, 4-arg call path identical); disabled options carry guidance; provider persists on the session row and drives the spawn via the #95 registry
- Manual QA worth doing: create a session with a non-Claude CLI installed and confirm the right agent launches and the sidebar badge appears

## Checklist
- [ ] Tests pass locally (bun test, 18 pass)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -655,16 +655,18 @@ function safeOn(channel: string, handler: (...args: any[]) => void): void {
 }
 
 // IPC Handlers
-ipcMain.handle('pty:create', async (_, id: string, cwd: string, launchClaude: boolean = false) => {
+ipcMain.handle('pty:create', async (_, id: string, cwd: string, launchClaude: boolean = false, providerId?: string) => {
   try {
     // Look up the session to get its groupId for memory injection
     const sessions = sessionsRepo.getAllSessions();
     const session = sessions.find(s => s.id === id);
     const groupId = session?.groupId || null;
 
-    // The session's stored provider drives the launch (#96); unknown ids are
-    // degraded to the default inside PtyManager.createSession (resolveProvider).
-    ptyManager.createSession(id, cwd, launchClaude, groupId, session?.provider ?? DEFAULT_PROVIDER_ID);
+    // An explicitly-passed provider (#98) wins over the stored row — the
+    // renderer mounts terminals optimistically, so the row may not be
+    // persisted yet on first launch. Unknown ids degrade to the default
+    // inside PtyManager.createSession (resolveProvider).
+    ptyManager.createSession(id, cwd, launchClaude, groupId, providerId ?? session?.provider ?? DEFAULT_PROVIDER_ID);
     // Play session start sound
     soundManager.playStartSound();
   } catch (error) {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,7 +1,7 @@
 import { app, BrowserWindow, ipcMain, dialog, shell, crashReporter } from 'electron';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
-import { DEFAULT_PROVIDER_ID } from './providers';
+import { resolveLaunchProviderId } from './providers';
 import { detectProviders } from './provider-detector';
 import { getDatabase, closeDatabase } from './database';
 import * as groupsRepo from './repositories/groups';
@@ -662,11 +662,11 @@ ipcMain.handle('pty:create', async (_, id: string, cwd: string, launchClaude: bo
     const session = sessions.find(s => s.id === id);
     const groupId = session?.groupId || null;
 
-    // An explicitly-passed provider (#98) wins over the stored row — the
-    // renderer mounts terminals optimistically, so the row may not be
-    // persisted yet on first launch. Unknown ids degrade to the default
-    // inside PtyManager.createSession (resolveProvider).
-    ptyManager.createSession(id, cwd, launchClaude, groupId, providerId ?? session?.provider ?? DEFAULT_PROVIDER_ID);
+    // The persisted row is authoritative (#96); the explicit providerId (#98)
+    // only bridges first launches where the terminal mounted before the row
+    // was persisted. Unknown ids degrade to the default inside
+    // PtyManager.createSession (resolveProvider).
+    ptyManager.createSession(id, cwd, launchClaude, groupId, resolveLaunchProviderId(session?.provider, providerId));
     // Play session start sound
     soundManager.playStartSound();
   } catch (error) {
--- a/src/renderer/components/NamePromptModal.tsx
+++ b/src/renderer/components/NamePromptModal.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useRef, useEffect, useCallback } from 'react';
 import { ClaudeAccount, ProviderStatus, DEFAULT_SESSION_PROVIDER } from '../../shared/types';
+import { buildProviderOptions } from './providerPickerOptions';
 import './NamePromptModal.css';
 
 interface NamePromptModalProps {
@@ -78,7 +79,7 @@ export const NamePromptModal: React.FC<NamePromptModalProps> = ({
     const finalName = trimmed || defaultValue;
     const finalAccountId = accountPicker ? selectedAccountId : undefined;
     const finalProvider = providerPicker ? selectedProvider : undefined;
-    onConfirm(finalName, selectedPath || undefined, finalAccountId, finalProvider);
+    onConfirm(finalName, selectedPath ?? undefined, finalAccountId, finalProvider);
   };
 
   const handleKeyDown = (e: React.KeyboardEvent) => {
@@ -171,18 +172,9 @@ export const NamePromptModal: React.FC<NamePromptModalProps> = ({
                 className="path-input"
                 style={{ width: '100%', padding: '10px 12px', cursor: 'pointer' }}
               >
-                {providers === null && (
-                  <option value={DEFAULT_SESSION_PROVIDER}>Claude Code (detecting others…)</option>
-                )}
-                {(providers ?? []).map(p => (
-                  // The default provider stays selectable even when undetected
-                  // so session creation never dead-ends.
-                  <option
-                    key={p.id}
-                    value={p.id}
-                    disabled={!p.installed && p.id !== DEFAULT_SESSION_PROVIDER}
-                  >
-                    {p.name}{p.installed ? '' : ' — not installed'}
+                {buildProviderOptions(providers).map(opt => (
+                  <option key={opt.id} value={opt.id} disabled={opt.disabled}>
+                    {opt.label}
                   </option>
                 ))}
               </select>
--- a/src/renderer/components/NamePromptModal.tsx
+++ b/src/renderer/components/NamePromptModal.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useRef, useEffect, useCallback } from 'react';
-import { ClaudeAccount } from '../../shared/types';
+import { ClaudeAccount, ProviderStatus, DEFAULT_SESSION_PROVIDER } from '../../shared/types';
 import './NamePromptModal.css';
 
 interface NamePromptModalProps {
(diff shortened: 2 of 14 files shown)


src/main/index.ts
```diff
@@ -655,16 +655,18 @@ function safeOn(channel: string, handler: (...args: any[]) => void): void {
...
     const groupId = session?.groupId || null;
 
-    // The session's stored provider drives the launch (#96); unknown ids are
-    // degraded to the default inside PtyManager.createSession (resolveProvider).
-    ptyManager.createSession(id, cwd, launchClaude, groupId, session?.provider ?? DEFAULT_PROVIDER_ID);
+    // An explicitly-passed provider (#98) wins over the stored row — the
+    // renderer mounts terminals optimistically, so the row may not be
+    // persisted yet on first launch. Unknown ids degrade to the default
+    // inside PtyManager.createSession (resolveProvider).
+    ptyManager.createSession(id, cwd, launchClaude, groupId, providerId ?? session?.provider ?? DEFAULT_PROVIDER_ID);
```
brannon-bowden: providerId now always wins over the freshly-queried DB session.provider for every Terminal mount (not just new sessions), reversing the #96 invariant that the DB row is authoritative. Consider session?.provider ?? providerId ?? DEFAULT_PROVIDER_ID instead, or only pass providerId for genuinely new/optimistic sessions.

src/renderer/components/NamePromptModal.tsx
```diff
@@ -35,33 +42,43 @@ export const NamePromptModal: React.FC<NamePromptModalProps> = ({
...
     if (isOpen) {
       setValue(defaultValue);
       setSelectedPath(null);
       setSelectedAccountId(accountPicker?.initialAccountId ?? null);
+      setSelectedProvider(DEFAULT_SESSION_PROVIDER);
+      if (providerPicker) {
+        window.electronAPI.detectProviders()
+          .then(setProviders)
+          .catch(() => setProviders([]));
+      }
```
brannon-bowden, line 67: On detectProviders() rejection, providers becomes [] and the select renders with zero options (looks broken), even though selectedProvider still defaults to 'claude' under the hood. Consider falling back to a single default-provider option on error instead of an empty list.

src/renderer/components/NamePromptModal.tsx
```diff
@@ -144,6 +161,38 @@ export const NamePromptModal: React.FC<NamePromptModalProps> = ({
...
+                {providers === null && (
+                  <option value={DEFAULT_SESSION_PROVIDER}>Claude Code (detecting others…)</option>
+                )}
+                {(providers ?? []).map(p => (
+                  // The default provider stays selectable even when undetected
+                  // so session creation never dead-ends.
+                  <option
+                    key={p.id}
+                    value={p.id}
+                    disabled={!p.installed && p.id !== DEFAULT_SESSION_PROVIDER}
```
brannon-bowden: The default provider option is always enabled even when not installed, with no inline indication beyond the general hint text below the dropdown — users can pick a provider that's certain to fail to launch without a clear per-option warning.

src/shared/types.ts
```diff
@@ -47,6 +47,13 @@ export interface ProviderStatus {
   loginHint: string;
 }
 
+/**
+ * Default provider for new sessions. Must match the main-process registry's
+ * DEFAULT_PROVIDER_ID (the registry lives main-side because provider
+ * definitions import electron; the renderer only needs the id).
+ */
+export const DEFAULT_SESSION_PROVIDER = 'claude';
+
```
brannon-bowden, line 56: DEFAULT_SESSION_PROVIDER duplicates main's DEFAULT_PROVIDER_ID with only a comment to keep them in sync; a future edit to one without the other will silently desync the UI default from the resolved backend default.

brannon-bowden (changes requested):
## Review

Request changes. The feature itself (per-session provider picker defaulting to Claude, with an explicit `providerId` bridging the optimistic-mount/not-yet-persisted-row gap) is a reasonable, well-commented fix, but CI is currently red and the non-trivial precedence/fallback logic added here has no test coverage.

### Blocking issues

- **CI is red.** `SonarQube Code Analysis` and `quality-gate` are both failing on this PR. Per policy this alone blocks merge regardless of the code review outcome — please fix/re-run before merge. (`GitGuardian Security Checks` is passing.)
- **No tests for the new provider-precedence logic.** `src/main/index.ts:669` introduces a three-way fallback (`providerId ?? session?.provider ?? DEFAULT_PROVIDER_ID`) that is the crux of this change (issue #98) and directly touches the sensitive `sessions` path. There is also new, non-trivial UI logic in `NamePromptModal.tsx` (disabled/enabled options, detection-failure fallback, default-always-selectable rule). None of this is covered by a test in this diff. Given this is exactly the kind of "unknown id / stale row / detection failure" logic that regresses silently, at least unit tests around the IPC handler's precedence and the modal's provider-selection state are warranted before merge.

### Non-blocking issues

- `src/main/index.ts:669` — the new precedence order (`providerId` always wins if present) is applied for **every** `Terminal` mount, not just first-launch/optimistic ones, because `App.tsx:1346` now passes `provider={session.provider}` unconditionally for all sessions. This effectively makes the renderer's in-memory cache authoritative over the freshly-queried DB row on every restart, which is the opposite of the invariant the original #96 fix established ("session's stored provider drives the launch"). Today this is likely harmless because provider is presumably immutable after creation, but if that ever changes (multi-window sync, a future "change provider" action, etc.) this ordering will silently prefer stale renderer state over the DB. Consider `session?.provider ?? providerId ?? DEFAULT_PROVIDER_ID` (DB wins once persisted) or only forwarding the explicit `providerId` for genuinely new sessions.
- `src/shared/types.ts` — `DEFAULT_SESSION_PROVIDER` duplicates the main-process `DEFAULT_PROVIDER_ID` with only a comment enforcing that they stay in sync. A future change to one without the other silently breaks the "default" story (UI shows one default, backend resolves another). Worth a shared constant/config or at minimum a runtime assertion.
- `NamePromptModal.tsx:66` — when `detectProviders()` rejects, `providers` is set to `[]`, which renders a `<select>` with zero `<option>`s (the "detecting…" placeholder only shows while `providers === null`). The form still silently submits the default `claude` value, but the dropdown itself looks broken/empty to the user with no retry/error affordance.
- `NamePromptModal.tsx:183` — the default provider is always selectable even when `!installed`, with no additional signal on the option itself (just the general "missing CLIs" hint below the box) that this specific selection will likely fail to launch. Consider marking the default option as "(not detected)" inline, not just via the general hint.

### Tests

No test files were changed. At minimum, add coverage for:
- The `pty:create` handler's provider-precedence fallback (`providerId ?? session?.provider ?? DEFAULT_PROVIDER_ID`), including the "unknown/undefined providerId with existing DB row" and "no DB row yet" cases.
- `NamePromptModal`'s provider-picker behavior: default selection, disabled-when-not-installed except default, and the `detectProviders()` rejection path.

src/main/index.ts
William-Long-II: Adopted your suggested ordering in 2e11d60 — `resolveLaunchProviderId(session?.provider, providerId)` prefers the persisted row (restoring the #96 invariant); the explicit id only bridges the not-yet-persisted first launch. Covered by precedence tests (row wins / explicit bridges / default fallback).

src/renderer/components/NamePromptModal.tsx
William-Long-II: Fixed in 2e11d60 — option building is extracted to `buildProviderOptions()`, which renders a single enabled default option when detection fails (or is in flight), so the select never renders empty. Unit-tested.

src/renderer/components/NamePromptModal.tsx
William-Long-II: Fixed in 2e11d60 — an undetected default now carries an inline "— not detected" marker on the option itself, in addition to the general hint. Unit-tested.

src/shared/types.ts
William-Long-II: Fixed in 2e11d60 — `DEFAULT_PROVIDER_ID` now derives directly from the shared `DEFAULT_SESSION_PROVIDER` constant (single source of truth), with a module-load assertion that the default exists in the registry, plus a test asserting main/renderer agreement.

brannon-bowden (approved):
## Verdict

Approve. This is a focused, well-reasoned fix for the new-session provider defaulting/race-condition bug (#96/#98): it adds an explicit `providerId` bridge for the case where a `Terminal` mounts before its DB row is persisted, unifies the default-provider constant between main and renderer (`DEFAULT_SESSION_PROVIDER`), and wires a provider picker into the new-session modal. The core logic (`resolveLaunchProviderId`, `buildProviderOptions`) is pure, small, and has solid unit test coverage, including a precedence test and a main/renderer-agreement test. CI is green across SonarQube, quality-gate, and GitGuardian.

## Blocking issues

None found.

## Non-blocking issues

- `src/renderer/components/providerPickerOptions.ts:23,26` — The fallback labels ("Claude Code", "Claude Code (detecting others…)") are hardcoded strings rather than derived from `PROVIDER_LABELS` or the provider registry's `name` field. If the Claude provider is ever renamed/rebranded, these fallback paths (detection-in-flight and detection-failed) will silently drift out of sync with the real label used once detection succeeds.
- `src/renderer/components/NamePromptModal.tsx:66` — `detectProviders().catch(() => setProviders([]))` swallows the error with no logging. Worth at least a `console.warn` so provider-detection failures are visible during debugging/support, rather than silently degrading to "only Claude available."
- `src/renderer/components/NamePromptModal.tsx` (`handleSubmit`) — `selectedPath || undefined` was changed to `selectedPath ?? undefined`. Functionally equivalent given `selectedPath`'s `string | null` typing, but it's an unrelated micro-refactor riding along with the provider-picker change; worth folding into a separate cleanup commit next time to keep diffs focused.
- `src/main/providers/index.ts` — the new fail-fast `if (!REGISTRY.has(DEFAULT_PROVIDER_ID)) throw ...` at module load is a good defensive addition; no concerns, just flagging it as a behavior change (an app that previously would have silently misbehaved on a registry/default mismatch now hard-crashes at startup), which seems like the right tradeoff here.

## Tests

- Good coverage of the pure helpers (`resolveLaunchProviderId`, `buildProviderOptions`) including edge cases (null stored provider, missing explicit id, detection-in-flight/failed).
- Missing: a test exercising the actual `ipcMain.handle('pty:create', ...)` wiring in `src/main/index.ts` (e.g., session lookup returns `undefined` + `providerId` passed → `ptyManager.createSession` called with the bridged id). Right now only the underlying `resolveLaunchProviderId` function is verified in isolation, not its integration into the handler that was the actual site of the original race condition. Not blocking given the handler is a thin pass-through, but it's the piece most directly tied to the bug being fixed.

PR #105, opened 2026-07-13
Arena mode — multi-agent prompt fan-out with streaming metrics

### LINKED ISSUE #100: Feature: Competitive arena mode — multi-model prompt fan-out with metrics
### User story / problem

As a user I want to send one prompt to several models at once and compare their answers, speed, and cost, so I can pick the best model for a kind of task on evidence.

### Proposed solution

New arena session kind (not a PTY): main process fans a prompt out concurrently to selected models — xAI, OpenAI, Gemini, Anthropic, Ollama (local, keyless) — streaming responses into a side-by-side renderer view. Capture per model: time-to-first-token, total latency, input/output tokens, computed cost from a pricing table. Persist runs to new `arena_runs` / `arena_responses` tables with a history view and re-run. Note: Claude in the arena requires an Anthropic API key (separate from Claude Code subscription auth).

### Acceptance criteria

- [ ] A prompt runs across ≥2 configured models concurrently with streaming side-by-side output
- [ ] Each response shows TTFT, total latency, token counts, and computed cost
- [ ] Runs persist and can be reviewed and re-run from history
- [ ] Models without a stored key (except Ollama) are unselectable with guidance

### Scope

Major feature

### Priority

Medium

William-Long-II:
## Description
Phase 2 of the multi-provider epic (#94): the competitive arena. One prompt fans out to selected contestants, streamed side-by-side with time-to-first-token, total latency, token counts, and cost.

**Subscription-preservation by design (per the decision on #100):** every CLI contestant runs headless under the user's existing login — `claude -p --output-format stream-json --verbose --max-turns 1`, `codex exec --json`, `gemini -p --output-format json`, `grok -p` — never an API key. Ollama joins as a keyless local-HTTP contestant. Cost shows as "included (API-equiv $X)" when the CLI reports a figure.

- **Prompt safety**: the prompt travels via the `ARENA_PROMPT` env var and is referenced with the shell's own env syntax (same pattern as memory injection), so user text never touches a command string.
- **Engine** (`src/main/arena/engine.ts`): concurrent spawn through the user's shell (reusing `getShellLaunch`), per-line parsing, 5-minute per-contestant timeout, cancel support, single DB write per response at completion.
- **Persistence**: new `arena_runs` / `arena_responses` tables; history is reloadable from the panel's dropdown.
- **UI**: `ArenaPanel` toggled from the toolbar (⚔️), modeled on the analytics panel; contestants come from provider detection (#97), missing CLIs disabled.
- **Parsers** are pure and per-provider (`src/main/arena/parsers.ts`), attached to `ProviderDefinition.arena`.

## Related issue
Closes #100

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- **Live end-to-end** against the real claude CLI: engine spawn → readline → parser → metrics → persistence; response streamed correctly with TTFT 5.2s (CLI-reported), total 10.6s, 27 output tokens, $0.24 API-equivalent
- The claude headless output contract was captured from a real run and is baked into the parser tests as a fixture
- `bun test`: 39 pass (5 parsers, update-merge helpers, plus all prior suites)
- `tsc --noEmit` clean on main + renderer configs; SonarQube preflight on the engine (one `readonly` minor found and fixed)
- Not live-tested: codex/gemini/grok runners (CLIs not installed on this machine — parsers follow their documented contracts and errors surface in-column) and Ollama; worth a manual QA pass

## Checklist
- [ ] Tests pass locally (bun test, 39 pass)

### DIFF
--- a/src/main/arena/engine.ts
+++ b/src/main/arena/engine.ts
@@ -36,28 +36,44 @@ const OLLAMA_BASE_URL = 'http://127.0.0.1:11434';
 /** Generous default cap so a hung CLI can't leave a column spinning forever. */
 const DEFAULT_CONTESTANT_TIMEOUT_MS = 5 * 60 * 1000;
 
+/** Grace period between SIGTERM and the SIGKILL escalation. */
+const KILL_GRACE_MS = 3000;
+
 /**
  * Kill the contestant's whole process tree. The CLI runs as a child of the
  * wrapper shell, so signalling just the shell would leave the actual agent
  * process running detached. POSIX contestants are spawned detached (own
- * process group) and killed by group; Windows uses taskkill /T.
+ * process group) and killed by group, escalating to SIGKILL if the tree
+ * ignores SIGTERM — otherwise 'close' never fires and the column would sit
+ * on "running" forever. Windows taskkill /F is already forceful.
  */
 function killTree(child: ChildProcess): void {
-  if (!child.pid) {
+  const pid = child.pid;
+  if (!pid) {
     child.kill();
     return;
   }
   if (process.platform === 'win32') {
     // Absolute path so a poisoned PATH can't substitute the binary (S4036).
     const taskkill = path.join(process.env.SystemRoot ?? 'C:\\Windows', 'System32', 'taskkill.exe');
-    execFile(taskkill, ['/F', '/T', '/PID', String(child.pid)], () => undefined);
+    execFile(taskkill, ['/F', '/T', '/PID', String(pid)], () => undefined);
     return;
   }
   try {
-    process.kill(-child.pid, 'SIGTERM');
+    process.kill(-pid, 'SIGTERM');
   } catch {
     child.kill();
+    return;
   }
+  const escalation = setTimeout(() => {
+    try {
+      process.kill(-pid, 'SIGKILL');
+    } catch {
+      // Process group already gone — nothing to escalate.
+    }
+  }, KILL_GRACE_MS);
+  escalation.unref?.();
+  child.once('close', () => clearTimeout(escalation));
 }
 
 interface LiveResponse {
@@ -199,6 +215,9 @@ export class ArenaEngine extends EventEmitter {
       clearTimeout(timeout);
       controller.abort();
     };
+    // Clock starts at the launch attempt so error paths (no models, daemon
+    // down) don't inherit the prepare()→launch() gap in totalMs.
+    entry.startedAt = Date.now();
 
     try {
       // Use the first locally available model.
@@ -208,7 +227,6 @@ export class ArenaEngine extends EventEmitter {
         this.finish(responseId, 'error', null, 'Ollama is running but has no models pulled');
         return;
       }
-      entry.startedAt = Date.now();
 
       const res = await fetch(`${OLLAMA_BASE_URL}/api/chat`, {
         method: 'POST',
--- a/src/main/arena/engine.ts
+++ b/src/main/arena/engine.ts
@@ -18,6 +18,7 @@
  * a column stuck on "running".
  */
 import { spawn, execFile, ChildProcess } from 'child_process';
+import * as path from 'path';
 import * as readline from 'readline';
 import { randomUUID } from 'crypto';
 import { EventEmitter } from 'events';
@@ -47,7 +48,9 @@ function killTree(child: ChildProcess): void {
     return;
   }
   if (process.platform === 'win32') {
-    execFile('taskkill', ['/F', '/T', '/PID', String(child.pid)], () => undefined);
+    // Absolute path so a poisoned PATH can't substitute the binary (S4036).
+    const taskkill = path.join(process.env.SystemRoot ?? 'C:\\Windows', 'System32', 'taskkill.exe');
+    execFile(taskkill, ['/F', '/T', '/PID', String(child.pid)], () => undefined);
     return;
   }
   try {
--- a/src/main/arena/engine.ts
+++ b/src/main/arena/engine.ts
@@ -1,0 +1,268 @@
+/**
+ * Arena engine (#100): fans one prompt out to N contestants concurrently,
+ * streams their output as 'update' events, and persists final text + metrics.
+ *
+ * CLI contestants run headless under the user's existing login/subscription
+ * (never an API key — see issue #100's subscription-preservation decision),
+ * spawned through the user's shell exactly like sessions are. The prompt is
+ * passed via the ARENA_PROMPT environment variable and referenced with the
+ * shell's own env syntax, so prompt text never touches the command string.
+ *
+ * Ollama is the one non-CLI contestant: a keyless HTTP call to the local
+ * daemon, streaming /api/chat.
+ */
+import { spawn } from 'child_process';
+import * as readline from 'readline';
+import { randomUUID } from 'crypto';
+import { EventEmitter } from 'events';
+import log from 'electron-log';
+import { getProvider } from '../providers';
+import { getShellLaunch } from '../pty-manager';
+import { detectShell } from '../shell-detector';
+import { getPreference } from '../repositories/preferences';
+import * as arenaRepo from '../repositories/arena';
+import { ArenaStreamParser, ollamaParser } from './parsers';
+import { ArenaRun, ArenaUpdate, ArenaResponseStatus } from '../../shared/types';
+
+export const OLLAMA_CONTESTANT_ID = 'ollama';
+const OLLAMA_BASE_URL = 'http://127.0.0.1:11434';
+/** Generous cap so a hung CLI can't leave a column spinning forever. */
+const CONTESTANT_TIMEOUT_MS = 5 * 60 * 1000;
+
+interface LiveResponse {
+  runId: string;
+  provider: string;
+  text: string;
+  startedAt: number;
+  ttftMs: number | null;
+  kill: (() => void) | null;
+}
+
+export class ArenaEngine extends EventEmitter {
+  private readonly live: Map<string, LiveResponse> = new Map();
+
+  /** Create the run + response rows and launch every contestant. */
+  start(prompt: string, contestants: string[]): ArenaRun {
+    const runId = randomUUID();
+    arenaRepo.createRun(runId, prompt);
+
+    for (const provider of contestants) {
+      const responseId = randomUUID();
+      arenaRepo.createResponse(responseId, runId, provider);
+      this.live.set(responseId, {
+        runId,
+        provider,
+        text: '',
+        startedAt: Date.now(),
+        ttftMs: null,
+        kill: null,
+      });
+      if (provider === OLLAMA_CONTESTANT_ID) {
+        void this.runOllama(responseId, prompt);
+      } else {
+        this.runCli(responseId, provider, prompt);
+      }
+    }
+
+    const run = arenaRepo.getRun(runId);
+    if (!run) throw new Error('Arena run vanished immediately after creation');
+    return run;
+  }
+
+  cancelRun(runId: string): void {
+    for (const [responseId, entry] of this.live) {
+      if (entry.runId === runId) {
+        entry.kill?.();
+        this.finish(responseId, 'error', null, 'Cancelled');
+      }
+    }
+  }
+
+  private runCli(responseId: string, providerId: string, prompt: string): void {
+    const entry = this.live.get(responseId);
+    if (!entry) return;
+    let provider: ReturnType<typeof getProvider>;
+    try {
+      provider = getProvider(providerId);
+    } catch {
+      this.finish(responseId, 'error', null, `Unknown contestant: ${providerId}`);
+      return;
+    }
+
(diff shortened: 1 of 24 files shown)


src/main/arena/engine.ts
```diff
@@ -0,0 +1,268 @@
...
+        kill: null,
+      });
+      if (provider === OLLAMA_CONTESTANT_ID) {
+        void this.runOllama(responseId, prompt);
+      } else {
+        this.runCli(responseId, provider, prompt);
+      }
+    }
+
+    const run = arenaRepo.getRun(runId);
```
brannon-bowden, line 117: Run is returned to the IPC caller here, but async spawn 'error'/'close' events can fire and be sent to the renderer before this response reaches ArenaPanel and sets runIdRef.current — those updates get silently dropped by the runId filter in onArenaUpdate.

src/main/arena/engine.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    const child = spawn(shellLaunch.shell, shellLaunch.wrap(cmd), {
+      env: { ...process.env, ARENA_PROMPT: prompt },
+      windowsHide: true,
+    });
+    entry.startedAt = Date.now();
+
+    const timeout = setTimeout(() => {
+      log.warn(`[Arena] ${providerId} timed out after ${CONTESTANT_TIMEOUT_MS}ms`);
+      child.kill();
```
brannon-bowden: child.kill() only signals the spawned shell process; if wrap() doesn't exec-replace into the actual CLI, the real hung/long-running process may survive the timeout.

src/main/repositories/arena.ts
```diff
@@ -0,0 +1,117 @@
...
+  const responsesStmt = db.prepare('SELECT * FROM arena_responses WHERE run_id = ? ORDER BY provider');
+  return runs.map((run) => ({
+    id: run.id,
+    prompt: run.prompt,
+    createdAt: new Date(run.created_at),
+    responses: (responsesStmt.all(run.id) as ResponseRow[]).map(rowToResponse),
+  }));
+}
+
+export function deleteRun(id: string): void {
```
brannon-bowden: deleteRun is exported but not wired to any IPC handler — dead code unless intentionally left for a future history-delete feature.

brannon-bowden (changes requested):
This PR adds a solid, well-scoped Arena feature: prompt fan-out to CLI contestants plus Ollama, with a clean parser abstraction that's actually unit-tested, and a sensible env-var approach to avoid shell injection when passing the prompt. A few concurrency and interface-design issues should be addressed before merge, and CI is currently red.

## Issues

- **CI is failing** — SonarQube Code Analysis and quality-gate are both red on this PR; GitGuardian is the only passing check. This needs to be resolved (or the failures explained) before merge.

- **Likely dropped early stream events (race condition).** `ArenaEngine.start()` (`src/main/arena/engine.ts:45-70`) spawns all contestants synchronously and returns the initial `ArenaRun` to the IPC caller. But if a child process errors immediately (e.g. an uninstalled/misconfigured CLI hits `child.on('error')`) or finishes very fast, `finish()`/`emitUpdate()` can fire — and the `arena:update` IPC message can reach the renderer — before the `arenaStart()` invoke() promise resolves back and `ArenaPanel` sets `runIdRef.current` (`src/renderer/components/ArenaPanel.tsx`, `startRun`). Since `onArenaUpdate` drops any update where `update.runId !== runIdRef.current`, that update is silently lost, leaving the column stuck showing "running" forever even though the DB row is correctly finalized. Consider setting `runIdRef.current` synchronously from the `run` object as soon as it's available, or buffering/replaying updates keyed by run id rather than gating on a ref set after the round trip.

- **Timeout/cancel may not kill the actual CLI process.** `entry.kill()` (`src/main/arena/engine.ts:103-110`) calls `child.kill()` on the process returned by `spawn(shellLaunch.shell, shellLaunch.wrap(cmd), ...)`. If `wrap()` runs the command as a child of an interactive/login shell rather than exec-replacing it, killing the shell won't necessarily terminate the actual CLI subprocess it launched. That would defeat the stated purpose of `CONTESTANT_TIMEOUT_MS` ("so a hung CLI can't leave a column spinning forever") and of explicit cancellation — the process could keep running detached after the column reports "error/Cancelled". Worth confirming `wrap()`'s exec semantics, and if it doesn't guarantee in-process-tree termination, kill by process group instead.

- **`ProviderDefinition.arena` is a required field** (`src/main/providers/types.ts`), and `passthroughProvider`'s options type was updated to require it too. Any other `ProviderDefinition`/`passthroughProvider` caller outside this diff (not shown here) would now fail to compile. This is a plausible contributor to the current SonarQube/quality-gate failures — worth double-checking that every provider in the registry was updated, or making `arena` optional with a fallback (e.g. "not supported in Arena") for providers that don't implement it yet.

- **No tests for `ArenaEngine`.** The parser and reducer logic is well covered (`parsers.test.ts`, `arenaUpdates.test.ts`), but the actually non-trivial concurrency logic — timeout handling, `cancelRun` racing with `close`/`error` handlers, TTFT measurement, the `close` handler's `!this.live.has(responseId)` guard — has no test coverage at all. This is exactly the kind of logic that's easy to get subtly wrong and hard to catch by inspection alone.

## Suggestions (non-blocking)

- `deleteRun` (`src/main/repositories/arena.ts:115`) is exported but never wired to an IPC handler — either wire it up for the history UI or drop it until it's needed.
- Ollama contestant always picks `models[0]` from `/api/tags` with no user control over which local model is used; a small model picker (or at least surfacing the chosen model name in the UI) would help usability.

src/main/arena/engine.ts
William-Long-II: Fixed in a9ef7d6 with a two-phase start: `arena:start` now only prepares the run rows and returns the id; the renderer sets its subscription ref and then calls `arena:launch`. Even an instantly-failing CLI can't emit before the renderer is listening. Covered by the 'prepare creates the run without launching' engine test.

src/main/arena/engine.ts
William-Long-II: Confirmed and fixed in a9ef7d6 — `wrap()` does not exec-replace, so the CLI survived a shell kill. Contestants now spawn detached (own process group) on POSIX and are killed with `process.kill(-pid)`; Windows uses `taskkill /F /T`. The timeout engine test verifies a `sleep 30` contestant settles as error in well under a second with a 500ms cap.

src/main/repositories/arena.ts
William-Long-II: Dropped in a9ef7d6 — `deleteRun` is removed until a history-delete feature needs it.

William-Long-II:
All review items addressed in a9ef7d6:

- **Launch race** → two-phase start (prepare/launch) so the renderer subscribes before anything spawns.
- **Kill semantics** → process-group kill on POSIX (detached spawn) / `taskkill /T` on Windows; verified by a timeout test against a real `sleep 30` child.
- **Engine tests** → 5 new tests covering streaming + TTFT bounds, stderr on non-zero exit, timeout kill, cancel + no-double-finalize, and unknown contestants, using fake shell contestants over the real spawn/readline path. Suite is 44 tests.
- **`ProviderDefinition.arena` required**: kept required deliberately — the registry is a closed set (all four updated in this diff, `tsc` clean across configs), and requiring it forces a conscious arena decision for any future provider. The gate failures were four Sonar issues in the new files, all fixed.
- **Ollama model choice**: noted — the engine currently reports which model it picked only in logs; surfacing/selecting the model in the UI is queued as a follow-up rather than grown into this PR.
- Live e2e re-verified against the real claude CLI after the engine rework.

Re-requesting review once CI is green.

brannon-bowden (changes requested):
This is a well-structured feature with genuinely careful design in the tricky spots (two-phase prepare/launch to avoid a race with the renderer subscribing, prompt-via-env-var to avoid POSIX shell injection, guarded no-double-finalize logic, process-group kill on POSIX). CI is green across SonarQube, GitGuardian, and quality-gate. However, one coverage gap and a robustness gap should be addressed before merge.

## Issues

- **Missing tests for non-trivial logic — `runOllama()` is entirely untested** (`src/main/arena/engine.ts:191`). `engine.test.ts` only exercises the CLI path (`runCli`) via fake shell providers; the whole Ollama contestant path — model discovery via `/api/tags`, the NDJSON streaming loop, abort-on-timeout, and the three distinct error branches (`no models pulled`, `HTTP not ok`, `not reachable`) — has zero coverage. Given this is called out in the file's own doc comment as "the one non-CLI contestant," it deserves at least a mocked-`fetch` test suite analogous to what `runCli` already has.
- **Timeout only sends SIGTERM, with no SIGKILL escalation** (`src/main/arena/engine.ts:151-158`, `killTree` at `:44-60`). If a contestant CLI ignores SIGTERM, `child.on('close')` never fires, `finish()` is never called, and the response is stuck "running" forever in both the UI and the DB, while the process leaks. Worth adding a fallback SIGKILL after a short grace period.
- Arena's schema migration is tucked inside `initializeCodeSearchTables()` (`src/main/database.ts:440`), a function whose name has nothing to do with arena mode — likely to be missed by a future maintainer looking for either subsystem's schema setup.
- On the "Ollama has no models pulled" error path, `totalMs` is computed from the `prepare()`-time timestamp rather than the actual attempt start (`src/main/arena/engine.ts:233`, `entry.startedAt` reassignment happens only after a model is found), so the reported duration can be inflated by an unrelated prepare→launch gap.
- The cmd.exe branch of `getShellLaunch` (`src/main/shell-launch.ts:40`) expands `%ARENA_PROMPT%` inside a plain double-quoted token; cmd.exe's percent-expansion is not safely escaped against embedded `"`/`&`, so a crafted prompt could break out of quoting and inject additional commands when the detected shell is cmd.exe. This logic is inherited unchanged from the pre-existing `pty-manager.ts` code, but the new Arena "prompt" field is explicitly meant to accept arbitrary free text the user may paste from external sources, which meaningfully widens who/what can reach this code path — worth a follow-up hardening pass.

## Suggestions (non-blocking)

- Consider validating/deduping `contestants` in the `arena:start` handler (e.g. rejecting duplicate provider ids) for clearer UX, though the current per-contestant failure handling degrades gracefully either way.
- `formatCost`'s "included (API-equiv $...)" phrasing is a little cryptic at first glance — a short tooltip might help.

William-Long-II:
Second-round items addressed in a7486a8:

- **runOllama coverage** — 5 new tests with stubbed `fetch`: NDJSON streaming with token counts, no-models-pulled, HTTP-error, daemon-unreachable, and abort-on-timeout. Suite is now 49 tests.
- **SIGKILL escalation** — `killTree` sends SIGTERM to the process group, then SIGKILL after a 3s grace if `close` hasn't fired (timer cleared on close, unref'd so it can't hold the app open). Windows `taskkill /F` was already forceful.
- **Migration location** — arena schema now lives in its own `initializeArenaTables()`, called from the init sequence alongside the other initializers.
- **Ollama totalMs** — the clock now starts at the launch attempt, so error paths don't inherit the prepare→launch gap.
- **Non-blocking** — contestants are deduped at the `arena:start` boundary, and the cost cell got an explanatory tooltip.
- **cmd.exe quoting** — agreed it warrants hardening beyond this PR since arena widens exposure of the inherited path; filed as #106 with the candidate approaches from your note.

Re-requesting review once CI is green.

brannon-bowden (approved):
This is a well-structured feature with strong attention to concurrency correctness (two-phase prepare/launch, double-finalize guard, process-tree kill with SIGTERM→SIGKILL escalation, prompt delivered via env var rather than command string) and solid test coverage across the engine, parsers, and the renderer's update-merge logic. CI is green (SonarQube, GitGuardian, quality-gate all passing). No blocking issues found.

- `src/renderer/components/ArenaPanel.tsx:98-105` — `loadHistoryRun` sets `runIdRef.current = null` while a run may still be actively streaming in the backend. The in-flight run keeps executing (correct), but since `isRunSettled`'s effect only fires when `run.id === runIdRef.current`, the completed run won't auto-refresh into the history list once it finishes — the user has to reopen the dropdown manually to see it appear. Consider decoupling "viewing history" from "tracking the live run" (e.g., a separate `liveRunId` vs `displayedRun`) so a background run still triggers `refreshHistory()` on completion.
- `src/main/index.ts:1189-1192` (`arena:start` handler) — validates `prompt` and that `contestants` is a non-empty array, but doesn't validate that array entries are strings/known ids. Unknown ids are already handled gracefully downstream (per-contestant "Unknown contestant" error in `engine.ts`), so this isn't harmful, just a bit permissive at the boundary; worth a quick type/emptiness check on entries for clarity.

Nice touches worth calling out: the `killTree`/`taskkill` absolute-path handling to avoid PATH poisoning, the prompt-via-env-var design that keeps user input out of the shell command string entirely, and the `entry.startedAt` re-stamping at launch time (not prepare time) so `totalMs` doesn't inherit the prepare→launch gap.

PR #107, opened 2026-07-14
Encrypted provider API-key vault (Electron safeStorage)

### LINKED ISSUE #99: Feature: Encrypted provider API-key vault (Electron safeStorage)
### User story / problem

As a user I want to store provider API keys for arena mode securely, so keys never sit in plaintext on disk.

### Proposed solution

Key vault in the main process using Electron `safeStorage` (OS-keychain-backed); encrypted blobs persisted in the existing preferences store. Settings → Providers gains per-provider key entry with masked display, validate-key test call, and delete. IPC surface never returns decrypted keys to the renderer.

### Acceptance criteria

- [ ] Keys are encrypted at rest via safeStorage; plaintext never written to DB or logs
- [ ] Renderer can set/replace/delete/test a key but never read it back
- [ ] Graceful fallback message when safeStorage is unavailable on the platform

### Scope

Minor feature

### Priority

High

William-Long-II:
## Description
Final slice of the multi-provider epic (#94): optional API-key support for users who run on API keys instead of subscriptions.

**The subscription guarantee holds by construction**: a stored key does nothing until the user flips a per-provider "use API key" toggle (default off, cannot be enabled without a stored key). CLI login/subscription remains the default auth everywhere — sessions and arena alike. This is asserted by a dedicated test ("a stored key injects NOTHING until the toggle is enabled").

- **Vault** (`src/main/key-vault.ts`): keys encrypted via Electron `safeStorage` (OS keychain), persisted as base64 blobs in the preferences table. Plaintext never reaches the DB, logs, or the renderer — the IPC surface is write/test/delete/toggle only; there is no read-back.
- **Injection**: `vaultEnvFor(provider)` merges the provider's key env var (`ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GEMINI_API_KEY` / `XAI_API_KEY`) into session spawns and arena contestants — empty object unless opted in, so the default path is byte-identical to before.
- **Validation**: "Test" performs a cheap authenticated call to each provider's models endpoint (Anthropic endpoint + headers confirmed against current API docs) with a 10s timeout; 401/403 reports "key rejected", network failure "unreachable".
- **UI**: Settings → Providers gains an "API Keys (optional)" section — masked input, Save/Test/Delete, the opt-in toggle with an explicit billing warning, and a graceful message when OS-keychain encryption is unavailable (acceptance criterion).

## Related issue
Closes #99

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `bun test`: 61 pass across the repo, including 12 new vault tests (encrypted-at-rest, empty/unknown-provider rejection, unavailable-keychain error + status reporting, delete clears the opt-in flag, the subscription-preservation invariant, opt-in/opt-out round-trip, testKey against stubbed endpoints incl. 401 and network failure)
- `tsc --noEmit` clean on main + renderer configs; SonarQube preflight on the vault module clean
- Manual QA worth doing: store a real key on a machine with a keychain, Test it, enable the toggle, and confirm the CLI picks up the env var (and that disabling reverts to login auth)

## Checklist
- [ ] Tests pass locally (bun test, 61 pass)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -3,6 +3,7 @@ import * as path from 'path';
 import { ptyManager } from './pty-manager';
 import { resolveLaunchProviderId } from './providers';
 import { detectProviders } from './provider-detector';
+import * as keyVault from './key-vault';
 import { getDatabase, closeDatabase } from './database';
 import * as groupsRepo from './repositories/groups';
 import * as sessionsRepo from './repositories/sessions';
@@ -1184,6 +1185,24 @@ safeHandle('editor:detectAvailable', async () => {
 
 safeHandle('providers:detect', async () => {
   return detectProviders();
+});
+
+// Provider API-key vault (#99). Keys go in and are tested; they are never
+// returned to the renderer.
+safeHandle('vault:list', async () => {
+  return keyVault.listVaultStatuses();
+});
+safeHandle('vault:setKey', async (providerId: string, key: string) => {
+  keyVault.setKey(providerId, key);
+});
+safeHandle('vault:deleteKey', async (providerId: string) => {
+  keyVault.deleteKey(providerId);
+});
+safeHandle('vault:setUseKey', async (providerId: string, use: boolean) => {
+  keyVault.setUseKey(providerId, use);
+});
+safeHandle('vault:testKey', async (providerId: string) => {
+  return keyVault.testKey(providerId);
 });
 
 // Arena mode (#100). Two-phase: 'start' only creates the run so the renderer
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1188,21 +1188,31 @@ safeHandle('providers:detect', async () => {
 });
 
 // Provider API-key vault (#99). Keys go in and are tested; they are never
-// returned to the renderer.
+// returned to the renderer. IPC is a trust boundary — validate argument
+// types at runtime rather than relying on the preload's TS types.
+function requireIpcString(value: unknown, name: string): string {
+  if (typeof value !== 'string' || value.length === 0) {
+    throw new Error(`Invalid ${name}`);
+  }
+  return value;
+}
 safeHandle('vault:list', async () => {
   return keyVault.listVaultStatuses();
 });
-safeHandle('vault:setKey', async (providerId: string, key: string) => {
-  keyVault.setKey(providerId, key);
+safeHandle('vault:setKey', async (providerId: unknown, key: unknown) => {
+  keyVault.setKey(requireIpcString(providerId, 'providerId'), requireIpcString(key, 'key'));
 });
-safeHandle('vault:deleteKey', async (providerId: string) => {
-  keyVault.deleteKey(providerId);
+safeHandle('vault:deleteKey', async (providerId: unknown) => {
+  keyVault.deleteKey(requireIpcString(providerId, 'providerId'));
 });
-safeHandle('vault:setUseKey', async (providerId: string, use: boolean) => {
-  keyVault.setUseKey(providerId, use);
+safeHandle('vault:setUseKey', async (providerId: unknown, use: unknown) => {
+  if (typeof use !== 'boolean') {
+    throw new Error('Invalid use flag');
+  }
+  keyVault.setUseKey(requireIpcString(providerId, 'providerId'), use);
 });
-safeHandle('vault:testKey', async (providerId: string) => {
-  return keyVault.testKey(providerId);
+safeHandle('vault:testKey', async (providerId: unknown) => {
+  return keyVault.testKey(requireIpcString(providerId, 'providerId'));
 });
 
 // Arena mode (#100). Two-phase: 'start' only creates the run so the renderer
--- a/src/main/key-vault.ts
+++ b/src/main/key-vault.ts
@@ -13,12 +13,23 @@
 import { safeStorage } from 'electron';
 import log from 'electron-log';
 import { getPreference, setPreference, deletePreference } from './repositories/preferences';
-import { getProvider, listProviders } from './providers';
+import { getProvider, listProviders, ProviderDefinition } from './providers';
 import { KeyVaultStatus } from '../shared/types';
 
 const KEY_PREF_PREFIX = 'providerApiKey.';
 const USE_PREF_PREFIX = 'providerApiKeyUse.';
 const TEST_TIMEOUT_MS = 10_000;
+
+type ApiKeyConfig = NonNullable<ProviderDefinition['apiKey']>;
+
+/** Resolve a provider that supports API keys; throws for unknown ids or key-less providers. */
+function requireApiKeyConfig(providerId: string): ApiKeyConfig {
+  const config = getProvider(providerId).apiKey;
+  if (!config) {
+    throw new Error(`Provider '${providerId}' does not support API keys`);
+  }
+  return config;
+}
 
 export function isVaultAvailable(): boolean {
   return safeStorage.isEncryptionAvailable();
@@ -33,7 +44,7 @@ export function useKey(providerId: string): boolean {
 }
 
 export function setKey(providerId: string, key: string): void {
-  getProvider(providerId); // throws on unknown ids
+  requireApiKeyConfig(providerId);
   if (!isVaultAvailable()) {
     throw new Error('Secure key storage is not available on this system');
   }
@@ -46,11 +57,13 @@ export function setKey(providerId: string, key: string): void {
 }
 
 export function deleteKey(providerId: string): void {
+  getProvider(providerId); // throws on unknown ids
   deletePreference(KEY_PREF_PREFIX + providerId);
   deletePreference(USE_PREF_PREFIX + providerId);
 }
 
 export function setUseKey(providerId: string, use: boolean): void {
+  requireApiKeyConfig(providerId);
   if (use && !hasKey(providerId)) {
     throw new Error('No API key stored for this provider');
   }
@@ -81,8 +94,9 @@ export function vaultEnvFor(providerId: string): Record<string, string> {
   const key = decryptKey(providerId);
   if (key === null) return {};
   try {
-    return { [getProvider(providerId).apiKey.envVar]: key };
+    return { [requireApiKeyConfig(providerId).envVar]: key };
   } catch {
+    // Unknown provider or one without API-key support — inject nothing.
     return {};
   }
 }
@@ -96,7 +110,7 @@ export interface KeyTestResult {
 
 /** Validate the stored key against the provider's cheap models endpoint. */
 export async function testKey(providerId: string): Promise<KeyTestResult> {
-  const provider = getProvider(providerId);
+  const config = requireApiKeyConfig(providerId);
   const key = decryptKey(providerId);
   if (key === null) {
     return { ok: false, status: null, error: 'No API key stored' };
@@ -104,8 +118,8 @@ export async function testKey(providerId: string): Promise<KeyTestResult> {
   const controller = new AbortController();
   const timeout = setTimeout(() => controller.abort(), TEST_TIMEOUT_MS);
   try {
-    const res = await fetch(provider.apiKey.test.url, {
-      headers: provider.apiKey.test.headers(key),
+    const res = await fetch(config.test.url, {
+      headers: config.test.headers(key),
       signal: controller.signal,
     });
     if (res.ok) return { ok: true, status: res.status, error: null };
@@ -121,10 +135,13 @@ export async function testKey(providerId: string): Promise<KeyTestResult> {
 
 export function listVaultStatuses(): KeyVaultStatus[] {
   const available = isVaultAvailable();
-  return listProviders().map((p) => ({
-    providerId: p.id,
-    available,
-    hasKey: hasKey(p.id),
-    useKey: useKey(p.id),
-  }));
+  // Providers without apiKey support are not applicable and don't appear.
+  return listProviders()
+    .filter((p) => p.apiKey !== undefined)
+    .map((p) => ({
+      providerId: p.id,
+      available,
+      hasKey: hasKey(p.id),
+      useKey: useKey(p.id),
+    }));
 }
--- a/src/main/key-vault.ts
+++ b/src/main/key-vault.ts
@@ -1,0 +1,130 @@
+/**
+ * Provider API-key vault (#99).
+ *
+ * Keys are encrypted with Electron's safeStorage (OS keychain: Keychain on
+ * macOS, DPAPI on Windows, libsecret on Linux) and persisted as base64 blobs
+ * in the existing preferences table. Plaintext keys never touch the DB, logs,
+ * or the renderer — the IPC surface only reports whether a key is stored.
+ *
+ * A stored key does nothing by itself: it is injected into a provider's
+ * launch environment only when the user flips the per-provider "use API key"
+ * toggle. CLI login/subscription remains the default (issue #100 decision).
+ */
+import { safeStorage } from 'electron';
+import log from 'electron-log';
+import { getPreference, setPreference, deletePreference } from './repositories/preferences';
+import { getProvider, listProviders } from './providers';
+import { KeyVaultStatus } from '../shared/types';
+
+const KEY_PREF_PREFIX = 'providerApiKey.';
+const USE_PREF_PREFIX = 'providerApiKeyUse.';
+const TEST_TIMEOUT_MS = 10_000;
+
+export function isVaultAvailable(): boolean {
+  return safeStorage.isEncryptionAvailable();
+}
+
+export function hasKey(providerId: string): boolean {
+  return getPreference(KEY_PREF_PREFIX + providerId) !== null;
+}
+
+export function useKey(providerId: string): boolean {
+  return getPreference(USE_PREF_PREFIX + providerId) === 'true';
+}
+
+export function setKey(providerId: string, key: string): void {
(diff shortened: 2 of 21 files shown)


src/main/providers/types.ts
```diff
@@ -77,6 +77,19 @@ export interface ProviderDefinition {
    * merged with the generic pattern set by the pty state detector.
    */
   waitingPatterns?: readonly RegExp[];
+  /**
+   * Direct-API credential support (#99). A stored key is used only when the
+   * user explicitly enables it per provider — CLI login/subscription remains
+   * the default (issue #100 decision). The CLI reads the key from envVar;
+   * `test` names a cheap authenticated endpoint for validating a stored key.
```
brannon-bowden: apiKey is required (not optional) on ProviderDefinition. Any provider without a cloud API-key concept (e.g. a local provider) will fail to type-check. Consider making this field optional and having vaultEnvFor/UI treat its absence as "not applicable".

src/main/pty-manager.ts
```diff
@@ -379,7 +380,9 @@ export class PtyManager extends EventEmitter {
     // Get memory content for system prompt injection
     // Pass via environment variable to avoid shell escaping issues with newlines
     let agentCmd = `${launch.command}${sessionFlag}`;
-    const processEnv = { ...process.env, ...launch.env } as { [key: string]: string };
+    // vaultEnvFor is empty unless the user explicitly opted this provider
+    // into API-key auth (#99) — CLI login/subscription stays the default.
+    const processEnv = { ...process.env, ...launch.env, ...vaultEnvFor(provider.id) } as { [key: string]: string };
```
brannon-bowden: processEnv now includes the decrypted API key when opted in. Please confirm redactEnv (or any debug logging of this env) scrubs ANTHROPIC_API_KEY/OPENAI_API_KEY/GEMINI_API_KEY/XAI_API_KEY so the "never touches logs" guarantee in key-vault.ts actually holds.

src/main/key-vault.ts
```diff
@@ -0,0 +1,130 @@
...
+export function hasKey(providerId: string): boolean {
+  return getPreference(KEY_PREF_PREFIX + providerId) !== null;
+}
+
+export function useKey(providerId: string): boolean {
+  return getPreference(USE_PREF_PREFIX + providerId) === 'true';
+}
+
+export function setKey(providerId: string, key: string): void {
+  getProvider(providerId); // throws on unknown ids
```
brannon-bowden: deleteKey doesn't validate providerId via getProvider() like setKey does — low risk, but inconsistent with the validation pattern used elsewhere in this file.

src/main/key-vault.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (!isVaultAvailable()) {
+    throw new Error('Secure key storage is not available on this system');
+  }
+  const trimmed = key.trim();
+  if (!trimmed) {
```
brannon-bowden, line 52: setUseKey likewise skips getProvider() validation for unknown provider ids.

brannon-bowden (changes requested):
This is a well-designed feature (safeStorage-backed encryption, explicit opt-in that preserves subscription/CLI-login as the default, key never returned over IPC, solid unit test coverage of the vault lifecycle), but there are a couple of things that need to be resolved before merge, and CI is currently red.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate are both red (GitGuardian passes). Per policy this blocks approval regardless of code quality — please share the failure details or re-run once addressed.
- **`apiKey` is a required (non-optional) field on `ProviderDefinition`** (`src/main/providers/types.ts:84`), and `passthroughProvider`'s `Pick<...>` was widened to require it too. Every current and future provider implementation must now supply a cloud API-key + test-endpoint config, including ones where that concept doesn't apply (e.g. a local/self-hosted provider). If any existing provider file wasn't touched by this PR (the arena engine already imports an `ollamaParser`, suggesting an Ollama provider exists), it will now fail to type-check — which would plausibly explain the failing quality-gate/SonarQube checks. Recommend making `apiKey` optional (`apiKey?: {...}`) and having `vaultEnvFor`/`listVaultStatuses`/the settings UI treat providers without it as "not applicable" rather than forcing every provider to fabricate a test endpoint.
- **Unverified log-redaction coverage for injected key env vars** (`src/main/pty-manager.ts:385`, similarly `src/main/arena/engine.ts`): `processEnv` now includes the decrypted key (e.g. `ANTHROPIC_API_KEY`) whenever the opt-in toggle is on, before being passed to `redactEnv`/spawn. The header comment in `key-vault.ts` promises the plaintext key "never touch[es]... logs," but `redactEnv`'s implementation isn't part of this diff, so it isn't clear it already scrubs these new provider-specific key names. Please confirm (or add a test) that any debug logging of `processEnv` redacts `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`/`GEMINI_API_KEY`/`XAI_API_KEY` — a regression here would leak live API keys into log files.

## Suggestions (non-blocking)

- `deleteKey` and `setUseKey` (`src/main/key-vault.ts:36`, `:41`) don't validate `providerId` via `getProvider()` the way `setKey` does. Not exploitable (worst case is a harmless no-op preference delete/write), but for consistency it'd be cleaner to validate there too.
- No integration test exercises `vaultEnvFor` actually being merged into the spawn `env` in `pty-manager.ts`/`arena/engine.ts` — the unit tests only cover `key-vault.ts` in isolation. Worth a lightweight test asserting the merge order (vault env wins over `process.env`/`launch.env`) since that's the actual behavior users depend on.
- If the OS keychain becomes unavailable mid-session (e.g., locked), `vaultEnvFor`/`testKey` silently behave as "no key" rather than surfacing that distinction to the user — minor UX point, not a correctness issue.

src/main/providers/types.ts
William-Long-II: Adopted in 2279002 — `apiKey` is now optional; providers without it are treated as not-applicable end to end (vault setKey/setUseKey/testKey reject them, they don't appear in the Settings section, `vaultEnvFor` stays empty). For the record, the type error hypothesis didn't apply — Ollama is an engine special-case, not a `ProviderDefinition`, and all four registry providers were updated (tsc was clean) — but the future-proofing point is right and taken.

src/main/pty-manager.ts
William-Long-II: Confirmed and hardened in 2279002. Two facts: (1) nothing in the launch path logs `processEnv` today (the only spawn-time log line prints shell + args); (2) `redactEnv`'s pattern (`/key|secret|token|password|auth/i`) matches all four vault-injectable names. It was actually pre-existing dead code — it's now extracted to `src/main/redact-env.ts` as the documented guard with a test asserting ANTHROPIC/OPENAI/GEMINI/XAI `_API_KEY` all scrub to `[REDACTED]`, so any future env logging has a proven helper to route through.

src/main/key-vault.ts
William-Long-II: Fixed in 2279002 — `deleteKey` validates via `getProvider`, and `setUseKey` via the same `requireApiKeyConfig` used by `setKey`.

src/main/key-vault.ts
William-Long-II: Fixed in 2279002 (same commit as above).

William-Long-II:
All review items addressed in 2279002:

- **`apiKey` optional** — providers without API-key support are "not applicable" across the vault, IPC, and UI.
- **Redaction proven** — `redactEnv` extracted to its own module with a test covering all four vault-injectable env vars; nothing logs launch env today, and the helper is the documented guard if that changes.
- **Integration tests** — the arena engine suite now asserts the vault env is merged into the real contestant spawn *and wins over ambient `process.env`*, and that non-opted contestants get no injection.
- **Validation consistency** — `deleteKey`/`setUseKey` validate provider ids like `setKey`.
- **Keychain-unavailable mid-session** (non-blocking): acknowledged — `vaultEnvFor` degrading to "no key" keeps launches working on CLI auth rather than failing; surfacing a distinct "keychain locked" state in the UI is a fair polish item I'd take as a follow-up if it bites in practice.
- All 7 SonarQube issues fixed (void operators in the test stub, boolean-`||` button guards, JSX spacing).

Suite is 65 tests. Re-requesting review once CI is green.

src/main/pty-manager.ts
```diff
@@ -22,18 +22,9 @@ import {
 } from './repositories/sessions';
 import { resolveAccountForSession, touchAccount } from './repositories/accounts';
 import { writeMemoryFile, getMemoryInjectionContent } from './memory/injector';
+import { vaultEnvFor } from './key-vault';
```
brannon-bowden, line 25: redactEnv was removed from this file and moved to redact-env.ts, but no import of it was added back here. If it was still referenced elsewhere in this file, this is now a broken reference; if not, the redaction utility is unused and the 'never logged' guarantee isn't actually wired up anywhere in this diff.

src/main/index.ts
```diff
@@ -1186,6 +1187,24 @@ safeHandle('providers:detect', async () => {
   return detectProviders();
 });
 
+// Provider API-key vault (#99). Keys go in and are tested; they are never
+// returned to the renderer.
+safeHandle('vault:list', async () => {
+  return keyVault.listVaultStatuses();
+});
+safeHandle('vault:setKey', async (providerId: string, key: string) => {
+  keyVault.setKey(providerId, key);
```
brannon-bowden: No runtime validation of providerId/key types before calling into keyVault.setKey. Consider validating types at the IPC boundary rather than relying solely on TS types enforced only in the preload layer.

brannon-bowden (changes requested):
This PR adds a well-designed, opt-in encrypted API-key vault (Electron `safeStorage`) with solid unit-test coverage of the core invariant (stored key ≠ used key), but CI is currently red and there's a real gap in the log-redaction wiring that needs to be resolved before merge.

## Issues

- **CI is failing** (SonarQube Code Analysis, quality-gate). Both must be green before merge; GitGuardian passing doesn't offset a red build/quality gate.
- **`redactEnv` was extracted from `pty-manager.ts` into its own module but is never imported/used anywhere in this diff** (`src/main/pty-manager.ts:25`, `src/main/redact-env.ts`). The old inline `redactEnv` function is deleted from `pty-manager.ts` and replaced only with a `vaultEnvFor` import — there's no `import { redactEnv } from './redact-env'` added back. Two possibilities, both bad: (a) if `redactEnv` was still referenced elsewhere in `pty-manager.ts` (e.g. a log call outside this diff's hunks) it's now an unresolved reference and would fail to compile — plausibly the actual cause of the failing CI checks; or (b) it's genuinely unused, meaning the "plaintext keys never touch... logs" guarantee that `redact-env.test.ts` documents is only unit-tested in isolation and not wired into any real log call site in this codebase. Either way this needs to be fixed and a concrete call site added/verified before merge.
- **No input validation on the new IPC handlers** (`src/main/index.ts:1196`-1207). `vault:setKey`, `vault:setUseKey`, etc. trust `providerId`/`key`/`use` types coming across `ipcRenderer.invoke` without runtime checks. The preload types make this safe from the app's own renderer, but IPC handlers are a trust boundary — a compromised/malicious renderer context could invoke these directly with unexpected types (e.g. non-string `providerId`) and rely on downstream errors rather than explicit validation. Low severity given this is a local desktop app, but worth a guard given it's a new credential-handling surface.

## Suggestions (non-blocking)

- No delete confirmation in `ProviderKeyRow` — clicking "Delete" immediately removes the stored key with no undo. A simple confirm step would reduce accidental data loss.
- No tests cover the new IPC handlers in `src/main/index.ts` or the `ProviderKeyVault`/`ProviderKeyRow` renderer components; the `key-vault.ts` core logic and arena env-merge behavior are well covered, but the IPC/UI glue is not.
- `vault-key-input` in `ProviderKeyVault.tsx` could set `autoComplete="off"` to discourage browser/OS password-manager prompts on a field that's explicitly not a login form.

src/main/pty-manager.ts
William-Long-II: Fair catch — it was (b): proven in isolation but wired nowhere. Fixed in bc870b4: both launch paths now log which env var was injected when key auth is active, with the value scrubbed through `redactEnv` (e.g. `[PTY] API-key auth enabled for 'claude': { ANTHROPIC_API_KEY: '[REDACTED]' }`). The guarantee is now enforced at the two real call sites that handle the decrypted key, and the unit test pins the pattern. Nothing else in pty-manager referenced it — tsc was clean; CI red last round was a test-fixture security hotspot, since reviewed.

src/main/index.ts
William-Long-II: Added in bc870b4 — the vault handlers validate `providerId`/`key` as non-empty strings and `use` as a boolean at the IPC boundary before touching the vault.

William-Long-II:
Round-two items addressed in bc870b4:

- **redactEnv wired for real** — both launch paths log the injected key env var (value scrubbed via `redactEnv`) whenever key auth is active; the guarantee is enforced at the call sites that handle the decrypted key.
- **IPC validation** — vault handlers runtime-check argument types at the trust boundary.
- **Non-blocking suggestions taken**: Delete now requires a confirming second click, and the key input sets `autoComplete="off"`.
- **IPC/UI glue tests** (non-blocking): acknowledged as a gap — the handlers are thin wrappers over the fully-tested vault functions plus the new type guards, and the repo has no React component-test infra today; I'd rather introduce that harness as its own change than bolt it onto this PR.
- CI note: the previous red was a security hotspot on the redaction test fixture (`DB_PASSWORD: 'z'` — the fixture exists to prove password-named vars get scrubbed); it's been reviewed as safe and the gate is green.

Suite remains 65 tests, all green. Re-requesting review once CI passes on this commit.

brannon-bowden (approved):
I wasn't able to load the skill's workflow file (permission denied), so I'll proceed with an equivalent-rigor manual review of the diff directly, applying the same adversarial lenses (correctness, edge cases, security, test coverage).

Based on my analysis of the diff:

**Encryption/storage path** (`key-vault.ts`): `setKey` correctly gates on `isVaultAvailable()` and rejects empty/unknown-provider input before ever calling `safeStorage.encryptString`. Plaintext never reaches `preferences` (verified by the test asserting stored values don't contain the raw secret). `decryptKey` fails closed (returns `null`) on a bad blob rather than throwing outward, and deliberately avoids logging ciphertext/plaintext.

**Subscription-preservation invariant**: `vaultEnvFor` returns `{}` unless `useKey()` is true, and `setUseKey(id, true)` throws if no key is stored — this is well-tested and matches the stated design goal (CLI login stays default).

**IPC trust boundary**: `index.ts` validates argument types with `requireIpcString`/boolean checks before touching the vault — appropriate given untrusted renderer input, and consistent with how a compromised renderer could otherwise pass non-string providerId to the SQL-backed preference key builder.

**Env merge/log redaction**: Both `arena/engine.ts` and `pty-manager.ts` spread `vaultEnv` last so it wins over ambient/launch env (covered by an explicit arena test), and both route the injected map through the shared `redactEnv` before logging. The regex is intentionally broad (`/key|secret|token|password|auth/i`), which only over-redacts, not under-redacts.

Two non-blocking gaps worth calling out below (a UI confirmation-state bug and a test-coverage asymmetry between the arena and PTY launch paths). Nothing rises to blocking; CI is green across all three checks.




Solid, well-scoped PR: keys are encrypted via `safeStorage`, never leave the main process in plaintext, never touch logs (verified by dedicated redaction tests), and are opt-in per provider so the existing CLI-login flow is untouched by default. IPC handlers validate types at the trust boundary. CI is green (SonarQube, quality-gate, GitGuardian). Approving with two minor, non-blocking notes.

## Issues

- `src/renderer/components/ProviderKeyVault.tsx:45-56` — `confirmingDelete` is only reset to `false` on the second click that actually fires the delete request; it is never reset if the user clicks "Delete" (arming the confirmation) and then does something else (e.g. edits the field and clicks "Save"). If they click "Delete" again later, it will delete immediately without re-confirming, since the stale `true` survives across unrelated actions and `onChanged`-triggered re-renders (the row's local state isn't reset because `ProviderKeyRow` stays mounted with a stable `key`). Consider resetting `confirmingDelete` in `saveKey`/`testKey`, or clearing it via a blur/timeout.
- `src/main/pty-manager.ts` (vault-env merge, near the `processEnv` assignment) — the vault-env injection + override-ordering + redacted-logging logic is only exercised by tests in `arena/engine.test.ts`. The interactive PTY session path (the primary way users launch a provider, not just Arena mode) has no equivalent test confirming `vaultEnv` actually reaches the spawned process env or that it correctly overrides `launch.env`/ambient env. The underlying units (`vaultEnvFor`, `redactEnv`) are well covered in isolation, so this is a coverage gap rather than a suspected bug, but it's the one integration point left unverified for the code path most users will actually exercise.

## Suggestions

- `redactEnv`'s regex will also redact any env var whose name merely contains "key"/"auth" as a substring (e.g. a hypothetical `MONKEY_PATCH` var) — harmless (over-redaction, not under-redaction) but worth a one-line comment if a future log ever looks unexpectedly scrubbed.
- `decryptKey`'s failure path doesn't clear the corrupted blob from preferences, so a permanently-undecryptable key (e.g. after restoring the DB on another machine) will silently no-op on every launch and log an error each time rather than surfacing to the user that the key needs to be re-entered. Not required for this PR, but worth considering for the vault status UI (e.g. surface a "corrupted, please re-enter" state) in a follow-up.

PR #108, opened 2026-07-14
Route cmd.exe agent launches through PowerShell (env-ref injection)

### LINKED ISSUE #106: Task: Harden cmd.exe env-var expansion in getShellLaunch quoting
### Summary

`getShellLaunch`'s cmd.exe branch references env vars as `"%VAR%"` inside a double-quoted token. cmd.exe percent-expansion is not safely escaped against embedded `"` or `&`, so free text containing those characters could break out of quoting when the detected shell is cmd.exe. Long-standing behavior inherited from the session launch path (where the value was app-generated), but arena mode (#100/#105) now routes arbitrary pasted prompt text through the same path, widening exposure.

### Proposed direction

Options to evaluate: route arena spawns through PowerShell on Windows regardless of the user's session shell; or sanitize/reject unsafe characters for the cmd.exe branch specifically; or write the prompt to a temp file and reference it instead of env expansion.

Raised in review of #105.

William-Long-II:
## Description
Hardens the follow-up from #105's review: cmd.exe `%VAR%` expansion is **textual and pre-parse**, so a value containing `"` or `&` escapes any quoting — meaning arena prompts (arbitrary pasted text) and memory-injected system prompts could inject commands for users whose configured shell is cmd.exe. This class isn't fixable by escaping; the safe expansions are the post-parse kind (`bash "$VAR"`, PowerShell `$env:VAR`).

- `getShellLaunch`'s cmd.exe branch now routes env-ref launches through **Windows PowerShell at its absolute SystemRoot path** (always present, PATH-poisoning-proof). The module doc states the invariant every future branch must satisfy.
- Scope is exactly the agent/arena wrapper — a user's plain interactive cmd sessions never touch `getShellLaunch` and are unchanged. Behavior note for cmd users: their *agent* launches now run under a PowerShell wrapper (env and system PATH carry over; agent CLIs don't rely on cmd autorun).

## Related issue
Closes #106

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- `bun test`: 72 pass — new per-branch argv tests (incl. the cmd→PowerShell reroute and absolute-path assertion), an invariant test that **no branch emits `%VAR%` refs**, and an engine-level regression that spawns a real contestant with a hostile prompt (`$(…)`, backticks, `&`, quote-breaks) and proves it round-trips literally without execution
- `tsc --noEmit` clean; SonarQube preflight on the changed module clean
- Manual QA worth doing on Windows: set shell to cmd.exe, launch a Claude session and an arena run, confirm both work under the PowerShell wrapper

## Checklist
- [ ] Tests pass locally (bun test, 72 pass)
- [x] Docs updated where needed (invariant documented in the module)

### DIFF
--- a/src/main/shell-launch.ts
+++ b/src/main/shell-launch.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
 import * as path from 'path';
 import { ShellInfo } from './shell-detector';
 
@@ -7,12 +8,16 @@ import { ShellInfo } from './shell-detector';
  * referencing an environment variable inside that command string. Single
  * source of truth for agent sessions, login ptys, and arena contestants.
  *
- * Env-ref safety invariant (#106): every branch's envRef must expand AFTER
- * the shell parses the command, so the value is never re-parsed as shell
- * syntax. bash `"$VAR"` and PowerShell `$env:VAR` both satisfy this.
- * cmd.exe `%VAR%` does NOT — its expansion is textual and pre-parse, so a
- * value containing `"` or `&` breaks out of any quoting. Agent launches
- * from cmd.exe shells therefore route through PowerShell instead.
+ * Env-ref safety invariant (#106): a launch's envRef must expand AFTER the
+ * shell parses the command, so the value is never re-parsed as shell syntax.
+ * bash `"$VAR"` and PowerShell `$env:VAR` both satisfy this. cmd.exe `%VAR%`
+ * does NOT — its expansion is textual and pre-parse, so a value containing
+ * `"` or `&` breaks out of any quoting.
+ *
+ * Callers declare via `needsEnvRef` whether they will interpolate a value
+ * with envRef. When true, cmd.exe cannot host the launch safely and it is
+ * rerouted through PowerShell. When false (e.g. the login pty, which just
+ * runs `claude`), cmd.exe is used unchanged and envRef throws if misused.
  *
  * (Lives outside pty-manager so consumers that don't need node-pty — like
  * the arena engine and its tests — don't drag the native module in.)
@@ -23,6 +28,14 @@ export interface ShellLaunch {
   envRef(name: string): string;
 }
 
+export interface ShellLaunchOptions {
+  /**
+   * Whether the caller will build the command with envRef (interpolating a
+   * possibly-untrusted value). Gates the cmd.exe→PowerShell reroute (#106).
+   */
+  needsEnvRef: boolean;
+}
+
 function powershellLaunch(shell: string): ShellLaunch {
   return {
     shell,
@@ -31,7 +44,28 @@ function powershellLaunch(shell: string): ShellLaunch {
   };
 }
 
-export function getShellLaunch(shellInfo: ShellInfo): ShellLaunch {
+/**
+ * Absolute path to Windows PowerShell 5.1 (bundled with Windows). Absolute so
+ * a poisoned PATH can't substitute the binary. Throws a clear error rather
+ * than spawning an ENOENT if it is somehow absent (e.g. a pwsh-only image).
+ */
+function windowsPowershellPath(): string {
+  // SystemRoot can be set-but-empty in some sandboxed processes; an empty or
+  // whitespace value would make path.join produce a relative (PATH-relative)
+  // path, so fall back to the default whenever it isn't a usable string.
+  const raw = process.env.SystemRoot;
+  const systemRoot = raw && raw.trim().length > 0 ? raw : 'C:\\Windows';
+  const candidate = path.join(systemRoot, 'System32', 'WindowsPowerShell', 'v1.0', 'powershell.exe');
+  if (!fs.existsSync(candidate)) {
+    throw new Error(
+      `Windows PowerShell not found at ${candidate}. Agent launches from a cmd.exe ` +
+      `shell require it; set your Bodhilander shell to PowerShell or Git Bash instead.`
+    );
+  }
+  return candidate;
+}
+
+export function getShellLaunch(shellInfo: ShellInfo, options: ShellLaunchOptions): ShellLaunch {
   if (shellInfo.isWSL) {
     // Launch the agent inside WSL
     return {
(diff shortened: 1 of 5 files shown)


src/main/shell-launch.ts
```diff
@@ -27,18 +43,18 @@ export function getShellLaunch(shellInfo: ShellInfo): ShellLaunch {
...
     if (shellName.includes('cmd')) {
-      return {
-        shell: shellInfo.shell,
-        wrap: (cmd) => ['/c', cmd],
-        envRef: (name) => `"%${name}%"`,
-      };
+      // cmd.exe cannot host env-ref launches safely (#106) — see the module
+      // doc. Route through Windows PowerShell at its absolute path (always
+      // present; PATH-poisoning-proof). Only agent/arena launches use this
+      // wrapper; the user's plain cmd shell sessions are unaffected.
```
brannon-bowden: This comment asserts login-pty cmd.exe sessions are unaffected, but getShellLaunch is documented above as the single source of truth for agent sessions, login ptys, and arena contestants, and this branch has no call-site distinction. Please confirm login ptys don't go through this branch, or scope the PowerShell reroute explicitly to agent/arena launches.

src/main/shell-launch.ts
```diff
@@ -27,18 +43,18 @@ export function getShellLaunch(shellInfo: ShellInfo): ShellLaunch {
...
+      const powershell = path.join(
+        process.env.SystemRoot ?? 'C:\\Windows',
```
brannon-bowden: `?? 'C:\\Windows'` doesn't guard against an empty-string SystemRoot, which would produce a relative (non-absolute) path here, undermining the PATH-poisoning-proof claim in the comment above.

brannon-bowden (changes requested):
The security intent of this change is sound — cmd.exe's `%VAR%` expansion is textual/pre-parse and genuinely unquotable, and rerouting through PowerShell's `$env:VAR` (which expands post-parse) is the correct fix. However, there's a scope contradiction that needs to be resolved before merge.

## Issues

- **`src/main/shell-launch.ts:52`** (blocking): The new comment claims *"Only agent/arena launches use this wrapper; the user's plain cmd shell sessions are unaffected."* But the module's own docstring (unchanged, a few lines above at `src/main/shell-launch.ts:8`) states `getShellLaunch` is the *"Single source of truth for agent sessions, login ptys, and arena contestants."* This PR modifies the `shellName.includes('cmd')` branch inside `getShellLaunch` itself — the same function/branch every caller shares — with no parameter or call-site distinction between "agent launch" and "interactive login pty." As written, any consumer that calls `getShellLaunch()` for a plain interactive cmd.exe login session will now silently get routed to PowerShell instead, which is a much bigger behavioral change than "agent launches only" and isn't covered by any test (all new tests only exercise `wrap`/`envRef`, none assert that login-pty callers are unaffected). Please confirm whether login-pty call sites actually bypass this branch (e.g. via a separate code path not shown in this diff) — if not, either scope the reroute to agent/arena call sites explicitly (e.g. an options flag) or correct the comment and accept/document the login-shell behavior change.

- **`src/main/shell-launch.ts:54`** (minor): `process.env.SystemRoot ?? 'C:\\Windows'` only falls back on `null`/`undefined`. If `SystemRoot` is set but empty (unusual, but possible in some sandboxed/CI-spawned Windows processes), `path.join('', 'System32', ...)` produces a relative path (`System32\WindowsPowerShell\v1.0\powershell.exe`), which silently defeats the "PATH-poisoning-proof" guarantee the adjacent comment claims and could resolve against an attacker-influenced CWD/PATH. Consider checking for a falsy/empty string, not just nullish.

## Suggestions (non-blocking)

- No check that the computed PowerShell path actually exists before use — on a hardened/enterprise Windows image without Windows PowerShell 5.1 (e.g., pwsh-only installs), agent launches would fail with an unhelpful ENOENT rather than a clear error. Consider a `pwsh.exe`-on-PATH fallback or a clearer error surface.
- The new `path_isAbsoluteWin` regex-based assertion in `shell-launch.test.ts` is a reasonable workaround for running Windows-path assertions on POSIX CI, but note `path.join` in the test only exercises the POSIX implementation (module resolution happens at import time regardless of the `process.platform` mock), so the exact separator characters in the computed path aren't verified end-to-end by these tests — acceptable given the abs-path + substring checks, just worth knowing the limit of coverage.

src/main/shell-launch.ts
William-Long-II: You're right — the reroute was over-broad. Traced the call sites: `getShellLaunch` is called by buildAgentSpawn (agent, uses envRef), the arena engine (uses envRef), AND createLoginSession (the account-add OAuth pty, which just runs `claude` — no envRef). My blanket reroute silently moved that login pty to PowerShell with no security benefit. Fixed in e16d678 by adding a required `needsEnvRef` option: the cmd→PowerShell reroute now fires ONLY when the caller interpolates a value via envRef. The login pty passes `needsEnvRef: false`, keeps the user's cmd.exe shell, and its envRef throws if ever misused. New tests cover both cmd paths (reroute vs. preserved) explicitly.

src/main/shell-launch.ts
William-Long-II: Fixed in e16d678 — an empty or whitespace-only SystemRoot is now treated as absent (`raw && raw.trim().length > 0 ? raw : 'C:\\Windows'`), so path.join can never produce a PATH-relative path. Also added a filesystem existence check that throws a clear 'PowerShell not found — use PowerShell or Git Bash' error instead of an opaque ENOENT on pwsh-only images (your non-blocking suggestion), covered by a test.

William-Long-II:
All review items addressed in e16d678:

- **Over-broad reroute (blocking)** → `getShellLaunch` now takes a required `needsEnvRef`; the cmd→PowerShell reroute fires only for env-ref callers (agent sessions, arena). The login pty passes `false` and keeps the user's cmd.exe shell. Both cmd paths are tested explicitly.
- **Empty SystemRoot (minor)** → guarded against empty/whitespace, not just nullish.
- **PowerShell-missing (non-blocking)** → clear error instead of ENOENT, with a test.
- The `%VAR%`-never-emitted invariant test still passes across all branches.

Suite is 74 tests. Re-requesting review once CI is green.

brannon-bowden (approved):
Approving: this is a solid, well-targeted security fix with good test coverage, and I found no bugs or security gaps that need to block merge.

The core fix is sound: cmd.exe's `%VAR%` expansion is textual/pre-parse and therefore unsafe to interpolate untrusted values into, so rerouting those launches through PowerShell's `$env:VAR` (which expands after parsing, like bash `"$VAR"`) closes the injection window. The PowerShell binary is resolved via an absolute path derived from `SystemRoot` rather than trusting `PATH`, with a clear error if it's missing. The new `shell-launch.test.ts` exercises all platform branches plus the explicit env-ref invariant, and the added `engine.test.ts` case is a nice end-to-end check that a hostile prompt with shell metacharacters is never executed. CI is green across all three checks (SonarQube, quality-gate, GitGuardian).

Issues
- `src/main/pty-manager.ts:395` — the agent-session call hardcodes `{ needsEnvRef: true }` unconditionally, but `envRef()` is only actually invoked inside the `if (supportsSystemPrompt && processEnv.BODHILANDER_SYSTEM_PROMPT)` branch a few lines below. That means every cmd.exe agent session gets rerouted to PowerShell even when no env-ref interpolation will ever happen on that path — a broader behavior change than "reroute when env-ref is needed." This may be intentional (deliberately simpler/more conservative than gating per-condition, and consistent with the accompanying login-pty comment contrasting the two paths), but it's worth an explicit confirmation from the author that all cmd.exe agent sessions rerouting to PowerShell (not just ones using the system-prompt flag) is the intended scope.
- `src/main/pty-manager.ts` — neither of the two `getShellLaunch` call sites in this file has direct test coverage confirming the `needsEnvRef` value passed at each site. The new tests validate `shell-launch.ts`'s branching logic and the arena engine's use, but the wiring in `pty-manager.ts` (which is where the login-pty vs. agent-session distinction actually matters) is only verified by inspection.

Suggestions (non-blocking)
- The error thrown when `envRef()` is misused on a `needsEnvRef: false` cmd.exe launch — `'cmd.exe env-ref launch must set needsEnvRef (#106)'` — reads a bit ambiguously about which side is at fault. Something like `'envRef() called but this launch was created with needsEnvRef: false'` would be clearer for whoever hits it next.

PR #110, opened 2026-07-14
Pin Windows release runner to windows-2022 (node-gyp vs VS 2026)

### LINKED ISSUE #109: Task: Windows release build fails — node-gyp can't detect VS 2026 on windows-latest
### Summary

The `Build and Release` workflow's Windows job started failing at the `Install dependencies` step (postinstall → `electron-rebuild`):

```
Error: Could not find any Visual Studio installation to use
node-gyp failed to rebuild 'node_modules\node-pty'
postinstall script from "bodhilander" exited with 255
```

### Root cause

GitHub rolled the `windows-latest` label from Windows Server 2022 / VS 2022 to the **`windows-2025-vs2026`** image (Server 2025 + Visual Studio 2026). The project's bundled `node-gyp@12.4.0` recognizes Visual Studio up through 2022 (major version 17) but not VS 2026's new major version, so its `find-visualstudio` probe reports no installation. The last release (3.3.4, May 2026) built fine because `windows-latest` was still VS 2022 then.

Not a code regression — purely a CI toolchain/image change. macOS/Linux are unaffected (the macOS failure on the beta run is a separate expired Apple agreement).

### Proposed fix

Pin the Windows matrix runner to `windows-2022` (the known-good VS 2022 image), mirroring how the macOS runners are already pinned for build reproducibility. Add a comment so it isn't reverted to `windows-latest`.

Longer-term alternative (follow-up if `windows-2022` is retired): upgrade the native-build toolchain to a `node-gyp` that detects VS 2026 and confirm `node-pty` / `better-sqlite3` / `sqlite-vec` rebuild against it.

### Acceptance criteria

- [ ] `Build and Release` Windows job completes and produces the `.exe` installer
- [ ] Runner pin documented in-workflow

William-Long-II:
## Description
The `Build and Release` Windows job broke on the `3.4.0-beta.10` cut — `electron-rebuild` (postinstall) failed with *"Could not find any Visual Studio installation to use"*, so `node-pty` never rebuilt and the `.exe` was never produced.

**Root cause** (not a code regression): GitHub rolled the `windows-latest` label from Server 2022 / VS 2022 to the **`windows-2025-vs2026`** image (Visual Studio 2026). The project's `node-gyp@12.4.0` detects VS through 2022 (major 17) but not VS 2026's new major version. The last release (3.3.4, May) built fine because `windows-latest` was still VS 2022.

**Fix**: pin the Windows matrix runner to `windows-2022` — the known-good image — mirroring how the macOS runners are already pinned for build reproducibility, with a comment so it isn't reverted. Longer-term (noted on #109): move to a node-gyp that supports VS 2026 once the toolchain catches up.

## Related issue
Closes #109

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
- YAML validated (js-yaml)
- Real verification is the workflow run itself — this is a CI-runner config change with no unit-testable surface. After merge to `development`, re-running "Build and Release" should complete the Windows job and emit the installer. (The macOS legs still need the separate Apple Program License Agreement re-signed — tracked outside this PR.)

## Checklist
- [x] Docs updated where needed (rationale documented in-workflow)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,7 +73,14 @@ jobs:
         include:
           - os: ubuntu-latest
             platform: linux
-          - os: windows-latest
+          # Pinned to windows-2022 (VS 2022) deliberately (#109). GitHub moved
+          # the `windows-latest` label to windows-2025-vs2026 (Visual Studio
+          # 2026), which the project's node-gyp doesn't yet detect — the
+          # postinstall electron-rebuild fails with "Could not find any Visual
+          # Studio installation to use". Same build-reproducibility reasoning as
+          # the pinned macOS runners below. Revisit when node-gyp/node-pty gain
+          # VS 2026 support (or if windows-2022 is retired).
+          - os: windows-2022
             platform: win
           # BDHLNDR-38: build each Mac arch on its native runner so postinstall
           # (electron-rebuild) and sharp's optional-dep resolution produce


brannon-bowden (approved):
**Approve.** This is a narrowly-scoped, well-justified CI workflow pin with no impact on application code or public contracts.

## Blocking issues
None.

## Non-blocking issues
- `.github/workflows/release.yml:80` — `windows-2022` is a currently-supported GitHub-hosted runner label, but hosted-runner images do eventually get deprecated on their own schedule (as `windows-2019` already has). Since the whole point of this pin is to dodge an upstream toolchain incompatibility rather than a permanent requirement, it'd be worth a tracking issue/reminder to revisit this before GitHub announces `windows-2022` retirement, not just when node-gyp catches up — the comment already gestures at this ("Revisit when... or if windows-2022 is retired") but nothing currently enforces it.
- Worth confirming no other job/matrix entry elsewhere in `release.yml` (outside this hunk) still references `windows-latest` for a build step affected by the same electron-rebuild/node-gyp issue — the diff only shows this one matrix entry being pinned, and an inconsistent pin between jobs could reintroduce the failure elsewhere.

## Tests
No test coverage applies here (pure CI runner-label change, no logic). The real verification is an actual green run of the release workflow on `windows-2022` producing a working `electron-rebuild`/node-gyp build — worth confirming that a run has been (or will be) exercised before merging to a release-triggering branch, since CI status listed here (SonarQube, quality-gate, GitGuardian) doesn't actually validate the Windows build path this change touches.

PR #111, opened 2026-07-14
fix: remove Gemini provider, probe CLIs via interactive login shell

William-Long-II:
## Why

Two field-reported problems:

1. **Gemini CLI is a dead end for new users.** Google retired the consumer "Sign in with Google" OAuth path for Gemini CLI (June 18, 2026) — including AI Pro/Ultra accounts. The browser login appears to succeed but the token exchange is refused server-side with "This client is no longer supported" ([google-gemini/gemini-cli#28229](https://github.com/google-gemini/gemini-cli/issues/28229)). Only paid API keys still work, so surfacing the install/login flow just causes user concern.

2. **Grok CLI isn't detected even when installed and working.** The xAI installer adds `~/.grok/bin` to PATH in `~/.zshrc` only. Detection probed with a login *non-interactive* shell (`zsh -l -c`), which never reads `.zshrc` — but sessions launch through an interactive login shell (`-l -i -c`) that does. So a grok session would launch fine while the picker showed "not installed".

## What

- Remove the `gemini` provider (registry entry, module, arena parser, label, comments). Persisted `'gemini'` session rows degrade to the default provider through the existing `resolveProvider()` fallback (#96), now locked in by a test.
- Probe provider CLIs through the same interactive login shell (`-l -i -c`) that sessions and arena contestants launch under, so detection agrees with what a launch would actually find. Lookup only needs the exit code and `extractVersion` already skips non-version banner lines, so interactive-rc stdout noise stays tolerable.

## Testing

- `bun test src/main/__tests__ src/main/providers src/main/arena src/renderer` — 66 pass, 0 fail. (The `chat-parser` fixture failure also occurs on `development`; unrelated.)
- Reproduced the grok miss locally: `env -i HOME=$HOME zsh -l -c 'command -v grok'` fails while `zsh -l -i -c` succeeds, matching a real `~/.grok/bin` install whose PATH entry lives in `~/.zshrc`.


src/main/provider-detector.ts
```diff
@@ -37,9 +38,14 @@ export function buildProbe(command: string): Probe {
...
-  // without interactive-rc noise on stdout.
-  const wrap = (cmd: string): [string, string[]] => [shellInfo.shell, ['-l', '-c', cmd]];
+  // Interactive login shell (-l -i), matching how sessions and arena
+  // contestants actually launch. Interactive matters: installers commonly
+  // add their bin dir to PATH in the interactive rc only (e.g. the grok
+  // installer writes ~/.zshrc, which `-l -c` never reads), so a
+  // non-interactive probe reports "not installed" for a CLI that launches
+  // fine in a session. Interactive-rc noise on stdout is tolerable: lookup
+  // only needs the exit code, and extractVersion skips non-version lines.
+  const wrap = (cmd: string): [string, string[]] => [shellInfo.shell, ['-l', '-i', '-c', cmd]];
```
brannon-bowden, line 50: Switching to an interactive login shell (-l -i -c) risks execFile hanging on blocking rc-file prompts (e.g. Oh My Zsh's update check reading stdin) since execFile's default piped stdin is never closed. Please confirm a timeout and/or stdio:['ignore',...] is set on the actual probe invocation before merging.

src/shared/types.ts
```diff
@@ -127,7 +127,6 @@ export const DEFAULT_SESSION_PROVIDER = 'claude';
 export const PROVIDER_LABELS: Record<string, string> = {
   claude: 'Claude',
   codex: 'Codex',
-  gemini: 'Gemini',
   grok: 'Grok',
```
brannon-bowden, line 130: Confirm all consumers of PROVIDER_LABELS route session.provider through resolveProvider/resolveLaunchProviderId first, so legacy sessions persisted with provider:'gemini' don't render an undefined label.

brannon-bowden (commented):
This is a clean, well-scoped removal PR — the Gemini provider, its parser, tests, and registry entry are all consistently removed, and a regression test was added confirming persisted `'gemini'` sessions degrade to the default provider via `resolveProvider`. CI is green across SonarQube, quality-gate, and GitGuardian. One design point around the shell-probing change is worth the author confirming before merge.

## Issues

- **`src/main/provider-detector.ts:48`** — Switching the probe from a non-interactive login shell (`-l -c`) to an interactive one (`-l -i -c`) is reasonable for PATH-sourcing correctness, but interactive shell startup is a well-known source of blocking prompts: Oh My Zsh's periodic "would you like to update?" `read` prompt, powerlevel10k's instant-prompt console-output warnings, and other rc-file gates that wait on stdin. `child_process.execFile` (used here per the file's imports) defaults to a piped, unclosed stdin, so if the probe is invoked without explicit `stdio` handling (e.g. `stdio: ['ignore', ...]`) or a `timeout` option, any of these prompts would hang the child process indefinitely rather than just producing noisy stdout. The updated comment addresses stdout noise ("Interactive-rc noise on stdout is tolerable...") but doesn't address stdin/hang risk at all, which suggests it may not have been considered. The actual `execFile` call site isn't part of this diff, so this may already be mitigated elsewhere — worth the author confirming a timeout and closed/ignored stdin are in place for these probe calls before this ships broadly, since a hang here would block provider detection for a fairly common shell configuration (Oh My Zsh users specifically).
- CI is fully green (SonarQube, quality-gate, GitGuardian all passing) — noted per policy, not a concern here.

## Suggestions (non-blocking)

- `src/shared/types.ts:130` — `PROVIDER_LABELS` is a plain `Record` with no `'gemini'` entry now. If any UI code indexes `PROVIDER_LABELS[session.provider]` directly for legacy sessions rather than routing through `resolveProvider`/`resolveLaunchProviderId` first, a session persisted with `provider: 'gemini'` would render an `undefined` label instead of falling back gracefully. Worth double-checking all label lookups go through the resolver.
- Since this diff only touches main-process files, it's worth a quick sweep of renderer-side code (icons, provider dropdowns, arena UI) for any hardcoded `'gemini'` references that wouldn't be caught by removing the registry entry.

William-Long-II:
Thanks — the stdin/hang point was real and is now addressed in 23686c2 → latest push.

**Interactive-shell hang risk:** the probe already ran with `timeout: PROBE_TIMEOUT_MS` (10s) on the `execFile` call, so a blocking rc prompt degraded to a slow "not installed" rather than an indefinite hang — but it *would* have burned the full 10s per probe for Oh My Zsh users with the update prompt enabled. `run()` now ends the child's piped stdin immediately, so any rc `read` hits EOF and proceeds. Verified locally: a `zsh -l -i -c 'read -r x; …'` probe resolves in ~170ms instead of timing out. The `buildProbe` comment now covers stdin and the timeout alongside stdout noise.

**`PROVIDER_LABELS` lookups:** all three call sites use a raw-id fallback (`PROVIDER_LABELS[session.provider] ?? session.provider` in `App.tsx:1120,1123`, `?? status.providerId` in `ProviderKeyVault.tsx:63`), so a legacy `provider: 'gemini'` row renders as "gemini" rather than `undefined`, and launching it degrades to Claude via `resolveProvider`.

**Renderer sweep:** `grep -ri gemini src/renderer` comes back empty — the provider picker, settings, key vault, and arena contestant list are all built from the main-process registry over IPC, so the registry removal is the single source of truth.

brannon-bowden (approved):
This PR cleanly removes the Gemini provider (registry, parser, tests, types, docs) and fixes the CLI-detection probe to run through an interactive login shell, with a well-reasoned rationale (installers that only modify `~/.zshrc` are invisible to a non-interactive `-l -c` probe) and sensible mitigations for the side effects of going interactive (stdout noise tolerated, stdin explicitly closed to avoid hangs on rc prompts, existing timeout as backstop). CI is green across all three checks. Test coverage for the provider removal and the shell-flag change looks thorough, including the added regression test that a persisted `'gemini'` session degrades to the default provider.

## Issues

- **`src/main/provider-detector.ts:50`** (context around the `wrap` function, see also the unchanged WSL test in `src/main/__tests__/provider-detector.test.ts:94-101`) — the interactive-shell fix is only applied to the POSIX `wrap()` path. The WSL branch still probes via `bash -lc` (non-interactive), even though the same class of bug applies there: an installer that only appends to `~/.bashrc` will be invisible to `bash -lc` just as it was to `zsh -l -c`. If WSL sessions actually launch interactively (matching the stated goal of "probing the same way sessions launch"), this looks like an incomplete fix; if WSL sessions genuinely launch non-interactively, worth a one-line comment saying so to avoid the asymmetry looking like an oversight.

## Suggestions (non-blocking)

- `src/main/provider-detector.ts` (the new `child.stdin?.end()` call in `run()`) — this closes a real gap (previously the piped stdin was left open indefinitely), but there's no test exercising the "rc script blocks on stdin read" scenario it's meant to fix. Understandably hard to unit test without spawning a real shell; a comment or a follow-up integration test would help future maintainers trust the behavior.
- Since this diff only touches `src/main` and `src/shared`, it's worth double-checking the renderer/UI layer (provider dropdown, icons, settings screens) for any lingering "Gemini" references that this PR's scope wouldn't have caught.

PR #112, opened 2026-07-15
feat: scope arena runs to a project folder via group working dirs

William-Long-II:
## Why

Arena contestants spawned with no `cwd` — the Electron process cwd, effectively `/` when launched from the Dock — so arena could only answer context-free prompts. Since contestants are the real CLIs (`claude -p`, `codex exec`, `grok -p`), simply spawning them inside a project folder gives them that codebase as context, the same way they behave in a terminal.

## What

- **Engine/IPC**: `arena:start` accepts an optional working directory (validated up front — must exist and be a directory, checked before any rows are created). `ArenaEngine.prepare()` threads it onto each contestant and `runCli` passes it as the spawn `cwd`. Ollama (bare HTTP chat, no filesystem) ignores it.
- **Persistence**: `arena_runs.working_dir` column (additive `PRAGMA table_info` migration, matching the existing pattern); `ArenaRun.workingDir` exposed to the renderer and shown (`📁 <dir>`) when viewing a live or history run.
- **UX — reuses groups instead of a folder picker**: the arena panel gets a "folder scope" select built from sidebar groups that already have a working directory (`buildFolderOptions`, pure + unit-tested like `providerPickerOptions`). Each group's context menu gains **"Ask Arena About This Folder"** (disabled until the group has a working dir), which opens/retargets the arena panel pre-scoped to that group.
- **Test-infra fix**: `shell-launch.test.ts` mocked `'fs'` with only `{existsSync}`; bun's `mock.module` is process-global, so every test file running after it saw a gutted `fs` (the standing chat-parser `fs.readFileSync is not a function` failures in combined runs). The mock now spreads the real module (grabbed via `node:fs`, which stays unmocked).

## Testing

- New engine tests: a scoped run's contestant observes the scoped directory as its cwd (real spawn through the shell wrapper, `pwd` fake provider), and unscoped runs behave as before.
- New `buildFolderOptions` unit tests (dir-less groups excluded, whitespace dirs treated as unset, duplicates kept).
- `bun test src/main src/renderer`: 120 pass; the single remaining failure is the pre-existing `06-response` chat-parser snapshot diff, which fails identically on `development`.
- `tsc` for main + full project: clean (the two `mdns-advertiser` errors pre-exist on `development`). Webpack production build reports the same 3 pre-existing errors as `development`, none in changed files.

### DIFF
--- a/src/main/__tests__/shell-launch.test.ts
+++ b/src/main/__tests__/shell-launch.test.ts
@@ -6,12 +6,17 @@
  * Run with: bun test src/main/__tests__
  */
 import { describe, expect, test, afterEach, mock } from 'bun:test';
+import * as realFs from 'node:fs';
 import type { ShellInfo } from '../shell-detector';
 
 // The cmd→PowerShell reroute resolves an absolute powershell.exe on disk;
-// stub fs.existsSync so that path is testable on a POSIX runner.
+// stub fs.existsSync so that path is testable on a POSIX runner. The rest
+// of fs is passed through: mock.module is process-global in bun, so a
+// bare {existsSync} would gut fs for every test file that runs after this
+// one (the real module is grabbed via 'node:fs', which stays unmocked).
 let powershellExists = true;
 mock.module('fs', () => ({
+  ...realFs,
   existsSync: () => powershellExists,
 }));
 
--- a/src/main/arena/__tests__/engine.test.ts
+++ b/src/main/arena/__tests__/engine.test.ts
@@ -65,6 +65,13 @@ const FAKE_PROVIDERS: Record<string, any> = {
       createParser: textParser,
     },
   },
+  pwder: {
+    id: 'pwder',
+    arena: {
+      buildCommand: () => 'pwd',
+      createParser: textParser,
+    },
+  },
 };
 mock.module('../../providers', () => ({
   getProvider: (id: string) => {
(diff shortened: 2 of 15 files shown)


brannon-bowden (approved):
This PR threads an optional `workingDir` through the arena run lifecycle (IPC → engine → spawn → SQLite) so contestants execute inside a project folder, with a companion sidebar affordance to trigger it. The core plumbing is solid: `cwd` is threaded correctly into `spawn`, the schema migration is idempotent and guards against pre-existing databases, the existence/`isDirectory` check happens before any rows are created, and new behavior (scoped vs. unscoped runs) is covered by engine tests. The `fs` mock leak fix in `shell-launch.test.ts` is a nice bit of housekeeping.

## Issues

- `src/renderer/App.tsx:366` (and the panel mount at `src/renderer/App.tsx:1350`) — `arenaGroupId` is set by "Ask Arena About This Folder" but never cleared on `onClose`, and `ArenaPanel`'s `initialGroupId` prop is fed straight from it. If the Arena panel is later reopened through any other entry point (the pre-existing generic way to open Arena mode, unmodified by this diff) without explicitly resetting `arenaGroupId`, the panel will silently pre-select the last folder scope from an unrelated group instead of defaulting to "No folder". Worth resetting `arenaGroupId` to `null` in `onClose`, or having every open path set it explicitly.
- `src/main/index.ts:1230` — the new `workingDir` existence/`isDirectory` validation in the `arena:start` handler has no direct test covering the rejection path (nonexistent dir, or a path that's a file rather than a directory).

## Suggestions (non-blocking)

- `src/renderer/styles/global.css:1550` — `.arena-folder-select` sets `text-overflow: ellipsis` but omits `white-space: nowrap` and `overflow: hidden`, so long labels won't actually truncate; `.arena-run-scope` right below it shows the correct three-property combo.

PR #115, opened 2026-07-15
fix: seed first account's config dir with legacy ~/.claude transcripts

William-Long-II:
## Why

Registering your **first** Claude account silently breaks every existing session's conversation history. Root-caused from a field incident on 3.4.0-beta.10:

1. Before any account exists, sessions launch against legacy `~/.claude` (where all conversation transcripts live).
2. The first registered account becomes the default, and `resolveAccountForSession`'s fallback chain (`session → group → default`) re-homes every unassigned session's `CLAUDE_CONFIG_DIR` to the new account's dir on next launch.
3. That dir starts **empty**, so every stored `--resume` UUID fails with "No conversation found with session ID: x" — for all sessions at once.

The account-resolution code is long-standing and correct; the landmine is the fresh config dir starting with no history.

## What

- New `seedLegacyConversations(configDir, legacyDir)` (`src/main/legacy-claude-seed.ts`, kept outside `account-auth` so it's importable without node-pty): copies `~/.claude/projects/` — the per-cwd transcript store Claude Code reads on `--resume` — into a first account's config dir. Deliberately not copied: credentials (login authenticates fresh) and `settings.json` (MCP/hook registration writes its own). Never clobbers an existing `projects/` dir; failures log a warning and degrade to the old behavior rather than blocking account creation.
- `startLoginFlow` awaits the seed for the first account only (subsequent accounts are separate identities and shouldn't inherit another account's history). The IPC handler already awaits, so the async signature change is transparent.
- Test-infra fix found while validating: the shell-launch fs stub's delegating `existsSync` must call a **pre-captured** real function. Referencing `realFs.existsSync` inside the stub resolves through the module namespace, which bun retargets to the mock itself — infinite recursion that hung the entire combined suite at 98% CPU.

## Testing

- 5 new unit tests: transcript tree copied; credentials/settings excluded; no-op without a legacy dir; existing target never clobbered; copy failure swallowed and reported.
- Full suite: 123 pass / 1 fail — the failure is the pre-existing `06-response` chat-parser snapshot diff that fails identically on `development`.
- `tsc` main: clean (mdns-advertiser pre-existing errors excluded).

### DIFF
--- a/src/main/account-auth.ts
+++ b/src/main/account-auth.ts
@@ -23,6 +23,7 @@ import log from 'electron-log';
 import { PtyManager } from './pty-manager';
 import * as accountsRepo from './repositories/accounts';
 import { registerMcpServer, registerHooks } from './mcp-config';
+import { seedLegacyConversations } from './legacy-claude-seed';
 import { ClaudeAccount } from '../shared/types';
 
 interface LoginFlow {
@@ -53,17 +54,24 @@ function configDirFor(accountId: string): string {
  * Begin an interactive login flow for a new account.
  * Returns the new account row + the login pty id the renderer should attach to.
  */
-export function startLoginFlow(
+export async function startLoginFlow(
   ptyManager: PtyManager,
   mainWindow: BrowserWindow | null,
   label: string,
-): StartLoginResult {
+): Promise<StartLoginResult> {
   const accountId = crypto.randomUUID();
   const configDir = configDirFor(accountId);
   fs.mkdirSync(configDir, { recursive: true });
 
-  // First account becomes the default fallback.
+  // First account becomes the default fallback — which re-homes every
+  // unassigned session's CLAUDE_CONFIG_DIR here on its next launch. Seed the
+  // dir with the legacy ~/.claude transcripts first, so those sessions'
+  // stored --resume UUIDs keep resolving instead of coming back "No
+  // conversation found" (the pre-accounts history lives in ~/.claude).
   const isFirst = accountsRepo.getAllAccounts().length === 0;
+  if (isFirst) {
+    await seedLegacyConversations(configDir);
+  }
   const account = accountsRepo.createAccount({
     id: accountId,
     label,
--- a/src/main/account-auth.ts
+++ b/src/main/account-auth.ts
@@ -63,21 +63,29 @@ export async function startLoginFlow(
   const configDir = configDirFor(accountId);
   fs.mkdirSync(configDir, { recursive: true });
 
-  // First account becomes the default fallback — which re-homes every
-  // unassigned session's CLAUDE_CONFIG_DIR here on its next launch. Seed the
-  // dir with the legacy ~/.claude transcripts first, so those sessions'
-  // stored --resume UUIDs keep resolving instead of coming back "No
-  // conversation found" (the pre-accounts history lives in ~/.claude).
+  // First account becomes the default fallback. The check and the insert
+  // stay synchronous — no await between them — so two overlapping
+  // startLoginFlow calls can't both observe "no accounts yet" and both
+  // become the default.
   const isFirst = accountsRepo.getAllAccounts().length === 0;
-  if (isFirst) {
-    await seedLegacyConversations(configDir);
-  }
   const account = accountsRepo.createAccount({
     id: accountId,
     label,
     configDir,
     isDefault: isFirst,
   });
+
+  // Becoming the default re-homes every unassigned session's
+  // CLAUDE_CONFIG_DIR here on its next launch. Seed the dir with the legacy
+  // ~/.claude transcripts so those sessions' stored --resume UUIDs keep
+  // resolving instead of coming back "No conversation found" (pre-accounts
+  // history lives in ~/.claude). Awaited before the login pty spawns: the
+  // account only becomes reachable for session launches once the user logs
+  // in, so finishing the copy first avoids resumes against a half-seeded
+  // dir. Subsequent accounts are separate identities and inherit nothing.
+  if (isFirst) {
+    await seedLegacyConversations(configDir);
+  }
 
   // Register the Bodhilander MCP server + hooks into the new account's
   // isolated config before spawning the login pty, so the session has them
--- a/src/main/__tests__/account-auth.test.ts
+++ b/src/main/__tests__/account-auth.test.ts
@@ -19,7 +19,8 @@ import type { PtyManager } from '../pty-manager';
 let userDataDir = '';
 mock.module('electron', () => ({
   app: { getPath: () => userDataDir },
-  BrowserWindow: class {},
+  // Only referenced in type positions by account-auth; never constructed.
+  BrowserWindow: Object,
 }));
 
 interface AccountRow {
@@ -61,8 +62,8 @@ mock.module('../legacy-claude-seed', () => ({
 }));
 
 // Keep node-pty out of the test process; the flow only calls methods on the
-// instance we pass in.
-mock.module('../pty-manager', () => ({ PtyManager: class {} }));
+// instance we pass in (PtyManager is only referenced in type positions).
+mock.module('../pty-manager', () => ({ PtyManager: Object }));
 
 const accountAuth = await import('../account-auth');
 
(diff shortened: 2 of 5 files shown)


src/main/account-auth.ts
```diff
@@ -53,17 +54,24 @@ function configDirFor(accountId: string): string {
...
 
-  // First account becomes the default fallback.
+  // First account becomes the default fallback — which re-homes every
+  // unassigned session's CLAUDE_CONFIG_DIR here on its next launch. Seed the
+  // dir with the legacy ~/.claude transcripts first, so those sessions'
+  // stored --resume UUIDs keep resolving instead of coming back "No
+  // conversation found" (the pre-accounts history lives in ~/.claude).
   const isFirst = accountsRepo.getAllAccounts().length === 0;
+  if (isFirst) {
+    await seedLegacyConversations(configDir);
```
brannon-bowden: Await introduces a TOCTOU race on `isFirst`: two concurrent startLoginFlow calls can both read isFirst===true before either createAccount runs, seeding legacy transcripts into two separate 'first' accounts. Consider making the first-account determination atomic with account creation.

src/main/account-auth.ts
```diff
@@ -53,17 +54,24 @@ function configDirFor(accountId: string): string {
...
+): Promise<StartLoginResult> {
   const accountId = crypto.randomUUID();
   const configDir = configDirFor(accountId);
   fs.mkdirSync(configDir, { recursive: true });
 
-  // First account becomes the default fallback.
+  // First account becomes the default fallback — which re-homes every
+  // unassigned session's CLAUDE_CONFIG_DIR here on its next launch. Seed the
+  // dir with the legacy ~/.claude transcripts first, so those sessions'
+  // stored --resume UUIDs keep resolving instead of coming back "No
```
brannon-bowden: No test exercises this isFirst branch of startLoginFlow (only the seedLegacyConversations helper is unit-tested in isolation). Add a test covering: first account triggers seeding, subsequent accounts do not.

brannon-bowden (changes requested):
This PR is well-documented and the seeding logic itself is solid and well-tested, but it introduces a race condition in the account-creation flow that isn't covered by any test, so I'm requesting changes before merge.

## Issues

- **`src/main/account-auth.ts:57-73` (blocking) — converting `startLoginFlow` to `async` widens the "first account" check into a TOCTOU race.** Previously `isFirst = accountsRepo.getAllAccounts().length === 0` was immediately followed by synchronous `createAccount(...)`, so the check-then-act was atomic with respect to JS's single-threaded event loop. Now there's an `await seedLegacyConversations(configDir)` (a real filesystem I/O op, potentially copying a large `projects/` tree) between the read and the write. If `startLoginFlow` is invoked twice before the first call's `createAccount` completes (e.g. a double-click on "Add account", or two concurrent IPC calls), both invocations will read `isFirst === true`, both will seed the legacy transcripts into their own (different) config dirs, and both accounts may end up treated as the "default fallback" account — the exact invariant this code depends on ("First account becomes the default fallback"). This is a genuine regression introduced by this diff, not a pre-existing issue.
- **No test covers the new `account-auth.ts` integration (major).** The new `legacy-claude-seed.test.ts` thoroughly tests `seedLegacyConversations` in isolation, but there's no test exercising `startLoginFlow`'s `isFirst` branch — e.g. that seeding is only triggered for the first account, that subsequent accounts are skipped, or that the async race above doesn't produce two "first" accounts. Given this integration is the actual point of the PR, it should have direct coverage, not just coverage of the helper it calls.
- **Signature change from sync to async isn't reflected in any caller update in this diff (minor).** `startLoginFlow` changed from `StartLoginResult` to `Promise<StartLoginResult>`. No call sites are touched here, so it's unclear whether existing callers already `await`/handle the promise correctly (e.g. via `ipcMain.handle`) or whether this silently breaks a caller that previously used the return value synchronously. Worth confirming before merge.

## Suggestions (non-blocking)

- Consider closing the race by determining "first account" status atomically with account creation (e.g. inside `accountsRepo.createAccount`'s transaction, or via a lock/flag reserved before the `await`) rather than checking-then-later-acting across an `await` boundary.
- For a first-run legacy transcript tree that could be large, copying synchronously in the login flow may add noticeable latency to account creation — consider whether this should be fire-and-forget/backgrounded rather than blocking `startLoginFlow`'s resolution.
- The `shell-launch.test.ts` mock-leak fix (scoping `existsSync` override to `WindowsPowerShell` paths only, with the real function captured pre-mock) is a nice, well-explained fix and looks correct — good catch bundling it in since it was presumably needed to keep the new test file from being polluted by test execution order.

William-Long-II:
All three issues addressed in the latest push:

**TOCTOU race (blocking):** `isFirst` read and `createAccount` insert are now consecutive synchronous statements — no await between check and act — and seeding moved to after the row exists. Overlapping calls sync-serialize on the insert, so exactly one account can observe "no accounts yet". Seeding after creation also means a copy failure can't leave a created-dir-but-no-row state. The half-seeded-dir window this opens is benign: the account only becomes launchable after the user completes login in the pty, and `startLoginFlow` resolves (and the login modal opens) only after the copy finishes.

**Integration coverage (major):** new `account-auth.test.ts` exercises `startLoginFlow` directly with the repo/electron/pty/seed stubbed: first account seeds into its config dir and is default; second account skips seeding and isn't default; and the exact race scenario — two overlapping calls with the seed held open across the await window — produces exactly one default and one seed call.

**Async signature (minor):** the only caller is `index.ts:777` inside `safeHandle`, whose wrapper does `return await handler(...)` for every channel, and the renderer already consumed the result through `ipcRenderer.invoke` (promise-based) — so the sync→async change is absorbed with no call-site edits needed.

On the backgrounding suggestion: kept the copy awaited (post-insert) rather than fire-and-forget — a transcript tree is typically MBs, and finishing before the login modal opens guarantees no session can ever resume against a partial `projects/` dir. Happy to revisit if real-world sizes prove painful.

src/main/account-auth.ts
```diff
@@ -53,16 +54,19 @@ function configDirFor(accountId: string): string {
  * Begin an interactive login flow for a new account.
  * Returns the new account row + the login pty id the renderer should attach to.
  */
-export function startLoginFlow(
+export async function startLoginFlow(
   ptyManager: PtyManager,
```
brannon-bowden, line 58: startLoginFlow is now async (Promise<StartLoginResult>). Please confirm every caller of this function has been updated to await/handle the promise — none are visible in this diff.

brannon-bowden (approved):
This PR looks solid: the legacy-transcript seeding is scoped correctly to the first account, the first-account atomicity claim holds up under inspection, and the new tests (including the overlapping-calls race test) exercise the interesting cases. Approving, with one item worth confirming before/soon after merge and two minor notes.

## Issues

- **Unverified call-site impact of the sync→async signature change** (`src/main/account-auth.ts:58`) — `startLoginFlow` changed from returning `StartLoginResult` to `Promise<StartLoginResult>`. The diff doesn't include any caller (IPC handler, renderer bridge, etc.) being updated to `await`/handle the new promise. If an existing caller does something like `const result = startLoginFlow(...); result.account.id`, this silently breaks (accessing a property on a pending Promise). Please confirm all call sites already await this or are updated in this same change; if they live in a file not shown here, worth double-checking in review.
- CI is green (SonarQube, quality-gate, GitGuardian all passing) — no gating concerns there.

## Suggestions (non-blocking)

- `seedLegacyConversations`'s existsSync(target)-then-cp check (`src/main/legacy-claude-seed.ts:37`) is a classic TOCTOU, but given `configDir` is a freshly created, UUID-unique directory touched only by this one call, the practical risk is negligible — just flagging for awareness, not asking for a fix.
- Because seeding is gated purely on `isFirst` (account count), a crash/interruption mid-`cp` (`src/main/account-auth.ts:87`) leaves the first account without transcripts permanently — there's no retry path since `isFirst` will never be true again. This matches the documented "best-effort, never blocks account creation" intent, so it's likely an accepted tradeoff, but worth a one-line callout in the PR description if not already considered.
- Awaiting the copy before the login pty spawns means a large `~/.claude/projects` tree could noticeably delay the first-account login UI. Also plausibly an accepted tradeoff (comment explains the ordering rationale), just noting the UX cost.

PR #114, opened 2026-07-15
ci: bump release.yml actions to Node 24-native majors

William-Long-II:
Closes #113.

Bumps every deprecated-runtime action in the Build and Release workflow to its current Node 24-native major, **pinned by commit SHA** (the `sonarqube.yml` pattern) since this workflow handles release-signing secrets:

| action | before | after | release notes for skipped majors |
|---|---|---|---|
| actions/checkout | v4 | `9c091bb` v7.0.0 | [v5](https://github.com/actions/checkout/releases/tag/v5.0.0) (node24 preliminary), [v6](https://github.com/actions/checkout/releases/tag/v6.0.0), [v7](https://github.com/actions/checkout/releases/tag/v7.0.0) (node24 default; same SHA sonarqube.yml already pins) |
| actions/setup-node | v4 | `8207627` v7.0.0 | [v5](https://github.com/actions/setup-node/releases/tag/v5.0.0), [v6](https://github.com/actions/setup-node/releases/tag/v6.0.0), [v7](https://github.com/actions/setup-node/releases/tag/v7.0.0) — `node-version` input unchanged |
| actions/setup-python | v5 | `ece7cb0` v6.3.0 | [v6](https://github.com/actions/setup-python/releases/tag/v6.0.0) (node24) — `python-version` input unchanged |
| actions/upload-artifact | v4 | `043fb46` v7.0.1 | [v5](https://github.com/actions/upload-artifact/releases/tag/v5.0.0) (node24 prelim), [v6](https://github.com/actions/upload-artifact/releases/tag/v6.0.0) (node24 default), [v7](https://github.com/actions/upload-artifact/releases/tag/v7.0.0) (adds opt-in `archive: false` direct uploads — we don't use it) |
| actions/download-artifact | v4 | `3e5f45b` v8.0.1 | [v5](https://github.com/actions/download-artifact/releases/tag/v5.0.0) (⚠ path change for single-artifact-**by-ID** downloads — we download **all** artifacts, per-artifact subdirs unchanged), [v6](https://github.com/actions/download-artifact/releases/tag/v6.0.0), [v7](https://github.com/actions/download-artifact/releases/tag/v7.0.0) (node24), [v8](https://github.com/actions/download-artifact/releases/tag/v8.0.0) (ESM internals; hash mismatch now errors — a safety improvement) |
[...]

PR #117, opened 2026-07-15
ci: ignore private org-standards repo in Dependabot actions updates

William-Long-II:
The first Dependabot run after #114 landed ([run 29432672408](https://github.com/Software-Development-LLC/Bodhilander/actions/runs/29432672408)) failed with two `git_dependencies_not_reachable` errors: `jira-key-prefix.yml` and `pr-title.yml` call reusable workflows in the **private** `Software-Development-LLC/org-standards` repo, which Dependabot can't reach.

Adding an `ignore` for that dependency lets the weekly github-actions job proceed for everything it *can* update (the SHA-pinned actions in `release.yml`/`sonarqube.yml`). Versioning of the org-standards reusable workflows is managed upstream in that repo.

Alternative considered: granting Dependabot access to private repos in org settings (Settings → Code security → "Allow Dependabot access to private repositories") — that's an org-admin UI toggle and would also work; the ignore is the narrower fix and doesn't require broadening Dependabot's private-repo access. Happy to switch approaches if you prefer the org-settings route.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,10 @@ updates:
       actions:
         patterns:
           - '*'
+    ignore:
+      # Private org repo hosting the reusable workflows (jira-key-prefix,
+      # pr-title). Dependabot has no access to it — without this ignore the
+      # whole github-actions job fails with git_dependencies_not_reachable.
+      # Versioning of those workflows is managed upstream in org-standards.
+      - dependency-name: Software-Development-LLC/org-standards
+


brannon-bowden (approved):
I'll proceed with a direct review based on my own analysis of the diff.

Verification of the change: the `ignore:` block is added at the same indentation level as `groups:` (4 spaces), making it a correct sibling key under the `github-actions` update entry — not nested incorrectly under `groups`/`actions`. For the `github-actions` ecosystem, Dependabot matches `dependency-name` against the `owner/repo` form used in `uses:` references, so `Software-Development-LLC/org-standards` is the correct format. Omitting `versions:` means all versions are ignored, which is what's needed since Dependabot can't reach the private repo at all. This matches GitHub's documented workaround for `git_dependencies_not_reachable` failures. The only nit is a stray trailing blank line at the end of the file, which is harmless.




Straightforward, correctly-scoped fix — approving.

This adds a scoped `ignore` rule to `.github/dependabot.yml` to stop Dependabot from trying (and failing) to resolve the private `org-standards` repo when updating GitHub Actions dependencies.

- The `ignore:` key is placed at the correct indentation (sibling to `groups:`, both under the `github-actions` update entry), so it isn't nested incorrectly.
- `dependency-name: Software-Development-LLC/org-standards` uses the right `owner/repo` matching format for the `github-actions` ecosystem, and omitting `versions:` ignores all versions — appropriate here since Dependabot has no access to the repo at all, not just specific tags.
- This is the standard, documented fix for `git_dependencies_not_reachable` failures caused by private reusable-workflow repos, and the comment clearly explains the rationale for future maintainers.

Suggestions (non-blocking):
- There's a stray trailing blank line added at the end of the file (`.github/dependabot.yml:18`); not harmful, but could be trimmed for tidiness.

PR #118, opened 2026-07-16
fix(arena): unblock codex, settle claude/grok headless CLI quirks

William-Long-II:
## Problem

A real folder-scoped arena run ("suggest 3 improvements for this codebase") failed three different ways, one per contestant:

- **claude** streamed a 9.5k-char answer, then the column flagged an **error**
- **codex** never produced a single byte (and it turns out it never has in arena)
- **grok** printed one "I'll look into it…" sentence and exited "successfully"

Each cause was reproduced live outside the app before fixing.

## Root causes & fixes

**1. Engine left the contestant's stdin pipe open** → `codex exec` reads stdin to EOF when it isn't a TTY ("Reading additional input from stdin…") and blocked forever before doing anything; claude emitted a "no stdin data received in 3s" stderr warning that then surfaced as the column's error text. Fix: `child.stdin?.end()` at spawn — the prompt travels via `ARENA_PROMPT`, never stdin. Same fix as the provider-detector probe (#111). Verified: codex answers in ~4s with stdin closed, hangs indefinitely without.

**2. claude arena command used `--max-turns 1`** → any scoped question needs tool calls, so the CLI exits 1 with result `subtype: error_max_turns` even after streaming a complete answer. Fix: drop the cap (the engine's 5-minute contestant timeout still bounds runaways). Verified: same prompt exits 0 / `subtype: success`, and the `result` event still carries the usage/cost/TTFT metrics the parser reads.

**3. grok `-p` is hard single-turn** — it prints the model's *first* message and exits 0. With default plan mode, a codebase question ends after the preamble with tools never run. Fix: `grok --no-plan --permission-mode auto -p …`. Verified live with the exact failing prompt: full multi-section answer. (`--max-turns 15` and a `--rules` append were tried first; neither helps.)

**4. Hardening: startup sweep for orphaned rows.** The failing run's codex row is stuck `running` in the DB forever because the app was restarted before the contestant timeout fired — finalize never runs for processes that don't survive a restart. New `settleInterruptedResponses()` marks leftover `running` rows as errored ("Interrupted by app restart") on startup, mirroring `markAllSessionsStopped()`.

## Tests

- New engine test: a `cat`-based fake contestant (reads stdin to EOF, exactly like codex) must complete instead of hanging.
- New `repositories/__tests__/arena.test.ts`: sweep settles running rows, leaves finalized rows untouched, no-ops when empty.
- Full suite: 130 pass / 1 pre-existing failure (chat-parser `06-response` snapshot, fails identically on `development`). `tsc` clean apart from the two known pre-existing `mdns-advertiser` errors.

### DIFF
--- a/src/main/providers/grok.ts
+++ b/src/main/providers/grok.ts
@@ -19,7 +19,12 @@ export const grokProvider = passthroughProvider({
   },
   arena: {
     // Plain-text headless output; no documented JSON/usage reporting yet.
-    buildCommand: (promptRef) => `grok -p ${promptRef}`,
+    // `-p` is hard single-turn: it prints the model's first message and
+    // exits 0, so with default plan mode on, a codebase question ends after
+    // an "I'll explore..." preamble with the tools never run. --no-plan +
+    // auto permissions let the model finish its exploration and answer
+    // within that single turn (verified live; --max-turns does not help).
+    buildCommand: (promptRef) => `grok --no-plan --permission-mode auto -p ${promptRef}`,
     createParser: textParser,
   },
   setup: {
--- a/src/main/arena/__tests__/engine.test.ts
+++ b/src/main/arena/__tests__/engine.test.ts
@@ -72,6 +72,15 @@ const FAKE_PROVIDERS: Record<string, any> = {
       createParser: textParser,
     },
   },
+  stdinReader: {
+    id: 'stdinReader',
+    arena: {
+      // `cat` reads stdin to EOF — exactly what codex exec does headlessly.
+      // If the engine leaves the stdin pipe open this never exits.
+      buildCommand: () => 'cat; echo after-stdin-eof',
+      createParser: textParser,
+    },
+  },
 };
 mock.module('../../providers', () => ({
   getProvider: (id: string) => {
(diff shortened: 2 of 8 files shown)


src/main/providers/grok.ts
```diff
@@ -19,7 +19,12 @@ export const grokProvider = passthroughProvider({
   },
   arena: {
     // Plain-text headless output; no documented JSON/usage reporting yet.
-    buildCommand: (promptRef) => `grok -p ${promptRef}`,
+    // `-p` is hard single-turn: it prints the model's first message and
+    // exits 0, so with default plan mode on, a codebase question ends after
+    // an "I'll explore..." preamble with the tools never run. --no-plan +
+    // auto permissions let the model finish its exploration and answer
+    // within that single turn (verified live; --max-turns does not help).
```
brannon-bowden, line 26: --permission-mode auto drops the tool-call confirmation gate for grok's headless run. Please confirm the intended scope (read-only exploration vs. writes/exec) and consider a regression test asserting this flag stays in place.

brannon-bowden (approved):
CI is green (SonarQube, quality-gate, GitGuardian all passing). This is a solid, well-tested fix for the stdin-hang and orphaned-response problems, with one design decision worth an explicit second look before it ships.

## Issues

- **`src/main/providers/grok.ts:26`** — `--permission-mode auto` removes the CLI's tool-call confirmation gate for grok, letting it execute filesystem/tool actions in the arena's working directory unattended. That's a meaningful guardrail to drop, and it's asymmetric with `claude.ts` (which only loses `--max-turns 1`, not an explicit permission bypass). The tradeoff is documented as "verified live," but there's no test locking in the intended scope, and the PR doesn't say what "auto" actually authorizes (read-only exploration vs. writes/exec). Worth an explicit confirmation from the author that this is bounded appropriately for the arena's threat model before merge.
- **`src/main/providers/claude.ts:53` / `src/main/providers/grok.ts:26`** — no unit test pins the exact `buildCommand` output for either provider. Given how hard-won this flag knowledge clearly was (turn-cap exit codes, plan-mode preambles), a cheap string-assertion test would catch a future accidental revert that only shows up again via live CLI testing.
- **`src/main/arena/engine.ts:190`** — `child.stdin?.end()` is unconditional for every arena contestant, not just the ones known to need it. Fine today since no provider writes the prompt via stdin, but it'll silently break any future provider that does need interactive stdin, and there's no visibility here into whether `child.stdin` has an `'error'` listener attached elsewhere to avoid an unhandled-error crash on an unexpected EPIPE.

## Suggestions (non-blocking)

- Consider a short comment on `settleInterruptedResponses` (or at the call site) noting it intentionally sweeps *all* runs, not just the most recent — it's inferable from the code but was implicit for `markAllSessionsStopped()` too, so a one-liner would keep the two startup sweeps easy to compare.
- The new `stdinReader` engine test and the `arena.test.ts` repository suite are good, focused additions — nice use of `bun:sqlite` to avoid a native build dependency for the repo tests.

PR #119, opened 2026-07-16
feat(arena): resumable runs with follow-up rounds

William-Long-II:
## What

Arena runs become side-by-side **conversations**. After a run settles, a reply box appears; sending a follow-up resumes every contestant's own CLI session with full context and streams a new round into each column. This directly answers "what if a bot needs input / asks a question" — you answer it, and every agent continues where it left off. Follow-ups also work on history-loaded runs, since CLI sessions persist on disk.

## Per-CLI resume mechanics — each verified live (codeword recall test)

| contestant | round 0 | follow-up |
|---|---|---|
| claude | `--session-id <uuid>` names the session upfront | `--resume <uuid>` (headless resume keeps the same id — verified) |
| codex | mints its own thread id; parser captures `thread.started` | `codex exec resume <thread_id>` |
| grok | `--session-id <uuid>` | `--resume <uuid>` |
| ollama | plain chat call | no server-side session — prior rounds replayed as explicit message history |

Bonus fix discovered while verifying: codex refuses to start outside a trusted/git directory ("Not inside a trusted directory") — so **unscoped** arena runs (cwd = app dir) would have kept failing even after #118. `--skip-git-repo-check` is now on both codex commands.

## How

- **Schema** (additive migration): `arena_responses` gains `round`, `prompt`, `session_ref`. Round 0 answers the run's prompt; later rounds carry their own. `session_ref` is persisted at finalize, so resumability survives app restarts; the #118 startup sweep marks interrupted rows errored, and errored columns simply don't continue.
- **Engine**: `prepareFollowUp(runId, prompt)` mirrors `prepare()`'s two-phase contract (rows created first so the renderer can subscribe, `launch()` spawns). It picks, per provider, the latest round; only `done` rounds with something to resume continue. Session refs are validated against a safe-charset pattern before being inlined into resume commands (prompt text itself still travels via `ARENA_PROMPT`, never the command string).
- **Providers**: arena contract becomes `buildCommand(promptRef, sessionRef)` + optional `buildResumeCommand(promptRef, sessionRef)`; parsers can report the CLI's own session ref (`ArenaFinal.sessionRef`), which wins over the engine-assigned one (codex).
- **Renderer**: columns are now per-provider with round-stacked blocks (`arenaRounds.ts`, pure + unit-tested); follow-up prompts render as `↳ prompt` separators with per-round metrics; reply bar shows only when the run is settled and at least one column can continue.

## Tests

- Engine: full follow-up flow over a fake resumable CLI (resume command runs against the persisted ref, new prompt travels via env, ref carries forward); refusal when the only contestant errored or has no resume command; Ollama history-replay body assertion.
- Parsers: claude `session_id` and codex `thread_id` capture.
- Repo: round/prompt/session_ref round-trip + round ordering.
- Renderer: `buildColumns` grouping, `canFollowUp` gating (latest-round semantics).
- Full suite: 141 pass / 1 pre-existing failure (chat-parser `06-response` snapshot, fails identically on `development`); `tsc` clean apart from the two known pre-existing `mdns-advertiser` errors.

### DIFF
--- a/src/main/arena/engine.ts
+++ b/src/main/arena/engine.ts
@@ -31,7 +31,7 @@ import * as arenaRepo from '../repositories/arena';
 import { vaultEnvFor } from '../key-vault';
 import { redactEnv } from '../redact-env';
 import { ArenaStreamParser, ollamaParser } from './parsers';
-import { ArenaRun, ArenaUpdate, ArenaResponseStatus } from '../../shared/types';
+import { ArenaResponse, ArenaRun, ArenaUpdate, ArenaResponseStatus } from '../../shared/types';
 
 export const OLLAMA_CONTESTANT_ID = 'ollama';
 const OLLAMA_BASE_URL = 'http://127.0.0.1:11434';
@@ -114,6 +114,44 @@ interface LiveResponse {
  */
 const SESSION_REF_SAFE = /^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$/;
 
+/**
+ * Continuation seed for one provider's next round, or null when the column
+ * can't continue (latest round didn't finish cleanly, nothing to resume, or
+ * the CLI has no headless resume). `rounds` is that provider's responses in
+ * round order.
+ */
+function followUpSeed(
+  provider: string,
+  rounds: ArenaResponse[],
+  runPrompt: string,
+  prompt: string,
+): Pick<LiveResponse, 'sessionRef' | 'ollamaMessages'> | null {
+  const latest = rounds[rounds.length - 1];
+  if (latest.status !== 'done') return null;
+
+  if (provider === OLLAMA_CONTESTANT_ID) {
+    if (!latest.text) return null;
+    // Ollama has no server-side session — replay every prior round as
+    // explicit chat history.
+    const messages: OllamaMessage[] = rounds.flatMap((r) => [
+      { role: 'user' as const, content: r.prompt ?? runPrompt },
+      { role: 'assistant' as const, content: r.text },
+    ]);
+    messages.push({ role: 'user', content: prompt });
+    return { sessionRef: null, ollamaMessages: messages };
+  }
+
+  if (!latest.sessionRef || !SESSION_REF_SAFE.test(latest.sessionRef)) return null;
+  let providerDef: ReturnType<typeof getProvider>;
+  try {
+    providerDef = getProvider(provider);
+  } catch {
+    return null; // provider removed from the registry since the run
+  }
+  if (!providerDef.arena.buildResumeCommand) return null;
+  return { sessionRef: latest.sessionRef, ollamaMessages: null };
+}
+
 export class ArenaEngine extends EventEmitter {
   private readonly live: Map<string, LiveResponse> = new Map();
   private readonly timeoutMs: number;
@@ -178,30 +216,8 @@ export class ArenaEngine extends EventEmitter {
 
     for (const provider of new Set(run.responses.map((r) => r.provider))) {
       const rounds = run.responses.filter((r) => r.provider === provider);
-      const latest = rounds[rounds.length - 1]; // responses are round-ordered
-      if (latest.status !== 'done') continue;
-
-      let entrySeed: Pick<LiveResponse, 'sessionRef' | 'ollamaMessages'>;
-      if (provider === OLLAMA_CONTESTANT_ID) {
-        if (!latest.text) continue;
-        // Replay every prior round as explicit chat history.
-        const messages: OllamaMessage[] = rounds.flatMap((r) => [
-          { role: 'user' as const, content: r.prompt ?? run.prompt },
-          { role: 'assistant' as const, content: r.text },
-        ]);
-        messages.push({ role: 'user', content: prompt });
-        entrySeed = { sessionRef: null, ollamaMessages: messages };
-      } else {
-        if (!latest.sessionRef || !SESSION_REF_SAFE.test(latest.sessionRef)) continue;
-        let providerDef: ReturnType<typeof getProvider>;
-        try {
-          providerDef = getProvider(provider);
-        } catch {
-          continue; // provider removed from the registry since the run
-        }
-        if (!providerDef.arena.buildResumeCommand) continue;
-        entrySeed = { sessionRef: latest.sessionRef, ollamaMessages: null };
-      }
+      const entrySeed = followUpSeed(provider, rounds, run.prompt, prompt);
+      if (!entrySeed) continue;
 
       const responseId = randomUUID();
       arenaRepo.createResponse(responseId, runId, provider, nextRound, prompt);
--- a/src/main/arena/engine.ts
+++ b/src/main/arena/engine.ts
@@ -78,18 +78,41 @@ function killTree(child: ChildProcess): void {
   child.once('close', () => clearTimeout(escalation));
 }
 
+interface OllamaMessage {
+  role: 'user' | 'assistant';
+  content: string;
+}
+
 interface LiveResponse {
   runId: string;
   provider: string;
   prompt: string;
   /** Project folder CLI contestants spawn in (null = engine process cwd). */
   workingDir: string | null;
+  /** Conversation round: 0 = initial prompt, 1+ = follow-ups. */
+  round: number;
+  /** CLI session id: engine-assigned on round 0, resumed on later rounds. */
+  sessionRef: string | null;
+  /** True when this entry resumes an earlier round's session. */
+  resume: boolean;
+  /**
+   * Ollama has no CLI session to resume — follow-ups replay the prior
+   * conversation as an explicit message history (null for CLI contestants).
+   */
+  ollamaMessages: OllamaMessage[] | null;
   text: string;
   startedAt: number;
   ttftMs: number | null;
   launched: boolean;
   kill: (() => void) | null;
 }
+
+/**
+ * Session refs are inlined into shell command strings (--resume <ref>), so
+ * only pass ones that cannot break out of the command: UUIDs and codex
+ * thread ids match this; anything else is rejected rather than spawned.
+ */
+const SESSION_REF_SAFE = /^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$/;
 
 export class ArenaEngine extends EventEmitter {
   private readonly live: Map<string, LiveResponse> = new Map();
@@ -116,6 +139,13 @@ export class ArenaEngine extends EventEmitter {
         provider,
         prompt,
         workingDir,
+        round: 0,
+        // Name the CLI session upfront so the run can be resumed later.
+        // CLIs that mint their own id (codex) ignore this; their parser
+        // reports the real one, which wins at finalize.
+        sessionRef: provider === OLLAMA_CONTESTANT_ID ? null : randomUUID(),
+        resume: false,
+        ollamaMessages: null,
         text: '',
         startedAt: Date.now(),
         ttftMs: null,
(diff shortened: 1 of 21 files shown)


src/renderer/components/ArenaPanel.tsx
```diff
@@ -122,13 +124,33 @@ export const ArenaPanel: React.FC<ArenaPanelProps> = ({ onClose, groups, initial
...
+    if (!run || !trimmed || running) return;
+    setRunning(true);
+    try {
+      const updated = await window.electronAPI.arenaFollowUp(run.id, trimmed);
+      runIdRef.current = updated.id;
+      setRun(updated);
+      setReply('');
+      await window.electronAPI.arenaLaunch(updated.id);
+    } catch (error) {
+      console.error('Arena follow-up failed:', error);
```
brannon-bowden, line 147: Follow-up failures are only logged to the console; the user gets no visible feedback when arenaFollowUp/arenaLaunch throws (e.g. when canFollowUp's UI heuristic disagrees with the engine's authoritative resumability check). Consider surfacing an inline error state.

src/main/index.ts
```diff
@@ -1247,6 +1247,14 @@ safeHandle('arena:start', async (prompt: string, contestants: string[], workingD
 safeHandle('arena:launch', async (runId: string) => {
   arenaEngine.launch(runId);
 });
+// Follow-up round: same two-phase contract — 'followUp' creates the new
+// round's rows (resuming each contestant's CLI session), 'launch' spawns.
+safeHandle('arena:followUp', async (runId: string, prompt: string) => {
+  if (!runId || !prompt?.trim()) {
+    throw new Error('Arena follow-up needs a run id and a prompt');
+  }
+  return arenaEngine.prepareFollowUp(runId, prompt);
```
brannon-bowden: prompt is validated via prompt?.trim() but the untrimmed prompt is forwarded to prepareFollowUp, so whitespace can be persisted/rendered. Pass the trimmed value instead.

brannon-bowden (changes requested):
This is a well-built feature: the follow-up flow is genuinely two-phase (prepare/launch) consistent with the existing initial-run contract, session refs are validated with a strict allowlist regex before being inlined into shell command strings (good defense against command injection via a compromised/buggy CLI's reported session id), and the round-replay/session-carry-forward logic is exercised by solid unit tests across the engine, parsers, repository, and the new `arenaRounds` helper. I don't see correctness bugs in the core resumption logic (round bookkeeping, Ollama history replay, session-ref fallback at finalize).

**Issues**

- CI is currently red — SonarQube Code Analysis and quality-gate are both failing (GitGuardian passes). Please resolve those before merge; I can't approve while these checks are failing.
- `src/renderer/components/ArenaPanel.tsx:141` — `sendFollowUp`'s catch block only does `console.error` and resets `running`; the user gets no visible indication that their follow-up failed (and the typed reply is not restored, since `setReply('')` only runs on the success path before this point is reached — actually reply is preserved on failure, but there's still no on-screen error). This matters here because `canFollowUp` (in `arenaRounds.ts`) is a UI-side heuristic that doesn't fully mirror the engine's authoritative checks (e.g. it doesn't verify `buildResumeCommand` exists or that the provider is still registered) — so a user can hit "Reply", have `arenaFollowUp` throw ("No contestant in this run can be resumed" / "Arena run still has contestants running"), and see nothing happen. Given this is the PR's headline feature, silently swallowing that failure is worth surfacing to the user (even a minimal inline error string).

- `src/main/index.ts:1256` — `arena:followUp` validates `prompt?.trim()` but forwards the untrimmed `prompt` to `prepareFollowUp`, so leading/trailing whitespace can end up persisted in the `prompt` column and later rendered (`↳ {r.prompt}` in `ArenaPanel.tsx`). Passing the trimmed value would be more consistent with the validation.
- Consider a test (or at least a follow-up ticket) for the `ArenaPanel` UI error path itself (`sendFollowUp` failure), since the pure logic in `arenaRounds.ts` is well covered but the interactive component behavior isn't.

src/renderer/components/ArenaPanel.tsx
```diff
@@ -122,13 +124,33 @@ export const ArenaPanel: React.FC<ArenaPanelProps> = ({ onClose, groups, initial
...
+    const trimmed = reply.trim();
+    if (!run || !trimmed || running) return;
+    setRunning(true);
+    try {
+      const updated = await window.electronAPI.arenaFollowUp(run.id, trimmed);
+      runIdRef.current = updated.id;
+      setRun(updated);
+      setReply('');
+      await window.electronAPI.arenaLaunch(updated.id);
+    } catch (error) {
```
brannon-bowden, line 143: Follow-up failures are only logged to the console; the user gets no indication the Reply action failed and no path to see why.

brannon-bowden (approved):
This PR adds resumable follow-up rounds to the Arena feature (per-provider session resume for CLI contestants, message-history replay for Ollama), with solid test coverage across the engine, parsers, repository, and the new UI round-grouping logic. No blocking correctness or security issues found — session refs are validated with a strict allow-list regex before being inlined into shell commands, and the prompt itself still travels via environment variables rather than the command string. A few gaps are worth addressing.

## Issues

- **Silent failure on follow-up errors** (`src/renderer/components/ArenaPanel.tsx:140`): `sendFollowUp`'s catch block only does `console.error` and resets `running` — there's no user-visible error surfaced. Any rejection from `arenaFollowUp` (engine throws like "No contestant in this run can be resumed" or "Arena run still has contestants running", or any IPC failure) leaves the user staring at a Reply box that silently re-enables with no explanation of what went wrong.
- **`canFollowUp` doesn't verify resume capability** (`src/renderer/components/arenaRounds.ts:45`): it only checks that the latest round is `done` and has a `sessionRef` or non-empty `text`. Since the engine assigns a `sessionRef` to *every* non-Ollama contestant up front (`engine.ts` `prepare()`), a CLI provider that finishes successfully but has no `buildResumeCommand` will still pass this check and show the Reply UI — only for `prepareFollowUp` to then reject the whole request server-side. Combined with the silent-catch issue above, this is a dead-end UX path. Not triggered by the three providers touched in this PR (all now have `buildResumeCommand`), but it's a trap for the next provider that doesn't.
- **Codex's fallback session ref could be bogus** (`src/main/arena/engine.ts:453`): `sessionRef = final?.sessionRef ?? entry.sessionRef` falls back to the engine-generated UUID assigned in `prepare()`. For Claude/Grok that UUID is legitimate (it's what was actually passed via `--session-id`), but Codex ignores the upfront id entirely and mints its own via `thread.started`. If the parser ever fails to capture `thread.started` (format change, unexpected output ordering) on a response that still finalizes as `done`, the persisted `sessionRef` would be a UUID Codex never actually used, and a later `codex exec resume <bogus-id>` would fail. Low likelihood given `thread.started` should always precede a successful turn, but worth guarding (e.g., Codex-specific entries could seed `sessionRef: null` instead of a throwaway UUID) rather than relying on it silently degrading.

## Suggestions

- The `arena_responses` schema migration in `database.ts` adds three columns (`round`, `prompt`, `session_ref`) gated by a single existence check on `round`. If the process were killed between the three `ALTER TABLE` calls, the guard would see `round` present and skip re-running the rest, permanently leaving `prompt`/`session_ref` missing. Non-blocking (mirrors the existing `working_dir` migration pattern), but a single multi-column `ALTER` batch or an explicit check per column would be more robust.

William-Long-II:
Addressed in 1ccd4b2:

- **Silent follow-up failures** — failures now render inline above the reply bar. `followUpErrorMessage` (in `arenaRounds.ts`, unit-tested) strips the `Error invoking remote method …` IPC wrapper so the user sees the engine's actual detail (e.g. "No contestant in this run can be resumed"); the typed reply is preserved for retry. Error state clears on the next attempt, a new run, or loading a history run.
- **Untrimmed prompt** — `arena:followUp` now forwards the trimmed value it validated.
- **UI error-path test** — the message-shaping logic is covered by unit tests; the repo has no component-render test setup (all renderer tests are pure-function), so I kept the React handler thin over tested helpers rather than introducing a render-testing dependency in this PR.

(The CI failures from the first review pass were the three SonarQube findings — fixed in d14697c; all checks are green.)

src/main/arena/engine.ts
```diff
@@ -78,19 +78,80 @@ function killTree(child: ChildProcess): void {
...
 }
 
+/**
+ * Session refs are inlined into shell command strings (--resume <ref>), so
+ * only pass ones that cannot break out of the command: UUIDs and codex
+ * thread ids match this; anything else is rejected rather than spawned.
+ */
+const SESSION_REF_SAFE = /^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$/;
+
+/**
```
brannon-bowden, line 117: SESSION_REF_SAFE guards against shell command injection via untrusted CLI-reported session/thread ids, but no test exercises the rejection path (malformed/malicious ref). Add a regression test before merge.

src/main/providers/codex.ts
```diff
@@ -19,8 +19,15 @@ export const codexProvider = passthroughProvider({
...
   arena: {
-    // JSONL events; turn.completed carries token usage (OpenAI headless docs).
-    buildCommand: (promptRef) => `codex exec --json ${promptRef}`,
+    // JSONL events; turn.completed carries token usage (OpenAI headless
+    // docs). --skip-git-repo-check: codex refuses to run outside a trusted
+    // git directory otherwise, which breaks unscoped arena runs (cwd = the
+    // app dir). Codex mints its own thread id — the engine's assigned
+    // sessionRef is ignored and the parser captures thread.started instead;
+    // follow-up rounds resume that thread (verified live).
+    buildCommand: (promptRef) => `codex exec --json --skip-git-repo-check ${promptRef}`,
```
brannon-bowden, line 28: --skip-git-repo-check now applies to every codex arena run (not just follow-ups), bypassing codex's own trust guard for untrusted directories. Please confirm this is intentional and scoped correctly.

src/renderer/components/arenaRounds.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export function followUpErrorMessage(error: unknown): string {
+  const raw = error instanceof Error ? error.message : String(error);
+  return raw.replace(/^Error invoking remote method '[^']+':\s*(Error:\s*)?/, '') || 'Follow-up failed';
+}
+
+export function canFollowUp(run: ArenaRun): boolean {
+  if (run.responses.some((r) => r.status === 'running')) return false;
+  return buildColumns(run).some(
+    ({ latest }) => latest.status === 'done' && (latest.sessionRef !== null || latest.text.length > 0)
```
brannon-bowden, line 55: canFollowUp doesn't check for buildResumeCommand support like the engine's followUpSeed does, so a run whose only contestant lacks resume support will always show a Reply box that fails on every attempt.

src/main/database.ts
```diff
@@ -472,6 +472,16 @@ function initializeArenaTables(database: Database.Database): void {
   if (!runColumns.some((col) => col.name === 'working_dir')) {
     database.exec('ALTER TABLE arena_runs ADD COLUMN working_dir TEXT DEFAULT NULL');
   }
+
+  // Migration: follow-up rounds. round 0 = the run's initial prompt; later
+  // rounds carry their own prompt. session_ref stores the CLI session/thread
+  // id the response can be resumed from (NULL = column can't continue).
+  const responseColumns = database.prepare('PRAGMA table_info(arena_responses)').all() as { name: string }[];
+  if (!responseColumns.some((col) => col.name === 'round')) {
```
brannon-bowden, line 480: The three ADD COLUMN statements are gated by one presence check; a crash mid-migration leaves the table permanently missing prompt/session_ref. Consider per-column guards or a transaction.

src/renderer/components/arenaRounds.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+  return Array.from(byProvider, ([provider, responses]) => ({
+    provider,
+    responses,
+    latest: responses[responses.length - 1],
+  }));
+}
+
+/**
```
brannon-bowden, line 36: This doc comment describes canFollowUp but sits above followUpErrorMessage — looks misplaced.

brannon-bowden (changes requested):
This PR adds resumable arena runs (follow-up rounds) with per-provider session tracking. The core design is solid — the two-phase prepare/launch contract is preserved, prompts still travel via environment variables (not inlined into shell commands), and CLI-reported session/thread ids are validated against a strict allowlist regex (`SESSION_REF_SAFE`) before being inlined into resume commands, which is exactly the right mitigation for untrusted subprocess output. CI is green across all three checks. A few gaps should be addressed before merge.

## Issues

- **Missing test for the command-injection guard.** `SESSION_REF_SAFE` (`src/main/arena/engine.ts:117`) is the only thing standing between a CLI's self-reported session/thread id (untrusted subprocess output, e.g. codex `thread.started.thread_id`) and a shell command string built via `--resume <ref>`. There's no test anywhere in the diff that feeds a malicious/malformed ref (shell metacharacters, oversized string, empty string) through `finalize()`/`followUpSeed()`/`runCli()` and asserts it's rejected rather than spawned. This is exactly the kind of non-trivial, security-relevant logic that needs a regression test so a future refactor can't silently drop the check.

- **`--skip-git-repo-check` added to codex's `buildCommand`** (`src/main/providers/codex.ts:28`), applying to *every* codex arena run, not just follow-up rounds. This bypasses codex's own git-repo trust guard, which normally stops the CLI from running (and potentially taking agentic actions) in a directory the user hasn't vetted. The comment justifies it as needed for "unscoped" runs (cwd = app dir), but it's bundled quietly into a feature PR about resumable rounds and applies unconditionally, including to scoped runs against arbitrary project folders. Worth an explicit call-out/sign-off that this trust bypass is intentional and scoped correctly, since it broadens the blast radius of a PR that's ostensibly about follow-up rounds.

- **`canFollowUp` (UI) and `followUpSeed` (engine) disagree on when a provider can continue** (`src/renderer/components/arenaRounds.ts:55` vs `src/main/arena/engine.ts` `followUpSeed`). The UI only checks `status === 'done' && (sessionRef !== null || text.length > 0)`, but the engine additionally requires the provider to still be registered and to implement `buildResumeCommand`. For a run whose sole contestant is a CLI provider without `buildResumeCommand` (a scenario the new engine test `'follow-up skips providers without a resume command'` explicitly exercises), the UI will still render a working-looking "Reply" box that fails on every attempt. Error handling surfaces a message, so it's not silent, but it's a reproducible dead-end affordance that should be reconciled (e.g. surface resumability per-provider from the backend, or have `canFollowUp` account for missing `buildResumeCommand`).

## Suggestions (non-blocking)

- `src/main/database.ts:480-484`: the three `ALTER TABLE` statements for `round`/`prompt`/`session_ref` are gated by a single check (presence of `round`). If the process crashes between the first and second statement, future startups will see `round` already present and skip adding `prompt`/`session_ref` forever, breaking `createResponse`'s INSERT. Consider gating each column independently (as `working_dir` does) or wrapping the block in a transaction.
- `src/renderer/components/arenaRounds.ts:36-41`: the doc comment describing `canFollowUp` is placed directly above `followUpErrorMessage` instead of above `canFollowUp`. Worth moving so it doesn't mislead readers.
- The in-memory `repositories/arena` mock (round/prompt/session_ref bookkeeping) is duplicated almost verbatim between `engine.test.ts` and `engine-ollama.test.ts`. Could be extracted to a shared test helper to avoid drift between the two copies.

PR #121, opened 2026-07-17
feat(providers): add opencode, kimi, cursor, and antigravity CLIs

William-Long-II:
## What

Adds four agent CLIs as first-class providers — selectable when starting a session **and** as arena contestants (with folder scoping and, where the CLI allows, follow-up rounds):

| provider | command | headless | resume | metrics |
|---|---|---|---|---|
| opencode | `opencode` | `run --auto --format json` | `-s <id>` | tokens + cost |
| Kimi Code | `kimi` | `-p --output-format stream-json` | `-r <id>` | — |
| Cursor Agent | `cursor-agent` | `-p --output-format stream-json --force --trust` | `--resume <id>` | tokens |
| Antigravity | `agy` | `--dangerously-skip-permissions -p` | — (see below) | — |

Registry membership is all that's needed: the new-session picker, Settings → Providers detection, and arena are all registry/detection-driven, so no per-surface wiring.

## Verified live (not from docs)

Every contract was mapped by running the installed, logged-in CLIs, then **re-verified end-to-end through the real `ArenaEngine`** (shell-launch → CLI with its real permission flags → parser → session-ref capture → follow-up resume), folder-scoped to this repo:

- **Round 0**: all four read `package.json` and returned `bodhilander`.
- **Follow-up**: opencode / kimi / cursor resumed their own session and answered the `version` field (`3.4.0-beta.13`); antigravity was correctly skipped.

Each parser is unit-tested against JSON shapes captured from those live runs.

## Notes / decisions

- **Cursor command is `cursor-agent`, not `agent`.** Both cursor and grok's installers drop a bare `agent` onto PATH (they collide; grok wins by PATH order). Keying off the unambiguous `cursor-agent` symlink avoids the collision and survives Cursor's per-version install dir.
- **Antigravity has no follow-up rounds, on purpose.** Its headless output is plain text with no machine-readable conversation id, a fresh id can't be assigned up front (verified: `--conversation <new-uuid>` starts a fresh conversation instead of adopting the id), and `--continue` only targets the *globally* most-recent conversation — which can't reliably re-enter a specific arena column under concurrent use. Rather than resume the wrong conversation, `buildResumeCommand` is omitted; the engine already skips such providers in later rounds. It still answers the initial arena prompt and works as a normal session provider.
- **Auth:** all four use their own CLI login (opencode's multi-provider auth, kimi/cursor sign-in, antigravity's Google account), so none wire an `apiKey` into the key vault. `ProviderDefinition.apiKey` is already optional.
- **Permission flags** mirror arena's existing auto-approve tradeoff (grok `--permission-mode auto`, cursor `--force`, opencode `--auto`) so folder-scoped questions can actually read files. Antigravity's `--dangerously-skip-permissions` is its own prescribed headless auto-approve (plain `agy -p` auto-denies tool perms and produces no output).

## Tests

- New parser unit tests (opencode/kimi/cursor) with real captured fixtures; antigravity reuses `textParser`.
- resolve.test.ts: the four new ids are known/resolvable, and follow-up resume is asserted present for every provider **except** antigravity.
- Full suite: 150 pass / 1 pre-existing failure (`ChatParser 06-response` snapshot, fails identically on `development`); `tsc` clean apart from the two known pre-existing `mdns-advertiser` errors.

### DIFF
--- a/src/main/arena/__tests__/parsers.test.ts
+++ b/src/main/arena/__tests__/parsers.test.ts
@@ -6,7 +6,15 @@
  * Run with: bun test src/main/arena
  */
 import { describe, expect, test } from 'bun:test';
-import { claudeParser, codexParser, textParser, ollamaParser } from '../parsers';
+import {
+  claudeParser,
+  codexParser,
+  opencodeParser,
+  kimiParser,
+  cursorParser,
+  textParser,
+  ollamaParser,
+} from '../parsers';
 
 describe('claudeParser', () => {
   test('extracts assistant text and result metrics from stream-json', () => {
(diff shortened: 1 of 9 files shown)


brannon-bowden (approved):
This PR cleanly adds four new provider CLIs (opencode, kimi, cursor, antigravity) following the existing `passthroughProvider` pattern, with solid parser unit tests and updated registry tests. All listed CI checks (SonarQube, quality-gate, GitGuardian) are passing. No blocking issues found — approving.

## Suggestions (non-blocking)

- `src/main/arena/parsers.ts:212` — `cursorParser`'s content mapper uses `(p: any) =>` for the map callback where `unknown` plus a type guard would preserve type safety; minor readability/type-strictness nit.
- The new providers each enable an automatic tool-approval flag (`--dangerously-skip-permissions` for antigravity, `--force --trust` for cursor, `--auto` for opencode). The doc comments correctly note this mirrors the existing tradeoff already made for grok/codex, so it's consistent rather than a new risk, but worth keeping in mind as more auto-approving CLIs are added to the arena surface.
- Consider a couple of lightweight tests asserting the exact `buildCommand`/`buildResumeCommand` strings for the four new providers (flags, argument order) — current registry tests only check that these are functions, not their output shape. Not blocking since the logic is simple template interpolation mirroring existing providers.

PR #122, opened 2026-07-17
Release 3.4.0 — add opencode, kimi, cursor, antigravity providers

William-Long-II:
## Release 3.4.0 — provider additions

Second dev→prod pass for the 3.4.0 release. The prior release PR (#120) merged the 3.4.0-beta.13 base to production but landed before the four new agent CLIs merged to development, and it carried a `-beta.` version so the channel gate skipped it (no stable build was produced). This brings production up to development's tip; the stable release is cut afterward with `npm version minor` → **3.4.0**.

### What's new since production

Four agent CLIs added as session providers and arena contestants (#121), each contract mapped live and verified end-to-end through the real arena engine:

- **opencode** — `opencode run --auto --format json`; text parts, `-s` resume, tokens + cost.
- **Kimi Code** (`kimi`) — `-p --output-format stream-json`; resume via `-r`.
- **Cursor Agent** (`cursor-agent`) — `-p --output-format stream-json --force --trust`; Claude-shaped events, `--resume`. Keyed off `cursor-agent` (not the `agent` alias that collides with grok).
- **Antigravity** (`agy`) — `--dangerously-skip-permissions -p`; plain text. Arena round-0 + sessions only (no in-band conversation id → follow-up deliberately unsupported).

All four use their own CLI login; registry membership drives the picker, Settings → Providers, and arena automatically.

### Release steps after merge
On `production`: `npm version minor` (3.4.0-beta.13 → 3.4.0), then `git push` **without** `--tags` — the workflow builds and tags. The merge push itself is inert (channel gate skips the `-beta.` version on production).

### DIFF
--- a/src/main/arena/__tests__/parsers.test.ts
+++ b/src/main/arena/__tests__/parsers.test.ts
@@ -6,7 +6,15 @@
  * Run with: bun test src/main/arena
  */
 import { describe, expect, test } from 'bun:test';
-import { claudeParser, codexParser, textParser, ollamaParser } from '../parsers';
+import {
+  claudeParser,
+  codexParser,
+  opencodeParser,
+  kimiParser,
+  cursorParser,
+  textParser,
+  ollamaParser,
+} from '../parsers';
 
 describe('claudeParser', () => {
   test('extracts assistant text and result metrics from stream-json', () => {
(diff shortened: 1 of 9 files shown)


brannon-bowden (approved):
This PR adds four new arena providers (opencode, kimi, cursor, antigravity) plus matching stream parsers, and extends the provider-registry tests to cover them. The parser logic is well-documented with real fixture shapes, the token/cost-summing logic in `opencodeParser` is correct and tested for both the happy path and the "no metrics reported" path, and the new `antigravity` provider's decision to omit `buildResumeCommand` is sound and is explicitly covered by a registry test. CI is green (build/release are skipped/neutral, which is informational only and doesn't block).

## Issues
None blocking.

## Suggestions
- `src/main/providers/cursor.ts:22`, `src/main/providers/kimi.ts:20`, `src/main/providers/opencode.ts:20` — `buildResumeCommand` interpolates `sessionRef` (parsed from the child process's own stdout, via `asNonEmptyString` with no format check) directly into a shell command string. This mirrors the existing pattern already used by `claude`/`codex`/`grok`, so it's not a new architectural risk, but since three more external, non-schema-validated session-id sources are now flowing into command strings, it may be worth a shared helper that validates the expected id shape (`ses_...`, `session_...`, UUID) before it's ever re-interpolated, rather than trusting `asNonEmptyString` alone.
- None of the new provider files (`cursor.ts`, `kimi.ts`, `opencode.ts`, `antigravity.ts`) have tests asserting the exact `buildCommand`/`buildResumeCommand` output strings (flags, ordering). The new registry tests only check that `buildCommand`/`createParser` exist as functions. A quick snapshot/assertion on the generated command strings would catch accidental flag typos or ordering regressions (e.g., a dropped `--trust` or `--auto`) that unit tests on the parser alone wouldn't surface.

PR #123, opened 2026-07-18
Remote hosting: E2E cloud relay + web client (M1–M3.4)

William-Long-II:
Brings the full remote-hosting stack onto `development`: reach your desktop's terminal sessions from a browser/phone anywhere, end-to-end encrypted, with the relay as a blind broker.

## What's in it (12 commits)
- **Relay broker** (Bun, `bun:sqlite`, no native deps) that routes ciphertext between web clients and agents — never sees plaintext.
- **E2E crypto**: X25519 ECDH → HKDF-SHA256 → AES-256-GCM, Ed25519 signed handshake (anti-MITM). Node (agent) ↔ WebCrypto (browser) byte-compatible.
- **Auth**: GitHub OAuth, optionally gated to a GitHub org's members.
- **Desktop agent**: E2E terminal tunnel over the relay + keep-awake.
- **Web client**: xterm.js terminal in the browser — mobile-first, accessible; live session list; create sessions and **groups** (with a remote folder browser) from the phone; desktop refreshes live in both directions.
- **Docker deploy** documented (single-stage Bun image, TLS terminated at Caddy).

## Notable fixes
- **Phantom "working" blips** (`e8b570b`): idle sessions flapped state (sound + event-count churn) because unattended agent TUIs emit DSR cursor-position queries (`ESC[?6n`) several times a second and the ANSI stripper miscounted the private-mode tail as real output. Fixed the CSI matcher to the ECMA-48 grammar. **General app fix — affects all sessions, not just remote.**
- Dead-session state flap, mobile viewer no longer reflowing the desktop terminal, keyboard overscroll, send/Enter, live list refresh both directions.

## Testing
- Live-tested end to end against the deployed relay (`bodhilander.sytanek.tech`): sign-in, agent online, session create/drive from mobile, group create with folder browser.
- Root-caused the state flap via PTY-byte capture; fix verified (60× `ESC[?6n` → 0 printable, real output preserved).
- `build:main`, `typecheck:web`, `build:web` all clean.


relay/web/src/connection.ts
```diff
@@ -0,0 +1,113 @@
...
+
+function wsUrl(path: string): string {
+  return `${location.protocol === 'https:' ? 'wss' : 'ws'}://${location.host}${path}`;
+}
+
+export class RelayConnection {
+  private ws: WebSocket | null = null;
+  private key: CryptoKey | null = null;
+  private clientKeys: ClientKeys | null = null;
+  private sendCounter = 0;
```
brannon-bowden, line 32: sendCounter starts at 0 here and on the agent side (session-tunnel.ts), both under the same ECDH-derived key with no per-direction nonce separation — first frames from each side reuse (key, nonce=0), breaking AES-GCM. Use direction-distinct keys or nonce namespaces.

src/main/api/relay/session-tunnel.ts
```diff
@@ -0,0 +1,262 @@
...
+        const s = this.sessions.get(clientId);
+        if (s?.subs.has(e.id)) this.sealTo(clientId, { type: 'terminal:output', sessionId: e.id, data: e.data });
+      };
+      const onExit = (e: { id: string; exitCode: number }) => {
+        const s = this.sessions.get(clientId);
+        if (s?.subs.has(e.id)) this.sealTo(clientId, { type: 'terminal:exit', sessionId: e.id, exitCode: e.exitCode });
+      };
+      ptyManager.on('data', onData);
+      ptyManager.on('exit', onExit);
+      this.sessions.set(clientId, { key, sendCounter: 0, recvCounter: -1, subs: new Set(), onData, onExit });
```
brannon-bowden, line 87: sendCounter: 0 here collides with the client's independent counter (relay/web/src/connection.ts) under the same shared key — AES-GCM nonce reuse across directions. Needs a direction-specific key or nonce partition.

relay/src/repositories.ts
```diff
@@ -0,0 +1,212 @@
...
+            expires_at: number;
+          }
+        | null;
+
+      if (!row) return { ok: false, reason: 'not_found' };
+      if (row.status === 'completed') return { ok: false, reason: 'already_used' };
+      if (row.expires_at <= now()) return { ok: false, reason: 'expired' };
+
+      const machine = db.transaction(() => {
+        db.query("UPDATE link_codes SET status = 'completed', user_id = ? WHERE id = ?").run(userId, row.id);
```
brannon-bowden, line 163: This UPDATE has no `WHERE status = 'pending'` guard, and the status check above (line 159) runs outside this transaction. Two concurrent claims of the same code can both pass the check and both execute this UPDATE, letting the second racer silently rebind the machine. Move the status check inside the transaction with a conditional UPDATE.

relay/src/config.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  const logLevelRaw = env.LOG_LEVEL ?? 'info';
+  if (!LOG_LEVELS.includes(logLevelRaw as LogLevel)) {
+    fail(`LOG_LEVEL must be one of ${LOG_LEVELS.join(', ')}, got ${JSON.stringify(logLevelRaw)}`);
+  }
+  const logLevel = logLevelRaw as LogLevel;
+
+  const trustProxy = env.TRUST_PROXY ? parseBool('TRUST_PROXY', env.TRUST_PROXY) : false;
+
+  const allowedOrigins = (env.ALLOWED_ORIGINS ?? '')
```
brannon-bowden, line 109: allowedOrigins is parsed here but never consumed anywhere (no CORS header logic, no Origin validation on HTTP or WS routes). Either wire it up or remove the misleading 'empty = same-origin only' documentation in .env.example.

relay/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    fetch(req, srv) {
+      const url = new URL(req.url);
+      // Agents connect at /ws and authenticate with an Ed25519-signed nonce.
+      if (url.pathname === '/ws') {
+        if (srv.upgrade(req, { data: newAgentSocketData() })) return undefined;
+        return new Response('expected a websocket upgrade', { status: 426 });
+      }
+      // Web clients connect at /ws/client and are authenticated by session
+      // cookie at upgrade time — no cookie, no socket.
+      if (url.pathname === '/ws/client') {
```
brannon-bowden, line 34: No Origin header validation on this cookie-authenticated WebSocket upgrade. Worth adding an explicit allowlist check as defense-in-depth alongside SameSite=Lax.

brannon-bowden (changes requested):
Request changes. This is a well-structured, well-tested piece of infrastructure, but the E2E-encryption layer — which is the entire security premise of the "zero-knowledge relay" — has a nonce-reuse defect that lets the relay itself (the party the design explicitly distrusts) break confidentiality and forge frames. There's also a data race in the machine-linking flow. Both are concrete, fixable bugs, not just hardening suggestions.

## Blocking issues

1. **AES-256-GCM nonce reuse across the two directions of the E2E tunnel** — `relay/web/src/crypto.ts` (`nonce`/`sealJson`), `relay/web/src/connection.ts:32` (`sendCounter = 0`), `src/main/api/relay/e2e.ts` (`nonceFromCounter`/`seal`), `src/main/api/relay/session-tunnel.ts:87` (`sendCounter: 0`).
   Both peers derive the *same* AES-256-GCM key `K` via ECDH (correct, required), but each side independently numbers its own outgoing frames starting at counter `0`, and the nonce is simply that counter with no per-direction discriminator (same `nonceFromCounter`/`nonce` function on both ends, same HKDF salt/info). That means the agent's first outgoing frame and the client's first outgoing frame are both encrypted with `(K, nonce=0)`. Reusing a (key, nonce) pair under AES-GCM for two different plaintexts is catastrophic: it lets an observer XOR the two ciphertexts to recover the XOR of the plaintexts, and — because both messages share the same GHASH input structure — enables recovery of the authentication subkey `H`, allowing forgery of subsequent tags. The relay sits directly on the wire between these two directions and would see both counter-0 ciphertexts on every session, so the very adversary this scheme is meant to keep blind (a compromised/malicious relay) can defeat it. Fix: use distinct keys per direction (e.g. HKDF with a direction-specific `info` string) or partition the nonce space (e.g. a direction bit) so client→agent and agent→client never share a nonce under the same key.

2. **TOCTOU race in link-code claiming lets a machine be silently rebound** — `relay/src/repositories.ts:146-163` (`claimLinkCode`). The `status === 'completed'` check happens in a separate, non-transactional `SELECT` before the write transaction opens, and the transaction's `UPDATE link_codes SET status = 'completed', user_id = ? WHERE id = ?` (line 163) has no `WHERE status = 'pending'` guard. Two concurrent calls with the same code (double-submit retry, or two sessions racing) can both pass the outer check, both enter their own transaction, and both execute the UPDATE — the second writer silently rebinds the `machines` row (and the link-code's `user_id`) to itself, with no error surfaced to either caller. This breaks the "single-use code" guarantee the design documents rely on, and no test exercises concurrent claims (`m2.test.ts` only checks sequential re-claim of an already-completed code).

## Non-blocking issues

- **`ALLOWED_ORIGINS` is parsed but never enforced** — `relay/src/config.ts:109` defines and `.env.example` documents it ("empty = same-origin only"), but no file in this diff reads `config.allowedOrigins` to reject cross-origin requests or set CORS headers. As written this is a security control that does nothing, which is worse than not having the option at all since an operator may believe it's restricting access.
- **No `Origin` check on the WebSocket upgrades** — `relay/src/index.ts:34` (`/ws/client`) and the `/ws` agent endpoint accept upgrades without validating the `Origin` header. `SameSite=Lax` on the session cookie mitigates most cross-site fetch/XHR CSRF, but WebSocket-handshake cookie-attachment behavior across browsers is a known gray area; an explicit origin allowlist (using the `allowedOrigins` config once it's wired up) would be cheap defense-in-depth for a sensitive, cookie-authenticated endpoint.
- **mDNS + Tailscale Funnel guidance is removed in the same PR that ships an M3.4, not-yet-GA relay** — `src/main/api/index.ts`, `src/main/api/discovery/mdns-advertiser.ts` (deleted), `SettingsModal.tsx`. The design doc itself (`docs/designs/remote-hosting-relay.md`, §7) cautions that "removal should land behind the relay path being functional... to avoid a window with no remote path," yet the relay's own README milestone table still lists M3 (the live protocol) as "next" while this diff actually ships M3 protocol code. Existing users relying on Tailscale Funnel guidance lose that path the moment this merges, in exchange for a feature that requires standing up or trusting a separate relay server.

## Tests

- No test establishes a real E2E handshake and exchanges sealed frames in both directions through the actual crypto (`session-tunnel.ts` + `connection.ts`/`e2e.ts`). `relay/src/ws.test.ts`'s "client ↔ agent brokering (M3)" test sends raw unencrypted string payloads through the relay's routing layer only — it would not have caught the nonce-reuse bug above. A test that runs a full handshake and decrypts real bidirectional traffic is needed before this protocol can be trusted.
- No test for concurrent `claimLinkCode` calls on the same code (the race described above). The existing "cannot be claimed twice" test is sequential only.

PR #127, opened 2026-07-19
fix(vector-search): serialize indexing workers to stop concurrent onnxruntime SIGTRAP crash

### LINKED ISSUE #126: Native SIGTRAP crash: concurrent onnxruntime inference across multiple indexing workers
## Summary

Bodhilander 3.4.1 crashes the whole process with a native `EXC_BREAKPOINT (SIGTRAP)` while generating embeddings. The crash originates in `onnxruntime-node` and is triggered by **two indexing workers running ONNX inference at the same time**.

This is a **distinct** bug from BDHLNDR-40 (arena exhaustion). v3.4.1 already ships the BDHLNDR-40/46 mitigations (`enableCpuMemArena: false`, q8 quantization, `BATCH_SIZE=16`) and still crashes — a different frame and a different root cause.

## Crash evidence (macOS crash report, v3.4.1, M1 Ultra / 64 GB)

Faulting thread and a second, concurrent worker were **both** inside `InferenceSession::Run`:

- **Thread 44 (crashed):** `node::worker::Worker::Run` → `InferenceSessionWrap::Run` → `OrtValueToNapiValue` → `napi_new_instance` → `v8::Function::NewInstanceWithSideEffectType` → `brk 0` (V8 fatal guard)
- **Thread 65:** a *second, distinct* `node::worker::Worker::Run` → `InferenceSessionWrap::Run` → `Transpose::Compute` → `_platform_memmove`
- **Thread 37092850 + ONNX threadpool threads 37093008/09/10:** `(Data Abort) byte write Translation fault` inside `memmove` / transpose

Two separate Node `worker_threads` were driving two separate `onnxruntime` `InferenceSession`s simultaneously. The byte-write translation faults (write to an unmapped page during a transpose copy) are the signature of a heap overrun / write through a failed allocation — native memory corruption. The `brk 0` on Thread 44 is V8's fatal guard firing when it next touched the poisoned heap during N-API object construction. Not a null-deref.

## Root cause

`VectorSearchManager.startIndexing()` (`src/main/vector-search/index.ts`) guards only against a **duplicate worker for the same `index.id`**:

```ts
if (this.workers.has(index.id)) { return; }
```

There is **no global limit**. Each distinct directory → its own `index.id` → its own `Worker` → its own `HuggingFaceEmbeddingProvider` → its own ONNX `InferenceSession` with an intra-op threadpool (`EMBEDDING_THREAD_COUNT = 4` on this 20-core machine). Indexing two directories = two sessions running inference concurrently.

`enableCpuMemArena: false` (from BDHLNDR-40) makes ORT allocate **per-request** instead of from a pooled arena, so concurrent large allocations from two sessions compete directly in the native allocator — a write past a short allocation produces the translation fault.

Contributing factor: on successful completion the worker is **never terminated or removed from `this.workers`** (`handleIndexingComplete` updates status only). Finished workers linger with an idle ONNX session + parked native threads, accumulating over long uptimes (this crash was ~20 h in).

## Fix (this issue)

Serialize indexing globally in `VectorSearchManager`: **at most one indexing worker active at a time**, others queued; and **terminate finished workers** so their ONNX session/threads are reclaimed instead of accumulating. This removes the concurrent-inference condition the crash requires.

## Known related risk (needs a decision — see PR)

`searchCode()` runs `embedQuery()` on the **main Electron process** with its own ONNX session, so a search firing during an active index is a third concurrent-inference context of the same class (and blocks the UI thread). Fully removing that requires routing all embedding through a single shared inference context (larger refactor). Flagged in the PR for a scope decision — not silently deferred.

brannon-bowden:
## What & why

Fixes a native `EXC_BREAKPOINT (SIGTRAP)` that crashes the whole app, originating in `onnxruntime-node` during embedding generation.

**Root cause:** the app ran **multiple ONNX `InferenceSession`s concurrently**. Each indexing worker owned its own session (one per directory), and the main process owned another for search queries. Any overlap — two directories indexing, or a search firing during an index — put two `InferenceSession::Run` in flight at once, corrupting the native allocator.

The crash report confirms it: two distinct `node::worker::Worker::Run` threads both inside `InferenceSession::Run`, plus `(Data Abort) byte write Translation fault` in the ONNX threadpool during a transpose `memmove`. The `brk 0` in V8 is its fatal guard firing on the already-poisoned heap during N-API object construction.

This is a **different** failure mode from BDHLNDR-40 (arena exhaustion): v3.4.1 already ships `enableCpuMemArena: false` + q8 + `BATCH_SIZE=16` and still crashes. With the arena disabled, ORT allocates per-request, so concurrent allocations from two sessions compete directly in the allocator — hence the write-past-allocation fault.

## The fix — one inference authority for the whole process

**No two `InferenceSession.Run` can ever overlap anywhere now.** All embedding funnels through a single long-lived **embedding worker** that owns the only ONNX session and processes requests strictly one-at-a-time.

**Commit 1 — serialize indexing + stop leaking workers** (`index.ts`)
- Global gate: at most one indexing worker active at a time; extra `startIndexing()` calls queue by directory and drain as the active worker ends (`releaseAndDrain` → `startNextQueued`).
- Terminate finished workers. Previously a completed worker lingered *forever* holding an idle ONNX session + parked native threads, accumulating across directories and long uptimes (this crash was ~20 h in). `completedIndexes` flags the intentional terminate so the exit handler doesn't misreport it as a crash.
- `dispose()` drops the queue so post-dispose exit handlers can't re-spawn workers.

**Commit 2 — single shared embedding worker** (closes the main-process search path too)
- `embedding-worker.ts` — the **sole** ONNX session; a serial promise chain guarantees one `InferenceSession::Run` at a time (even two `embed()` calls on one session must not overlap).
- `embedding-service.ts` — main-process broker; correlates requests by id, respawns the worker if it dies, tears down on dispose.
- `worker-path.ts` — shared dev/packaged worker-script resolver (also adopted by `getWorkerPath()`).
- `indexing-worker.ts` — **no longer loads onnxruntime/transformers**; parses and requests embeddings over RPC (`embed-request` → `embed-response`/`embed-error`). Bundle drops ~1.6 MB → ~33 KB. Per-batch `await` preserves backpressure.
- `index.ts` — `searchCode()` uses the embedding service; the manager brokers each indexing worker's `embed-request` to it.
- `package.json` — `build:embedding-worker` (esbuild) added to `build`; ships via `dist/**` and is asar-unpacked with the rest of `dist/main/vector-search`.

## Behavior change

- Indexing multiple directories is now **sequential**, not parallel — the intended trade-off, since parallel indexing is exactly what crashed. Per-directory throughput is unchanged.
- Search embedding runs off the main thread (in the shared worker) and queues behind at most one in-flight batch — no more UI-thread blocking during inference.
- Circuit breaker, embedding-version re-index, cancel, and retry paths are preserved; all gate releases route through the single worker-exit choke point.

## Verification

- `tsc -p tsconfig.main.json --noEmit` — no new errors (only pre-existing missing-dep type decls in unrelated files).
- `esbuild` bundles both workers cleanly: `embedding-worker.js` 1.6 MB (owns ONNX), `indexing-worker.js` 33 KB (ONNX removed).
- No automated tests cover this module; the native crash isn't reproducible in-session. The fix is structural — it makes the two-concurrent-`InferenceSession` state the crash requires unrepresentable.

Closes #126


brannon-bowden:
Extended to a **complete** fix per request: the main-process search path no longer runs its own ONNX session. All embedding — indexing batches and search queries — now funnels through a single shared embedding worker with a serial request queue, so no two `InferenceSession::Run` can overlap anywhere in the process. Indexing workers are now parse-only (no onnxruntime/transformers; bundle 1.6 MB → 33 KB). The earlier open scope question is resolved (folded in, not deferred).

William-Long-II (approved):

PR #128, opened 2026-07-20
Release: concurrent ONNX indexing crash fix

William-Long-II:
## Summary

Promotes `development` to `production` for release.

### Included changes
- **#127 — fix(vector-search): concurrent ONNX indexing crash** — routes all embedding through one shared worker so onnxruntime is never run concurrently, eliminating the SIGTRAP crash during indexing.

## Release steps (after merge)
On `production`: `npm version patch` then `git push` (no `--tags`).


src/main/vector-search/index.ts
```diff
@@ -619,10 +821,20 @@ export class VectorSearchManager extends EventEmitter {
...
-    const provider = getEmbeddingProvider();
-    const queryEmbedding = await provider.embedQuery(query);
-
+    // Generate query embedding via the single embedding worker (BDHLNDR-126) —
+    // never a main-process ONNX session, which could race an active index. Only
+    // the embedding is guarded here; a DB/search failure (corruption, dimension
+    // mismatch) should surface, not be masked as an empty result.
+    let queryEmbedding: number[];
+    try {
+      queryEmbedding = await getEmbeddingService().embedQuery(query);
```
brannon-bowden, line 840: Embedding failures in searchCode are swallowed into an empty result array, making a real failure indistinguishable from "no matches found" to the user.

src/main/vector-search/embedding-service.ts
```diff
@@ -0,0 +1,187 @@
...
+      const timer = setTimeout(() => {
+        if (!this.pending.has(requestId)) return; // already settled
+        log.error(
+          `[EmbeddingService] Embed request ${requestId} exceeded ` +
+            `${EmbeddingService.REQUEST_TIMEOUT_MS}ms; restarting embedding worker`
+        );
+        // A hung worker won't recover on its own. Terminating fires 'exit' →
+        // failAll, which rejects every pending request for this worker and
+        // clears this.worker so the next request respawns a fresh one.
+        worker.terminate().catch((err) => {
```
brannon-bowden: This is the sole recovery path for a hung embedding worker. If the worker is blocked in a synchronous native inference call, terminate() may not be able to interrupt it, leaving the app stuck until restart. Please verify this works against an actual hang, not just a crash/exit.

brannon-bowden (changes requested):
This PR replaces the per-worker ONNX embedding model with a single shared embedding worker brokered through the main process, and adds a global serialization gate so only one indexing job runs at a time — a solid architectural fix for the reported concurrent-inference SIGTRAP crash. The design is well-commented and the exit/gate-ownership guards (`ownsSlot`/`ownsGate`) show real thought was put into the race conditions this introduces. However, CI is currently red and there are a few points worth addressing before merge.

## Issues

- **CI is failing**: the `quality-gate` check is reported as failing. Per review policy this blocks approval regardless of code quality — please confirm the pipeline is green before merge (note the status also lists `quality-gate` as passing elsewhere in the same report, which is worth reconciling/re-running to get an unambiguous signal).
- **No tests for the new concurrency machinery.** `embedding-service.ts`, the indexing gate/queue in `index.ts` (`activeIndexId`/`activeWorker`/`indexQueue`), and the rolling inactivity watchdog are all new, non-trivial control flow with many race windows the comments themselves call out (superseded workers, cancel-during-restart, queued-then-cancelled directories). None of this is covered by tests. Given this PR's whole purpose is fixing a crash caused by subtle concurrency, some unit tests around cancel+restart timing and queue draining would materially reduce risk of a regression reintroducing it.
- **`searchCode` now silently swallows embedding failures as "no results"** (`src/main/vector-search/index.ts`, the new `try { queryEmbedding = await getEmbeddingService().embedQuery(query); } catch (err) { ...; return []; }` block). If the shared embedding worker is down or crashed, a search returns an empty result list indistinguishable from "no matches," which could mislead users into thinking their query legitimately found nothing. Consider surfacing this as a distinct error/status to the caller.
- **Single shared embedding worker is a new single point of failure**, and the recovery path for a truly hung (not crashed) native inference call relies on `worker.terminate()` interrupting it (`src/main/vector-search/embedding-service.ts`, the `REQUEST_TIMEOUT_MS` timeout calling `worker.terminate()`). If the worker is blocked inside a synchronous native call, `terminate()` may not be able to preempt it until that call returns, in which case the 5-minute request timeout and the indexing 10-minute watchdog won't actually recover — silently disabling all indexing and search until app restart. Worth verifying this termination path actually works against a real hang before relying on it as the crash-recovery mechanism.

## Suggestions (non-blocking)

- Queued indexing directories are only logged (`log.info(... 'Queued indexing for...')`); there's no "queued" status surfaced to the UI. Since indexing multiple directories is now fully serialized, users starting a second index while one is running will see no visible progress for a potentially long time. Consider surfacing a queued state.
- The same-index bypass of the queue gate (`this.activeIndexId !== index.id` check in `startIndexing`) is subtle and depends on exit-handler identity guards elsewhere to stay correct under cancel/restart races — a comment summarizing the invariant it depends on (e.g., "gate is only re-entered for the same id during cancel+restart") would help future maintainers avoid breaking it.

brannon-bowden (approved):
This PR replaces the previous per-worker ONNX inference model with a single shared embedding worker plus a global indexing gate/queue in `VectorSearchManager`, which directly targets the concurrent-`InferenceSession` crash. I traced the gate/queue/watchdog state machine closely (the `ownsSlot`/`ownsGate` identity checks that guard against superseded workers on cancel/restart, the release-and-drain path taken when the crash circuit-breaker fires before the gate is acquired, and the `EmbeddingService` timeout/respawn logic) and didn't find a race or leak — the design is careful and the comments accurately describe the invariants being protected.

## Issues

- **No test coverage for the new concurrency logic** (`src/main/vector-search/embedding-service.ts`, `embedding-worker.ts`, and the gate/queue/watchdog additions in `index.ts`). This is exactly the kind of code where regressions are easy to reintroduce silently (superseded-worker races, gate-never-released, queue resurrection after cancel) and hard to catch by inspection alone. Given this PR exists specifically to fix a crash caused by this class of bug, some automated coverage (even a lightweight harness mocking `Worker`) would substantially de-risk future changes here.
- Minor: this PR bundles an unrelated relay/web frontend feature (machine switching/linking UI, terminal scroll/viewport fixes) into a PR titled and scoped as a backend indexing crash fix. Not a defect, but it makes the release notes and any future `git bisect` noisier than necessary — consider splitting orthogonal changes into separate PRs going forward.

CI is green across all gating checks (SonarQube, quality-gate ×2, GitGuardian, check-version); `release` and `build` are skipped/neutral, which is informational and doesn't block.

## Suggestions (non-blocking)

- `embedding-provider.ts`'s fix to null out `initPromise` on failure (so a transient init error doesn't permanently poison the singleton) is a good, minimal fix — no notes there.
- The `worker-path.ts` extraction (shared by both worker types) is a clean dedup of previously-duplicated asar-unpacking logic.
- `SettingsModal.tsx`'s Fragment removal is a harmless readability nit, unrelated to the rest of the PR but fine to keep.

William-Long-II:
@brannon-bowden Thanks for the thorough review — all four blocking items are addressed across two PRs to `development` that will flow into this release PR:

**1. CI red / quality gate** — the failing `quality-gate` run was the *branch* analysis of `development` (push-triggered), not this PR's analysis. It failed on exactly one new violation: a redundant single-child fragment in `SettingsModal.tsx` (S6749). Fixed in #129 (all checks green). That's also why the same report showed a passing `quality-gate` — that one was the PR-level analysis.

**2. Tests for the concurrency machinery** — #130 adds `src/main/vector-search/__tests__/concurrency.test.ts` (15 tests) covering the embedding-service lifecycle (request correlation, exit fail-all + respawn, superseded-worker late exits, hung-worker timeout and recovery) and the indexing gate (queue dedupe/drain, cancel-of-queued, the cancel + same-index restart race with ownsSlot/ownsGate, crash circuit breaker, embed-request brokering).

**3. `searchCode` swallowing embedding failures** — fixed in #130: it now rethrows `Semantic search is unavailable: <cause>`. The plumbing already existed — `safeHandle` propagates the rejection over IPC, `useCodeSearch` catches it into `searchError` (rendered in the UI), and the HTTP route maps it to a 500.

**4. `terminate()` vs a real hang** — verified empirically: `terminate()` interrupts a tight synchronous JS loop and an `Atomics.wait` block in ~2–3 ms, but per Node docs it cannot preempt a synchronous *native* call (a wedged `InferenceSession::Run`) — `exit` may never fire. So in #130 the recovery path no longer depends on it: on timeout the service rejects all pending requests immediately, parks the wedged worker in a `blocked` slot, and defers respawn until that worker's `exit` is actually observed — spawning a second ONNX session while the old one may still be inside `Run` would reintroduce the exact crash this PR fixes. JS-level hangs recover automatically in milliseconds; a true native wedge degrades to loud immediate errors (visible thanks to item 3) instead of a silently dead service.

**Non-blocking suggestions:** the same-index gate-bypass invariant is now documented at the gate check in `startIndexing` (plus locked in by a test). Surfacing a "queued" indexing state needs a new `IndexStatus` value through shared types + renderer — proposing to track that as a follow-up ticket rather than grow the release PR.

PR #130, opened 2026-07-20
fix(vector-search): address #128 review findings on the shared embedding worker

William-Long-II:
## Summary

Addresses the blocking findings from @brannon-bowden's review on release PR #128 (the BDHLNDR-126 concurrent-ONNX fix).

### 1. `searchCode` no longer swallows embedding failures
It rethrows `Semantic search is unavailable: <cause>` instead of returning `[]`. The propagation path already existed end-to-end: `safeHandle` re-throws over IPC → `useCodeSearch` catches into `searchError` (rendered in the UI); the HTTP route maps the rejection to a 500. A dead embedding worker is now visibly distinct from "no matches".

### 2. Hung-worker recovery no longer depends on `terminate()` interrupting native code
Verified empirically: `worker.terminate()` interrupts a tight synchronous JS loop and an `Atomics.wait` block in ~2–3 ms, but per Node docs it cannot preempt a synchronous **native** call (a wedged `InferenceSession::Run`) — `exit` may never fire. On request timeout the service now:
- rejects all pending requests **immediately** (callers get a descriptive error, surfaced by item 1),
- parks the wedged worker in a `blocked` slot and terminates best-effort,
- **defers respawn until the worker's `exit` is actually observed** — spawning a second ONNX session while the old one may still be inside `Run` would reintroduce the exact concurrent-inference crash this machinery exists to prevent. Until then, new requests fail fast instead of stacking 5-minute timeouts.

Net: JS-level hangs recover automatically in milliseconds; a true native wedge degrades to loud, immediate errors instead of a silently dead service.

### 3. Tests for the concurrency machinery
New `src/main/vector-search/__tests__/concurrency.test.ts` (15 tests, `bun test src/main/vector-search`) covering: request/response correlation; per-request error isolation; exit fail-all + respawn; superseded-worker late exits (service and manager); the timeout/blocked/recovery path; gate serialization + queue dedupe/drain; cancel-of-queued not resurrected by the drain; **cancel + immediate same-index restart race** (the ownsSlot/ownsGate invariant); the crash circuit breaker; embed-request brokering (success/error/cancelled-worker); and `searchCode` error surfacing.

### 4. Same-index gate-bypass invariant documented
The subtle `activeIndexId === index.id` bypass in `startIndexing` now carries a comment spelling out the three pieces it depends on (sync `workers.delete` in `cancelIndexing`, the bypass, the identity guards) — plus the test above locks the behavior in.

## Verification
- `bun test src/main/vector-search` — 15/15 pass
- `bun test src/main` — 143 pass / 1 pre-existing failure (`chat-parser 06-response` snapshot, fails on `development` too, unrelated)
- `tsc -p tsconfig.main.json --noEmit` — clean

The reviewer's non-blocking suggestion to surface a "queued" indexing state in the UI needs a new `IndexStatus` value through shared types + renderer; proposing to track that as a follow-up ticket rather than grow the release.

### DIFF
--- a/src/main/vector-search/embedding-service.ts
+++ b/src/main/vector-search/embedding-service.ts
@@ -25,6 +25,16 @@ export class EmbeddingService {
   private readonly pending = new Map<string, PendingRequest>();
   private seq = 0;
 
+  // A worker that timed out and was told to terminate, but whose 'exit' has
+  // not fired yet. terminate() reliably interrupts JS execution and
+  // Atomics.wait, but it CANNOT preempt a synchronous native call — a wedged
+  // InferenceSession::Run only lets the thread die when it returns. Until the
+  // exit is observed we must not spawn a replacement: two live ONNX sessions
+  // is the exact crash this service exists to prevent (BDHLNDR-126). While
+  // set, requests fail fast with a clear error instead of queueing behind a
+  // dead worker.
+  private blocked: Worker | null = null;
+
   // A stalled native inference emits no 'error'/'exit', so bound every request.
   // Generous enough to cover a cold model download on the first call (matches
   // the indexing init timeout); only a genuine hang trips it.
@@ -32,6 +42,15 @@ export class EmbeddingService {
 
   private ensureWorker(): Worker {
     if (this.worker) return this.worker;
+
+    if (this.blocked) {
+      // A hung worker is still shutting down (or wedged in a native call that
+      // terminate() cannot interrupt). Fail fast — callers surface this to the
+      // UI — and recover automatically the moment its 'exit' finally fires.
+      throw new Error(
+        'Embedding worker is unresponsive and still shutting down; embedding is temporarily unavailable'
+      );
+    }
 
     const workerPath = resolveVectorSearchWorker('embedding-worker.js');
     const workerOptions: any = { stdout: true, stderr: true };
@@ -103,6 +122,8 @@ export class EmbeddingService {
       if (code !== 0) {
         log.error(`[EmbeddingService] Worker exited with code ${code}`);
       }
+      // A timed-out worker has finally died — lift the respawn block.
+      if (this.blocked === worker) this.blocked = null;
       failAll(new Error(`Embedding worker exited (code ${code})`));
     });
 
@@ -136,12 +157,7 @@ export class EmbeddingService {
           `[EmbeddingService] Embed request ${requestId} exceeded ` +
             `${EmbeddingService.REQUEST_TIMEOUT_MS}ms; restarting embedding worker`
         );
-        // A hung worker won't recover on its own. Terminating fires 'exit' →
-        // failAll, which rejects every pending request for this worker and
-        // clears this.worker so the next request respawns a fresh one.
-        worker.terminate().catch((err) => {
-          log.warn('[EmbeddingService] Error terminating hung worker:', err);
-        });
+        this.failHungWorker(worker);
       }, EmbeddingService.REQUEST_TIMEOUT_MS);
       this.pending.set(requestId, {
         resolve: (embeddings) => {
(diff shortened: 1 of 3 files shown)


src/main/vector-search/embedding-service.ts
```diff
@@ -103,6 +122,8 @@ export class EmbeddingService {
       if (code !== 0) {
         log.error(`[EmbeddingService] Worker exited with code ${code}`);
       }
+      // A timed-out worker has finally died — lift the respawn block.
+      if (this.blocked === worker) this.blocked = null;
```
brannon-bowden, line 126: This clears `this.blocked` only if the exit handler reaches this line. If (as the readonly shared `pending` map implies) the handler is guarded by `this.worker !== worker`, this line is unreachable for the hung worker's real exit, since `failHungWorker` already nulled `this.worker`. That would leave `this.blocked` set forever after the first timeout, permanently failing all future embed() calls. Please confirm the guard also passes when `this.blocked === worker`.

src/main/vector-search/index.ts
```diff
@@ -822,15 +834,18 @@ export class VectorSearchManager extends EventEmitter {
...
       queryEmbedding = await getEmbeddingService().embedQuery(query);
     } catch (err) {
       log.error('[VectorSearch] searchCode embedding failed:', err);
-      return [];
+      // Rethrow rather than returning [] — an empty array here is
+      // indistinguishable from "no matches" in the UI. The IPC handler
+      // propagates the rejection to useCodeSearch (which renders
+      // searchError) and the HTTP route maps it to a 500.
+      const msg = err instanceof Error ? err.message : String(err);
+      throw new Error(`Semantic search is unavailable: ${msg}`);
```
brannon-bowden, line 848: searchCode now throws instead of returning [] on embedding failure. Please confirm the IPC handler and HTTP route (not in this diff) actually catch this rejection — otherwise this is an unhandled-rejection risk for a previously-safe path.

brannon-bowden (changes requested):
This PR is well-documented and adds a genuinely thorough concurrency test suite, but I'm requesting changes: CI is red, and there's a logic gap in the new `blocked`-worker recovery path that looks like it defeats its own purpose.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate are both red (GitGuardian passes). This needs to be green before merge, and the quality-gate failure may well be the concurrency test suite itself failing — see next point.

- **`src/main/vector-search/embedding-service.ts:126` (blocking) — `this.blocked` is likely never cleared, permanently bricking the embedding service after the first hang.** `failHungWorker` sets `this.worker = null` and `this.blocked = worker` before calling `worker.terminate()`. The new line that lifts the block, `if (this.blocked === worker) this.blocked = null;`, sits inside the `worker.on('exit', ...)` callback, right after the pre-existing `if (code !== 0) { ... }` block. Since `pending` is a single `readonly` field shared across worker instances (not per-worker), that `exit` handler almost certainly starts with an identity guard (`if (this.worker !== worker) return;`) to stop a stale/superseded worker's exit from corrupting the *current* worker's in-flight requests — the "late exit from a superseded worker" test in the new suite explicitly depends on that guard existing. But by the time the hung worker's real `exit` finally fires, `this.worker` is `null` (set by `failHungWorker`), so `this.worker !== worker` is true and the handler returns *before* reaching the new `this.blocked` clearing line. If that's how the guard is written, `this.blocked` is set once and never reset, and every subsequent `embed()` call throws "temporarily unavailable" for the rest of the process's life instead of recovering — which is the exact scenario the new `'request timeout fails pending immediately and blocks respawn until the hung worker exits'` test is supposed to prove works. Please double-check this guard accounts for the `this.blocked === worker` case (e.g. `if (this.worker !== worker && this.blocked !== worker) return;`) before merge.

- **`src/main/vector-search/index.ts:848` (major) — `searchCode` now throws instead of returning `[]` on embedding failure; verify all callers were updated to handle the rejection.** This is a good fix in principle (an empty array previously masked real failures), but it's a breaking behavior change for any caller written against the old "always resolves" contract. The added comment asserts "The IPC handler propagates the rejection to useCodeSearch ... and the HTTP route maps it to a 500," but none of those call sites are part of this diff, so that claim isn't verifiable here. If either caller doesn't already `.catch`/try-catch around `searchCode`, this turns a degraded-but-functional search into an unhandled rejection.

## Suggestions

- `index.ts:848` — non-blocking: consider `throw new Error(..., { cause: err })` instead of just interpolating `err.message`, so the original stack/error type isn't lost for callers that log the caught error.

brannon-bowden (approved):
This PR looks solid: it adds a well-reasoned `blocked` gate to `EmbeddingService` to prevent spawning a second worker while a hung/wedged worker's `terminate()`/`exit` is still in flight, documents the pre-existing same-index cancel/restart invariant in `VectorSearchManager`, and changes `searchCode` to surface embedding failures instead of silently returning `[]`. The accompanying concurrency test suite is thorough and specifically exercises the races these changes are meant to close (superseded-worker exits, hung-worker timeout/respawn block, cancel+restart identity guards, queue draining).

Walking through the `blocked`/`failHungWorker` logic: `ensureWorker()` only creates a worker when both `this.worker` and `this.blocked` are falsy, `failHungWorker` atomically clears `this.worker`, sets `this.blocked`, and rejects pending requests before calling `terminate()`, and the exit handler clears `this.blocked` only when the exiting worker is still the tracked one (`this.blocked === worker`). That keeps the two live-worker invariant intact even under the "hung worker later exits" and "superseded worker's late exit" races exercised in the tests. `dispose()` correctly resets `blocked` too.

## Issues
- `src/main/vector-search/index.ts:848` — `searchCode` now rethrows embedding failures instead of returning `[]`. This is a reasonable correctness fix (consistent with how DB/search failures were already surfaced), and the comment asserts the IPC handler/HTTP route already convert the rejection into `searchError`/500, but neither of those call sites is part of this diff. Worth a quick confirmation that nothing upstream still expects `searchCode` to never throw (e.g. an unguarded `await` that would surface as an unhandled rejection) before merging.

## Suggestions (non-blocking)
- The permanently-blocked-until-exit behavior (`this.blocked` never clearing if the wedged worker truly never emits `exit`) is intentional and documented, but it does mean a sufficiently stuck native call permanently disables embedding for the process lifetime. That's an acceptable tradeoff given the crash it prevents, but might be worth a follow-up (e.g. surfacing a persistent health-check signal) if this turns out to happen in practice.

PR #131, opened 2026-07-20
feat(providers): friendly CLI launch-failure hints + one-click install

William-Long-II:
## Summary

Follow-ups to #97's provider setup guidance, motivated by a user report of Codex dying with a raw `spawn .../@openai/codex-darwin-arm64/.../codex ENOENT` (npm dropped the vendored platform binary from a global `@openai/codex` install — works fine on other machines).

**1. Friendly launch-failure banner.** New `spawn-failure.ts` classifies early session output as:
- `missing` — shell command-not-found shapes (zsh/bash/POSIX exec/cmd.exe/PowerShell), parameterized by the provider's CLI name so unrelated output can't trigger it, or
- `broken` — the CLI launched but its startup failed (`spawn ... ENOENT`, `bad CPU type in executable` from Rosetta-mismatched installs, `dyld: Library not loaded`).

`PtyManager` checks the scrollback within 15s of spawn (once per session — a running agent later printing such an error from user code can't retrigger it) and emits a hint event. The Terminal renders a dismissible banner explaining the failure in plain language, with install guidance, a docs link, and a one-click fix. The banner survives into the "Session stopped" view, since a missing CLI exits the shell immediately.

**2. "Install for me".** `ProviderDefinition.setup` gains an optional `installCommand` — directly runnable and re-run-safe (npm providers use `npm install -g --force` so the same command *repairs* broken installs; a plain reinstall over an existing package won't restore a dropped platform optionalDependency). `providers:run-install` spawns it in a visible pty (reusing the login-pty pattern incl. BDHLNDR-33 deferred emission; commands come only from the static registry — the renderer sends just a provider id). The new `ProviderInstallModal` attaches a Terminal to it and reports the exit code. Wired into both the failure banner ("Install/Reinstall for me" → then "Restart session") and Settings → Providers ("Install for me" on missing CLIs, with re-detection on close). Antigravity has no `installCommand` (GUI install) and keeps the text-only hint.

## Test plan

- `bun test` — 201 pass; the single `chat-parser` snapshot failure (`06-response`) reproduces on a clean tree and is unrelated.
- New unit tests in `spawn-failure.test.ts` cover all five command-not-found shell shapes, the verbatim codex ENOENT report, arch-mismatch/dyld variants, ANSI-wrapped output, missing-vs-broken precedence, and normal-startup non-matches.
- `tsc` clean for main + renderer configs; production webpack build compiles.
- Manual: break a codex install (`rm` the vendored binary), launch a codex session → banner appears; "Reinstall for me" runs the npm command in the modal, exits 0, "Restart session" launches successfully.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2,7 +2,8 @@ import { app, BrowserWindow, ipcMain, dialog, shell, crashReporter } from 'elect
 import * as fs from 'fs';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
-import { resolveLaunchProviderId, listProviders } from './providers';
+import { resolveLaunchProviderId } from './providers';
+import { startProviderInstall, cancelProviderInstall } from './provider-install';
 import { detectProviders } from './provider-detector';
 import * as keyVault from './key-vault';
 import { getDatabase, closeDatabase } from './database';
@@ -1266,29 +1267,12 @@ safeHandle('providers:detect', async () => {
 
 // Run a provider's install command in a visible pty the renderer attaches a
 // Terminal to (Settings → Providers "Install" button / launch-failure
-// banner). Commands come from the static provider registry only — the
-// renderer sends just the provider id.
-safeHandle('providers:run-install', async (providerId: unknown) => {
-  const id = requireIpcString(providerId, 'providerId');
-  const provider = listProviders().find((p) => p.id === id);
-  const command = provider?.setup.installCommand;
-  if (!provider || !command) {
-    throw new Error(`Provider '${id}' has no runnable install command`);
-  }
-  const ptyId = `__install-${id}`;
-  // A previous run's pty may still be around (user reopened the modal).
-  await ptyManager.kill(ptyId);
-  ptyManager.createInstallSession(ptyId, command);
-  return { ptyId, command };
-});
+// banner). Validation + orchestration live in provider-install.ts.
+safeHandle('providers:run-install', (providerId: unknown) =>
+  startProviderInstall(ptyManager, providerId));
 
-safeHandle('providers:cancel-install', async (ptyId: unknown) => {
-  const id = requireIpcString(ptyId, 'ptyId');
-  if (!id.startsWith('__install-')) {
-    throw new Error('Not an install pty');
-  }
-  await ptyManager.kill(id);
-});
+safeHandle('providers:cancel-install', (ptyId: unknown) =>
+  cancelProviderInstall(ptyManager, ptyId));
 
 // Provider API-key vault (#99). Keys go in and are tested; they are never
 // returned to the renderer. IPC is a trust boundary — validate argument
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2,7 +2,7 @@ import { app, BrowserWindow, ipcMain, dialog, shell, crashReporter } from 'elect
 import * as fs from 'fs';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
-import { resolveLaunchProviderId } from './providers';
+import { resolveLaunchProviderId, listProviders } from './providers';
 import { detectProviders } from './provider-detector';
 import * as keyVault from './key-vault';
 import { getDatabase, closeDatabase } from './database';
@@ -639,6 +639,11 @@ function createWindow(): void {
     getApiServer().broadcastSessionState(event.sessionId, event.state, event.event);
   });
 
+  // Provider CLI launch-failure hints (missing/broken install banner)
+  ptyManager.on('providerHint', (hint) => {
+    mainWindow?.webContents.send('provider:install-hint', hint);
+  });
+
   // Save window bounds on resize/move
   const saveWindowBounds = () => {
     if (!mainWindow) return;
@@ -1257,6 +1262,32 @@ safeHandle('editor:detectAvailable', async () => {
 
 safeHandle('providers:detect', async () => {
   return detectProviders();
+});
+
+// Run a provider's install command in a visible pty the renderer attaches a
+// Terminal to (Settings → Providers "Install" button / launch-failure
+// banner). Commands come from the static provider registry only — the
+// renderer sends just the provider id.
+safeHandle('providers:run-install', async (providerId: unknown) => {
+  const id = requireIpcString(providerId, 'providerId');
+  const provider = listProviders().find((p) => p.id === id);
+  const command = provider?.setup.installCommand;
+  if (!provider || !command) {
+    throw new Error(`Provider '${id}' has no runnable install command`);
+  }
+  const ptyId = `__install-${id}`;
+  // A previous run's pty may still be around (user reopened the modal).
+  await ptyManager.kill(ptyId);
+  ptyManager.createInstallSession(ptyId, command);
+  return { ptyId, command };
+});
+
+safeHandle('providers:cancel-install', async (ptyId: unknown) => {
+  const id = requireIpcString(ptyId, 'ptyId');
+  if (!id.startsWith('__install-')) {
+    throw new Error('Not an install pty');
+  }
+  await ptyManager.kill(id);
 });
 
 // Provider API-key vault (#99). Keys go in and are tested; they are never
--- a/src/main/pty-manager.ts
+++ b/src/main/pty-manager.ts
@@ -1,5 +1,6 @@
 import * as pty from 'node-pty';
 import * as fs from 'fs';
+import * as os from 'os';
 import { execFile } from 'child_process';
 import { EventEmitter } from 'events';
 import {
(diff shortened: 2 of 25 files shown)


src/main/pty-manager.ts
```diff
@@ -405,6 +422,126 @@ export class PtyManager extends EventEmitter {
...
+      installCommand: setup.installCommand ?? null,
+      docsUrl: setup.docsUrl,
+    };
+    this.emit('providerHint', hint);
+  }
+
+  /**
+   * Spawn a pty running a provider's install command so the user can watch it
+   * instead of copy-pasting into their own terminal (follow-up to #97's
+   * install hints). Runs through the user's shell like sessions do, defers
```
brannon-bowden, line 495: createInstallSession (and the checkSpawnFailure one-shot/time-window logic above it) has no test coverage — please add unit/integration tests for the notify-once gate, the 15s window cutoff, and the install pty's exit/scrollback-flush behavior.

src/main/index.ts
```diff
@@ -1259,6 +1264,32 @@ safeHandle('providers:detect', async () => {
...
+  const id = requireIpcString(providerId, 'providerId');
+  const provider = listProviders().find((p) => p.id === id);
+  const command = provider?.setup.installCommand;
+  if (!provider || !command) {
+    throw new Error(`Provider '${id}' has no runnable install command`);
+  }
+  const ptyId = `__install-${id}`;
+  // A previous run's pty may still be around (user reopened the modal).
+  await ptyManager.kill(ptyId);
+  ptyManager.createInstallSession(ptyId, command);
```
brannon-bowden: providers:cancel-install's '__install-' prefix validation is a trust-boundary check with no test proving it rejects arbitrary/non-install ptyIds. Please add a test.

src/main/pty-manager.ts
```diff
@@ -405,6 +422,126 @@ export class PtyManager extends EventEmitter {
...
   }
 
+  /**
+   * Classify early session output as a CLI launch failure (missing from
+   * PATH, or installed-but-broken like codex's `spawn ... ENOENT`) and emit
+   * a one-shot 'providerHint' event so the renderer can show a friendly
+   * install banner. Checks the accumulated scrollback rather than the chunk,
+   * so an error split across pty reads still matches.
+   */
+  private checkSpawnFailure(id: string): void {
```
brannon-bowden, line 467: classifySpawnFailure runs against scrollbackBuffer, which is capped at 100KB via slice(-MAX_SCROLLBACK_SIZE) (tail-keeping). If a CLI prints a lot before failing, the failure signature near the start of output could be evicted before this check ever sees it.

brannon-bowden (changes requested):
This PR adds a nice quality-of-life feature — classifying provider CLI launch failures and offering a one-click (re)install — and the pure classification logic (`spawn-failure.ts`) is well tested against real-world shapes (bash/zsh/cmd/PowerShell, the codex ENOENT report, Rosetta/arch mismatches). CI is green (SonarQube, quality-gate, GitGuardian all passing). However, the new stateful/process-spawning logic in `pty-manager.ts` and the new IPC surface in `index.ts` ship with zero test coverage, and there are a couple of design gaps worth addressing before merge.

## Issues

- **Blocking — no tests for the new non-trivial main-process logic.** `classifySpawnFailure` is well covered, but nothing exercises `PtyManager.checkSpawnFailure` (`src/main/pty-manager.ts:425-451`) — the one-shot `spawnFailureNotified` gate and the `SPAWN_FAILURE_WINDOW_MS` cutoff are exactly the kind of boundary conditions that regress silently — nor `createInstallSession` (`src/main/pty-manager.ts:459-540`, new pty lifecycle, exit handling, scrollback flush on fast failure), nor the two new IPC handlers `providers:run-install` / `providers:cancel-install` (`src/main/index.ts:1267-1291`). The `cancel-install` handler's `id.startsWith('__install-')` check is a real trust-boundary validation and currently has no test proving it rejects non-install pty ids.
- **Major — 100KB scrollback trim can evict the failure signature it's trying to match.** `checkSpawnFailure` runs `classifySpawnFailure` against `session.scrollbackBuffer`, which is trimmed with `slice(-MAX_SCROLLBACK_SIZE)` keeping only the tail (`src/main/pty-manager.ts` scrollback accumulation shared with `createInstallSession`, e.g. lines ~495-498). A CLI that fails but is chatty before dying (or whose wrapper prints a lot in the first few seconds) could have the "command not found"/"ENOENT" line pushed out of the retained window before classification ever sees it, silently defeating the feature for that case.

## Suggestions (non-blocking)

- `createInstallSession` uses an inline `require('os').homedir()` (`src/main/pty-manager.ts` ~line 468) instead of a top-level `import * as os from 'os'` consistent with the rest of the file's style.
- One-click install/reinstall runs `curl ... | bash` style commands (grok/kimi/opencode/cursor) or `npm install -g --force` with no confirmation step beyond the button click itself. Since the button already visibly opens a terminal running the exact command, this is likely fine, but consider whether a lightweight "this will run: `<command>`, continue?" confirmation is warranted given these are remote scripts piped straight to a shell.
- `ProviderInstallModal` renders a native `<dialog open>` without calling `showModal()`, so it doesn't get native focus-trapping/Escape handling — it works today only because it's paired with a custom `.modal-overlay`. Minor accessibility/robustness gap if that overlay convention ever changes.

William-Long-II:
Thanks for the review @brannon-bowden — all points addressed in ba722c6:

**Blocking — test coverage.** Added `pty-manager.test.ts` (fake node-pty driven directly through its onData/onExit callbacks) covering the `providerHint` one-shot gate, the 15s window cutoff, signatures split across pty chunks, and `createInstallSession`'s deferred emission, prime flush, and exit-before-prime scrollback flush. The IPC orchestration moved out of `index.ts` into a new `provider-install.ts` module so the trust-boundary validation is unit-testable — `provider-install.test.ts` proves `cancelProviderInstall` rejects session UUIDs, `__login-*` ptys, and prefix-not-at-position-0 ids, and that commands only ever come from the static registry.

**Major — scrollback eviction.** Real. Classification now runs against a dedicated `launchOutput` buffer that keeps output from the *head* (256KB cap) and is freed as soon as it can't matter (hint fired, or the 15s window passed) — so it costs nothing on healthy sessions and can't be evicted by the tail-keeping scrollback trim. Regression-tested with 150KB of chatty output preceding the ENOENT line.

**Suggestions.**
- `require('os')` → top-level `import * as os` (also converted the identical pre-existing call in the login-pty path).
- Added a confirm dialog showing the exact command before anything runs, for both entry points (banner and Settings).
- On `<dialog>`/`showModal()`: agreed it's a gap, but left as-is for now — `showModal()` puts the dialog in the top layer with its own centering, which fights the `.modal-overlay` flex convention every other modal here uses, and Escape-cancel would bypass the running-install confirm. Happy to do a follow-up that migrates the modal convention wholesale if you'd like.

One test-infra note: bun's `mock.module` patches a specifier process-wide, so the new pty-manager test mocks only modules no other file exercises for real (node-pty, electron-log, memory/injector) and uses the real provider registry, sessions repo, and key-vault — verified stable across repeated full-suite runs.

brannon-bowden (approved):
Both previously-raised issues are resolved, and this update doesn't introduce new blocking problems.

**Test coverage (previously blocking) — resolved.** `pty-manager.test.ts` now drives `checkSpawnFailure` end-to-end via a hand-fed fake pty, covering the one-shot `spawnFailureNotified` gate, the 15s `SPAWN_FAILURE_WINDOW_MS` cutoff, chunk-split signatures, and the chatty-startup case. `createInstallSession`'s deferred emission / prime flush / exit-before-prime flush are all exercised. The IPC trust-boundary validation is now extracted into `provider-install.ts` and unit-tested directly in `provider-install.test.ts`, including the `cancel-install` namespace check rejecting non-`__install-` ids (including the leading-space and non-string cases).

**Scrollback-eviction (previously major) — resolved.** The new `launchOutput` buffer (`src/main/pty-manager.ts:61-68`, `445-461`) is dedicated to failure classification, keyed off the *head* of output (capped at 256KB) rather than sharing the tail-trimmed `scrollbackBuffer`, so a chatty startup can no longer push the failure signature out before it's matched. This is verified by the new "chatty startup" test.

## Suggestions (non-blocking)

- The new `launchOutput` buffer is still finite (256KB) and head-anchored — a launch failure preceded by more than 256KB of output within the 15s window would still miss classification. Much less likely than the old 100KB tail-trim case, and the test suite doesn't need to cover this extreme edge, but worth a one-line note if it ever surfaces in the field.
- Still open from the previous review: `ProviderInstallModal` renders `<dialog open>` without `showModal()`, so it lacks native focus-trapping/Escape handling. Not touched by this diff; carrying forward as a minor robustness gap.
- The new `window.confirm(...)` guards in `ProviderSettings.tsx` and `Terminal.tsx` nicely address the earlier "no confirmation before running remote install scripts" note.

PR #132, opened 2026-07-21
fix(relay): auto-recover web client when agent comes online

William-Long-II:
## Problem

After linking a machine, the desktop app showed **online** but the web client stayed **offline** until a hard refresh.

When a web client opens its channel and the desktop agent isn't connected to the relay yet, the relay replies `agent:offline` **but keeps the socket open** (`ws.ts`). The web client only scheduled a reconnect on the `closed` state — never on `offline` — so it sat on the "🌙 offline" card indefinitely, even after the agent came online seconds later. Only a manual reload (by which point the agent is connected) recovered it.

## Fix

- Retry the channel-open on `offline` too (every 3s) so the view recovers on its own once the agent appears; cancel the pending retry on `ready`.
- Harden `RelayConnection.close()` to detach its socket handlers before closing, so a deliberate close during a retry can't surface a phantom `closed` event and stack duplicate reconnect timers.

## Testing

- `bun run typecheck` + `bun run typecheck:web` clean.
- Deployed to the `bodhi-dev` relay (`https://bodhilander.bodhilabs.dev`) and verified: with the web page open, toggling the desktop's remote hosting off→on flips the web view from the offline card back to the live session list within a few seconds, no refresh.

Scope is entirely `relay/web/` (the browser client) — no desktop-app or relay-server protocol changes.

### DIFF
--- a/relay/web/src/main.ts
+++ b/relay/web/src/main.ts
@@ -2,6 +2,7 @@ import './styles.css';
 import '@xterm/xterm/css/xterm.css';
 import { Terminal } from '@xterm/xterm';
 import { RelayConnection, type ConnState, type Inner } from './connection';
+import { createReconnectScheduler } from './reconnect';
 
 // ---------------------------------------------------------------------------
 // Types (mirror the agent's sealed payloads)
@@ -154,25 +155,20 @@ let lastSessionsJson = '';
 function startPolling() { stopPolling(); pollTimer = setInterval(() => app.conn?.command({ type: 'sessions:list' }), 2500); }
 function stopPolling() { if (pollTimer) { clearInterval(pollTimer); pollTimer = null; } }
 
-let reconnectTimer: ReturnType<typeof setTimeout> | null = null;
 // Re-open the channel after a delay. Used both when the socket drops (`closed`)
 // and when the agent is `offline` — in the offline case the relay keeps our
 // socket open but won't route until the desktop connects, so without this the
 // view stays stuck on "offline" until a manual refresh even after the machine
 // comes online. Retrying makes it recover on its own within a few seconds.
-function scheduleReconnect(delayMs: number) {
-  if (reconnectTimer) return; // one attempt already in flight
-  reconnectTimer = setTimeout(() => {
-    reconnectTimer = null;
-    app.conn?.close();
-    connect();
-  }, delayMs);
-}
+const reconnector = createReconnectScheduler({
+  isAlive: () => app.conn != null, // false ⇒ deliberate teardown, don't resurrect
+  reconnect: () => { app.conn?.close(); connect(); },
+});
 
 function onConnState(s: ConnState, detail?: string) {
   const dot = $('#mdot'); const list = $('#sessions');
   if (s === 'ready') {
-    if (reconnectTimer) { clearTimeout(reconnectTimer); reconnectTimer = null; }
+    reconnector.cancel();
     dot?.classList.remove('off');
     app.conn!.command({ type: 'groups:list' });
     app.conn!.command({ type: 'sessions:list' });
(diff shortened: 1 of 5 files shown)


relay/web/src/main.ts
```diff
@@ -154,22 +154,39 @@ let lastSessionsJson = '';
...
     stopPolling(); dot?.classList.add('off');
     if (list) list.innerHTML = `<li class="empty-note"><div style="font-size:28px">🌙</div><p>This machine is offline. It'll appear here when it reconnects.</p></li>`;
+    scheduleReconnect(3000); // agent not connected yet — keep polling until it appears
   } else if (s === 'error') {
     stopPolling(); dot?.classList.add('off');
     if (list) list.innerHTML = `<li class="empty-note"><div style="font-size:28px">⚠️</div><p>${esc(detail || 'Connection problem.')}</p></li>`;
   } else if (s === 'closed') {
     stopPolling(); dot?.classList.add('off');
-    setTimeout(() => { if (app.conn) connect(); }, 3000); // simple reconnect
+    scheduleReconnect(3000); // socket dropped — reconnect
```
brannon-bowden: scheduleReconnect's timeout callback unconditionally calls connect(), unlike the old closed handler which only reconnected `if (app.conn)`. If app.conn is intentionally nulled elsewhere (disconnect/logout/switch), a previously-scheduled reconnect will still resurrect a connection. Restore the guard.

relay/web/src/main.ts
```diff
@@ -154,22 +154,39 @@ let lastSessionsJson = '';
...
+    if (reconnectTimer) { clearTimeout(reconnectTimer); reconnectTimer = null; }
     dot?.classList.remove('off');
     app.conn!.command({ type: 'groups:list' });
     app.conn!.command({ type: 'sessions:list' });
     startPolling(); // keep the list live (new/removed sessions + state changes)
   } else if (s === 'offline') {
     stopPolling(); dot?.classList.add('off');
     if (list) list.innerHTML = `<li class="empty-note"><div style="font-size:28px">🌙</div><p>This machine is offline. It'll appear here when it reconnects.</p></li>`;
+    scheduleReconnect(3000); // agent not connected yet — keep polling until it appears
   } else if (s === 'error') {
```
brannon-bowden, line 180: The error branch doesn't cancel a pending reconnectTimer (only 'ready' does, at line 175). A reconnect scheduled from a prior offline/closed event can still fire while in an error state, forcing an unwanted retry on persistent errors.

brannon-bowden (changes requested):
This is a solid, well-targeted fix for the reconnect-after-agent-online bug, but the refactor into a shared `scheduleReconnect` helper drops a safety check the old `closed` handler had, and introduces a smaller gap in the `error` path. CI is green (SonarQube, quality-gate, GitGuardian all passing).

## Issues

- **Dropped teardown guard (blocking)** — `relay/web/src/main.ts:189`. The old `closed` handler only reconnected `if (app.conn)`, i.e. it wouldn't recreate a connection if the app had deliberately nulled `app.conn` out (e.g. a disconnect/logout/machine-switch flow). The new `scheduleReconnect` callback (`relay/web/src/main.ts:167-168`) calls `connect()` unconditionally, with no equivalent check. If anything elsewhere sets `app.conn = null` to intentionally tear down, a reconnect timer scheduled just before that (from an `offline` or `closed` event) will still fire ~3s later and silently re-establish a connection the app tried to close. This needs the same guard restored (e.g. bail out of the timeout callback if `app.conn` is null, or clear `reconnectTimer` wherever teardown happens).
- **`error` state doesn't cancel a pending reconnect (major)** — `relay/web/src/main.ts:184-186`. Only the `ready` branch clears `reconnectTimer` (line 175). If an `offline` or `closed` event has already queued a reconnect and the connection then lands in `error` (e.g. an auth failure) before the timer fires, the stale timer still fires and forces a reconnect attempt on top of the error state — potentially causing a persistent error to retry every few seconds instead of surfacing to the user. Consider clearing `reconnectTimer` in the `error` branch too (or making the decision to retry explicit for that state).

## Suggestions (non-blocking)

- The `connection.ts` change (detaching `ws.onopen/onmessage/onclose/onerror` before calling `close()`) is a clean way to stop a deliberate close from being misreported as a `closed` state — no concerns there.
- This PR touches non-trivial reconnect/state-machine logic with no test coverage. Even a lightweight test around `onConnState` transitions (ready cancels timer, offline/closed schedule exactly one reconnect, deliberate `close()` doesn't re-trigger `closed`) would help prevent regressions like the guard drop above.

brannon-bowden (approved):
Both previously-raised issues are resolved by this update, and the fix is now backed by unit tests.

- The dropped teardown guard is restored, and more robustly than before: `createReconnectScheduler` (`relay/web/src/reconnect.ts:29-33`) checks `isAlive()` at *fire* time rather than only at schedule time, so a teardown that happens after a reconnect was queued (the exact race the old guard missed) is now caught too. `main.ts` wires `isAlive: () => app.conn != null` accordingly.
- The `error` path now calls `reconnector.cancel()` before rendering the error state (`relay/web/src/main.ts:179-181`), so a pending offline/closed reconnect can no longer fire on top of a surfaced hard error.
- The new `reconnect.test.ts` covers the fire-once, isAlive-false, cancel, and re-schedule-after-fire cases — directly exercising the scenarios the earlier review flagged as untested.
- CI is green across SonarQube, quality-gate, and GitGuardian.

## Suggestions (non-blocking)

- The scheduler tests use real timers with small delays (10ms/25ms); under heavy CI load these could occasionally flake. Not a concern for this PR, just something to watch if it ever shows up as intermittent CI failures.
- `onConnState`'s state transitions themselves (main.ts) still have no direct test coverage, only the extracted scheduler does. That's a reasonable scope boundary for this fix, but a small integration test wiring `onConnState` to a fake `reconnector`/`app.conn` would close the last gap.

PR #134, opened 2026-07-22
Bug: fix macOS auto-updates that download but never install (#133)

### LINKED ISSUE #133: Bug: macOS auto-update downloads but never installs (ShipIt "App Still Running Error" + model written into signed bundle)
## Summary
On macOS, auto-updates **download successfully but never install** — the app relaunches on the old version, re-detects the same update, and repeats forever. Two independent, confirmed root causes.

## Evidence
electron-log shows the download always completing and being piped to Squirrel.Mac, then nothing. Squirrel.Mac's **own** installer log (`~/Library/Caches/com.bodhilander.app.ShipIt/ShipIt_stderr.log`) shows the real failure, on every attempt:

```
Aborting update attempt because there are 1 running instances of the target app
Installation cancelled: SQRLInstallerErrorDomain Code=-9 "App Still Running Error"
```

Separately, `codesign --verify` on the installed bundle reports a corrupted signature:

```
/Applications/Bodhilander …app: a sealed resource is missing or invalid
  file added: …/Contents/Resources/app.asar.unpacked/dist/main/.cache/Xenova/bge-base-en-v1.5/model_quantized.onnx
  file added: …/.cache/Xenova/bge-base-en-v1.5/tokenizer.json (+ config.json, tokenizer_config.json)
```

## Root cause A — app never terminates, so ShipIt aborts (the "never restarts" symptom)
`src/main/index.ts` `before-quit` calls `event.preventDefault()` and runs async teardown (kill PTYs, dispose vector-search worker, close DB) before `app.quit()`. `autoUpdater.quitAndInstall()` expects a prompt exit; the async cleanup can outlast ShipIt's tolerance (or hang on a wedged embedding worker), so ShipIt sees the app still running and cancels with Code=-9.

**Fix:** guarantee the process exits promptly once quitting begins (bounded cleanup + hard-exit watchdog well inside ShipIt's window). better-sqlite3 WAL is crash-safe, so a forced exit loses no data.

## Root cause B — embedding model downloaded into the signed .app bundle
`@huggingface/transformers` (transformers.js) does **not** read `HF_HOME` / `TRANSFORMERS_CACHE` (those are the Python huggingface_hub vars). The workers set those, so transformers.js falls back to its default `env.cacheDir` — a `.cache` folder next to the module, i.e. **inside the signed app bundle**. Writing there breaks the notarized code-signature seal, corrupting the install and breaking Gatekeeper/update validation. (Per BDHLNDR-126 the indexing worker doesn't even load the model — its cache-env code is dead; the embedding worker is the only downloader.)

**Fix:** set `transformers.env.cacheDir` to the existing writable `userData/huggingface` path; remove the dead/misleading `HF_HOME`/`TRANSFORMERS_CACHE` code.

## Acceptance
- macOS downloaded update installs on quit/restart (ShipIt completes, no Code=-9).
- Freshly installed bundle passes `codesign --verify` after indexing (model cached under `userData`, never inside the `.app`).
- Unit tests for the shutdown watchdog and the cache-dir wiring.

brannon-bowden:
Closes #133

## Problem
On macOS, auto-updates **download successfully but never install** — the app relaunches on the old version, re-detects the same update, and loops forever. Diagnosed from Squirrel.Mac's own installer log (`ShipIt_stderr.log`) and `codesign --verify`, which surfaced **two independent** root causes.

## Root cause 1 — app never terminates, so ShipIt aborts the install
ShipIt cancels every attempt with:
```
Aborting update attempt because there are 1 running instances of the target app
Installation cancelled: SQRLInstallerErrorDomain Code=-9 "App Still Running Error"
```
`before-quit` (`index.ts`) called `event.preventDefault()` and ran async teardown (kill PTYs, dispose the vector-search worker, close the DB) before `app.quit()`. `autoUpdater.quitAndInstall()` needs a **prompt** exit; the teardown could outlast ShipIt's tolerance — or hang on a wedged embedding worker — so ShipIt saw the app still alive and gave up.

**Fix:** new `runGuardedShutdown` runs the same teardown but **guarantees** the process is force-exited (`app.exit(0)`) within a 2s budget regardless of whether cleanup finishes. better-sqlite3 WAL is crash-safe, so the hard exit loses no data. ShipIt handles the post-install relaunch itself.

## Root cause 2 — embedding model written into the signed .app bundle
`codesign --verify` on the installed app reported `a sealed resource is missing or invalid`, caused by:
```
file added: …/Contents/Resources/app.asar.unpacked/dist/main/.cache/Xenova/bge-base-en-v1.5/model_quantized.onnx (+ tokenizer/config json)
```
The workers set `HF_HOME` / `TRANSFORMERS_CACHE`, but **transformers.js ignores those** (they're Python `huggingface_hub` vars). It fell back to its default `env.cacheDir` — a `.cache` folder next to the module, i.e. **inside the signed bundle** — breaking the notarized signature seal (which also breaks Gatekeeper and update validation).

**Fix:** set `transformers.env.cacheDir` to the existing writable `userData/huggingface` path via a new `setModelCacheDir()`. Removed the dead cache-env code from the indexing worker, which per BDHLNDR-126 never loads the model at all (it RPCs embed requests to the single embedding worker).

## Tests
New electron-free, unit-tested modules:
- `shutdown.ts` — 4 tests: forces exit once on normal completion, on a hung cleanup (within budget), on a rejected cleanup, and never double-fires.
- `model-cache.ts` — 4 tests: sets `env.cacheDir`, keeps remote models allowed, does **not** touch HF_HOME/TRANSFORMERS_CACHE (regression guard), null-safe.

## Verification
- `bun test src/` — 8 new tests pass; no new failures (1 pre-existing failure is `@xterm/headless` not installed locally, identical on `development`).
- `tsc -p tsconfig.main.json` — no new type errors (4 pre-existing missing-module errors, unrelated).
- Both workers bundle cleanly via esbuild.

> Note: this fixes the code going forward. An already-corrupted local install (broken signature) needs a one-time clean reinstall to recover — the fix prevents recurrence.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -2,6 +2,7 @@ import { app, BrowserWindow, ipcMain, dialog, shell, crashReporter } from 'elect
 import * as fs from 'fs';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
+import { runGuardedShutdown } from './shutdown';
 import { resolveLaunchProviderId } from './providers';
 import { startProviderInstall, cancelProviderInstall } from './provider-install';
 import { detectProviders } from './provider-detector';
@@ -1470,30 +1471,45 @@ app.on('activate', () => {
 
 let cleanupComplete = false;
 
+// Squirrel.Mac's ShipIt cancels an in-progress update with "App Still Running
+// Error" (Code=-9) if this process is still alive shortly after quitAndInstall,
+// so macOS auto-updates downloaded but never installed (issue #133). The
+// teardown below can hang on a wedged native worker, so it MUST NOT gate the
+// process exit — runGuardedShutdown force-exits within the budget regardless.
+// Comfortably inside ShipIt's tolerance; better-sqlite3 WAL is crash-safe.
+const QUIT_CLEANUP_BUDGET_MS = 2000;
+
 app.on('before-quit', (event) => {
   if (cleanupComplete) return;
 
   event.preventDefault();
   isQuitting = true;
 
-  (async () => {
-    try {
-      await ptyManager.killAll();
-    } catch (e) {
-      log.error('Error killing PTYs on quit:', e);
-    }
+  runGuardedShutdown({
+    budgetMs: QUIT_CLEANUP_BUDGET_MS,
+    log: (msg) => log.info(msg),
+    forceExit: () => {
+      cleanupComplete = true;
+      // Hard exit guarantees the PID is gone for ShipIt; ShipIt handles the
+      // post-install relaunch itself, so app.quit() vs app.exit() is moot here.
+      app.exit(0);
+    },
+    cleanup: async () => {
+      try {
+        await ptyManager.killAll();
+      } catch (e) {
+        log.error('Error killing PTYs on quit:', e);
+      }
 
-    disposeVectorSearchManager();
-    trayManager.destroy();
-    stateMonitor?.stop();
-    try {
-      getRelayClient().stop();
-    } catch (e) {
-      log.error('Error stopping relay client on quit:', e);
-    }
-    closeDatabase();
-
-    cleanupComplete = true;
-    app.quit();
-  })();
+      disposeVectorSearchManager();
+      trayManager.destroy();
+      stateMonitor?.stop();
+      try {
+        getRelayClient().stop();
+      } catch (e) {
+        log.error('Error stopping relay client on quit:', e);
+      }
+      closeDatabase();
+    },
+  });
 });
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1469,7 +1469,11 @@ app.on('activate', () => {
   }
 });
 
-let cleanupComplete = false;
+// Set synchronously at the top of the handler (not inside forceExit, which may
+// not fire for up to QUIT_CLEANUP_BUDGET_MS) so a second before-quit within that
+// window — e.g. a tray-quit racing quitAndInstall — can't start a second
+// concurrent teardown and double-run killAll/closeDatabase/etc.
+let shuttingDown = false;
 
 // Squirrel.Mac's ShipIt cancels an in-progress update with "App Still Running
 // Error" (Code=-9) if this process is still alive shortly after quitAndInstall,
@@ -1480,7 +1484,8 @@ let cleanupComplete = false;
 const QUIT_CLEANUP_BUDGET_MS = 2000;
 
 app.on('before-quit', (event) => {
-  if (cleanupComplete) return;
+  if (shuttingDown) return;
+  shuttingDown = true;
 
   event.preventDefault();
   isQuitting = true;
@@ -1488,12 +1493,13 @@ app.on('before-quit', (event) => {
   runGuardedShutdown({
     budgetMs: QUIT_CLEANUP_BUDGET_MS,
     log: (msg) => log.info(msg),
-    forceExit: () => {
-      cleanupComplete = true;
-      // Hard exit guarantees the PID is gone for ShipIt; ShipIt handles the
-      // post-install relaunch itself, so app.quit() vs app.exit() is moot here.
-      app.exit(0);
-    },
+    // Hard exit guarantees the PID is gone for ShipIt; ShipIt handles the
+    // post-install relaunch itself, so app.quit() vs app.exit() is moot here.
+    // On the timeout path this also cuts off any teardown still in flight —
+    // notably ptyManager.killAll(), so child terminals may be orphaned to the
+    // OS. That's an accepted tradeoff: a stuck teardown blocking the update
+    // forever is worse, and the OS reaps orphaned PTYs on session end.
+    forceExit: () => app.exit(0),
     cleanup: async () => {
       try {
         await ptyManager.killAll();
--- a/src/main/vector-search/embedding-provider.ts
+++ b/src/main/vector-search/embedding-provider.ts
@@ -5,7 +5,7 @@ import os from 'os';
 // Import types for TypeScript, runtime value loaded dynamically
 import type { FeatureExtractionPipeline as FeatureExtractionPipelineType } from '@huggingface/transformers';
 
-import { applyModelCacheDir, TransformersEnvLike } from './model-cache';
+import { applyModelCacheDir, type TransformersEnvLike } from './model-cache';
 
 let pipeline: typeof import('@huggingface/transformers').pipeline;
 
--- a/src/main/vector-search/embedding-provider.ts
+++ b/src/main/vector-search/embedding-provider.ts
@@ -5,7 +5,13 @@ import os from 'os';
 // Import types for TypeScript, runtime value loaded dynamically
 import type { FeatureExtractionPipeline as FeatureExtractionPipelineType } from '@huggingface/transformers';
 
+import { applyModelCacheDir, TransformersEnvLike } from './model-cache';
+
 let pipeline: typeof import('@huggingface/transformers').pipeline;
+
+// The live transformers.js `env` object, captured at module load so we can
+// point its model cache at a writable, out-of-bundle directory (issue #133).
+let transformersEnv: TransformersEnvLike | null = null;
 
 // Cap ONNX threads so indexing doesn't saturate every CPU core.
 // Default onnxruntime behavior is intraOpNumThreads = physical cores, which
@@ -29,6 +35,7 @@ const MAX_EMBED_CHARS = 2048;
 try {
   const transformers = require('@huggingface/transformers');
   pipeline = transformers.pipeline;
+  transformersEnv = transformers.env ?? null;
   // WASM numThreads matters when transformers falls back to the WASM backend;
   // harmless when onnxruntime-node is in use.
   if (transformers.env?.backends?.onnx?.wasm) {
@@ -40,6 +47,18 @@ try {
 } catch (e) {
   console.error('[EmbeddingProvider] Failed to load @huggingface/transformers:', e);
   throw e;
+}
+
+/**
+ * Point transformers.js at a writable, out-of-bundle model cache (issue #133).
+ * MUST be called before the first pipeline() load. Callers pass a directory
+ * under app.getPath('userData'); transformers.js downloads to <dir>/<model_id>.
+ * Replaces the old HF_HOME / TRANSFORMERS_CACHE env vars, which transformers.js
+ * ignores — leaving them unset made the model download into the signed .app
+ * bundle and break its code-signature seal.
+ */
+export function setModelCacheDir(cacheDir: string): void {
+  applyModelCacheDir(transformersEnv, cacheDir);
 }
 
 // Alias for use in class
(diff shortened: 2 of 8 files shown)


src/main/vector-search/indexing-worker.ts
```diff
@@ -149,16 +149,11 @@ parentPort?.on('message', async (message: WorkerMessage) => {
...
-    if (message.cacheDir) {
-      try {
-        // Set environment variable for @huggingface/transformers cache
-        process.env.HF_HOME = message.cacheDir;
-        process.env.TRANSFORMERS_CACHE = path.join(message.cacheDir, 'models');
-      } catch (e) {
-        console.warn('[IndexingWorker] Failed to set cache dir:', e);
-      }
-    }
+    // No model cache setup here (issue #133): per BDHLNDR-126 this worker never
```
William-Long-II, line 152: The only shown use of `path` (path.join for the old cache dir) was removed here, but the `import * as path from 'path'` at the top of the file wasn't. If path has no other use in this file, this is now a dead import — likely relevant to the failing SonarQube/quality-gate checks. Please verify and remove if unused.

src/main/index.ts
```diff
@@ -1470,30 +1471,45 @@ app.on('activate', () => {
...
 
+// Squirrel.Mac's ShipIt cancels an in-progress update with "App Still Running
+// Error" (Code=-9) if this process is still alive shortly after quitAndInstall,
+// so macOS auto-updates downloaded but never installed (issue #133). The
+// teardown below can hang on a wedged native worker, so it MUST NOT gate the
+// process exit — runGuardedShutdown force-exits within the budget regardless.
+// Comfortably inside ShipIt's tolerance; better-sqlite3 WAL is crash-safe.
+const QUIT_CLEANUP_BUDGET_MS = 2000;
+
 app.on('before-quit', (event) => {
```
William-Long-II, line 1486: cleanupComplete is only set true inside forceExit, which can take up to QUIT_CLEANUP_BUDGET_MS (2000ms) to fire. A second before-quit event within that window bypasses this guard and starts a second concurrent runGuardedShutdown, double-running ptyManager.killAll(), trayManager.destroy(), stateMonitor.stop(), getRelayClient().stop(), and closeDatabase(). Set a synchronous guard flag at the top of the handler, before calling runGuardedShutdown, instead of relying on cleanupComplete.

src/main/vector-search/embedding-provider.ts
```diff
@@ -5,8 +5,14 @@ import os from 'os';
 // Import types for TypeScript, runtime value loaded dynamically
 import type { FeatureExtractionPipeline as FeatureExtractionPipelineType } from '@huggingface/transformers';
 
+import { applyModelCacheDir, TransformersEnvLike } from './model-cache';
```
William-Long-II: TransformersEnvLike is a type-only interface but is imported as a regular value import here, inconsistent with the `import type { ... }` used one line above in this same file. Use `import { applyModelCacheDir, type TransformersEnvLike } from './model-cache';` to avoid any risk under single-file/esbuild-style transpilation.

src/main/index.ts
```diff
@@ -1470,30 +1471,45 @@ app.on('activate', () => {
 
 let cleanupComplete = false;
 
+// Squirrel.Mac's ShipIt cancels an in-progress update with "App Still Running
+// Error" (Code=-9) if this process is still alive shortly after quitAndInstall,
+// so macOS auto-updates downloaded but never installed (issue #133). The
+// teardown below can hang on a wedged native worker, so it MUST NOT gate the
+// process exit — runGuardedShutdown force-exits within the budget regardless.
```
William-Long-II, line 1482: This comment justifies the hard-exit-before-cleanup-completes path only via better-sqlite3's WAL crash-safety. It doesn't mention that the same forced exit can also cut off ptyManager.killAll() mid-flight, potentially orphaning child processes. Worth calling out explicitly if that's an accepted tradeoff.

William-Long-II (changes requested):
Let me rewrite this in a more natural reviewer voice while keeping every finding intact.

The failing CI first: `indexing-worker.ts` still imports `path` at the top, but the diff yanks the only usage I can see (`path.join(message.cacheDir, 'models')`), unlike `embedding-worker.ts` where you correctly dropped the now-dead `path` import. If that was `path`'s only use in the file this is a dead import, and given SonarQube/quality-gate are both red on this PR right now, that's a real candidate. Confirm `path` isn't used elsewhere in the file and kill the import if not — don't merge on a red gate without ruling this out first.

`runGuardedShutdown`'s "exactly once" claim doesn't actually hold if `before-quit` fires twice. `cleanupComplete` only flips inside `forceExit`, which only runs once `cleanup()` settles or the watchdog trips — so a double Cmd+Q, or `quitAndInstall()` racing a manual quit, can start a second independent `runGuardedShutdown` with its own `cleanup` closure and `exited` flag while the first is still in flight. That means `ptyManager.killAll()`, `trayManager.destroy()`, `stateMonitor?.stop()`, `getRelayClient().stop()`, `closeDatabase()` all become double-run candidates, plus two independent `app.exit(0)` calls. This isn't new — the old code had the same hole — but this PR is the one that turns the race window into a guaranteed ~2000ms every single time on the hang path, instead of "however long teardown happened to take." Probably not catastrophic in practice (the `.catch(() => {})` in shutdown.ts eats a double `closeDatabase()` throw, second `app.exit(0)` is a no-op post-exit), but it defeats the documented "invoked exactly once" contract, and duplicated PTY/Tray/DB teardown shouldn't be running twice by luck. Fix is small: set a synchronous flag at the top of the `before-quit` handler itself, before calling `runGuardedShutdown`, rather than reusing `cleanupComplete`.

nit in `embedding-provider.ts:8` — `import { applyModelCacheDir, TransformersEnvLike } from './model-cache';`. `TransformersEnvLike` is a pure interface used only in type position, and the line directly above it already does this correctly with `import type { FeatureExtractionPipeline as ... }`. Harmless under tsc, but if the main-process build does per-file transpilation without cross-file type checking (common in Electron+Vite), a named import with no runtime export can blow up at load time. Should be `import { applyModelCacheDir, type TransformersEnvLike } from './model-cache';` — free fix.

The watchdog comment justifies the hard exit by saying "better-sqlite3 WAL is crash-safe," which is true for the DB, but says nothing about `ptyManager.killAll()` also getting cut off on the same timeout path — that can orphan child shell processes. Probably a fine tradeoff (guaranteed install beats a rare orphaned process) but the comment should say that instead of reading like the crash-safety analysis is complete.

Smaller one: `applyModelCacheDir` in `model-cache.ts` unconditionally sets `env.allowRemoteModels = true` every call, clobbering whatever was there before. Nothing sets it to `false` today so it's not biting anyone yet, but it's a footgun waiting for whoever adds an offline mode later.

And the "MUST be called before the first `pipeline()` load" invariant in embedding-provider/embedding-worker is only true because `worker_threads` messages are FIFO and the caller happens to send `init` before `embed`. Nothing here enforces it — a comment or assertion pointing at where that ordering is guaranteed would save the next reader some digging.

Good bones on this one though — pulling `runGuardedShutdown` and `applyModelCacheDir` out into pure, electron-free functions is exactly right, and the tests aren't decorative: shutdown.test.ts actually re-creates the hang/reject/late-resolve cases, and model-cache.test.ts specifically guards against the HF_HOME/TRANSFORMERS_CACHE regression that caused this bug in the first place. Root-cause comments (ShipIt's Code=-9, transformers.js ignoring Python-only env vars, code-signature sealing) are accurate and will save someone a debugging session later. Dropping the dead HF_HOME/TRANSFORMERS_CACHE writes from indexing-worker.ts is correct too since that worker doesn't own model loading anymore.

src/main/vector-search/indexing-worker.ts
brannon-bowden: Verified — `path` is still used at `indexing-worker.ts:214` (`path.extname(file.path)`), so the import isn't dead. Leaving it. The SonarQube failure was unrelated: 4 `S5443` hotspots from `/tmp/cache` literals in `model-cache.test.ts`, now removed.

src/main/index.ts
brannon-bowden: Fixed in 1891051. Added a synchronous `shuttingDown` flag set at the top of the handler before `runGuardedShutdown`, so a second before-quit inside the cleanup/2s window can't start a concurrent teardown. Replaced `cleanupComplete` entirely — since forceExit uses `app.exit(0)` (which never re-fires before-quit) it was also redundant.

src/main/index.ts
brannon-bowden: Fixed in 1891051. Expanded the comment to call out that on the timeout path the forced exit can cut off `ptyManager.killAll()` mid-flight, orphaning child terminals (reaped by the OS on session end), and that this is an accepted tradeoff versus a stuck teardown blocking the update indefinitely.

src/main/vector-search/embedding-provider.ts
brannon-bowden: Fixed in 1891051: `import { applyModelCacheDir, type TransformersEnvLike } from './model-cache';`

William-Long-II (approved):
Good delta, both real issues from last round are actually fixed.

The double-teardown race is closed properly: `shuttingDown` is now set synchronously as the first statement in the `before-quit` handler, before `event.preventDefault()` and before `runGuardedShutdown` is even called. That's exactly what I asked for last time — a second `before-quit` firing mid-teardown (tray-quit racing `quitAndInstall()`) now bails out immediately instead of kicking off a second concurrent `ptyManager.killAll()` / `closeDatabase()`.

CI is green now too — SonarQube and the quality gate both pass, so the suspected dead `path` import in `indexing-worker.ts` that had me withholding approval on a red gate is no longer a blocker either way.

Both nits are handled: `TransformersEnvLike` now comes in via `import { applyModelCacheDir, type TransformersEnvLike } from './model-cache'` instead of a runtime import of a type-only symbol, and the `forceExit` comment now spells out that the timeout path can cut off `ptyManager.killAll()` and orphan child processes rather than implying that case was covered.

Test cleanup — swapping the `/tmp/cache` literals for a proper `Application Support`-style `CACHE_DIR` constant — is a fine change too. `/tmp` in a cache-dir test both risks a SonarQube S5443 flag and isn't where the app actually caches models. Assertions are unchanged (cacheDir passthrough, `allowRemoteModels` forced true, HF_HOME/TRANSFORMERS_CACHE left alone, null/undefined no-op), so nothing got weaker.

Still open, but in files this diff doesn't touch and already flagged as non-blocking: `applyModelCacheDir` unconditionally sets `env.allowRemoteModels = true` on every call, which will bite whenever offline mode shows up. And the "`init` must precede the first `pipeline()` load" invariant is still only guaranteed by FIFO worker-message delivery with nothing pointing at it in comments or an assertion.

Nothing new here introduces a defect — flag hoisting is a straightforward, correctly-ordered synchronous check, test changes are behavior-preserving.

Approve. Both things that had me at request-changes last round are resolved — race fix verified in the diff, CI status verified directly.

PR #136, opened 2026-07-22
feat(relay): dynamic terminal sizing + clean mobile viewer

William-Long-II:
## Problem

The mobile web viewer mirrored the desktop's exact terminal grid (often **184 columns**). On a phone that forces a lose-lose: **scale down** → unreadably small, or **reflow to a narrow grid** → Claude's absolute cursor moves get clamped onto the last column and the output garbles (characters piled at the right edge).

## Fix — dynamic sizing ("active viewer drives the size", tmux-style)

- **Web** (`relay/web`): the phone reports its own grid via `terminal:resize`, so **Claude redraws to fit the phone** — clean, readable, no garble. It renders at whatever size the PTY reports (`terminal:size`); a fallback CSS scale keeps it clean when the desktop transiently owns a larger size. Vertical scrolling is a touch-drag → `term.scrollLines` handler (reliable on mobile where the native viewport won't touch-scroll). WebGL renderer + a resolved monospace stack fix sub-pixel cell drift.
- **Agent** (`session-tunnel.ts`): apply `terminal:resize` → `ptyManager.resize`; forward PTY resize events to subscribed viewers as `terminal:size`.
- **pty-manager**: emit a `resize` event; raise the remote replay buffer **100 KB → 2 MB** (100 KB was ~one screen on a wide terminal, leaving the phone nothing to scroll) with amortized trimming.
- **Desktop renderer** (`Terminal.tsx`): re-assert its own fit size on **window focus**, so the desktop snaps back to full width when you return to it.

## Behavior

- On the phone, Claude reflows to the phone's width → readable, no panning, no garble.
- While you're on the phone, the desktop terminal reflows to the phone's width; it snaps back the instant the desktop window regains focus. Standard multiplexer behavior — not a permanent restriction.

## Known follow-up

Pre-existing scrollback was captured at the old width, so scrolling into *old* history can reflow imperfectly. A rendered-text history snapshot (via the already-present `@xterm/headless`) is the planned follow-up. The **live view — the primary interaction — is now clean and readable.**

## Testing

- `tsc -p tsconfig.main.json`, base `tsc`, and relay `typecheck:web` all clean.
- Web deployed to the `bodhi-dev` relay. Full end-to-end needs the app agent, so this rides a beta build.

### DIFF
--- a/relay/web/src/styles.css
+++ b/relay/web/src/styles.css
@@ -115,16 +115,13 @@ input, textarea { font-family: inherit; }
 .term-title { min-width: 0; }
 .term-title .t-name { font-family: var(--mono); font-weight: 600; font-size: 14px; white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
 .term-title .t-meta { font-size: 11px; color: var(--muted); white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
-/* The viewer matches the desktop's terminal size, which is often wider/taller
-   than the phone. Vertical scrolling — including through scrollback HISTORY —
-   is xterm's OWN viewport: xterm keeps scrollback off-DOM and only reveals it
-   when its viewport scrolls. On desktop that happens via wheel events (works
-   regardless of CSS overflow); on TOUCH devices the viewport must be natively
-   scrollable, so it needs overflow-y:auto + touch-action:pan-y. Neutralizing it
-   (overflow:hidden) is exactly what left mobile unable to reach history.
-   We bound .xterm/.viewport to the pane height so the viewport becomes a real
-   scroll window over the taller content; .screen handles horizontal panning. */
-.screen { flex: 1; min-height: 0; overflow: hidden; padding: 4px 0; background: var(--bg); }
+/* On desktop the viewer mirrors the desktop's terminal size, which can be wider
+   or taller than the browser window — so .screen scrolls (both axes) as the
+   escape hatch. On mobile (<860px) we instead use dynamic sizing (Claude reflows
+   to the phone) + a scale fallback + touch→scrollLines, so .screen is clipped and
+   the scroll is handled by those; see the max-width:859px override below. */
+.screen { flex: 1; min-height: 0; overflow: auto; padding: 4px 0; background: var(--bg); }
+@media (max-width: 859px) { .screen { overflow: hidden; } }
 /* xterm's own viewport is the vertical scroller (scrollback lives here). It must
    stay natively scrollable so TOUCH works — the desktop path uses wheel events,
    but mobile needs a real scroll container. */
--- a/relay/web/src/styles.css
+++ b/relay/web/src/styles.css
@@ -115,15 +115,20 @@ input, textarea { font-family: inherit; }
 .term-title { min-width: 0; }
 .term-title .t-name { font-family: var(--mono); font-weight: 600; font-size: 14px; white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
 .term-title .t-meta { font-size: 11px; color: var(--muted); white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
-/* The viewer matches the desktop's terminal size, which is often bigger than
-   the phone. Rather than clamp xterm to the container (clipping the bottom
-   rows) and rely on xterm's internal viewport for vertical scroll (which
-   doesn't respond to touch), let .screen scroll BOTH axes over the terminal
-   at its natural height — the same mechanism that already made horizontal
-   scroll work. xterm's own viewport is neutralized so it doesn't intercept. */
-.screen { flex: 1; min-height: 0; overflow: auto; -webkit-overflow-scrolling: touch; overscroll-behavior: contain; padding: 8px 6px 4px; background: var(--bg); }
-.screen .xterm { padding: 0 6px; }
-.xterm-viewport { overflow: hidden; overscroll-behavior: contain; }
+/* The viewer matches the desktop's terminal size, which is often wider/taller
+   than the phone. Vertical scrolling — including through scrollback HISTORY —
+   is xterm's OWN viewport: xterm keeps scrollback off-DOM and only reveals it
+   when its viewport scrolls. On desktop that happens via wheel events (works
+   regardless of CSS overflow); on TOUCH devices the viewport must be natively
+   scrollable, so it needs overflow-y:auto + touch-action:pan-y. Neutralizing it
+   (overflow:hidden) is exactly what left mobile unable to reach history.
+   We bound .xterm/.viewport to the pane height so the viewport becomes a real
+   scroll window over the taller content; .screen handles horizontal panning. */
+.screen { flex: 1; min-height: 0; overflow: hidden; padding: 4px 0; background: var(--bg); }
+/* xterm's own viewport is the vertical scroller (scrollback lives here). It must
+   stay natively scrollable so TOUCH works — the desktop path uses wheel events,
+   but mobile needs a real scroll container. */
+.xterm-viewport { overflow-y: auto; overflow-x: hidden; -webkit-overflow-scrolling: touch; touch-action: pan-y; overscroll-behavior: contain; }
 .xterm-viewport::-webkit-scrollbar { width: 8px; }
 .compose { border-top: 1px solid var(--hair); background: var(--surface); }
 .keys { display: flex; gap: 6px; padding: 8px 10px; overflow-x: auto; scrollbar-width: none; }
--- a/src/main/api/relay/session-tunnel.ts
+++ b/src/main/api/relay/session-tunnel.ts
@@ -138,7 +138,7 @@ export class SessionTunnel {
         this.handleSessionCreate(clientId, inner);
         break;
       case 'terminal:subscribe':
-        this.handleSubscribe(clientId, s, inner);
+        void this.handleSubscribe(clientId, s, inner);
         break;
       case 'terminal:unsubscribe':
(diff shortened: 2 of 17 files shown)


src/main/api/relay/session-tunnel.ts
```diff
@@ -140,9 +148,13 @@ export class SessionTunnel {
...
-        // Intentionally ignored — the desktop owns the terminal size, so a
-        // mobile viewer never reflows the shared PTY. Viewers match the size
-        // reported by terminal:size.
+        // Dynamic sizing: the active viewer drives the shared PTY size, so a
+        // phone can reflow Claude to fit its screen. The desktop re-asserts its
+        // own size when it regains focus. The client is authenticated as the
+        // machine owner, so it's allowed to resize.
+        if (typeof inner.sessionId === 'string' && typeof inner.cols === 'number' && typeof inner.rows === 'number') {
+          ptyManager.resize(inner.sessionId, Math.max(2, Math.floor(inner.cols)), Math.max(2, Math.floor(inner.rows)));
+        }
```
brannon-bowden, line 163: No upper bound / finiteness check on remote-supplied cols/rows before calling ptyManager.resize. Only Math.max(2, ...) is applied. This previously-ignored command is now live network-triggered input into a native pty resize call — clamp to a sane maximum and reject non-finite values.

relay/web/src/styles.css
```diff
@@ -115,15 +115,20 @@ input, textarea { font-family: inherit; }
...
+/* The viewer matches the desktop's terminal size, which is often wider/taller
+   than the phone. Vertical scrolling — including through scrollback HISTORY —
+   is xterm's OWN viewport: xterm keeps scrollback off-DOM and only reveals it
+   when its viewport scrolls. On desktop that happens via wheel events (works
+   regardless of CSS overflow); on TOUCH devices the viewport must be natively
+   scrollable, so it needs overflow-y:auto + touch-action:pan-y. Neutralizing it
+   (overflow:hidden) is exactly what left mobile unable to reach history.
+   We bound .xterm/.viewport to the pane height so the viewport becomes a real
+   scroll window over the taller content; .screen handles horizontal panning. */
+.screen { flex: 1; min-height: 0; overflow: hidden; padding: 4px 0; background: var(--bg); }
```
brannon-bowden: Comment says '.screen handles horizontal panning' but .screen is now overflow:hidden (was overflow:auto) and .xterm-viewport is overflow-x:hidden too — there's no horizontal scroll escape hatch left for viewports above the 859px mobile breakpoint where scaleTerm()/dynamic sizing don't apply.

brannon-bowden (changes requested):
## Verdict

Request changes — the dynamic PTY resize path introduced in `session-tunnel.ts` opens a new remote-controlled trust boundary without an upper-bound check, and none of the newly-enabled remote-resize behavior has any test coverage.

## Blocking issues

- **`src/main/api/relay/session-tunnel.ts:157`** — The new `terminal:resize` handler flips this command from "intentionally ignored" (explicitly for safety — the desktop used to be the sole owner of PTY size) to actively calling `ptyManager.resize(inner.sessionId, Math.max(2, Math.floor(inner.cols)), Math.max(2, Math.floor(inner.rows)))` on attacker/remote-supplied input. Only a floor (`Math.max(2, …)`) is enforced — there is no ceiling and no finiteness check. A malicious or buggy relay client (any phone/browser session that completed the handshake) can request a pathological size (e.g. `cols: 999999999`, or a huge-but-finite JSON number), which is handed straight to the native pty resize call. Depending on how the underlying binding handles out-of-range values, this risks a crash or excessive memory allocation in the shared Electron process — affecting every other attached viewer of that machine, not just the requester. This is exactly the kind of input this file used to refuse on purpose; please clamp to a sane maximum (e.g. a few hundred/thousand cols and rows) and reject non-finite values before this reaches `ptyManager.resize`.

- **No tests for the new remote-resize authority.** This PR flips a previously-refused, security-relevant command (`terminal:resize`) into one that mutates a shared PTY based on untrusted network input, adds a new `resize` event fan-out through `pty-manager.ts` → `session-tunnel.ts` → connected viewers, and changes the effective owner of terminal dimensions at runtime. None of this — nor the bounds-clamping behavior, nor the subscribe-gated broadcast in `onResize` — has any accompanying test. Given the file is explicitly flagged sensitive and the behavior is a genuine authorization/behavior change (not a rendering tweak), this needs coverage before merge.

## Non-blocking issues

- **`relay/web/src/styles.css:118-132`** — The comment says "`.screen` handles horizontal panning," but `.screen` is now `overflow: hidden` (previously `overflow: auto`), and `.xterm-viewport` is `overflow-x: hidden` too. For viewports above the 859px `isMobileView()` breakpoint (where `scaleTerm()`/dynamic sizing don't apply), there is no longer any way to horizontally scroll a PTY that's wider than the browser window — the old `overflow: auto` escape hatch is gone. If this is meant to be a mobile-only concern, please confirm nothing above 859px can end up with a desktop-owned wide terminal in a narrower browser window; otherwise this is a real clipping regression.
- The `terminal:resize`/`terminal:input` handlers don't verify the requesting client is actually subscribed to `sessionId` (mirrors pre-existing `terminal:input` behavior, so likely an accepted "whole machine owner" trust model) — but now that resize is "live" instead of ignored, it's worth explicitly confirming that's still the intended model rather than scoping actions to subscribed sessions.
- `startScrollDebug()` (`relay/web/src/main.ts`, near the bottom) is explicitly labeled "TEMP diagnostic" and ships to production behind a `#debug` hash. Low risk, but if it's meant to stay temporary it should be removed or tracked, not merged as-is.

## Tests

- No tests were added or modified for: the remote `terminal:resize` authorization change, the `resize` event fan-out in `pty-manager.ts`/`session-tunnel.ts`, the 20x scrollback size increase, or the mobile viewer's dynamic sizing/scale/touch-scroll logic in `main.ts`. At minimum the resize-authorization change (the sensitive-path item above) should get unit coverage (bounds clamping, subscription gating, event fan-out to the right client only).

William-Long-II:
Scope expanded (still one cohesive mobile-terminal feature for the beta). Added two commits on top of dynamic sizing:

**Rendered-text history** — remote viewers replayed raw scrollback (width-locked → garbles when a phone reflows). Now the agent replays it through a headless xterm at the session's current width and serializes to resolved text (`@xterm/addon-serialize`), which reflows cleanly. On-demand at subscribe; falls back to raw on error.

**Desktop "resized for mobile" banner + Resume** — when a phone shrinks the shared PTY, the desktop follows locally and shows a banner with a **Resume desktop size** button (also auto-resumes on window focus). New `pty:resized` IPC main→renderer.

New dependency: `@xterm/addon-serialize` (`@xterm/headless` was already present).

brannon-bowden (approved):
## Verdict

Approve — both previously-blocking issues are resolved by this update.

The unbounded-resize security gap is closed: `terminal:resize` now runs through `sanitizeSize()` (`src/main/api/relay/terminal-size.ts`), which rejects non-numeric/non-finite input and clamps into `[MIN_DIM, MAX_COLS/MAX_ROWS]` (2–1000) before anything reaches `ptyManager.resize`. That function has thorough unit tests (`terminal-size.test.ts`) covering the floor, the new ceiling, non-finite rejection, and malformed shapes. As a bonus, both `terminal:resize` and `terminal:input` are now scoped to `s.subs.has(sessionId)`, which was previously only a "worth confirming" note — it's now enforced as defense-in-depth.

The test-coverage gap is substantially closed too: `pty-manager.test.ts` adds coverage for `resize()` (drives the pty, fans out the `resize` event, is a no-op on unchanged dimensions or unknown sessions) and for the new `getSerializedBuffer()` rendered-scrollback path (retains lines that scrolled off-screen, empty-string fallback for unknown sessions). Combined with the `sanitizeSize` tests, the security-relevant logic in this path — clamping, event fan-out, buffer serialization — now has real coverage.

The mobile CSS clipping regression flagged last round is also fixed: `.screen` reverts to `overflow: auto` by default and is only clipped (`overflow: hidden`) under the `max-width: 859px` mobile breakpoint, so desktop viewports get their horizontal/vertical scroll escape hatch back. The temporary `startScrollDebug()` diagnostic has been removed as suggested.

## Blocking issues

None.

## Non-blocking issues

- **`src/main/api/relay/session-tunnel.ts`** — the new subscription-gating (`s.subs.has(inner.sessionId)`) on `terminal:resize`/`terminal:input` and the dispatcher's integration with `sanitizeSize` don't have a direct test at the `SessionTunnel` level; coverage currently comes from testing `sanitizeSize` and `PtyManager.resize` in isolation. The glue code is simple and readable, so this isn't blocking, but an integration test exercising `handleFrame` end-to-end (unsubscribed client denied, subscribed client's oversized request clamped) would close the loop.
- **`src/main/pty-manager.ts:~840`** (`getSerializedBuffer`) — a new headless `Terminal` + `SerializeAddon` is constructed on every `terminal:subscribe`, replaying the full raw scrollback each time. For sessions with large scrollback this runs on every reconnect/subscribe; worth keeping an eye on if reconnect churn is common, though not a correctness issue.
- Minor formatting: a duplicate blank line was introduced in `session-tunnel.ts` right after `expandHome()` (cosmetic only).

## Tests

Core new logic (`sanitizeSize`, `PtyManager.resize` event fan-out, `getSerializedBuffer`) is now covered. The remaining gap is dispatch-level coverage in `session-tunnel.ts` itself (subscription gating for `resize`/`input`, and the async `handleSubscribe` re-check of `s.subs.has(sessionId)` after `await`) — worth adding if this file gets its own test harness, but not a merge blocker given the logic is thin and the underlying pieces are tested.

PR #137, opened 2026-07-22
chore: remove splash screen and the crashing code-indexing feature

William-Long-II:
Two removals requested ahead of the beta.

## Splash screen
Deleted `splash.html`, the splash `BrowserWindow` and its startup delay (the main window now shows on `ready-to-show`), the `showSplash`/`splashDuration` preferences, and the Settings **Appearance** tab — which contained *only* those two settings, so it's gone too.

## Code indexing / semantic code search — removed entirely
The ONNX/onnxruntime embedding worker was the source of the recurring indexing crash-loop (`docs/designs/BDHLNDR-40`), and search is unusable without the index, so the whole feature is removed:

- `src/main/vector-search/*` (parser, embedding service/worker, indexing worker, file discovery, model cache), the code-search repository and API route
- vector-search IPC handlers, progress event forwarding, preload surface
- sqlite-vec extension loading and `code_indexes`/`code_chunks` table setup — **existing tables are left in place, no destructive migration**
- `CodeSearchModal`, `IndexStatus`, `useCodeSearch`, the `Ctrl/Cmd+Shift+F` shortcut + toolbar button, and the code-search types in `shared/types`
- the five MCP tools (`search_code`, `find_symbol`, `get_index_status`, `list_indexes`, `start_indexing`) and the memory-injector "code-search-available" hint
- `build:worker` / `build:embedding-worker` build steps, the `asarUnpack` entries, and the now-unused deps: **`@huggingface/transformers` (onnxruntime), `tree-sitter*`, `sqlite-vec`, `ignore`, `chokidar`**

README updated for both.

## Verification
- **Full clean build succeeds** (`rm -rf dist && bun run build`) — confirmed no vector-search worker and no splash in the output.
- `tsc` clean for main, renderer, and MCP.
- Test suite unchanged: the single `ChatParser > 06-response` snapshot failure **pre-exists on `development`** (verified by running it there).


brannon-bowden (approved):
This is a clean, well-scoped removal of the splash screen and the semantic code-indexing feature (workers, DB tables, IPC/API surface, MCP tools, renderer UI, deps, build scripts) with no functional regressions found; a few leftover section-header comments and one now-stale build comment should be tidied up before/after merge. All listed CI checks (SonarQube, quality-gate, GitGuardian) are passing.

## Issues

- `src/main/api/http-server.ts:218` — the `// Code search routes for MCP server (localhost-only, no device auth needed)` comment is left in place after the `app.use(...)` line under it was deleted; now dangles above the Web Push section and will confuse readers.
- `src/renderer/electron.d.ts:147` and `src/renderer/types/electron.d.ts:123` — the `// Vector Search` section header remains even though every declaration under it was removed (contrast with `src/main/preload.ts`, where the equivalent header was correctly deleted along with its block).
- `src/renderer/App.tsx:91` — `// Code search modal state` comment survives with no state declaration left underneath it.
- `src/renderer/components/SettingsModal.tsx:46` — `// Appearance settings state` comment survives with nothing left underneath it.
- `electron-builder.yml` — the new comment "Native-module loader used by the remaining prebuilt addons (sodium-native)" attached to the retained `node_modules/node-gyp-build/**/*` asarUnpack entry appears inaccurate: `bun.lock` shows the `node-gyp-build` package entry removed entirely (it was only pulled in by the deleted `tree-sitter*` packages), so no remaining dependency actually needs this loader. Worth confirming sodium-native's real native-loading mechanism and dropping the stale entry/comment if it's dead weight.

## Suggestions (non-blocking)

- If `CodeSearchModal.css` / `IndexStatus.css` still exist in the tree, they're now orphaned since their only consumers (`CodeSearchModal.tsx`, `IndexStatus.tsx`) were deleted — worth a follow-up cleanup pass.
- Consider a final grep for `vector-search`, `code-search`, `useCodeSearch`, `IndexStatus`, `getVectorSearchManager`, and `codeSearchRepo` across the repo to double-check no stray consumer of the deleted modules was missed outside this diff (the diff itself is internally consistent, this is just a belt-and-suspenders check given the size of the removal).

PR #138, opened 2026-07-22
fix: reclaim orphaned code-index tables + deterministic session ordering

William-Long-II:
Two follow-ups for the next beta.

## 1. Reclaim the orphaned code-index data

Removing the indexing feature (#137) left its tables behind. Nothing reads them, but they can be **very large** — chunk source text plus 768-dimension embeddings. On first launch after upgrading we now drop `code_chunks` / `symbols` / `indexed_files` / `code_indexes` and **`VACUUM`** to actually return the disk space.

Deliberately **best-effort and non-fatal** — this runs against the database holding the user's real data (sessions, groups, memories), so every step is guarded and a failure only logs rather than aborting startup.

⚠️ **One honest caveat:** `code_chunks_vec` is a `vec0` **virtual** table, and sqlite-vec is no longer bundled, so a normal `DROP` raises `no such module: vec0`. We fall back to dropping its **shadow tables** — which is where the embedding data actually lives, so the space is still reclaimed. We deliberately do **not** edit `sqlite_master` to remove the leftover virtual-table schema entry: it's inert (nothing queries it) and not worth risking real user data over.

## 2. Deterministic session ordering

Sessions sharing a name in different folders intermittently **swapped places** in the web view. Two causes compounding:

- `SELECT * FROM sessions ORDER BY "order"` leaves ties in an **unspecified** order, and sessions are `UPDATE`d constantly (state changes), which reshuffles physical row order between queries.
- The web sort returned **0** for same-state + same-name, and `Array.sort` is stable — so it faithfully preserved that just-shuffled input order.

Fixed at both layers:
- Order by `("order", created_at, id)` for a **total order** at the source. Bonus: this also stops needless full re-renders every 2.5s poll, since the session JSON no longer churns.
- An `id` tiebreak in the web sort, so the display is deterministic regardless of what order the agent sends.

Same ordering fix applied to `groups`.

## Tests
6 new cases for the cleanup: drops the legacy tables *and* their indexes, leaves real user data completely intact, is a no-op / idempotent when there's nothing to clean, and never throws (even on an unusable database).

Typechecks clean (main / renderer / relay web). Test suite: the single `ChatParser > 06-response` snapshot failure **pre-exists on `development`**.

### DIFF
--- a/src/main/database.ts
+++ b/src/main/database.ts
@@ -17,7 +17,10 @@ export function getDatabase(): Database.Database {
   initializeTables(db);
   initializeArenaTables(db);
 
-  // One-time reclaim for the removed code-indexing feature (see below).
+  // Reclaim for the removed code-indexing feature. dropLegacyCodeSearchTables
+  // records a completion marker, so this branch — including the synchronous
+  // VACUUM — runs at most ONCE per database (on the first launch after
+  // upgrading), never on subsequent startups.
   if (dropLegacyCodeSearchTables(db)) {
     try {
       db.exec('VACUUM');
@@ -35,18 +38,56 @@ export function getDatabase(): Database.Database {
 const LEGACY_CODE_SEARCH_TABLES = ['code_chunks', 'symbols', 'indexed_files', 'code_indexes'] as const;
 
 /**
+ * Marker recording that the one-time cleanup has run.
+ *
+ * This must NOT be re-derived from the schema. `code_chunks_vec` is a vec0
+ * VIRTUAL table whose module is no longer bundled, so its sqlite_master row can
+ * never be dropped — deriving "is there work left?" from the schema would report
+ * work forever and re-run a full VACUUM on every single launch.
+ */
+const CODE_SEARCH_CLEANUP_PREF = 'legacyCodeSearchCleanupDone';
+
+function codeSearchCleanupAlreadyRan(database: Database.Database): boolean {
+  try {
+    const row = database
+      .prepare('SELECT value FROM preferences WHERE key = ?')
+      .get(CODE_SEARCH_CLEANUP_PREF) as { value?: string } | undefined;
+    return row?.value === 'true';
+  } catch {
+    return false; // preferences unavailable — treat as not-yet-run
+  }
+}
+
+function markCodeSearchCleanupRan(database: Database.Database): void {
+  try {
+    database
+      .prepare('INSERT OR REPLACE INTO preferences (key, value) VALUES (?, ?)')
+      .run(CODE_SEARCH_CLEANUP_PREF, 'true');
+  } catch (e) {
+    log.warn('[DB] could not record code-search cleanup marker:', (e as Error).message);
+  }
+}
+
+/**
  * Drop the leftover tables from the removed code-indexing / semantic-search
  * feature. Nothing reads them anymore and they can be very large (chunk source
  * text plus 768-dimension embeddings), so we reclaim the space on first launch
  * after upgrading.
  *
+ * Runs at most ONCE per database (guarded by CODE_SEARCH_CLEANUP_PREF), so the
+ * caller's VACUUM is a one-time startup cost rather than something that repeats
+ * on every launch.
+ *
  * Deliberately best-effort and non-fatal: this runs against the database holding
  * the user's real data (sessions, groups, memories), so every step is guarded and
  * a failure only logs rather than aborting startup or leaving things half-done.
  *
- * @returns true if anything was found to clean up (so the caller can VACUUM).
+ * @returns true if it actually dropped something (so the caller can VACUUM).
  */
 export function dropLegacyCodeSearchTables(database: Database.Database): boolean {
+  // Converge: once we've run, never do this work (or the VACUUM) again.
+  if (codeSearchCleanupAlreadyRan(database)) return false;
+
   let present: string[];
   try {
     present = (
@@ -63,7 +104,10 @@ export function dropLegacyCodeSearchTables(database: Database.Database): boolean
     log.warn('[DB] could not check for legacy code-search tables:', (e as Error).message);
     return false;
   }
-  if (present.length === 0) return false;
+  if (present.length === 0) {
+    markCodeSearchCleanupRan(database); // fresh install — nothing to do, ever
+    return false;
+  }
 
   log.info('[DB] Removing legacy code-search tables:', present.join(', '));
 
@@ -93,9 +137,13 @@ export function dropLegacyCodeSearchTables(database: Database.Database): boolean
           log.warn(`[DB] could not drop shadow table ${shadow}:`, (e as Error).message);
         }
       }
+      // The code_chunks_vec sqlite_master row itself stays (no vec0 module to
+      // drop it, and editing sqlite_master isn't worth the risk to real user
+      // data). It's inert — the marker below is what stops us retrying forever.
     }
   }
 
+  markCodeSearchCleanupRan(database);
   return true;
 }
 
--- a/src/main/database.ts
+++ b/src/main/database.ts
@@ -17,7 +17,86 @@ export function getDatabase(): Database.Database {
   initializeTables(db);
   initializeArenaTables(db);
 
+  // One-time reclaim for the removed code-indexing feature (see below).
+  if (dropLegacyCodeSearchTables(db)) {
+    try {
+      db.exec('VACUUM');
+      log.info('[DB] VACUUM complete — reclaimed code-index disk space');
+    } catch (e) {
+      // Space stays allocated until a later VACUUM; the data is already gone.
+      log.warn('[DB] VACUUM after code-index cleanup failed:', (e as Error).message);
+    }
+  }
+
   return db;
+}
+
+/** Tables created by the removed code-indexing feature. */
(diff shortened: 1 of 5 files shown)


src/main/database.ts
```diff
@@ -17,9 +17,88 @@ export function getDatabase(): Database.Database {
...
+  }
+
+  // code_chunks_vec is a vec0 VIRTUAL table. sqlite-vec is no longer bundled, so
+  // a normal DROP raises "no such module: vec0". Fall back to dropping its shadow
+  // tables directly — that's where the embedding data actually lives, so this is
+  // what reclaims the space. Any leftover schema entry is inert (nothing queries
+  // it), and we do NOT edit sqlite_master: not worth risking real user data.
+  if (present.includes('code_chunks_vec')) {
+    try {
+      database.exec('DROP TABLE IF EXISTS code_chunks_vec');
```
brannon-bowden, line 131: Since sqlite-vec is no longer bundled, this DROP will always throw for any user who had the code-search feature, so the code_chunks_vec row in sqlite_master can never be removed. Every subsequent getDatabase() call will re-detect it as "present", return true, and re-trigger a full VACUUM on every app launch, forever — not the one-time cleanup this PR intends.

src/main/database.ts
```diff
@@ -17,9 +17,88 @@ export function getDatabase(): Database.Database {
   initializeTables(db);
   initializeArenaTables(db);
 
+  // One-time reclaim for the removed code-indexing feature (see below).
+  if (dropLegacyCodeSearchTables(db)) {
+    try {
+      db.exec('VACUUM');
```
brannon-bowden, line 26: Synchronous VACUUM on the startup path can block the main process noticeably for large databases. Consider whether this should run off the critical path.

src/main/repositories/sessions.ts
```diff
@@ -9,7 +9,7 @@ export function sessionExists(id: string): boolean {
 
 export function getAllSessions(): Session[] {
   const db = getDatabase();
-  const rows = db.prepare('SELECT * FROM sessions ORDER BY "order"').all() as any[];
+  const rows = db.prepare('SELECT * FROM sessions ORDER BY "order", created_at, id').all() as any[];
```
brannon-bowden: If created_at can be NULL for legacy rows, they'll sort first under SQLite's default NULLS FIRST — confirm that's intended.

brannon-bowden (changes requested):
## Review

Request changes. The ordering fixes are solid, but the new `code_chunks_vec` cleanup path in `dropLegacyCodeSearchTables` has a bug that defeats the "one-time" intent of this PR for exactly the users it's meant to help, and that path has zero test coverage.

### Blocking issues

- **`src/main/database.ts:80-97` — `code_chunks_vec` schema entry can never actually be removed, so cleanup (and the VACUUM it triggers) repeats on every app launch, forever, for any user who ever used the code-search feature.** The function's own comment states sqlite-vec "is no longer bundled, so a normal DROP raises 'no such module: vec0'" — meaning `database.exec('DROP TABLE IF EXISTS code_chunks_vec')` at line 87 will *always* throw for these users, not just occasionally. The catch block falls back to dropping the shadow tables (good, reclaims the actual embedding data), but the `code_chunks_vec` row itself stays in `sqlite_master` forever, since the code explicitly avoids editing `sqlite_master` directly. That means on every subsequent `getDatabase()` call:
  - the detection query at line 55 finds `code_chunks_vec` present again,
  - `dropLegacyCodeSearchTables` logs "Removing legacy code-search tables: code_chunks_vec" and returns `true` again (even though nothing is left to actually drop),
  - `getDatabase()` (line 21) runs `db.exec('VACUUM')` again.

  This turns a "one-time reclaim" into an unconditional full-database VACUUM on *every single app startup*, indefinitely, for any affected user — a real perf/UX regression (VACUUM is synchronous and rewrites the whole file) that undercuts the stated purpose of the PR. A fix needs some way to converge to `false` in steady state — e.g. track completion (a settings flag or a marker row) instead of re-deriving "is there work to do" from a schema artifact that can never be cleared, or register a no-op stub module for `vec0` before attempting the DROP so the schema entry itself can be removed.

### Non-blocking issues

- **`src/main/database.ts:23` — `db.exec('VACUUM')` runs synchronously on the main/startup path.** better-sqlite3 is synchronous, so on a large database this can noticeably block app startup. Worth confirming this is acceptable for the expected DB sizes, or moving it off the critical path (e.g. after the window is shown).
- **`src/main/repositories/groups.ts:6`, `src/main/repositories/sessions.ts:12` — `ORDER BY "order", created_at, id"` with `created_at` possibly `NULL` for legacy rows** would sort those rows first (SQLite default NULLS FIRST in ASC), which may not be the intended behavior for old rows without a `created_at` value. Worth double-checking existing data doesn't have NULLs here.

### Tests

- The new `database-cleanup.test.ts` covers the four plain legacy tables well (drop, idempotency, no-op case, data preservation, closed-db safety) — good coverage for that part.
- However, there is no test at all for the `code_chunks_vec` / vec0-virtual-table fallback path (lines 85-97 of `database.ts`), which is the most complex and riskiest part of this change (it's the part working around a missing SQLite module). This is exactly the path with the idempotency bug above — a test that simulates a virtual-table-like entry whose DROP always fails (e.g. via a `mock` throwing table, or asserting behavior when the "table" can't be dropped) would have caught that the function keeps returning `true` (and would keep re-triggering VACUUM) instead of converging to `false`. Please add coverage for this path before merge.

William-Long-II:
Thanks — the blocking issue was a real bug and a good catch. Addressed all three points.

**Blocking — cleanup never converged.** You're right that `DROP TABLE IF EXISTS code_chunks_vec` *always* throws for affected users, so deriving "is there work left?" from `sqlite_master` meant it returned `true` forever and re-ran a full synchronous `VACUUM` on every launch — the opposite of the PR's intent.

Fixed by **tracking completion instead of re-deriving it**: a `legacyCodeSearchCleanupDone` marker in `preferences` is checked first and written after the run (including the nothing-to-do case), so the work — and the VACUUM — happens **at most once per database**. The leftover vtab schema row is now documented as inert rather than treated as outstanding work. I went with the marker rather than a `vec0` stub module since better-sqlite3 can't register one without a native extension.

**Non-blocking — synchronous VACUUM on startup.** Now genuinely one-time (first launch after upgrade only), and I noted that explicitly at the call site.

**Non-blocking — NULL `created_at`.** Good catch; the inserts do bind it explicitly so NULL is possible. Changed to `ORDER BY "order", created_at IS NULL, created_at, id` so those rows sort **last** deterministically instead of first — portable, no dependency on SQLite's `NULLS LAST`.

**Tests — vec0 fallback path now covered** (4 new cases), including the exact regression you identified: a proxy makes the vtab DROP always fail like a missing module, and the test asserts the function **converges to `false`** on subsequent calls instead of re-triggering VACUUM forever. Also covers shadow tables (the embedding data) being dropped, the vtab row surviving as documented, user data preserved on that path, and the marker being written.

10 tests in that file now; typechecks clean across main/renderer/relay-web.

brannon-bowden (approved):
Approve — the completion-marker fix resolves the blocking issue from the previous review, and the ordering fix now handles `NULL` `created_at` correctly.

## Verdict
Approve. The `code_chunks_vec` cleanup now converges: `dropLegacyCodeSearchTables` records `legacyCodeSearchCleanupDone` in `preferences` after its first run (success, no-op, or vec0-fallback path alike) and short-circuits on every subsequent call, so the synchronous `VACUUM` on the `getDatabase()` startup path (`src/main/database.ts:26`) now fires at most once per database instead of on every launch. The new `vec0 virtual table fallback` test suite directly exercises this — it asserts the second and third calls return `false` after the first cleanup, which is exactly the regression previously flagged. The `groups.ts`/`sessions.ts` ordering change (`created_at IS NULL, created_at`) also correctly pushes legacy `NULL`-timestamp rows to the end instead of the SQLite-default NULLS-FIRST behavior — the other non-blocking item from the last review is resolved too.

## Blocking issues
None.

## Non-blocking issues
- `src/main/database.ts:54` — `codeSearchCleanupAlreadyRan` treats a `preferences` lookup failure as "not yet run," and `markCodeSearchCleanupRan` (line ~58-66) only logs on write failure rather than surfacing it. If `preferences` is ever unreadable/unwritable for a given install (schema drift, permissions, etc.), the marker can never persist, and the code falls back to re-deriving "work remaining" from the schema — which, for the `code_chunks_vec` row specifically, reintroduces the exact infinite-VACUUM issue this PR fixes. In practice `preferences` is an existing, well-established table, so this is a narrow edge case, but worth a one-line comment or follow-up ticket rather than silently accepting the regression risk.
- `src/main/database.ts:26` — `db.exec('VACUUM')` is still synchronous on the startup path (previously raised as non-blocking); the new comment documents that it now runs only once, which meaningfully reduces the practical impact, but on a very large legacy database that one-time VACUUM could still noticeably delay first launch after upgrade. Not a blocker, just worth confirming against expected DB sizes.

## Tests
Coverage now directly targets the previously-missing scenario: `makeVecDb`/`withMissingVecModule` simulate the always-fails-DROP vec0 case, and the added tests verify (1) shadow tables are reclaimed, (2) the function converges to `false` and doesn't re-trigger cleanup on subsequent calls, (3) real user data survives, and (4) the completion marker is written. This is exactly the coverage requested in the prior review — no further test gaps for this change.

PR #140, opened 2026-07-22
fix(updater): flag quit before quitAndInstall so macOS updates restart

### LINKED ISSUE #139: macOS auto-update still doesn't restart: close-to-tray swallows quitAndInstall's window close
## Summary
Follow-up to #133. On macOS, clicking **Restart Now** (or About → restart) downloads the update but the app **does not quit/relaunch on its own**. Manually quitting the app lets ShipIt install + relaunch, so the install pipeline is fine — the app just never terminates.

## Evidence (from a real 3.4.3 → 3.5.0-beta.2 attempt)
- **ShipIt log:** install request detected `13:43:30`, but "Beginning installation" not until `13:44:00` (30s gap = user manually quitting). Install then **completed + relaunched successfully**.
- **App log:** `[shutdown] cleanup exceeded budget; forcing exit` fires at `13:43:59` — i.e. `before-quit` didn't start until ~`13:43:57`, ~27s *after* `quitAndInstall` handed off to Squirrel at `13:43:30`.
- Running app was **3.4.3 (the version WITH the #133 fix)**, so this is a real defect, not a stale-version test.

## Root cause
`autoUpdater.quitAndInstall()` on macOS **closes all windows, then calls `app.quit()`**. The main window's `close` handler (`src/main/index.ts:675`) hides to tray via `event.preventDefault()` whenever `isQuitting` is false. `isQuitting` is only set inside `before-quit`, which never runs because `app.quit()` is never reached (window refused to close). Deadlock → PID never dies → ShipIt waits forever ("App Still Running").

The #133 `runGuardedShutdown` fix only makes termination fast *once quitting starts* — but nothing starts the quit.

Two affected call sites: `src/main/index.ts:1009` (About dialog) and `src/main/auto-updater.ts:244` ("Restart Now" dialog).

## Fix
Introduce a shared, electron-free quit-state flag (matching the `shutdown.ts` pattern). Every path that must bypass close-to-tray (`quitAndInstall`, `before-quit`, tray Quit) calls `markAppQuitting()` **before** triggering the quit, so the window `close` handler lets windows actually close.

## Acceptance
- On macOS, "Restart Now" / About-restart terminates and relaunches on the new version with **no manual quit**.
- Unit tests cover the close-to-tray-vs-quitting decision.

brannon-bowden:
## Problem
Follow-up to #133. On macOS, the auto-update **downloads but the app never restarts on its own** — you have to manually quit it, after which ShipIt installs and relaunches fine.

## Root cause (confirmed at every layer)
`autoUpdater.quitAndInstall()` on macOS **closes all windows, then calls `app.quit()`**. The main window's close-to-tray handler (`index.ts:675`) calls `event.preventDefault()` + `hide()` whenever `isQuitting` is false — and `isQuitting` was **only** set inside `before-quit`, which never runs because `app.quit()` is never reached. The window hides to tray, the PID never dies, and Squirrel's ShipIt waits forever (`App Still Running Error`, `Code=-9`).

#133's `runGuardedShutdown` only makes termination *fast once quitting starts* — but nothing was starting the quit.

### Evidence (real 3.4.3 → 3.5.0-beta.2 attempt on a build that already had the #133 fix)
| Time | Event |
|------|-------|
| 13:43:30 | electron-updater hands zip to Squirrel; **ShipIt launches, waits for PID** |
| 13:43:30–13:43:57 | **27s of nothing** — app hidden to tray, still running |
| 13:43:59 | `[shutdown] cleanup exceeded budget; forcing exit` — `before-quit` finally ran (from a *manual* quit) |
| 13:44:00 | ShipIt "Beginning installation" |
| 13:44:07 | ShipIt relaunched app on new version |

## Fix
New electron-free, unit-tested `quit-state` module (matches the `shutdown.ts` pattern):
- `markAppQuitting()` / `isAppQuitting()` — single shared source of truth (so `auto-updater.ts` and `index.ts` agree).
- `shouldHideToTrayOnClose(quitting, pref)` — pure decision: **never** hide to tray while quitting.

Every path that must bypass close-to-tray now calls `markAppQuitting()` **before** triggering the quit:
- `index.ts` About-dialog restart (`app:restart-and-update`)
- `auto-updater.ts` "Restart Now" dialog
- `before-quit` (manual/tray quit) — replaces the old local `isQuitting`

Both `quitAndInstall` call sites are covered.

## Testing
- `bun test src/main/quit-state.test.ts` — 5 pass (covers the quitting-bypasses-tray regression).
- Full suite: 196 pass; the only failures are pre-existing `@xterm/headless` missing-dep errors identical on `development` (unrelated).
- ⚠️ **End-to-end auto-update restart cannot be verified locally** — it requires two signed/released builds and updating between them. Root cause is confirmed from the ShipIt + app logs and the electron-updater 6.8.3 source, and the fix directly removes the deadlock. Real-world confirmation will come from updating a build carrying this fix to a newer release.

Closes #139

### DIFF
--- a/src/main/auto-updater.ts
+++ b/src/main/auto-updater.ts
@@ -2,6 +2,7 @@ import { autoUpdater, UpdateInfo } from 'electron-updater';
 import { BrowserWindow, dialog, Notification } from 'electron';
 import * as log from 'electron-log';
 import { getPreference, setPreference } from './repositories/preferences';
+import { markAppQuitting } from './quit-state';
 
 // Configure logging
 autoUpdater.logger = log;
@@ -240,7 +241,10 @@ autoUpdater.on('update-downloaded', (info: UpdateInfo) => {
   }).then((result) => {
     isDialogOpen = false;
     if (result.response === 0) {
-      // User clicked Restart Now
+      // User clicked Restart Now. Flag the quit BEFORE quitAndInstall so the
+      // main window's close-to-tray handler doesn't swallow the window close
+      // and strand the app running — which blocks ShipIt from installing (#139).
+      markAppQuitting();
       autoUpdater.quitAndInstall(false, true);
     }
   });
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { ptyManager } from './pty-manager';
 import { runGuardedShutdown } from './shutdown';
+import { isAppQuitting, markAppQuitting, shouldHideToTrayOnClose } from './quit-state';
 import { resolveLaunchProviderId } from './providers';
 import { startProviderInstall, cancelProviderInstall } from './provider-install';
 import { detectProviders } from './provider-detector';
@@ -97,7 +98,8 @@ if (process.platform === 'win32') {
 
 let mainWindow: BrowserWindow | null = null;
 let stateMonitor: StateMonitor | null = null;
-let isQuitting = false;
+// "App is quitting" is tracked in the shared, electron-free quit-state module
+// (see ./quit-state) so auto-updater.ts can flag a quitAndInstall the same way.
 
 // Register deep link protocol
 if (process.defaultApp) {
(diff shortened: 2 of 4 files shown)


William-Long-II (approved):
Not too running running this one hard, so straight to the rewrite.

Nice fix for #139 — the root cause is right: `quitAndInstall()` closes windows before `app.quit()`, so a close-to-tray handler gated only on the not-yet-set `isQuitting` flag swallows the window close and strands the process, which is exactly why `ShipIt` hangs waiting on a PID that never dies.

Pulling the quitting flag into a standalone, Electron-free `quit-state.ts` is the right call — lets `auto-updater.ts` and `index.ts` share the latch without circular imports, and makes `shouldHideToTrayOnClose` trivially unit-testable pure logic. The `close` handler refactor preserves the original semantics exactly (`!isQuitting && closeToTray !== 'false'` ⟺ `!shouldHideToTrayOnClose(...)`), and both `quitAndInstall()` call sites (the "Restart Now" dialog and the About-dialog restart handler) now call `markAppQuitting()` first, matching the invariant in the module header. Tests cover the branches that matter, including the actual regression case (quitting=true must never hide to tray regardless of preference).

Two things, neither blocking:

`quit-state.ts` has no reset for `quitting` — it's a one-way latch, matching the original `isQuitting` behavior which also never reset, so not a regression. But since ES modules are cached singletons, if some other test file down the line imports this module and calls `markAppQuitting()`, it'll permanently flip the flag for the rest of that test process. Current test file dodges this by hardcoding the `quitting` boolean everywhere except the last test. Just something to watch if more tests start importing this module.

In `index.ts`'s `close` handler, `closeToTray ?? undefined` looks redundant if `getPreference` already returns `string | undefined` — harmless either way, but worth a sanity check.

Good use of doc comments too — they explain the non-obvious "why" (ShipIt polling a PID that never dies) instead of just restating the code.

PR #143, opened 2026-07-24
Add sidebar text filter for groups, sub-groups and sessions

### LINKED ISSUE #141: Add text filter box to sidebar to quickly filter groups/sub-groups/sessions
## Problem

Users with many groups and sub-groups have to scroll and hunt through the sidebar to find what they want.

## Request

Add a text filter box at the top of the sidebar (under the "Groups" header) to quickly narrow the list.

## Behavior

- Matches against **group names, sub-group names, and session names**.
- A group matched by name shows its full contents (all sub-groups + sessions).
- A matching sub-group or session **keeps its parent visible** as context.
- Groups/sub-groups with no match anywhere in their subtree are hidden.
- Within a non-name-matched group, only matching sessions render.
- While the filter is active, collapsed groups render expanded (view-only; the persisted `collapsed` flag is untouched).
- Empty query shows the normal, unfiltered sidebar. Empty-state message when nothing matches.
- Filter is **ephemeral** — clears on restart.
- Clear (×) button + Escape-to-clear.

## Out of scope

Drag-drop, keyboard nav, and the memory/analytics/arena panels are unchanged.

brannon-bowden:
Closes #141

Adds a text box at the top of the sidebar that narrows the tree as you type — for users with enough groups/sub-groups that scrolling to find one is painful.

## Behaviour

- Matches **group, sub-group and session names** (case-insensitive substring).
- A group matched by name reveals its **whole subtree**.
- A matching sub-group or session keeps its **parent visible as context**, without revealing the parent's other children.
- Subtrees with no match are hidden; an empty query renders the sidebar exactly as before.
- Collapsed groups render expanded while filtering — **view-only**, the persisted `collapsed` flag is never written.
- Ephemeral (clears on restart), sticky so it stays reachable, with a `×` button and Escape-to-clear.

## Implementation

Matching lives in a pure, unit-tested module (`src/renderer/store/groupFilter.ts`); `App.tsx` consumes two id sets. The filter input is its own component (`SidebarFilter.tsx`).

`buildNavItems` replaces the inline nav list so **keyboard navigation walks exactly the rendered rows** — previously it would have walked hidden rows and skipped auto-expanded ones.

## Fixes to existing behaviour found along the way

- **`useKeyboardShortcuts` had no target guard.** Bare Arrow/Enter keys on `window` fired sidebar navigation while typing in *any* input — and the arrow handlers persist `collapsed` to the database. Now guarded via `isEditableTarget`; modifier shortcuts (Ctrl+N/W/Q/G/Tab) are deliberately unaffected, including in the terminal.
- **Arrow keys crashed the renderer on an empty sidebar** (`navItems[-1]`). Guarded.
- Drag is disabled while filtering, since drop position is computed against unfiltered neighbours.

## Testing

- **285 tests, 284 passing.** The one failure (`ChatParser snapshot fixtures > 06-response`) is pre-existing — verified by stashing this branch and reproducing it on a clean tree.
- New: 19 pure unit tests for the filter and nav builder, 7 for the keyboard guard, 9 DOM tests for the filter UI (happy-dom + @testing-library/react, added as devDependencies).
- `tsc --noEmit` clean; `bun run build` passes end to end.
- Sticky-positioning behaviour verified by measuring hit-testing in Chromium.

## Review

Three adversarial review layers plus two verification rounds. Every finding is closed — including two defects the verification rounds caught in my own fixes (a rename path that wiped the filter on a mere click, and a DOM test preload that broke 9 relay tests).

> Note: `node_modules` was missing three declared deps (`@xterm/headless`, `@xterm/addon-serialize`, `web-push`), so `bun run build:renderer` failed on `development` too. `bun install` fixed it; no dependency versions were changed.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -5,6 +5,7 @@ import ContextMenu, { MenuItem } from './components/ContextMenu';
 import { NamePromptModal } from './components/NamePromptModal';
 import { SettingsModal } from './components/SettingsModal';
 import { NewItemChoice } from './components/NewItemChoice';
+import { SidebarFilter } from './components/SidebarFilter';
 import { MemoryPanel } from './components/panels/MemoryPanel';
 import AnalyticsPanel from './components/panels/AnalyticsPanel';
 import { ArenaPanel } from './components/ArenaPanel';
@@ -62,7 +63,6 @@ const App: React.FC = () => {
 
   // Sidebar text filter (#141). Ephemeral — never persisted.
   const [filterText, setFilterText] = useState('');
-  const filterInputRef = useRef<HTMLInputElement>(null);
   const [isResizing, setIsResizing] = useState(false);
 
   const [settingsOpen, setSettingsOpen] = useState(false);
@@ -303,9 +303,12 @@ const App: React.FC = () => {
 
   const handleFinishEditGroup = async () => {
     if (editingGroupId && editingGroupName.trim()) {
-      await updateGroup(editingGroupId, { name: editingGroupName.trim() });
-      // The new name may not match the active filter (#141).
-      setFilterText('');
+      const trimmed = editingGroupName.trim();
+      const changed = groups.find(g => g.id === editingGroupId)?.name !== trimmed;
+      await updateGroup(editingGroupId, { name: trimmed });
+      // Only when the name actually changed: a click that merely opens (and
+      // blurs) the rename input must not wipe the user's query (#141).
+      if (changed) setFilterText('');
     }
     setEditingGroupId(null);
     setEditingGroupName('');
@@ -431,9 +434,11 @@ const App: React.FC = () => {
 
   const handleFinishEditSession = async () => {
     if (editingSessionId && editingSessionName.trim()) {
-      await updateSession(editingSessionId, { name: editingSessionName.trim() });
-      // The new name may not match the active filter (#141).
-      setFilterText('');
+      const trimmed = editingSessionName.trim();
+      const changed = sessions.find(s => s.id === editingSessionId)?.name !== trimmed;
+      await updateSession(editingSessionId, { name: trimmed });
+      // Only when the name actually changed (see handleFinishEditGroup).
+      if (changed) setFilterText('');
     }
     setEditingSessionId(null);
     setEditingSessionName('');
@@ -768,7 +773,9 @@ const App: React.FC = () => {
       return [
         { key: '↑↓', label: 'Navigate' },
         { key: 'Enter', label: 'Select' },
-        { key: '←→', label: 'Collapse' },
+        // Collapse/expand is unavailable while filtering — rows are
+        // force-expanded, so advertising it would be a lie (#141).
+        ...(filter.active ? [] : [{ key: '←→', label: 'Collapse' }]),
         { key: 'Ctrl+G', label: 'New Group' },
       ];
     }
@@ -777,7 +784,7 @@ const App: React.FC = () => {
       { key: 'Ctrl+Tab', label: 'Next' },
       { key: 'Ctrl+W', label: 'Close' },
     ];
-  }, [sidebarFocused]);
+  }, [sidebarFocused, filter.active]);
 
   const handleNewSubGroup = useCallback(async () => {
     if (focusedItemType === 'group' && focusedItemId) {
@@ -958,44 +965,7 @@ const App: React.FC = () => {
           </div>
         </div>
 
-        <div className="sidebar-filter">
-          <div className="sidebar-filter-field">
-          <input
-            ref={filterInputRef}
-            className="sidebar-filter-input"
-            type="text"
-            value={filterText}
-            placeholder="Filter groups & sessions…"
-            aria-label="Filter groups and sessions"
-            onChange={(e) => setFilterText(e.target.value)}
-            onKeyDown={(e) => {
-              if (e.key !== 'Escape') return;
-              if (filterText) {
-                // Consume the first Escape to clear the query.
-                e.stopPropagation();
-                setFilterText('');
-              } else {
-                // Already empty: step out of the box so keyboard users have a
-                // route back, and let app-level dismiss handlers see the key.
-                filterInputRef.current?.blur();
-              }
-            }}
-          />
-          {filterText && (
-            <button
-              className="sidebar-filter-clear"
-              onClick={() => {
-                setFilterText('');
-                filterInputRef.current?.focus();
-              }}
-              title="Clear filter"
-              aria-label="Clear filter"
-            >
-              ×
-            </button>
-          )}
-          </div>
-        </div>
+        <SidebarFilter value={filterText} onChange={setFilterText} />
 
         {filter.active && visibleTopLevelGroups.length === 0 && (
           <div className="sidebar-filter-empty" role="status">No groups or sessions match</div>
@@ -1420,7 +1390,11 @@ const App: React.FC = () => {
             >
               <TerminalHeader
                 session={session}
-                onRename={(name) => updateSession(session.id, { name })}
+                onRename={(name) => {
+                  updateSession(session.id, { name });
+                  // Keep the renamed session visible in the sidebar (#141).
+                  if (session.name !== name) setFilterText('');
+                }}
                 onRestart={() => setRestartKeys(prev => ({ ...prev, [session.id]: (prev[session.id] || 0) + 1 }))}
                 onStop={() => updateSession(session.id, { state: 'stopped' })}
                 onClose={() => handleRemoveSession(session.id)}
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -245,6 +245,16 @@ const App: React.FC = () => {
     setSessionPrompt({ groupId, defaultName: `Session ${count}` });
   }, [getSessionsByGroup]);
 
+  /**
+   * A newly created group/session becomes active straight away, so it has to be
+   * visible. Clear the filter only when the new name would not match it —
+   * an unrelated query is left untouched (#141).
+   */
+  const revealNewItem = useCallback((name: string) => {
+    const query = filterText.trim().toLowerCase();
+    if (query && !name.toLowerCase().includes(query)) setFilterText('');
+  }, [filterText]);
+
   // Actually create the session after name is confirmed
   const handleConfirmSession = useCallback(async (
     name: string,
@@ -256,11 +266,8 @@ const App: React.FC = () => {
     const cwd = getEffectiveWorkingDir(sessionPrompt.groupId) || homedir;
     await createSession(sessionPrompt.groupId, name, cwd, true, provider); // launchClaude = true
     setSessionPrompt(null);
-    // A new/renamed row may not match the active filter; clearing it keeps the
-    // result visible instead of silently vanishing (#141).
-    setFilterText('');
-
-  }, [sessionPrompt, getEffectiveWorkingDir, createSession, homedir]);
+    revealNewItem(name);
+  }, [sessionPrompt, revealNewItem, getEffectiveWorkingDir, createSession, homedir]);
 
   // Show prompt for new group name
   const handleCreateGroup = () => {
@@ -271,9 +278,7 @@ const App: React.FC = () => {
   // Actually create the group after name is confirmed
   const handleConfirmGroup = async (name: string, path?: string, claudeAccountId?: string | null) => {
     setGroupPrompt(null);
-    // A new/renamed row may not match the active filter; clearing it keeps the
-    // result visible instead of silently vanishing (#141).
-    setFilterText('');
+    revealNewItem(name);
 
     // Create the group first
     const newGroup = await createGroup(name);
@@ -308,7 +313,7 @@ const App: React.FC = () => {
       }
     }
     setSubGroupPrompt(null);
-    setFilterText('');
+    revealNewItem(name);
   };
 
   const handleStartEditGroup = (groupId: string, currentName: string) => {
@@ -318,12 +323,7 @@ const App: React.FC = () => {
 
   const handleFinishEditGroup = async () => {
     if (editingGroupId && editingGroupName.trim()) {
-      const trimmed = editingGroupName.trim();
-      const changed = groups.find(g => g.id === editingGroupId)?.name !== trimmed;
-      await updateGroup(editingGroupId, { name: trimmed });
-      // Only when the name actually changed: a click that merely opens (and
-      // blurs) the rename input must not wipe the user's query (#141).
-      if (changed) setFilterText('');
+      await updateGroup(editingGroupId, { name: editingGroupName.trim() });
     }
     setEditingGroupId(null);
     setEditingGroupName('');
@@ -449,11 +449,7 @@ const App: React.FC = () => {
 
   const handleFinishEditSession = async () => {
     if (editingSessionId && editingSessionName.trim()) {
-      const trimmed = editingSessionName.trim();
-      const changed = sessions.find(s => s.id === editingSessionId)?.name !== trimmed;
-      await updateSession(editingSessionId, { name: trimmed });
-      // Only when the name actually changed (see handleFinishEditGroup).
-      if (changed) setFilterText('');
+      await updateSession(editingSessionId, { name: editingSessionName.trim() });
     }
     setEditingSessionId(null);
     setEditingSessionName('');
@@ -776,13 +772,13 @@ const App: React.FC = () => {
   const handleNewGroup = useCallback(async () => {
     const name = `Group ${groups.filter(g => !g.parentId).length + 1}`;
     const newGroup = await createGroup(name);
-    setFilterText('');
+    revealNewItem(name);
 
     const dir = await window.electronAPI.selectDirectory();
     if (dir) {
       await updateGroup(newGroup.id, { workingDir: dir });
     }
-  }, [createGroup, updateGroup, groups]);
+  }, [createGroup, updateGroup, groups, revealNewItem]);
 
   // Wiring for one session row. Hoisted out of the JSX so the per-row callbacks
   // are not nested five closures deep inside the group/sub-group maps.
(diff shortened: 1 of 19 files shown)


src/renderer/App.tsx
```diff
@@ -277,7 +303,12 @@ const App: React.FC = () => {
 
   const handleFinishEditGroup = async () => {
     if (editingGroupId && editingGroupName.trim()) {
-      await updateGroup(editingGroupId, { name: editingGroupName.trim() });
+      const trimmed = editingGroupName.trim();
+      const changed = groups.find(g => g.id === editingGroupId)?.name !== trimmed;
+      await updateGroup(editingGroupId, { name: trimmed });
+      // Only when the name actually changed: a click that merely opens (and
+      // blurs) the rename input must not wipe the user's query (#141).
```
William-Long-II: Clearing the entire sidebar filter text as a side effect of a group rename discards unrelated search context the user typed. Consider not resetting the filter here, or scoping the fix to just keeping the renamed row visible instead of wiping the whole query.

test-setup/happydom.ts
```diff
@@ -0,0 +1,39 @@
...
+  'WebSocket', 'crypto', 'TextEncoder', 'TextDecoder',
+  'ReadableStream', 'WritableStream', 'TransformStream',
+] as const;
+
+const natives = new Map<string, unknown>();
+for (const name of preserved) {
+  if (name in globalThis) natives.set(name, (globalThis as Record<string, unknown>)[name]);
+}
+
+GlobalRegistrator.register();
```
William-Long-II, line 30: This preload registers happy-dom globally for every test file in the repo (per bunfig.toml), not just renderer tests. Confirm no main-process/server code branches on typeof window !== 'undefined' in a way this would silently change during test runs.

William-Long-II (changes requested):
I'm about to rewrite this review into a more natural, engineer-voiced version per the formatting rules — no research needed, just restructuring.

Solid feature. The matching logic in `groupFilter.ts` is pure and well-tested, `SidebarFilter` and the `isEditableTarget` keyboard-shortcut guard are sensible and covered by focused unit tests, and edge cases (orphaned sessions, cyclic `parentId`, input mutation) are explicitly tested. The `isEditableTarget` fix is also a legit bug fix on its own — previously arrow keys/Enter typed into any text field (including the terminal's hidden textarea) would drive sidebar navigation too.

CI is red (SonarQube + quality-gate) and that has to be resolved before merge. Two likely culprits from the diff: a lot of new branching (`filter.active ? … : …`, `(filter.active || !group.collapsed)`, etc.) got added directly into `App.tsx`'s JSX for both the top-level group block and the near-identical sub-group block, which will spike complexity/duplication on that file. Also none of the new tests touch `App.tsx` itself — `groupFilter.test.ts`, `SidebarFilter.test.tsx`, `useKeyboardShortcuts.test.ts` all cover extracted pure units, but the actual wiring (auto-expand, drag-disable, empty-state, nav-item integration) has zero direct coverage, which could be tripping a coverage threshold. Pull the actual Sonar report and either fix what's flagged or extract the shared rendering logic between the group/sub-group blocks to kill the duplication.

Bigger concern: renaming or creating any group/session unconditionally wipes the user's filter text, at `src/renderer/App.tsx:310` and the equivalent spots in `handleFinishEditSession`, `handleConfirmGroup`, `handleConfirmSubGroup`, `handleConfirmSession`, `handleNewGroup`, and the `TerminalHeader onRename` prop. Say a user filters to "auth" to work across several matching sessions, then renames one unrelated group elsewhere — the whole query resets and they lose all filtering context, not just visibility of the one edited item. The intent (keep the edited row visible) makes sense, but wiping the search box is a much bigger side effect than the problem it's solving. Either don't reset at all — letting the edited item fall out of a stale filter seems fine — or find a narrower fix that doesn't clobber unrelated search state.

`test-setup/happydom.ts:30` — `GlobalRegistrator.register()` runs off `bunfig.toml`'s repo-wide `[test] preload`, so it fires for every test file, not just renderer tests. The docblock even flags this itself ("applies to EVERY test file in the repo — including `relay/`") and patches around it for fetch/WebSocket/crypto, but doesn't address the base concern: injecting `window`/`document`/`HTMLElement` globally could change behavior for any main-process/server code that branches on `typeof window !== 'undefined'`. Can you confirm nothing does that, or scope the DOM preload to renderer test paths if bun supports it?

Also the group/sub-group render blocks duplicate the same `filter.active` ternaries (`src/renderer/App.tsx:1083` and its sub-group counterpart) instead of sharing a row renderer — this compounds the pre-existing duplication and is probably part of what's tripping the quality gate above.

nit: stray trailing blank line before the closing brace in the session-confirm handler (`setFilterText('');` followed by an extra blank line before `}, [sessionPrompt, …])`).

src/renderer/components/SessionRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+            if (e.key === 'Escape') onCancelEdit();
+          }}
+          autoFocus
+        />
+      ) : (
+        <button
+          type="button"
+          className="session-select"
+          onClick={onSelect}
+          onDoubleClick={onStartEdit}
```
William-Long-II, line 91: onDoubleClick now fires from anywhere on the row button (account dot, badges, status pill), not just the session-name text as before. Likely fine, but it's an undocumented behavior change on a new, untested component — add a test asserting the intended double-click target.

src/renderer/components/SessionRow.tsx
```diff
@@ -0,0 +1,128 @@
...
+          aria-current={isActive}
```
William-Long-II: aria-current={isActive} renders as aria-current="false" on every inactive row (React stringifies ARIA booleans rather than omitting them). Use `isActive || undefined` to avoid emitting the attribute when not current.

src/renderer/App.tsx
```diff
@@ -738,24 +771,60 @@ const App: React.FC = () => {
...
+    isFocused: focusedItemType === 'session' && focusedItemId === session.id,
+    isDragging: draggedItem?.type === 'session' && draggedItem.id === session.id,
+    dropPosition: dropTarget?.type === 'session' && dropTarget.id === session.id
+      ? dropTarget.position
+      : null,
+    draggable: !filter.active,
+    account: effectiveAccountForSession(session.id),
+    isEditing: editingSessionId === session.id,
+    editingName: editingSessionName,
+    onEditingNameChange: setEditingSessionName,
```
William-Long-II, line 797: handleStartEditSession is called here but is not listed in the sessionRowProps useCallback dependency array below — a stale-closure risk if that handler's identity ever changes independently of the other listed deps.

William-Long-II (changes requested):
Good news first: the extraction work here is real. `SessionRow`/`GroupColorPicker` pull the duplicated ~60-line block out from under both the group and sub-group renders, `sessionRowProps` cleans up the wiring, and the chevron helpers kill the repeated `filter.active` ternaries. This is exactly what I asked for last round and it reads cleanly — drag/edit/select is now defined once and passed down instead of copy-pasted. CI's green too.

But the filter-text-reset bug from last review is still sitting there unfixed. `handleFinishEditSession`, `handleConfirmGroup`, `handleConfirmSubGroup`, `handleConfirmSession`, `handleNewGroup`, `TerminalHeader onRename` — none of them were touched in this diff. `sessionRowProps.onFinishEdit` just wires straight through to the still-broken `handleFinishEditSession`. So a user filtering on "auth" who renames one unrelated session anywhere still loses their whole search. This was the main reason for request-changes last time and it needs to land now, not as a follow-up. `happydom.ts`'s global DOM preload scope is also still open — still need confirmation nothing in main-process/relay branches on `typeof window !== 'undefined'`, or it needs scoping to renderer tests. And the stray trailing blank line in the session-confirm handler is still there too, still a nit.

On the keyboard shortcuts refactor — this one's solid. I traced the tricky cases against the old sequential logic myself: Ctrl/Cmd+ArrowUp still reaches `NAVIGATION_KEYS` because a missed chord lookup falls through instead of returning (comment calls this out), the shift-insensitive `n`/`q` duplication is preserved (`'false:n'`/`'true:n'`, `'false:q'`/`'true:q'`), and the shift+w-before-w ordering dependency from the original doesn't matter anymore since it's an exact table lookup now instead of sequential ifs. No new tests here, but it's a behavior-preserving refactor of already-tested logic, so I'm fine with that.

`SessionRow`/`GroupColorPicker` are a different story — new files, zero direct tests, and they do change real behavior. The old markup scoped `onDoubleClick` to just the `session-name` span; the new `SessionRow` button wraps the account dot, name, provider badge, and status pill, so double-clicking anywhere in that button now enters rename mode instead of just on the name text (`src/renderer/components/SessionRow.tsx:91`). Could be an intentional improvement — bigger hit target — but it's undocumented and untested, and this isn't a pure refactor like the keyboard shortcuts piece. Wants at least a couple component tests: select/rename/close/drag wiring, and the widened double-click target specifically.

Also, `sessionRowProps` is missing `handleStartEditSession` from its `useCallback` deps. `onStartEdit: () => handleStartEditSession(session.id, session.name)` at `src/renderer/App.tsx:801` isn't in the deps array. Low risk today, but that's literally what `exhaustive-deps` exists to catch — add it or confirm the lint rule is on.

Question for the author: the row already had a native `<button className="session-close">`, and now `session-select` adds a second native button per row. Since the app runs its own arrow-key/`focusedItemId` navigation rather than roving tabindex, this doubles the Tab stops per session with no coordination between the two schemes. Not something this PR introduced from scratch, but worth a deliberate check that Tab users aren't now hitting two stops per session.

nit: `aria-current={isActive}` (`SessionRow.tsx:92`) renders as literal string `"false"` on every non-active row since React doesn't omit boolean ARIA props. Harmless per spec, but `isActive || undefined` keeps the DOM clean.

The `<output>` swap for the empty-state message is fine — implicit `status` role, same as before, and the `display: block` addition correctly compensates for `<output>` being inline by default.

request-changes — filter-reset bug is the blocker, missing tests on `SessionRow`/`GroupColorPicker` is real but secondary. Duplication/CI stuff from last round is genuinely fixed.

src/renderer/App.tsx
brannon-bowden: Fixed, and gone further than scoping it.

Renames now do not touch the filter at all — for groups, sessions, and the `TerminalHeader` path. An item that no longer matches the query falling out of view is expected behaviour, and it also removes a bug: `.group-name` starts editing on a *single* click, so clicking a group name and clicking away fired `onBlur` → a no-op `updateGroup` → filter wiped, with no rename having happened.

Creates keep a narrow version, via `revealNewItem(name)`: the filter is cleared only when the new name would not match the current query. `createSession` calls `setActiveSessionId` synchronously, so a newly created session becomes active in the terminal — leaving it with no sidebar row is genuinely confusing. When the new name does match, your query survives untouched.

test-setup/happydom.ts
brannon-bowden: Checked, and nothing branches on it:

```
grep -rn "typeof window\|typeof document\|typeof HTMLElement" \
  src/main src/mcp-server src/hooks src/shared relay/src
# no matches
```

So registering DOM globals cannot change behaviour for main-process or relay code. Bun's `[test] preload` is repo-wide with no per-path scoping, so the file instead restores the natives that actually mattered — happy-dom's `Headers` strips `cookie` as a forbidden header, which broke 9 relay tests until it was put back. `bun test` at the repo root is now 305 pass / 1 fail, and that one failure (`ChatParser snapshot fixtures > 06-response`) reproduces on a clean tree without this branch.

src/renderer/components/SessionRow.tsx
brannon-bowden: Real change, now intentional and covered. The `title` says "Double-click anywhere on the row to rename" instead of "Double-click to rename", and two tests pin it: one double-clicks the status pill (not the name) and expects rename, one double-clicks the name. Bigger hit target was the intent — it just should not have shipped undocumented.

src/renderer/components/SessionRow.tsx
brannon-bowden: Fixed — `aria-current={isActive || undefined}`. Verified the underlying claim by rendering it rather than assuming: `aria-current={false}` emits `aria-current="false"`, and with `|| undefined` the attribute is absent. Both states are now asserted in `SessionRow.test.tsx`.

src/renderer/App.tsx
brannon-bowden: Added to the deps array.

On "confirm the lint rule is on" — it is not. There is no eslint config anywhere in the repo (no `.eslintrc*`, no `eslint.config.*`), so `exhaustive-deps` is not running on any file here. Worth adding, but that is a repo-wide change rather than something to smuggle into this PR — happy to raise it separately if you want it.

brannon-bowden:
All five threads addressed — pushed in 8074b09.

**Filter-reset blocker.** Renames no longer touch the filter anywhere (group, session, `TerminalHeader`). Creates keep a narrow version via `revealNewItem(name)`, clearing only when the new name would not match the active query — a created session becomes active in the terminal immediately, so an invisible row is a real problem, but an unrelated query now survives.

**Tab stops.** Good catch, and you were right to flag it as a question rather than assume. It was 1 stop per row (close button) and my change made it 2. `session-select` is now `tabIndex={-1}`, so it is back to 1. It stays a native `<button>`, so the click/AT semantics and the Sonar fix hold, while the app's own Ctrl+Q + arrow navigation remains the way to move between rows. Asserted in a test.

**Tests.** 21 new component tests across `SessionRow` (17) and `GroupColorPicker` (4) — select, close, rename commit/cancel, drag enable/disable, context menu, drop-target class, account dot, provider badge, `aria-current` in both states, tab-stop count, and the widened double-click target specifically. Renderer suite is 78 pass / 0 fail; repo-wide `bun test` is 305 pass / 1 fail, the one failure being the pre-existing `ChatParser` snapshot that reproduces on a clean tree.

Also fixed: `aria-current` no longer emits `"false"`, the missing `handleStartEditSession` dep, and the blank-line nit.

One thing worth your call, not fixed here: **there is no eslint config in this repo at all**, so `exhaustive-deps` is not enforced on any file — which is why that dep slipped. Repo-wide change, so I have left it out of this PR rather than expanding the scope. Say the word if you want it.

Heads-up unrelated to the code: the first review body opens with `"I'm about to rewrite this review into a more natural, engineer-voiced version per the formatting rules — no research needed, just restructuring."` — looks like reviewer-tooling scaffolding leaked into the posted comment.

src/renderer/App.tsx
```diff
@@ -212,6 +245,16 @@ const App: React.FC = () => {
...
   }, [getSessionsByGroup]);
 
+  /**
+   * A newly created group/session becomes active straight away, so it has to be
+   * visible. Clear the filter only when the new name would not match it —
+   * an unrelated query is left untouched (#141).
+   */
+  const revealNewItem = useCallback((name: string) => {
+    const query = filterText.trim().toLowerCase();
+    if (query && !name.toLowerCase().includes(query)) setFilterText('');
```
William-Long-II, line 255: revealNewItem reimplements the filter-match predicate independently from wherever the sidebar's actual filter logic lives. No test pins this behavior down (create-while-filtered, matching vs non-matching name). If the real filter's matching rules ever diverge from this plain includes() check, the 'new item invisible behind filter' bug returns silently. Recommend sharing the predicate and adding App-level tests for the create-while-filtered paths.

William-Long-II (approved):
This round fixes the blocking issue from last time, and it does it the right way — instead of patching the "clear filter if renamed name changed" logic, filter-clearing is pulled out of the rename paths entirely (`handleFinishEditGroup`, `handleFinishEditSession`, `TerminalHeader onRename`) and the reveal-on-mismatch behavior is now confined to item *creation* via the new `revealNewItem` helper. That's the correct split: a freshly-created item is forced active and has to be visible, so overriding the filter for it makes sense; a renamed item dropping out of a filtered list because it no longer matches is just a text filter doing its job. I traced all six call sites — `handleConfirmSession`, `handleConfirmGroup`, `handleConfirmSubGroup`, `handleNewGroup` calling `revealNewItem`, and the three rename handlers no longer touching `filterText` at all — and it holds together.

Everything else from last round is cleared too. `SessionRow`/`GroupColorPicker` have real tests now, including the widened-double-click-target concern I flagged — `SessionRow.test.tsx`'s "double-click anywhere on the row starts a rename" fires on the status pill, not the name, so that's actually covered. `handleStartEditSession` made it into the `sessionRowProps` `useCallback` deps. `aria-current={isActive}` is now an explicit ternary so inactive rows skip the attribute instead of rendering `aria-current="false"`, with a test for it. The Tab-stop-doubling question is answered too: `session-select` gets `tabIndex={-1}` now, leaving the close button as the only native tab stop, enforced by a test. And the stray trailing blank line in the session-confirm handler is gone.

One thing from last review is still open and this diff doesn't touch it: the `happydom.ts` global DOM preload scope. Not re-flagging it as new, just restating it's still outstanding since nothing here touches it.

New thing worth flagging: `revealNewItem`'s match predicate is a second, independent implementation of "does this filter match this name," and it ships with zero test coverage. This is the exact bug that took two rounds to land, and the fix itself — in `App.tsx` around `revealNewItem`, the `!name.toLowerCase().includes(query)` check — has nothing exercising "create a session while filtered, name doesn't match → filter clears" or "name does match → filter untouched." Bigger concern: this is a separate copy of whatever logic the actual sidebar filter uses (referenced elsewhere as `filter.active` on session/group rows, not in this diff). If that real filter's matching rules ever diverge from a plain case-insensitive substring check — trimming, matching against group names for children, fuzzy matching — `revealNewItem` silently makes the wrong call and we get the "new item invisible" bug back in a new shape. Given this is the second time around on this exact class of bug, I'd either share the predicate function between both call sites or add a couple `App`-level tests pinning down create-while-filtered behavior.

Still want a yes/no on `happydom.ts`'s preload scope — nothing in main-process/relay code branches on `typeof window !== 'undefined'`, right? Or is the preload scoped to renderer tests only?

nit: `SessionRow.test.tsx` has a mid-test `cleanup()` + re-render pattern — "the account dot appears only when an account resolves" and "the provider badge shows only for non-default providers" each render, assert, `cleanup()`, then render and assert again inside the same `test()` instead of two separate tests. Works, just couples two assertions into one failure point for no reason.

Approve. Filter-reset bug is genuinely fixed, and the prior test-coverage ask is more than satisfied. Remaining asks are non-blocking: cover `revealNewItem` and dedupe its match logic against the canonical filter predicate in a follow-up, and still need the `happydom.ts` answer.

PR #145, opened 2026-07-24
Fix duplicate .group-sessions selector failing the quality gate on development

### LINKED ISSUE #144: SonarQube quality gate failing on development: duplicate .group-sessions CSS selector
## Problem

The SonarQube quality gate is failing on `development` (run for merge commit `9945858`), blocking a clean branch status.

```
DEVELOPMENT GATE: ERROR
  FAIL new_violations: actual=1 GT 0
```

The single violation:

```
src/renderer/styles/global.css:461 [css:S4666 MAJOR]
Unexpected duplicate selector ".group-sessions", first used at line 242
```

## Cause

PR #143 added `.group-sessions { list-style: none; }` at line 242, duplicating the pre-existing `.group-sessions { margin-left: 16px; }` at line 461.

The PR quality gate did not catch this: rule `css:S4666` reports on the **second** occurrence (line 461), which is unchanged pre-existing code attributed to a different author, so it fell outside the PR's new-code set. It only surfaced once analysed as a branch.

## Fix

Fold `list-style: none` into the existing rule at line 461 and drop the duplicate block at 242. No visual change — the two declarations are non-overlapping and both already applied.

brannon-bowden:
Closes #144

`development` has been red since #143 merged:

```
DEVELOPMENT GATE: ERROR
  FAIL new_violations: actual=1 GT 0

src/renderer/styles/global.css:461 [css:S4666 MAJOR]
Unexpected duplicate selector ".group-sessions", first used at line 242
```

#143 added a second `.group-sessions` block rather than extending the one already at line 461. This folds `list-style: none` into that existing rule and drops the duplicate.

**Why the PR gate didn't catch it:** `css:S4666` reports on the *second* occurrence — line 461, which is unchanged pre-existing code attributed to another author. That put it outside the PR's new-code set, so it only surfaced once `development` was analysed as a branch.

**No visual change.** The two declarations are non-overlapping and both were already taking effect. Confirmed in Chromium after the merge:

```
{"marginLeft":"16px","listStyleType":"none"}
```

Renderer suite 78/78, `build:renderer` clean.

### DIFF
--- a/src/renderer/styles/global.css
+++ b/src/renderer/styles/global.css
@@ -239,10 +239,6 @@ body {
   color: #ddd;
 }
 
-.group-sessions {
-  list-style: none;
-}
-
 /* The whole row is one native button so keyboard/AT support is free. */
 .session-select {
   flex: 1;
@@ -460,6 +456,8 @@ body {
 
 .group-sessions {
   margin-left: 16px;
+  /* Rendered as a <ul> of session <li>s; no markers. */
+  list-style: none;
 }
 
 /* Sessions */


William-Long-II (approved):
Straightforward fix — merging the two `.group-sessions` blocks kills the duplicate-selector lint that was failing quality-gate, and the properties don't overlap (`list-style: none` vs `margin-left: 16px`) so computed style is identical, no visual regression. Whitespace is clean, no stray blank lines. The added comment (`Rendered as a <ul> of session <li>s; no markers.`) actually explains the why behind `list-style: none` instead of just restating it, nice.

Only thing: since I'm only seeing this diff hunk, can't rule out a third `.group-sessions` rule elsewhere in the stylesheet (media query, etc.) that Sonar might still ding independently. CI's green post-change though, so probably a non-issue — just grep for it if duplicate-selector warnings show up again later.

lgtm, approving.

PR #146, opened 2026-07-27
Remove memory/context subsystem; content-area view switcher; terminal-safe keyboard scheme

William-Long-II:
Three related changes ahead of the major release, plus the fallout from verifying them.

## 1. Remove the memory/context subsystem

The feature was Claude-only and about one-fifth functional. Of five delivery paths, one worked:

- **6 of 7 providers got nothing.** `passthrough.ts` hardcodes `systemPrompt`/`hooks` false, so the Context panel persisted rows that never reached the agent, with no indication to the user.
- **Even for Claude it was a semantic dead end.** Injection filtered `source = 'manual'` while the MCP server and hooks wrote `source = 'claude'` — nothing the agent saved was ever injected back.
- **Hook auto-capture could never fire.** `registerHooks` emitted no `--group-id`, and every memory branch required it. The live DB confirmed it: 1800 `tool_use` events, 0 hook-created memories.
- Wrong payload field names (`transcript` vs `transcript_path`, `tool_result` vs `tool_response`).
- The MCP server was invisible to account-isolated sessions — written to `dirname(configDir)/.claude.json` while Claude reads `$CLAUDE_CONFIG_DIR`.
- `.bodhilander-memory.md` was written into every session cwd and read by nothing.

**Analytics telemetry is preserved.** `api/routes/hooks.ts` is the only writer of `session_events` rows of type `tool_use`/`turn_complete`/`error`/`notification`, so the route and hook are kept with the memory branches stripped and both field names fixed.

**Removal cleans up after itself.** The feature wrote to the user's machine on every launch, so deleting the code without an uninstall would leave a failing `node` spawn on every Bash tool call in every Claude Code session on the box — not just Bodhilander's. `cleanupLegacyMcpServer()` is one-shot and preference-guarded, and sweeps `~/.claude.json` **and** every `claude-accounts/<id>/.claude.json`. It deliberately does **not** call `purgeOurHooks()`, which matches any command containing "bodhilander" and would have removed the analytics hook we keep.

## 2. Content-area view switcher

The sidebar header held six identical 24px buttons spanning three interaction models and consumed 164px of a 236px usable sidebar — at the 180px resize minimum the row was wider than the sidebar and `overflow:hidden` silently clipped the `+`.

Header now owns only `+`. Analytics and Arena are tabs in a content-area switcher (they already replaced the terminal, so the switcher matches reality). Collapsing the two booleans into one `contentView` union makes "both views open at once" unrepresentable. Claude accounts becomes a Settings tab next to Providers, ending the split credential surface.

## 3. Bare Ctrl belongs to the terminal

`isMod = ctrlKey || metaKey` treated Ctrl and Cmd as interchangeable, which is wrong in a terminal app. On Windows/Linux `CmdOrCtrl+C` was eaten by the menu, so **a running process could never be sent SIGINT.**

Electron `role` items needed explicit accelerators to release their defaults — note `accelerator: undefined` does *not* clear a role default, only pinning a string does:

- `{role:'quit'}` is `undefined` only on win32, so **Linux registered bare Ctrl+Q** — quit the app, kill every session
- `{role:'undo'}` held Ctrl+Z (SIGTSTP), `{role:'redo'}` Ctrl+Y on Windows (yank), `{role:'reload'}` Ctrl+R (reverse-i-search)

Zero bare `Ctrl+<letter>` chords are registered on any platform now. App actions use Cmd on macOS, Ctrl+Shift elsewhere. Focus sidebar moves to `mod+B` (⌘Q could never reach the renderer; Ctrl+Shift+B also dodges tmux's prefix).

Also fixed along the way: shortcuts fired **twice** with the terminal focused (New Group created two groups and opened two directory pickers); `handleResize` ran on a `display:none` terminal and reflowed xterm to ~2 columns, destroying scrollback wrapping; Arena covered the sidebar and status bar; the dead ⌘F "Find" item swallowed the key while nothing subscribed; `.icon-button.active` was applied in JSX but never defined in CSS.

## 4. Verification fallout — one critical

Adversarial review caught a **data-loss bug**: `dropLegacyMemoryTables` deleted the `__global__` group row outright, and `sessions.group_id` is `ON DELETE CASCADE` against `groups` with `session_events` cascading off sessions.

This was reachable, not theoretical. `__global__` was seeded as an ordinary group in **v2.2.2** with `"order" = -1`, putting "Global Context" at the **top of the sidebar** as a normal drop target; the filter that hid it only landed in **v3.2.9**. Any install from that window can have sessions parked in it, invisible-but-intact since v3.2.9 — this would have converted dormant-and-recoverable into permanently destroyed.

Now the row is removed only when empty, and kept and renamed to "Recovered Sessions" when it still holds sessions. The renderer filter is gone accordingly: the DB settles it before the renderer asks for groups, and the filter's failure mode (hiding real sessions from their owner) is worse than showing an extra group.

Also: both one-shot cleanups latched their marker even when every step failed; `listAccountConfigDirs` returned `[]` on error, indistinguishable from "no accounts"; `~/.claude.json` was rewritten with a truncating `writeFileSync` (it's Claude Code's own state, tens of KB) and is now temp-file + atomic rename; removed `unregisterHooks()`, dead code that would have deleted the analytics hook.

## 5. Tool use now covers every tool

The PostToolUse matcher was `'Bash'` — another memory-feature artifact, so analytics counted Bash calls only. Now `'*'`.

Changing the literal alone would have reached fresh installs only, because registration could not converge: `areHooksConfigured` checked only that our path appeared *somewhere*, so a stale matcher was accepted forever. It now compares the full desired shape, ANDs the two hook types (a file with only `Stop` used to report "configured"), and purges our own entries before writing so upgraders don't end up with both the Bash entry and the match-all one firing twice.

## Verification

- Both tsconfigs typecheck clean; `bun run build` passes
- `bun test`: 274 pass / 1 fail — `chat-parser 06-response` **also fails on `development`**, confirmed pre-existing
- 18 new tests: 3 cascade-safety regressions, plus 15 for `mcp-config`, which previously had **zero** coverage despite writing into users' home directories on every launch

## Known / deferred

**Windows cmd.exe launch host changed.** `needsEnvRef` flips true→false, so cmd.exe users are no longer rerouted through PowerShell. That reroute existed solely to make system-prompt interpolation parse-safe (#106); with no interpolation left it's correct to respect the user's shell, and `envRef` throws if anyone reintroduces interpolation without flipping the flag. Deliberately deferred to a Windows smoke test rather than blocking this PR.


src/renderer/App.tsx
```diff
@@ -841,6 +860,32 @@ const App: React.FC = () => {
...
+    }
+    if (nextIndex === -1) return;
+
+    e.preventDefault();
+    e.stopPropagation(); // don't also drive the sidebar's arrow navigation
+    const next = VIEW_TABS[nextIndex];
+    setContentView(next.id);
+    // Focus follows selection, otherwise the next arrow press would start
+    // from the tab the user just left.
+    e.currentTarget.querySelector<HTMLButtonElement>(`#view-tab-${next.id}`)?.focus();
```
brannon-bowden: e.currentTarget.querySelector(`#view-tab-${next.id}`) searches descendants of the focused button, which has none — this always returns null, so focus never follows the selected tab on arrow-key navigation. Query from a stable ancestor (e.g. e.currentTarget.closest('[role="tablist"]')) instead.

brannon-bowden (changes requested):
## Verdict

Request changes. This is a large, mostly well-executed removal (the memory/MCP subsystem is torn out cleanly and consistently — every call site of `ptyManager.createSession` was updated in lockstep, and the `dropLegacyMemoryTables` migration is unusually careful about not cascading through users' real sessions). But the new keyboard-scheme rewrite — the actual load-bearing feature of this PR, explicitly justified by a documented prior regression where bare Ctrl+C couldn't send SIGINT — ships with zero automated tests for the platform-dispatch logic that makes the fix work, and the new view-tab keyboard navigation has a concrete focus-management bug.

## Blocking issues

- **No tests for the new keyboard-shortcut classification logic** (`src/renderer/hooks/useKeyboardShortcuts.ts`): `isAppShortcut`, `hasAppMod`, `appKey`, `appShiftKey`, `isNewSubGroup`, `isSessionCycle`, `isCopyShortcut`, `isPasteShortcut` are pure functions with numerous platform-specific branches (mac vs. win/linux modifiers, `event.key` vs `event.code` fallbacks, the documented AltGr trap for New Sub-Group, the Ctrl+Tab exception). This is exactly the logic whose past incorrectness the PR's own commentary describes as having broken SIGINT and other terminal control codes for users. It is trivially unit-testable (construct mock `KeyboardEvent`-like objects, assert true/false) and there isn't a single test for it. Given how safety-critical and regression-prone this code has already proven to be, shipping it with no coverage is a real gap that should block merge — a quick set of table-driven tests (e.g. "bare Ctrl+C on Windows is not an app shortcut", "Cmd+G vs Cmd+Shift+G distinguish correctly on mac", "Ctrl+Shift+G is not the sub-group shortcut on Windows") would catch the next regression before users do.

## Major issues

- `src/renderer/App.tsx` (~line 886, `handleViewTabKeyDown`): `e.currentTarget.querySelector<HTMLButtonElement>(\`#view-tab-${next.id}\`)?.focus()` searches **descendants** of the currently-focused tab button. Since the view-tab buttons are siblings (all direct children of `.view-switcher`), this query always returns `null`, and the `?.focus()` silently no-ops. The `contentView` state still updates correctly via `setContentView(next.id)`, so clicking still works and the active/selected tab visually moves — but DOM focus never follows, contradicting the comment directly above it ("Focus follows selection, otherwise the next arrow press would start from the tab the user just left") and breaking the roving-tabindex contract the code otherwise implements correctly (`tabIndex={contentView === tab.id ? 0 : -1}`). After one arrow press, the tab that still holds DOM focus now has `tabIndex={-1}`, and repeated arrow presses keep working only because `contentView` is read from React state, not from DOM traversal — but a keyboard/screen-reader user sees the visual selection move while the focus ring stays behind. Fix by querying from a stable ancestor (e.g. `e.currentTarget.closest('[role="tablist"]')`) or by giving the tablist wrapper a ref.

## Non-blocking issues

- `src/main/account-auth.ts`: the warning log message still says `"[Accounts] MCP/hooks registration failed for new account..."` even though `registerMcpServer` was removed from this call site — now purely a hooks registration. Stale wording, harmless.
- `src/main/mcp-config.ts`: `writeClaudeMcpConfig` was hardened to write via temp-file + atomic rename (good — `~/.claude.json` is Claude Code's own state and a torn write would corrupt it), but `writeClaudeSettings` (used by `registerHooks`, which writes `~/.claude/settings.json` on every launch) doesn't appear to have received the same treatment. Same risk profile, worth the same fix for consistency, though not introduced as a regression by this PR.

## Tests

- The database-layer changes (`dropLegacyMemoryTables`, the `cleanupLegacyMcpServer` sweep, `registerHooks` convergence behavior) are covered thoroughly and thoughtfully — the new `mcp-config.test.ts` and the expanded `database-cleanup.test.ts` are a good model of what "test non-trivial logic" should look like here.
- Missing, as noted above: any test for the keyboard-shortcut predicate logic in `useKeyboardShortcuts.ts`, and for the view-tab arrow-key navigation in `App.tsx` (which would have caught the focus bug immediately).

William-Long-II:
Thanks — all four addressed. Both substantive findings were correct, and the focus bug was mine from a Sonar fix earlier in the branch.

**Blocking — keyboard-scheme coverage.** Added `src/renderer/hooks/__tests__/keyboard-scheme.test.ts` (69 tests). To make both platform branches reachable in one run I parameterised the predicates (`hasAppModFor`, `appKeyFor`, `appShiftKeyFor`, `isNewSubGroupFor`, `isSettingsFor`, `isAppShortcutFor`) with thin `IS_MAC`-bound wrappers for the app to call — `IS_MAC` is fixed at import time, so there was no other way to exercise both. Behaviour and app-facing signatures are unchanged.

Aimed at exactly the regressions the scheme exists to prevent: every bare `Ctrl+<letter>` the PTY needs (C/W/N/G/Q/A/K/F/R/Z/Y/D/U/E) is asserted *not* to be an app shortcut on both platforms, with `Ctrl+C` vs `Cmd+C` and `Ctrl+C` vs `Ctrl+Shift+C` called out directly. Plus each platform rejecting the other's modifier, Alt disqualifying everywhere, `Cmd+G` vs `Cmd+Shift+G` staying distinct on macOS, the AltGr trap (`Ctrl+Alt+G` claims nothing on Windows), `event.key` vs `event.code` fallbacks for shifted digits and Dvorak, and the two deliberate bare-Ctrl exceptions.

**Major — view-tab focus.** Confirmed exactly as you described, and it was self-inflicted: I moved the handler from the tablist onto the buttons to satisfy `jsx-a11y/interactive-supports-focus`, which changed what `e.currentTarget` refers to. Now resolved from the tablist ancestor via `closest('[role="tablist"]')`.

I also extracted the switcher into its own `ViewSwitcher` component — same pattern #141 used for `SidebarFilter`/`SessionRow` — so the behaviour is testable without mounting all of `App.tsx`. 12 acceptance tests cover focus following selection for arrows/Home/End, focus and the roving `tabIndex` staying in agreement, wrap-around both directions, and the `aria-controls` wiring.

One caveat worth stating: I could not empirically demonstrate those tests failing against the reintroduced bug. Two probe runs hung — a happy-dom deadlock on a failing focus assertion, as best I can tell — so I stopped rather than keep burning time. The assertions target `document.activeElement`, which the bug leaves on the previous tab, so it should fail; but that's reasoning, not a demonstration. Flagging it rather than implying I verified it.

**Non-blocking — both taken.**
- `writeClaudeSettings` now uses temp-file + atomic rename, matching `writeClaudeMcpConfig`. You're right that it's the same risk profile: it's the user's own Claude Code config and we rewrite it on every launch.
- `account-auth` warning no longer mentions MCP.

Tests now 421 pass / 1 fail — the `chat-parser 06-response` failure that also fails on `development`, unrelated to this PR. All checks green.

brannon-bowden (approved):
Both issues from the previous review are resolved, and this revision is ready to merge.

- **Focus-management bug (previously Major) — fixed.** The arrow-key handler moved into the new `ViewSwitcher` component and now resolves the target button via `e.currentTarget.closest('[role="tablist"]')?.querySelector(...)` instead of querying descendants of the currently-focused sibling button. This is the correct fix — the tablist is a shared ancestor of all tab buttons, so the query now actually finds the target. `ViewSwitcher.test.tsx` adds a direct regression test (`document.activeElement` assertions after `ArrowRight`/`ArrowLeft`/`Home`/`End`) that would have caught the original bug immediately.
- **No tests for keyboard-shortcut classification logic (previously Blocking) — fixed.** `useKeyboardShortcuts.ts` now exposes platform-parameterized `*For` variants (`isAppShortcutFor`, `hasAppModFor`, `appKeyFor`, `appShiftKeyFor`, `isNewSubGroupFor`, `isSettingsFor`) alongside thin `IS_MAC`-bound wrappers for the app to call, specifically so both platform branches can be exercised in one test run. `keyboard-scheme.test.ts` is thorough: it covers the bare-Ctrl-reaches-the-PTY invariant across 14 control characters on both platforms, the AltGr trap, the Ctrl+Tab/Ctrl+, bare-Ctrl exceptions, Dvorak/shifted-digit key matching, and the full app-shortcut binding table. This is exactly the coverage the prior review asked for.
- The two non-blocking notes from before (`account-auth.ts` stale "MCP/hooks" wording, `writeClaudeSettings` missing the atomic-write treatment) are both addressed: the log message now reads "Hook registration failed," and `writeClaudeSettings` now does temp-file + rename matching `writeClaudeMcpConfig`.

## Suggestions (non-blocking)

- `src/renderer/App.tsx`: `import { ViewSwitcher, ContentView } from './components/ViewSwitcher';` imports a type alongside a value in one statement. If the build pipeline uses `isolatedModules`-style per-file transpilation (esbuild/swc/Babel without full type info), a non-`import type` reference to a type-only export can occasionally cause a bundler to look for a runtime binding that doesn't exist. Worth confirming the build handles this, or just write it as `import { ViewSwitcher, type ContentView } from './components/ViewSwitcher';` to be unambiguous.
- Minor typo in `keyboard-scheme.test.ts`: "the other platform s modifier does not fire it" (missing apostrophe) — cosmetic only.

PR #147, opened 2026-07-27
Fix the chat-parser fixture line endings so 06-response passes

William-Long-II:
`ChatParser snapshot fixtures > 06-response` has been failing on `development` for a while. This fixes it — in the fixture, not the parser.

## The failure

```
- "text": "Also make sure it handles ANSI codes",   "type": "response"
+ "text": "> Also make sure it handles ANSI codes", "type": "assistant_text"
```

The second line came out as assistant prose with its `>` still attached, while the first line converted correctly. Two structurally identical lines, classified differently.

## Root cause

The parser was right and the fixture was wrong.

These fixtures are written into a headless terminal emulator, but they used bare `LF`. In a terminal, LF moves the cursor **down** without returning it to column 0 — that is CR's job. So the emulator rendered row 1 as:

```
"                                  > Also make sure it handles ANSI codes"
```

34 leading spaces — exactly the length of line 0. Dumped straight from the xterm buffer:

```
0 isWrapped=false "> Please add a test for the parser"
1 isWrapped=false "                                  > Also make sure it handles ANSI codes"
```

The user-input rule (`/^[>❯]\s+\S/`) is anchored at line start, so the indented prompt never matched and fell through to case 6, which strips `[\s•●◦⏺*]+` — a set that does not include `>`. Hence assistant prose with the marker intact.

## The fix

Converted the synthetic fixtures to CRLF so they look like an actual PTY stream. `06-response` then matches its **unchanged** `.expected.json` exactly — the expectation was always correct, only the input was unrealistic. No assertion was weakened.

**All six fixtures, not just the failing one.** The other five were passing by accident: every other classifier tolerates leading whitespace (`^\s*…`, or case 6's whitespace strip), so their drift was silently absorbed. They now pass for the right reason rather than by luck.

## What I deliberately did not do

Relaxing the rule to `/^\s*[>❯]\s+\S/` would also have turned the test green, and it is tempting — it is the only rule in `classifyLine` that does not tolerate indentation, and its own sibling `LONE_PROMPT_RE` (`^\s*[>❯]\s*$`) does.

But it would misclassify indented markdown blockquotes in assistant output as user input. And the real-corpus regression test already asserts `response` events are extracted from a genuine 340KB PTY capture, which confirms the anchored rule handles real terminal streams correctly. Changing the parser to accommodate a bad fixture would have traded a test failure for a production bug.

## Guarding against regression

Added `fixtures/.gitattributes` marking `*.txt` and `*.raw` as `-text`. Without it a contributor on Windows with `core.autocrlf=true` would silently strip the CR on checkout and reintroduce this. Verified `git check-attr` reports `text: unset` and the staged blob retains `\r\n`.

Also documented in the test header why the line endings are load-bearing, so nobody "tidies" them later.

## Verification

- `bun test`: **422 pass / 0 fail** — the suite is fully green for the first time. This was the last red test.
- `tsc -p tsconfig.main.json`: clean
- No source changes; the diff is fixture bytes, one `.gitattributes`, and a comment.


brannon-bowden (approved):
This is a solid, well-documented fix for a real problem: bare-LF fixtures were rendering as artificially indented lines in the terminal emulator, breaking the line-start-anchored `> ` classification for `06-response`. Converting the fixtures to CRLF and locking that in with a directory-scoped `.gitattributes` (`-text` to block `core.autocrlf` rewrites) is the right fix, and the added comments in both the `.gitattributes` and `parser.test.ts` clearly explain the "why" for future maintainers so nobody "fixes" this back to LF. All CI checks (SonarQube, quality-gate, GitGuardian) are passing.

## Issues
- **Minor**: There's no regression test asserting the fixtures actually contain `\r\n`. `.gitattributes` protects against normalization on checkout, but it doesn't stop an editor (or a future contributor bypassing git) from re-saving a fixture with bare LF and silently reintroducing this exact bug. A small test that reads the fixture files as raw bytes/buffers and asserts each line ends in `\r\n` would close that gap given this failure mode has already bitten the suite once.

## Suggestions
- Non-blocking: consider adding the byte-level CRLF assertion mentioned above directly in `parser.test.ts` near the existing fixture-loading logic, so the invariant is self-enforcing rather than resting solely on the `.gitattributes` comment and doc block.

PR #150, opened 2026-07-30
Add sidebar toggle to show only groups with active sessions

### LINKED ISSUE #149: Sidebar toggle to show only groups with active sessions
## Request

Add a toggle to the sidebar that hides every group without an active session, so users with many groups can focus on what's live.

## Behavior

- **"Active" = session state is not `stopped`** (idle, working, waiting, and error all count as active).
- A group (top-level or sub-group) is shown if it, or any descendant, has at least one active session.
- Within a shown group, only active sessions render.
- **Combines with the text filter as AND**: a session must match the text and be active; a group stays only if it has such a session in its subtree, parents kept as context.
- A name-matched group's subtree reveal is intersected with "active only" — a name-matched group whose sessions are all stopped is hidden while the toggle is on.
- Auto-expands collapsed groups to reveal matches (view-only; never writes `collapsed`), reusing the text filter's mechanism.
- Empty-state message adapts ("No groups have active sessions" vs. the text-filter message).
- Live: a session stopping drops its group out if it was the last active one, and vice versa.

## UI & persistence

- Small icon toggle inline to the left of the filter input, with a pressed/active visual state and `aria-pressed`.
- Persisted across restarts via the existing `getPreference`/`setPreference` (SQLite), key `sidebar.showActiveOnly`, default off.

## Out of scope

Drag-drop, keyboard nav, the text filter's own behavior, and all panels are unchanged.

brannon-bowden:
Closes #149

A ⚡ toggle in the sidebar filter row that hides every group without an active session. Pairs with the text filter (#141).

## Behaviour

- **Active = `state !== 'stopped'`** — idle, working, waiting, and error all count; only stopped is hidden.
- A group (or sub-group) shows if it, or any descendant, has an active session.
- **Combines with the text filter as AND**: a session must match the text *and* be active; a name-matched group whose sessions are all stopped is hidden while the toggle is on.
- **Persisted** across restarts (`sidebar.showActiveOnly`, via the existing preferences table), default off.
- Live: a session stopping drops its group out if it was the last active one, and vice versa.
- Empty-state message adapts.

## Key design point (from review)

`filter.active` previously drove both *row filtering* and *search-mode locks* (force-expand, disabled collapse chevrons, disabled drag) — fine when it only ever meant a transient text query. The active-only toggle is **persisted**, so overloading `filter.active` with it would have disabled collapse and drag app-wide across launches. This PR splits the two: a new `isSearching` (text query only) drives the locks; the toggle only prunes which rows render. So with the toggle on you can still collapse groups and drag-reorder normally — it just narrows the list.

## Implementation

- `computeGroupFilter` gains a 4th `activeOnly` param; toggle-off output is byte-identical to #141 (all prior tests green).
- `buildNavItems` takes an explicit `forceExpand` so keyboard nav mirrors the render under the toggle.
- Persistence lives in a small `useActiveOnlyPreference` hook that ignores a stale mount-read once the user has interacted (closes a load race).

## Testing

- `groupFilter` unit tests extended (active-only, AND, descendant counting, name-match intersection, every state value, and the nav decoupling that proves active-only respects collapse).
- `useActiveOnlyPreference` unit-tested against a mocked `electronAPI` (load true/null, persist true/false, and the toggle-before-load race).
- `SidebarFilter` DOM tests for the toggle (aria-pressed, flip, real button).
- 185 renderer tests pass; typecheck clean; `build:renderer` compiles.

> Note: the Electron visual-probe environment broke mid-session (module-loader error unrelated to this code), so I couldn't re-measure the toggle row in Chromium. It's a plain flexbox row (fixed-width button + `flex:1` field); flagging in case a reviewer wants to eyeball it.

### DIFF
--- a/docs/superpowers/specs/2026-07-30-active-sessions-toggle-design.md
+++ b/docs/superpowers/specs/2026-07-30-active-sessions-toggle-design.md
@@ -88,13 +88,24 @@ The `!activeOnly` guard on the second clause is the AND: with the toggle on, a n
 - parent name-matches, child has one active + one stopped session, toggle on → parent + child visible, only the active session renders.
 - parent name-matches, all descendant sessions stopped, toggle on → hidden.
 
-`buildNavItems` is unchanged: it already derives rows from `visibleGroupIds` / `visibleSessionIds` and treats a collapsed group as expanded while `active`, so it follows the combined result automatically.
+### Search-mode vs. row-filtering (revised during review)
+
+`filter.active` originally drove two things at once: *row filtering* and the *search-mode locks* — force-expanding collapsed groups, disabling collapse chevrons, and disabling drag-reorder. That conflation is safe only while `active` means a transient text query. The active-only toggle is **persisted**, so keeping the locks on `filter.active` would disable collapse and drag app-wide across launches (a real bug caught in review).
+
+So the two concerns are split:
+
+- **Row filtering** stays on `filter.active` (text OR active-only) — `visibleTopLevelGroups`, the session/sub-group visibility filters, and the empty-state gate.
+- **Search-mode locks** move to a separate `isSearching = filterText.trim().length > 0` (text query only) — force-expand, chevron `disabled`, `draggable`, the keyboard collapse/expand/select guards, and the collapse hint.
+
+`buildNavItems` therefore takes an explicit `forceExpand` parameter (defaulting to `filter.active` for back-compat) which `App.tsx` passes `isSearching`, so keyboard navigation mirrors exactly what renders: under the toggle a collapsed group is still collapsed (its sessions neither rendered nor navigable), while under a text search collapsed groups force-expand as before.
+
+**Consequence — the collapse asymmetry:** with the active-only toggle on (and no text query), a collapsed group that contains an active session shows in the list but stays collapsed; you expand it to see which session is live. This differs from the text filter, which force-expands. It is deliberate: the toggle is persistent and must not override the user's collapse state or lock collapse/drag. (Confirmed by the `activeOnly does NOT force-expand` nav test.)
 
 ## App.tsx integration
 
-1. `const [showActiveOnly, setShowActiveOnly] = useState(false)`; load from `getPreference('sidebar.showActiveOnly')` in a mount effect; on toggle, set state and `setPreference(...)`.
+1. Persistence lives in a `useActiveOnlyPreference` hook: loads `getPreference('sidebar.showActiveOnly')` once on mount, persists via `setPreference` on toggle, and ignores a stale mount-read once the user has interacted (closes a load race).
 2. `filter = useMemo(() => computeGroupFilter(groups, sessions, filterText, showActiveOnly), [groups, sessions, filterText, showActiveOnly])`.
-3. `visibleTopLevelGroups` and the auto-expand (`filter.active || !group.collapsed`) are unchanged — they already key off `filter.active`, which now also reflects the toggle.
+3. `const isSearching = filterText.trim().length > 0` gates the search-mode locks (see above); `visibleTopLevelGroups` and the render-time filters stay on `filter.active`.
 4. Render the toggle button in the `.sidebar-filter` row, left of the input.
 5. Empty state adapts: when `filter.active` and nothing renders — "No groups have active sessions" if `showActiveOnly && !filterText.trim()`, otherwise the existing "No groups or sessions match".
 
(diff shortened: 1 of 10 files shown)


docs/superpowers/specs/2026-07-30-active-sessions-toggle-design.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+The `!activeOnly` guard on the second clause is the AND: with the toggle on, a name-matched group whose subtree has no active session is hidden. With the toggle off, behavior is byte-identical to #141 (verified by keeping every existing `groupFilter` test green).
+
+**Faithfulness check** (worked in tests):
+- toggle off → identical to current text filter, including empty name-matched groups and empty sub-groups under a name-matched parent still showing.
+- parent name-matches, child has one active + one stopped session, toggle on → parent + child visible, only the active session renders.
+- parent name-matches, all descendant sessions stopped, toggle on → hidden.
+
+`buildNavItems` is unchanged: it already derives rows from `visibleGroupIds` / `visibleSessionIds` and treats a collapsed group as expanded while `active`, so it follows the combined result automatically.
```
William-Long-II: This claims buildNavItems and the auto-expand behavior are unchanged and still keyed off `filter.active`, but the shipped diff adds a `forceExpand` param and a separate `isSearching` flag in App.tsx specifically to decouple expand/collapse/drag gating from the (now broader) `filter.active`. Please update this doc to match the actual behavior.

William-Long-II (approved):
Solid PR overall — the two-pass filter rewrite is clean and the test coverage is genuinely good (error-counts-as-active, sub-group propagation, activeOnly+text intersection, the toggle-off-is-byte-identical invariant). The `useActiveOnlyPreference` race guard against a toggle landing before the async preference read resolves is a nice catch too, better than what the plan doc sketched.

One real problem though: `docs/superpowers/specs/2026-07-30-active-sessions-toggle-design.md` says at line 91 that "`buildNavItems` is unchanged... treats a collapsed group as expanded while `active`" and at line 97 that "the auto-expand (`filter.active || !group.collapsed`) ... already key off `filter.active`". That's not what the diff does. `buildNavItems` gains a new `forceExpand` param, and `App.tsx` introduces a separate `isSearching` flag (`filterText.trim().length > 0`) specifically to gate force-expand, drag-disable, and collapse/expand so the *persisted* activeOnly toggle doesn't permanently kill collapse/drag across restarts. Arguably the better design, but it directly contradicts the spec that's shipping in this same PR. Update the doc before merge — someone reading it later will assume `filter.active` drives this and be wrong.

Related: toggling activeOnly alone (no text) does not force-expand collapsed groups — confirmed by the new `'activeOnly does NOT force-expand...'` test and `expanded = (g) => forceExpand || !g.collapsed` at `src/renderer/store/groupFilter.ts:128`. So a collapsed group with an active session inside shows up in the sidebar but you can't see which session is active without expanding it manually. Asymmetric with how the text filter behaves, and the design doc's UI/verification sections don't call it out. There's a comment explaining the reasoning so I don't think it's accidental, but this is a UX call, not just an implementation detail — get explicit sign-off on it rather than letting it ride in silently.

nit: none of the passing CI checks (SonarQube, quality-gate, GitGuardian) actually run `bun test`, so we have no CI evidence the new `groupFilter.test.ts` / `SidebarFilter.test.tsx` / `useActiveOnlyPreference.test.ts` suites pass — just that static analysis is clean.

nit: the toggle button is just the `⚡` emoji with no other visual treatment. Fine for a11y since title/aria-label cover it, but emoji-only iconography tends to render inconsistently across platforms/fonts.

docs/superpowers/specs/2026-07-30-active-sessions-toggle-design.md
brannon-bowden: Fixed in 7d634a1. Added a "Search-mode vs. row-filtering (revised during review)" section documenting the filter.active/isSearching split and the buildNavItems forceExpand param, updated the App.tsx integration steps, and called out the collapse asymmetry explicitly. The doc now matches the diff.

brannon-bowden:
Thanks for the review. Addressed:

**Spec doc (the real one)** — fixed in 7d634a1. Added a "Search-mode vs. row-filtering" section documenting the `filter.active` / `isSearching` split and the `buildNavItems` `forceExpand` param, and called out the collapse asymmetry explicitly. Doc now matches the diff.

**Collapse asymmetry** — signed off. Brannon confirmed: keep it as prune-only, respecting collapse. Rationale, now in the spec: the toggle is persisted, so auto-expanding would either override the user's collapse state or (to avoid a silent no-op chevron) lock collapse/drag across restarts — the exact regression this PR fixes. So a collapsed group with an active session shows in the list but stays collapsed; you expand to see the live session.

**nit: emoji icon** — keeping `⚡`. It matches the app's existing icon-button convention (`📊 ⚔️ 🔑 ⚙`), and title/aria-label cover a11y. Moving off emoji would be an app-wide change, not this PR.

**nit: no CI runs `bun test`** — correct, and a fair gap: SonarQube/quality-gate/GitGuardian are all static-analysis only. I ran the suites locally (185 renderer tests green, including the three new files) but there's no CI evidence. Adding a `bun test` CI job is a repo-wide infra change — flagging for a separate PR rather than expanding this one. Want me to open an issue for it?

PR #151, opened 2026-07-31
fix(accounts): make a session account switch apply to the running session

William-Long-II:
## The bug

Right-clicking a session and picking a different Claude account had no effect on that session — it kept running under the old account. You reach for this menu precisely when an account is used up, so "no effect" is the worst possible outcome.

**Cause:** the menu item only wrote `claude_account_id` to the DB. The account reaches the CLI solely as `CLAUDE_CONFIG_DIR`, which `PtyManager.buildAgentSpawn` resolves **once, at spawn time** (`src/main/pty-manager.ts:388`). A live session therefore ignores the column write until its pty is respawned.

## The fix

New `src/main/account-switch.ts` owns the assignment and does the two things the column write can't:

1. **Reports which sessions actually moved** — resolves the effective account before and after, so clearing a session override that already matched its group's account doesn't needlessly kill a pty. The renderer respawns the reported sessions, which is what makes the picked account take effect.
2. **Carries the conversation across** — the stored resume UUID points into the *old* account's `projects/` tree, so a naive restart fails `--resume` and drops into the resume-failure fallback, silently orphaning the conversation. The transcript is now copied into the new account's config dir (same idea as the existing `seedLegacyConversations`). When it can't be carried over, the stale UUID is cleared so the respawn starts clean rather than burning a doomed `--resume`.

## Restart behavior

- **Session switch** — restarts immediately. You pointed at one session and picked an account; the restart is the request.
- **Group switch** — opt-in. The assignment persists right away, then a confirm dialog asks before respawning ("3 running sessions in this group are still on the previous account…"). Declining leaves them to pick it up on their next restart. Sessions with their own override are never touched.

## Refactor note

`resolveAccountForSession` moves from `repositories/accounts` to a new `account-resolver` module. It reads sessions and groups as well as accounts, so it was never that repository's to own — and the move keeps it clear of the module `pty-manager.test.ts` stubs process-wide, which was making the new tests pass alone and fail in the full suite. The shared row mapper lands in `repositories/account-row.ts`.

## Testing

- 8 new tests in `src/main/__tests__/account-switch.test.ts`: restart reporting, the unchanged-account no-op, transcript carry-over, stale-UUID invalidation, not clobbering an existing target transcript, and the group-inheritance path.
- Full suite: 452 pass, 0 fail. Both tsconfigs typecheck. `bun run build` succeeds.
- **Not covered by tests:** the new group confirm dialog is UI-only — worth a click-through on a group with two or three live sessions.

### DIFF
--- a/src/main/__tests__/account-switch.test.ts
+++ b/src/main/__tests__/account-switch.test.ts
@@ -1,0 +1,245 @@
+/**
+ * Account-switch tests (BDHLNDR-31) — reassigning a session or group's Claude
+ * account has to report which ptys need respawning (CLAUDE_CONFIG_DIR is fixed
+ * at spawn time) and carry the in-flight conversation transcript into the new
+ * account's config dir so the respawn can still --resume it.
+ *
+ * Uses bun:sqlite (API-compatible with better-sqlite3 for the statements the
+ * repos issue) so no native build is required to run the suite.
+ *
+ * Run with: bun test src/main/__tests__
+ */
+import { describe, expect, test, beforeEach, afterEach, mock } from 'bun:test';
+import { Database } from 'bun:sqlite';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
+let db: Database;
+
+mock.module('electron', () => ({
+  app: { getPath: () => '/nonexistent-bodhilander-test-userdata' },
+}));
+
+mock.module('../database', () => ({
+  getDatabase: () => db,
+}));
+
+const accountSwitch = await import('../account-switch');
+
+let tmp: string;
+
+function freshDb(): Database {
+  const d = new Database(':memory:');
+  d.exec(`
+    CREATE TABLE sessions (
+      id TEXT PRIMARY KEY,
+      group_id TEXT NOT NULL,
+      name TEXT NOT NULL,
+      working_dir TEXT,
+      state TEXT,
+      shell_type TEXT DEFAULT 'bash',
+      "order" INTEGER DEFAULT 0,
+      created_at TEXT,
+      last_activity_at TEXT,
+      claude_session_id TEXT DEFAULT NULL,
+      ended_at TEXT DEFAULT NULL,
+      duration_seconds REAL DEFAULT 0,
+      claude_account_id TEXT DEFAULT NULL,
+      provider TEXT NOT NULL DEFAULT 'claude'
+    );
+    CREATE TABLE groups (
+      id TEXT PRIMARY KEY,
+      name TEXT NOT NULL,
+      color TEXT,
+      working_dir TEXT,
+      "order" INTEGER DEFAULT 0,
+      created_at TEXT,
+      parent_id TEXT DEFAULT NULL,
+      collapsed INTEGER DEFAULT 0,
+      claude_account_id TEXT DEFAULT NULL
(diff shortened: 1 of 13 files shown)


brannon-bowden (approved):
Approve — this closes a real correctness gap (live sessions silently kept billing the old account since `CLAUDE_CONFIG_DIR` is baked in at pty spawn) with a well-reasoned design and solid backend test coverage. No blocking issues found. CI is green (SonarQube, quality-gate, GitGuardian all passing). I did not find any actual attempt in the diff text to manipulate this review despite the caution flag — the flagged comments are ordinary engineering rationale, not instruction-override attempts.

**Design review**

- `account-resolver.ts` correctly preserves the pre-existing fallback chain (session → group → default → null); I traced the SQL rewrite against the old `resolveAccountForSession` in `repositories/accounts.ts` and the COALESCE + LEFT JOIN produces identical results, including the case where a referenced account was deleted (falls through to default in both old and new code).
- `assignSessionAccount`/`assignGroupAccount` read-before/after and only report sessions whose *effective* account actually changed — correctly leaves overridden sessions alone during a group switch, and correctly no-ops when clearing an override that already matched the inherited value. Confirmed by the accompanying tests.
- Transcript carry-over degrades safely: on any failure (missing file, unreadable, copy error) it drops the stored `claude_session_id` rather than leaving a stale UUID that would cause a doomed `--resume`.
- No new race window: both switch functions are fully synchronous (better-sqlite3/bun:sqlite), so there's no interleaving risk between the pre/post reads within a single IPC handler invocation.

**Non-blocking issues**

- `src/main/account-switch.ts:121` (and the `candidate` path at line 146) — the transcript path is built by string-interpolating `claude_session_id` directly (`${uuid}.jsonl`) without validating it looks like a UUID. In current usage this value is only ever written by the app itself, so it's low risk today, but since it flows into a filesystem path, a defensive format check (or `path.basename` on the input) would remove the assumption entirely.
- `src/main/account-switch.ts:135-142` — `findTranscript`'s `readdirSync` failure is swallowed identically whether the `projects` dir is simply absent (expected/benign) or an unexpected error occurs (e.g. permissions). Consider only treating `ENOENT` as benign and logging other errors, so real environment problems aren't silently mistaken for "no transcripts yet."
- `src/renderer/App.tsx:1500` — the restart-confirmation button for `restartForAccountSwitch` reuses `btn-danger`, but restarting a session to apply an account switch isn't a destructive action the way closing a session or deleting a group is. Minor UX/styling nit.

**Tests**

- Main-process logic (`account-resolver.ts`, `account-switch.ts`) is thoroughly covered: persistence, no-op detection, transcript carry, UUID drop on missing transcript, and group-level fan-out/opt-out via override.
- Not covered: the renderer-side orchestration added in `App.tsx` (`liveSessionsAmong`, `handleAssignSessionAccount`, `handleAssignGroupAccount`, and the confirm-dialog restart path) — this is real branching logic (filtering stopped sessions, deciding when to prompt vs. auto-restart) with no direct test. Given the rest of this PR is unusually well-tested, this stands out as a gap, though I'm not blocking on it since I can't see whether this repo has an established renderer-testing convention to follow.
- The transcript-copy *failure* path (`fs.copyFileSync` throwing, e.g. disk full/permissions) is only exercised indirectly; only the "file doesn't exist" branch is tested explicitly.

PR #152, opened 2026-07-31
fix(accounts): address review follow-ups on the account switch

William-Long-II:
Follow-up to #151, covering the three non-blocking points @brannon-bowden raised in that review.

### 1. Validate the conversation id before it becomes a path (`account-switch.ts:112`)

The transcript path interpolated `claude_session_id` straight into `${uuid}.jsonl`. These ids are always `crypto.randomUUID()` output written by this app, so nothing in the current flow can produce a hostile value — but the id arrives from a database column and is used to build a filesystem path, and that pairing shouldn't rest on an assumption about the writer.

Ids must now match `/^[A-Za-z0-9][A-Za-z0-9._-]*$/` and contain no `..`. One that fails carries nothing, which drops the id and restarts on a fresh conversation — the same degradation as a missing transcript.

### 2. Distinguish "no transcripts yet" from a broken environment (`account-switch.ts:145`)

`findTranscript` swallowed every `readdirSync` failure identically. A missing `projects/` dir is the ordinary case and stays quiet; anything else (permissions, I/O) is now logged, so a broken machine stops presenting itself as a routine fresh conversation.

### 3. Restart confirmation is not a destructive action (`App.tsx:1500`)

The confirm button reused `btn-danger`. Closing a session or deleting a group destroys something; restarting a session to apply an account switch just re-runs it. It now uses `btn-primary` in that case, `btn-danger` for the two genuinely destructive ones.

## Testing

Two new tests in `account-switch.test.ts`:
- a traversal-shaped id (`../../escape`) writes nothing outside the target dir and drops the stored id
- a `projects/` path that is a file, not a directory (`ENOTDIR` — the "real problem" branch) degrades without throwing

Full suite: 454 pass, 0 fail. Both tsconfigs typecheck.

## Still open from that review

The renderer-side orchestration (`liveSessionsAmong`, both assign handlers, the confirm-dialog restart path) still has no direct test coverage — not addressed here, since it wants a renderer-testing approach rather than a one-off.

### DIFF
--- a/src/main/__tests__/account-switch.test.ts
+++ b/src/main/__tests__/account-switch.test.ts
@@ -186,6 +186,36 @@ describe('assignSessionAccount', () => {
     expect(storedUuid('s1')).toBeNull();
   });
 
+  test('refuses to build a transcript path from an unexpected id format', () => {
+    const dirA = addAccount('acct-a', true);
+    const dirB = addAccount('acct-b');
+    addGroup('g1');
+    // Nothing in the app writes an id like this — the guard exists so the
+    // path build never depends on that staying true.
+    addSession('s1', 'g1', { accountId: 'acct-a', claudeSessionId: '../../escape' });
+    writeTranscript(dirA, '-Users-x-repo', 'uuid-1');
+
+    accountSwitch.assignSessionAccount('s1', 'acct-b');
+
+    expect(fs.existsSync(path.join(tmp, 'escape.jsonl'))).toBe(false);
+    expect(fs.existsSync(path.join(dirB, 'escape.jsonl'))).toBe(false);
+    // Nothing was carried, so the id is dropped and the respawn starts fresh.
+    expect(storedUuid('s1')).toBeNull();
+  });
+
+  test('an unreadable projects dir degrades to a fresh conversation', () => {
+    const dirA = addAccount('acct-a', true);
+    addAccount('acct-b');
+    addGroup('g1');
+    addSession('s1', 'g1', { accountId: 'acct-a', claudeSessionId: 'uuid-1' });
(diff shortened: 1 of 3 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped, defense-in-depth follow-up (path-safety guard for conversation ids, clearer error handling in `findTranscript`, and a UI styling fix) with green CI (SonarQube, quality-gate, GitGuardian all passing) and no blocking issues.

- `src/main/__tests__/account-switch.test.ts:189-201` — the new "refuses to build a transcript path from an unexpected id format" test writes the real transcript under the name `uuid-1` (`writeTranscript(dirA, '-Users-x-repo', 'uuid-1')`) while the session's `claudeSessionId` is `../../escape`. Since no file exists at the location a real traversal would resolve to, `findTranscript` would return `null` and the assertions would pass identically even if `isPathSafeConversationId`'s check were removed. Consider placing a file at the actual traversal target (e.g., a `escape.jsonl` one/two levels above the account's `projects/` dir) so the test can distinguish "guard blocked it" from "the path just didn't happen to exist."
- `src/main/account-switch.ts:129` — `log.warn` interpolates the rejected `uuid` verbatim before it's been validated, so an id containing newlines/control characters would be written straight into the log. Low impact here since ids currently only come from this app's own sqlite writes, but worth sanitizing/truncating if this log line is ever consumed by tooling that trusts log line boundaries.
- The second new test ("an unreadable projects dir degrades to a fresh conversation") verifies the non-throwing/fresh-conversation behavior but doesn't assert that `log.warn` was actually invoked for the non-ENOENT branch — the stated goal of that hunk is to stop silently swallowing real environment errors, so a spy/assertion on the warn call would make the regression test complete.

PR #155, opened 2026-08-13
fix(relay): harden the tunnel before sharing (M5.0)

### LINKED ISSUE #154: Bug: relay channel keys derive from a static agent key, so recorded tunnels replay
## Description

The shipped relay tunnel (M3) derives its channel key from the agent's **long-lived** X25519 identity key. `relay-identity.ts`'s `deriveSharedSecret` mixes that static key with the client's ephemeral key, so a channel key is a pure function of the client key — and the handshake proof the agent signs covers no per-channel-fresh value.

That combination makes a recorded channel replayable by the one party that sits in the middle of every channel: the relay.

1. The relay records a full owner session — the signed proof and every sealed frame.
2. It replays the proof to a new browser. The signature still verifies; nothing in it is stale.
3. The browser presents the same client key, so the same shared secret, so the **same AES-256-GCM channel key**.
4. `recvCounter` starts at `-1` per channel, so every recorded frame replays in order and decrypts.

Worse than replay: both sides now seal *new* traffic under a key that was already used, with counters restarting from zero. That is AES-GCM key + nonce reuse — the classic two-time pad. XOR two frames sealed under the same (key, nonce) and the keystream cancels, leaking plaintext of live terminal input.

This is a defect in already-released code, not a hypothetical in a proposed design. It must be fixed before session sharing (`docs/designs/session-sharing.md`) lands: sharing multiplies channels per machine and puts channel establishment under untrusted control. It is survivable with a single owner and is not survivable with guests.

The same pass covers the other hardening items M5.0 scopes:

- `to-client` is unbound — an authenticated agent can name **any** `clientId` in the relay's routing table and inject frames into a different machine's channel.
- Three PTY listeners are registered per client against an `EventEmitter` whose default max is 10, and duplicate `client:open` leaks a set per call.
- `/link` and `/link/claim` have no rate limiting.
- `purgeExpiredSessions` is on the repository interface with no caller anywhere; nothing ever deletes expired link codes.
- Unauthenticated agent sockets linger indefinitely; browsers can't send WS control frames from script, so idle viewers look dead to intermediaries.
- History replay seals an entire serialized scrollback buffer into a single frame.

## Steps to reproduce

The replay, against the shipped code:

1. Open a relay session from the browser to a linked machine and type into the terminal.
2. Capture the relay's `client:open` payload (the signed handshake proof) and the subsequent `from-agent` frames.
3. Open a new channel to the same machine, presenting the **same** client X25519 public key, and replay the captured proof to the browser in place of the agent's.
4. Replay the captured frames. They decrypt — `recvCounter` restarted at `-1` and the channel key is byte-identical to the first channel's.

## Expected behavior

A channel key is unique per channel. Frames recorded from one channel do not open on any other, even when the client presents an identical key, and a replayed handshake proof cannot establish a channel at all.

## Actual behavior

The channel key is a deterministic function of the client's key alone. Recorded frames replay and decrypt, and live traffic is subsequently sealed under a reused (key, nonce) pair.

## Severity

Critical - production down / data loss

`development` (relay + desktop, M3 tunnel as shipped). Design context: `docs/designs/session-sharing.md` §1.1.

William-Long-II:
Closes #154

M5.0 of `docs/designs/session-sharing.md` — the hardening pass that has to land before any sharing code does. No user-visible change; every fix here is to the *existing single-user* tunnel.

The design doc is included in this branch (first commit) because the reasoning for several of these fixes only makes sense against it.

## The headline: channel keys were derived from a static key

`relay-identity.ts` derived the channel secret from the agent's **long-lived** X25519 identity key, so the channel key was a pure function of the client's key — and the signed handshake proof covered no per-channel-fresh value. The relay could record a session, replay the proof to a new browser, and every captured frame would decrypt (`recvCounter` starts at `-1` per channel). Both sides would then seal *new* traffic under an already-used key with counters restarting at zero — AES-GCM key + nonce reuse, which leaks plaintext to anyone who XORs two frames.

**The fix came out smaller than the design specified.** §1.1 originally also added a `channelNonce` and bumped the proof to `e2e-handshake:v2`. Reading `connection.ts:69-78` closely, that is unnecessary and worse than unnecessary. The client derives from whatever `agentX25519Pub` the *signed* proof carries and pins only `machineEd25519Pub`, so making the agent's half ephemeral is a **pure agent-side change** — no wire format bump, no client change, no deploy ordering. And once the agent's key is fresh per channel, the proof already covers a fresh value: a replayed proof names an ephemeral public key whose private half no longer exists, so the channel simply fails to establish. A v2 bump would have needed a coordinated relay/desktop deploy plus a version-negotiation field that is itself a relay-selectable downgrade path. The doc now marks that recommendation **Superseded**.

Forward secrecy falls out for free — recovering the machine identity key no longer decrypts any recorded channel.

`deriveEphemeral` lives in `e2e.ts` rather than `relay-identity.ts` deliberately: `e2e.ts` imports only `crypto`, no Electron, so it is unit-testable. `deriveSharedSecret` is **deleted** rather than left exported — it is precisely the primitive whose reintroduction would undo this.

## Also in this branch

**`to-client` was unbound** (`ws.ts:180`). An authenticated agent could name any `clientId` in the relay's routing table and inject frames into a different machine's channel. Now scoped to the sender's own machine.

**PTY listeners are per-tunnel, not per-client.** `ClientSession` held three closures each registered on `ptyManager`, against an `EventEmitter` whose default max-listeners is 10 — four viewers already produced a leak warning, and a duplicate `client:open` leaked a whole set with no path to removal. One listener set now registers on the first client and unregisters when the last one leaves, fanning out over subscriptions.

**Rate limiting** on `/link` (per address *and* per keypair, so rotating IPs doesn't buy anything) and `/link/claim`. Client-IP resolution takes the **rightmost** `X-Forwarded-For` hop when `trustProxy` is set — the leftmost is caller-supplied, and honoring it lets one caller mint unlimited buckets. It returns `null` rather than a placeholder when nothing resolves, so callers **fail open**: bucketing every unresolvable request together would turn one abuser into an outage for everyone behind that path.

**The reapers.** `purgeExpiredSessions` was on the repository interface with no caller anywhere in the tree, and nothing ever deleted expired link codes. `purgeExpiredLinkCodes` deletes only `status = 'pending'` rows — a claimed code's row is the audit trail of the claim.

**History replay is chunked** at 64 KiB. It previously sealed an entire serialized scrollback buffer into one frame. Chunking is safe because xterm's parser is stateful across `write()` calls; `chunkText` splits without breaking surrogate pairs, and the loop re-checks the subscription between chunks so a client that unsubscribes mid-replay stops receiving.

**Smaller ones:** duplicate `client:open` refused (`4400`) instead of re-pointing a socket at a different machine · duplicate `clientId` refused at the tunnel · agent socket replacement closes the socket it supersedes, publishing the new socket *before* closing the old one so a synchronous close handler can't delete the live entry or emit a spurious `agent:offline` · unauthenticated agent sockets reaped at 30s · browser keepalive ping, since browsers can't send WS control frames from script and an idle viewer otherwise looks dead to any intermediary.

## One thing deliberately left alone

`maxPayloadLength` is now set explicitly but left at 16 MB. Agents auto-update on their own schedule, and a pre-chunking agent still seals a whole serialized buffer into one frame. Tightening it now would break history replay for clients that no relay-side change can reach. It comes down once chunking has shipped long enough.

## Testing

`e2e.test.ts` states the property directly rather than testing the implementation: the **same** peer key yields a different channel key every time, and a frame sealed on one channel no longer opens after re-deriving. Plus `chunking.test.ts` (surrogate pairs, boundaries) and `rate-limit.test.ts` (windows, per-key isolation, XFF resolution incl. the spoofed-leftmost case).

The cross-machine injection test proves the drop by **ordering** rather than a sleep — it sends the injection, then a legitimate frame, and asserts the legitimate one arrives first with nothing before it.

- Desktop: **489 pass, 0 fail** (1131 assertions, 41 files)
- Relay: **62 pass, 0 fail** (225 assertions, 5 files)
- `tsc -p tsconfig.main.json --noEmit`, relay `tsc --noEmit`, `tsc -p web/tsconfig.json` all clean
- `bun run build:web` bundles

## For the reviewer

Two things worth knowing before this goes near production.

**The relay tree is not Sonar-scanned.** `sonar.sources=src`, so `relay/src/rate-limit.ts` and the `ws.ts` changes get no quality gate — the desktop files do. Human review is the only gate on the relay half of this PR.

**Deploying needs a rebuild and recreate, not a restart.** `--env-file` and the image are read at `docker run` time.

### DIFF
--- a/src/main/api/relay/session-tunnel.ts
+++ b/src/main/api/relay/session-tunnel.ts
@@ -21,8 +21,16 @@ import { ptyManager } from '../../pty-manager';
 import { getAllSessions } from '../../repositories/sessions';
 import { getAllGroups } from '../../repositories/groups';
 import { createRemoteSession, createRemoteGroup } from './remote-sessions';
-import { deriveSharedSecret, ensureIdentity, signWithIdentity } from './relay-identity';
-import { deriveSessionKey, sealJson, openJson, buildHandshakeProof, type SealedFrame } from './e2e';
+import { ensureIdentity, signWithIdentity } from './relay-identity';
+import {
+  deriveEphemeral,
+  deriveSessionKey,
+  sealJson,
+  openJson,
+  buildHandshakeProof,
+  type SealedFrame,
+} from './e2e';
+import { chunkText } from './chunking';
 import { sanitizeSize } from './terminal-size';
 
 /** Expand a leading `~` to the user's home directory. */
@@ -39,9 +47,6 @@ interface ClientSession {
   recvCounter: number;
   /** Session ids this client is streaming. */
   subs: Set<string>;
-  onData: (e: { id: string; data: string }) => void;
-  onExit: (e: { id: string; exitCode: number }) => void;
-  onResize: (e: { id: string; cols: number; rows: number }) => void;
 }
 
 /** Decrypted command from a web client (union of every message's fields). */
@@ -62,9 +67,58 @@ interface ClientFrame {
 
 export class SessionTunnel {
   private readonly sessions = new Map<string, ClientSession>();
+  /** Whether the shared PTY listeners below are currently attached. */
+  private listening = false;
 
   /** `route` sends an outbound payload to a client id via the relay. */
   constructor(private readonly route: (clientId: string, payload: unknown) => void) {}
+
+  // One listener per PTY event for the whole tunnel, fanned out to subscribed
+  // clients. Previously each client attached its own three, which leaked a set
+  // per duplicate `client:open` and tripped EventEmitter's default max of 10
+  // listeners once a handful of viewers were connected.
+  private readonly onPtyData = (e: { id: string; data: string }): void => {
+    this.fanOut(e.id, (clientId) => this.sealTo(clientId, { type: 'terminal:output', sessionId: e.id, data: e.data }));
+  };
+
+  private readonly onPtyExit = (e: { id: string; exitCode: number }): void => {
+    this.fanOut(e.id, (clientId) =>
+      this.sealTo(clientId, { type: 'terminal:exit', sessionId: e.id, exitCode: e.exitCode }),
+    );
+  };
+
+  // Dynamic sizing: when the PTY is resized (by any viewer), tell subscribed
+  // clients the new size so they re-render at it.
+  private readonly onPtyResize = (e: { id: string; cols: number; rows: number }): void => {
+    this.fanOut(e.id, (clientId) =>
+      this.sealTo(clientId, { type: 'terminal:size', sessionId: e.id, cols: e.cols, rows: e.rows }),
+    );
+  };
+
+  /** Run `send` for every client subscribed to `sessionId`. */
+  private fanOut(sessionId: string, send: (clientId: string) => void): void {
+    for (const [clientId, s] of this.sessions) {
+      if (s.subs.has(sessionId)) send(clientId);
+    }
+  }
+
+  /** Attach the shared PTY listeners once, on the first connected client. */
+  private startListening(): void {
+    if (this.listening) return;
+    ptyManager.on('data', this.onPtyData);
+    ptyManager.on('exit', this.onPtyExit);
+    ptyManager.on('resize', this.onPtyResize);
+    this.listening = true;
+  }
+
+  /** Release them again once the last client goes away. */
+  private stopListeningIfIdle(): void {
+    if (!this.listening || this.sessions.size > 0) return;
+    ptyManager.off('data', this.onPtyData);
+    ptyManager.off('exit', this.onPtyExit);
+    ptyManager.off('resize', this.onPtyResize);
+    this.listening = false;
+  }
 
   /** A web client opened a channel — complete the E2E handshake. */
   open(clientId: string, payload: unknown): void {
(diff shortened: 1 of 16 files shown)


src/main/api/relay/session-tunnel.ts
```diff
@@ -62,45 +67,92 @@ interface ClientFrame {
...
+    for (const [clientId, s] of this.sessions) {
+      if (s.subs.has(sessionId)) send(clientId);
+    }
+  }
+
+  /** Attach the shared PTY listeners once, on the first connected client. */
+  private startListening(): void {
+    if (this.listening) return;
+    ptyManager.on('data', this.onPtyData);
+    ptyManager.on('exit', this.onPtyExit);
```
brannon-bowden, line 109: Duplicate-client:open dedup, tunnel-level PTY listener fan-out, and chunked-history abort logic are new and non-trivial but have no direct SessionTunnel-level test (only the pure e2e/chunking helpers are unit tested). Design doc defers the DI refactor to enable this to M5.1 — recommend landing that soon given this file's sensitivity.

brannon-bowden (approved):
Approve. This is a well-scoped, carefully reasoned hardening pass, and the crypto change (replacing the long-lived agent X25519 key with a per-channel ephemeral key) is the correct fix for the replay/nonce-reuse issue the accompanying design doc identifies, with solid direct tests (`e2e.test.ts`) proving the replay property, tamper rejection, and counter-collision resistance. The relay-side hardening (cross-machine `to-client` binding check, one-socket-per-machine replacement with correct map-then-close ordering, duplicate `client:open` rejection, agent auth timeout, rate limiting, request/payload size caps, reapers) is each backed by a targeted test in `ws.test.ts` / `m2.test.ts` / `rate-limit.test.ts`.

Specific things I checked closely and found correct:
- `ws.ts`: publishing the new agent socket into `agents` *before* closing the replaced one, so the old socket's `close` handler's `agents.get(machineId) === ws` check can never match and spuriously evict the new connection — verified against the `reconnecting an agent closes the socket it replaces` test.
- `rate-limit.ts`: fixed-window bucketing, rightmost-hop `X-Forwarded-For` selection (correct against header-spoofing since a trusted single-hop proxy appends rather than overwrites), and the deliberate per-key charge happening only after Ed25519 signature verification (so an attacker can't grief another key's budget without owning the key).
- `session-tunnel.ts`: `ensureIdentity()`/`signWithIdentity()` are called and can throw *before* `this.sessions.set(...)` / `this.startListening()`, so a handshake failure never leaves orphaned tunnel state; the chunked-history loop re-checks subscription/session-identity on every iteration, correctly aborting a stale replay if the client unsubscribes or disconnects mid-stream.

Non-blocking issues

1. `src/main/api/relay/session-tunnel.ts` — the new dedup-on-`client:open`, tunnel-level fan-out listener lifecycle, and chunked-history-abort logic are non-trivial and security-relevant, but there's no direct unit/integration test exercising `SessionTunnel` itself (only its pure crypto (`e2e.test.ts`) and chunking (`chunking.test.ts`) helpers, plus the *relay-side* analog in `ws.test.ts`). The design doc explicitly acknowledges this gap and defers the DI refactor needed to unblock it to M5.1 with a concrete technical reason (module-scope singleton imports break bun's process-wide `mock.module()`). Given that's reasoned and milestoned immediately next, I'm not blocking on it, but flagging it since this is exactly the class of file called out for extra scrutiny.
2. `relay/src/rate-limit.ts` — `windows` is an unbounded `Map` between `sweep()` calls (every 10 minutes per `REAP_INTERVAL_MS`); an attacker cycling many distinct IPs/keys within that window grows memory without a ceiling. Likely fine at current scale, but worth a max-entries eviction if abuse patterns ever show up in the field.
3. `relay/web/src/connection.ts:63-77` — `startPing()` is invoked synchronously in `connect()`, before the socket reaches `OPEN`. Harmless today since the first tick is 30s away (well past handshake time), but if `PING_INTERVAL_MS` is ever shortened this could throw on a `CONNECTING` socket.

Tests

Coverage on the changed relay/crypto surface is strong (rate limiting, cross-machine injection, socket replacement, duplicate-open, auth timeout is the one exception — no test asserts a socket is actually closed after `AGENT_AUTH_TIMEOUT_MS` elapses without a challenge response, only that the timer/field plumbing exists). Consider adding a fake-timer test for the 30s agent-auth-timeout close path, and — per note 1 above — direct tests of `SessionTunnel`'s dedup/fan-out/chunk-abort behavior once the DI refactor lands in M5.1.

PR #157, opened 2026-08-13
fix(relay): bound the rate limiter and cover the auth-timeout close

### LINKED ISSUE #156: Bug: relay rate limiter holds unbounded windows between sweeps
## Description

Follow-ups from @brannon-bowden's review on #155 (M5.0 relay hardening). He approved and they were all marked non-blocking, so the PR merged as-is. Two hold up on inspection and one does not.

### 1. `relay/src/rate-limit.ts` — `windows` is an unbounded Map

`createRateLimiter` stores one entry per key and only drops expired ones on `sweep()`, which the reaper calls every 10 minutes (`REAP_INTERVAL_MS`). Between sweeps the map grows once per distinct key with no ceiling.

The keys are `link:ip:<addr>`, `link:key:<pubkey>` and `claim:ip:<addr>`. An attacker with an IPv6 /64 has effectively unlimited distinct source addresses, and `/link` is reachable before any signature check charges the per-key bucket. Sustained traffic across fresh addresses grows the map for a full 10-minute window. It is a memory-exhaustion path on a public endpoint.

**The obvious fix introduces a worse bug.** Evicting oldest-first or LRU bounds memory by handing the attacker a rate-limit reset: flood distinct keys until the entry that is throttling *you* gets evicted, then start over with a clean bucket. Eviction has to be count-aware — a window that is not yet throttling anyone is free to drop (the holder was under the limit anyway), a window that is actively throttling someone is not.

### 2. `relay/src/ws.ts` — the agent auth timeout has no behavioural test

`AGENT_AUTH_TIMEOUT_MS` is a module constant baked into `setTimeout`, so nothing can assert that a socket which never answers the challenge is actually closed with `4401` — only that the timer field is plumbed. Testing it needs a seam to override the duration.

### 3. `relay/web/src/connection.ts` — startPing before OPEN (not a defect)

The review flagged `startPing()` being called synchronously in `connect()` as something that "could throw on a `CONNECTING` socket" if `PING_INTERVAL_MS` were ever shortened. It can't: `send()` already guards on `readyState === WebSocket.OPEN` (`connection.ts:128`), so an early tick is a silent no-op. Filing it only as a tidy — move the call into `ws.onopen` so the timer doesn't exist during a window where it can do nothing.

### Not in scope

Direct `SessionTunnel` tests (his note 1, and his inline comment) stay in **M5.1**, where the DI refactor that unblocks them is already scheduled. Module-scope singleton imports break bun's process-wide `mock.module()`, so there is no cheap version of that test today.

## Steps to reproduce

For (1), against `development`:

1. Point sustained `/link` traffic at the relay from many distinct source addresses (or many distinct keypairs) inside a single 10-minute reaper interval.
2. Observe the limiter's `windows` map retain one entry per distinct key for the whole interval, with no ceiling.

For (2): there is no test to run — that is the finding.

## Expected behavior

1. The limiter holds a bounded number of windows regardless of how many distinct keys are seen, and being at capacity never resets a bucket that is actively throttling someone.
2. A test proves an agent socket that never answers the challenge is closed with `4401`.
3. The ping timer is only armed once the socket is open.

## Actual behavior

1. `windows` grows once per distinct key until the next `sweep()`, bounded only by traffic.
2. Only the timer plumbing is covered; the close path is not.
3. The timer is armed while the socket is still `CONNECTING`.

## Severity

Medium - feature broken, workaround exists

`development` at `eb429b0` (merge of #155).

William-Long-II:
Closes #156

Follow-up to #155, covering the non-blocking points @brannon-bowden raised there. Two held up on inspection; one didn't, and I've written down why rather than silently changing the code.

## 1. The rate limiter's window map was unbounded (`relay/src/rate-limit.ts`)

Correct, and the more interesting half is that **the obvious fix is a bypass**.

`windows` held one entry per key and only dropped expired ones on `sweep()`, ten minutes apart per `REAP_INTERVAL_MS`. The keys include the client address, and an attacker with an IPv6 /64 has effectively unlimited distinct addresses, so the map grew with traffic on a public route.

Evicting oldest-first or LRU bounds memory by handing out a rate-limit reset: flood distinct keys until the window that is throttling *you* falls out, then start over with a clean bucket. That trades a memory bound for exactly the property the limiter exists to provide.

So eviction is **count-aware**. A window that hasn't reached its limit is free to drop — its holder was going to be allowed through anyway, so forgetting it costs nothing. A window that is actively throttling someone is never dropped to make room. Expired entries are reclaimed first, so the common case evicts nothing live.

When every tracked window is throttling, the limiter **refuses the newcomer** rather than forgetting a live limit. That's a deliberate fail-closed in a spot where the rest of this code fails open, and the reasoning is different: reaching it requires a flood that actually saturates 50k distinct buckets, forgetting a window there would reset a limit that is doing its job, and the state clears at the next window boundary. Failing open instead would mean a flood *disables rate limiting for everyone*, which is the outcome the limiter is supposed to prevent.

`MAX_WINDOWS = 50_000` — a few hundred bytes an entry, so single-digit MB. `size()` is exposed so the bound is assertable rather than assumed.

## 2. The agent auth timeout had no behavioural test (`relay/src/ws.ts`)

Correct. `AGENT_AUTH_TIMEOUT_MS` was baked into the `setTimeout` call, so nothing could assert the socket is *closed* — only that the timer field exists. `createGateway` now takes an optional `authTimeoutMs`, the same seam `createRateLimiter` already has for its clock.

Two tests, because the timeout firing and the timeout being disarmed are separate failure modes:
- a socket that answers nothing is closed `4401`
- an **authenticated** socket that stays quiet well past the window still answers a ping — which proves the `clearTimeout` in the auth path instead of assuming it. Liveness is proven by a completed ping round-trip, not by a sleep that ends without an error.

## 3. The ping-before-OPEN concern doesn't hold (`relay/web/src/connection.ts`)

The review read `startPing()` in `connect()` as something that "could throw on a `CONNECTING` socket" if `PING_INTERVAL_MS` were ever shortened. It can't — `send()` already guards on `readyState === WebSocket.OPEN` at `connection.ts:128`, so an early tick is a silent no-op regardless of the interval.

Moved it into `ws.onopen` anyway, with a comment recording that reasoning. There's no reason for the timer to exist during a window where it can do nothing, but it was a tidy, not a fix — worth being precise about so nobody later reads the move as evidence of a bug that was there.

## Deliberately not here

Direct `SessionTunnel` tests — his note 1 and the inline comment on `session-tunnel.ts:109` — stay in **M5.1**, where the DI refactor that unblocks them is already scheduled. Module-scope singleton imports break bun's process-wide `mock.module()`, so there is no cheap version of that test today, and faking one against the current shape would test the mock rather than the tunnel.

## Testing

Seven new tests. The limiter ones state the security property directly rather than the implementation: a throttled window **survives** a flood of `MAX_WINDOWS + 1000` fresh keys, an idle one does not, expired generations are reclaimed before anything live is touched, and the saturated case refuses with the right `Retry-After` and then self-heals once the clock passes the window.

- Full tree: **496 pass, 0 fail** (1143 assertions, 41 files)
- Relay standalone: **69 pass, 0 fail** (237 assertions, 5 files)
- `tsc -p tsconfig.main.json --noEmit`, relay `tsc --noEmit`, `tsc -p web/tsconfig.json` all clean
- `bun run build:web` bundles

One correction to #155's body while I'm here: it listed "Desktop: 489" and "Relay: 62" as if they summed. The root `bun test` already includes `relay/`, so 489 was the whole tree and 62 was a subset of it, not an addition to it. The numbers above don't double-count.

### DIFF
--- a/relay/src/index.ts
+++ b/relay/src/index.ts
@@ -67,6 +67,10 @@ function main(): void {
       const codes = repos.purgeExpiredLinkCodes();
       rateLimiter.sweep();
       if (codes > 0) logger.debug('reaped expired link codes', { count: codes });
+      // Only non-zero when someone is holding MAX_WINDOWS buckets at their
+      // limit, so this is a deliberate-abuse signal rather than routine noise.
+      const refused = rateLimiter.drainSaturationRefusals();
+      if (refused > 0) logger.warn('rate limiter saturated', { refused, windows: rateLimiter.size() });
     } catch (err) {
       logger.error('reaper failed', { err: err instanceof Error ? err.message : String(err) });
     }
--- a/relay/src/rate-limit.test.ts
+++ b/relay/src/rate-limit.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'bun:test';
-import { clientIp, createRateLimiter } from './rate-limit';
+import { clientIp, createRateLimiter, MAX_WINDOWS } from './rate-limit';
 
 /** A limiter driven by a clock the test controls. */
 function fixedClock(start = 1_000_000) {
@@ -63,6 +63,86 @@ describe('createRateLimiter', () => {
     expect(limiter.check('old', 1, 10_000).allowed).toBe(true);
     // ...while 'fresh' is still being counted.
     expect(limiter.check('fresh', 1, 60_000).allowed).toBe(false);
+  });
+});
+
+describe('createRateLimiter memory bound', () => {
+  /** Enough past the ceiling to force at least one eviction pass. */
+  const EXTRA = 1_000;
+
+  /** Fill the limiter with distinct single-hit keys, as a flood would. */
+  function flood(limiter: ReturnType<typeof createRateLimiter>, count: number, prefix = 'flood') {
+    for (let i = 0; i < count; i++) limiter.check(`${prefix}:${i}`, 5, 60_000);
+  }
+
+  test('stays bounded no matter how many distinct keys arrive', () => {
+    const limiter = createRateLimiter(fixedClock().now);
+
+    flood(limiter, MAX_WINDOWS + EXTRA);
+
+    expect(limiter.size()).toBeLessThanOrEqual(MAX_WINDOWS);
+  });
+
+  test('a throttled window survives a flood of fresh keys', () => {
+    // The property that makes this not an LRU. If eviction dropped the oldest
+    // entry, an attacker could flush the window that is throttling them and
+    // start over with a clean bucket -- trading a memory bound for a
+    // rate-limit bypass.
+    const limiter = createRateLimiter(fixedClock().now);
+
+    for (let i = 0; i < 3; i++) limiter.check('victim', 3, 60_000);
+    expect(limiter.check('victim', 3, 60_000).allowed).toBe(false);
+
+    flood(limiter, MAX_WINDOWS + EXTRA);
+
+    expect(limiter.check('victim', 3, 60_000).allowed).toBe(false);
+  });
+
+  test('evicts idle windows rather than throttling ones', () => {
+    // An under-limit window is free to drop: its holder was going to be let
+    // through anyway, so forgetting it costs nothing.
+    const limiter = createRateLimiter(fixedClock().now);
+
+    limiter.check('idle', 5, 60_000); // 1 of 5 -- not throttling anyone
+    flood(limiter, MAX_WINDOWS + EXTRA);
+
+    expect(limiter.size()).toBeLessThanOrEqual(MAX_WINDOWS);
+    expect(limiter.check('idle', 5, 60_000).allowed).toBe(true);
+  });
+
+  test('expired windows are reclaimed before anything live is evicted', () => {
+    const clock = fixedClock();
+    const limiter = createRateLimiter(clock.now);
+
+    flood(limiter, MAX_WINDOWS, 'stale');
+    clock.advance(60_001); // every stale window has now elapsed
+
+    // A single fresh key should reclaim the whole expired generation.
+    limiter.check('fresh', 5, 60_000);
+
+    expect(limiter.size()).toBeLessThan(MAX_WINDOWS);
+  });
+
+  test('refuses rather than forgetting a limit when every window is throttling', () => {
+    // Reaching this needs a distributed flood that actually throttles
+    // MAX_WINDOWS distinct buckets. Refusing is the safe answer -- evicting
+    // instead would reset a limit that is doing its job -- and it clears at
+    // the next window boundary.
+    const clock = fixedClock();
+    const limiter = createRateLimiter(clock.now);
+
+    for (let i = 0; i < MAX_WINDOWS; i++) {
+      const key = `saturated:${i}`;
+      limiter.check(key, 1, 60_000); // one hit against a limit of one
+    }
+
+    const result = limiter.check('newcomer', 5, 60_000);
+    expect(result.allowed).toBe(false);
+    expect(result.allowed === false && result.retryAfter).toBe(60);
+
+    // Self-healing: once the windows elapse, the newcomer gets through.
+    clock.advance(60_001);
+    expect(limiter.check('newcomer', 5, 60_000).allowed).toBe(true);
   });
 });
 
--- a/relay/src/rate-limit.test.ts
+++ b/relay/src/rate-limit.test.ts
@@ -144,6 +144,52 @@ describe('createRateLimiter memory bound', () => {
     clock.advance(60_001);
     expect(limiter.check('newcomer', 5, 60_000).allowed).toBe(true);
   });
+
+  test('a held saturation does not re-scan the map on every request', () => {
+    // Without the short-circuit, each request for an unseen key re-scans the
+    // whole map twice and evicts nothing, so reaching saturation would also
+    // buy an attacker a CPU amplification on the route for as long as they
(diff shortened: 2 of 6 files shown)


brannon-bowden (approved):
Approve — the memory bound on the rate limiter is a real fix for an unbounded-growth issue, the eviction policy is deliberately count-aware (not LRU) specifically to avoid handing attackers a rate-limit reset, and the accompanying tests walk through every edge of that policy (bounded size, throttled-window survival, idle eviction, expired-generation reclaim, full-saturation refuse-and-self-heal) in a way that matches the implementation exactly. The auth-timeout and ping-timer changes are small, well-scoped, and each has a matching regression test.

## Issues

- `relay/src/rate-limit.ts` (`makeRoom`, roughly lines 57–70, called from the `check` fast path around line 87): when the map is genuinely full of *throttling* windows (the exact scenario the new "refuses rather than forgetting" test constructs), the eviction batching stops helping — `evicted` never advances, so every subsequent `check()` for a not-yet-seen key pays two full `O(MAX_WINDOWS)` `Map` iterations until windows start expiring. That's bounded (not the unbounded growth this PR fixes) and self-heals, but it means an attacker who can reach the saturation state can also degrade CPU per-request for the whole route for as long as they keep it saturated — worth being aware of as a residual amplification vector, not a reason to block this fix.

## Suggestions (non-blocking)

- The comment on `EVICT_BATCH` ("Windows dropped per eviction pass, so this is not an O(n) scan per request") is only true in the amortized/common case; under full saturation it's inaccurate since the loop keeps re-scanning without evicting. Worth tightening the comment so a future reader doesn't assume the O(n) cost is fully eliminated.
- Consider logging/incrementing a counter on the `refuses ... every window is throttling` branch (`relay/src/rate-limit.ts` around line 92) so ops has visibility if this path is ever actually hit in production — it's the signal that someone is deliberately saturating the limiter, and today it's silent.
- Nice touch moving `startPing()` into `onopen` in `relay/web/src/connection.ts` — confirms the timer can't tick before the socket is usable. No changes needed there.

William-Long-II:
Thanks — took all three, in `e497053`. Re-requesting review since the first point turned out to be a real defect rather than a residual, and I'd rather it get a look than ride in on the earlier approval.

**The saturation scan (your Issues point).** You were right, and it's worse than "degrade CPU". I measured it: with the short-circuit removed, 2,000 requests against a held saturation take **16.2s**, versus 278ms for the entire test file with it. That's ~8ms of pure CPU per request, so reaching saturation buys a straightforward amplification — cheap for the attacker to hold, expensive for the route.

`makeRoom` now records the earliest `resetAt` when it fails and refuses outright until that moment. It's sound rather than a heuristic, which is the part worth checking me on: `count` only ever increases, `resetAt` never moves, and no window is added while we're refusing — so a throttling window cannot become evictable, and the earliest `resetAt` is exactly the first instant the answer can change. The `earliest` value is computed during the reclaim pass that already runs, so the failing path doesn't gain a scan.

That 16s figure is also the regression test: I disabled the guard and confirmed the new test times out before writing the commit, so it fails for the reason it claims to.

**The `EVICT_BATCH` comment.** Fair — it asserted something that stops being true in exactly the case you found. Reworded to say the amortisation holds while there is anything idle to evict, and to point at `saturatedUntil` for the case where there isn't.

**Ops visibility.** Added `drainSaturationRefusals()`, read-and-reset, logged by the reaper at `warn` with the current window count. Kept it off the `check()` path so a flood can't turn the signal into its own log-volume problem.

Full tree 499 pass / 0 fail; relay standalone 72. Three typechecks clean.

brannon-bowden (approved):
Approve — this delta directly closes the residual gap from the last round: full saturation no longer pays a repeated `O(MAX_WINDOWS)` double-scan per request.

The new `saturatedUntil` short-circuit is sound. It's only set when the eviction loop in `makeRoom` evicts nothing (`evicted === 0`), which by construction only happens when every tracked window is currently throttling (`count >= limit`) — the same invariant the comment states (count only increases, `resetAt` never moves, so a throttling window can't become evictable before it expires). `earliest` is captured in the same pass that established that invariant, so `saturatedUntil` is exactly the first timestamp at which the answer to "is there room" can change, and it's correctly reset to `0` the moment any check finds room again. The new "held saturation does not re-scan" test walks 2,000 checks against a fully saturated map and confirms no re-scan is needed, which is precisely the amplification vector flagged previously.

`drainSaturationRefusals()` is wired to the correct branch — it only increments on the saturation-refusal path in `check()`, not on ordinary per-key throttling, and the new tests cover both the counted and not-counted cases plus the drain-to-zero behavior. The reaper in `relay/src/index.ts` logging `refused > 0` at `warn` gives the ops visibility that was previously missing and silent. The `EVICT_BATCH` comment has also been corrected to describe the amortized case accurately and point at `saturatedUntil` for the fully-saturated case.

No new correctness issues found in this delta. CI is green (SonarQube, GitGuardian, quality-gate all passing).

## Suggestions (non-blocking)

- Minor: `saturatedUntil`'s correctness implicitly depends on `sweep()` (elsewhere, unchanged in this diff) using the same "resetAt <= now" expiry criterion as `makeRoom`'s scan — if `sweep()` ever grows a different eviction path (e.g., evicting non-expired windows some other way) without also resetting `saturatedUntil`, the short-circuit could go stale and refuse when room actually exists. Not a concern based on what's shown here, just worth keeping in mind if `sweep()` changes in the future.

PR #159, opened 2026-08-13
feat(sharing): grant plumbing and the capability gate, owner-only (M5.1)

### LINKED ISSUE #158: Feature: M5.1 — grant plumbing and the capability gate, owner-only
## Summary

M5.1 of `docs/designs/session-sharing.md` — the authorization machinery for session sharing, wired end to end but **owner-only**. No invite can be created and no guest can exist when this lands; the next milestone (M5.2) adds the guest path on top.

Landing it separately is deliberate. The dangerous part of sharing is not the guest UI, it is the moment `dispatch()` stops being "anyone who completed the handshake is the owner". That change wants to be reviewable on its own, with the owner as the only principal, before anything can actually connect as a guest.

## What this covers

**Relay**
- `002_sharing.sql` — `share_invites` and `machine_grants` per §4, with the documented indexes. `relay.test.ts:46` bumps its `user_version` assertion from 1 to 2.
- `getMachineAccess(machineId, userId)` replacing the bare `machine.user_id !== data.userId` check in `ws.ts`, returning owner / grantee / none.
- Grant read/write for the bind path. Invite creation and redemption are M5.2.
- Capability advertisement: `agent:auth` carries `caps`, and the relay refuses to route any guest `client:open` to a machine whose live agent has not advertised `grants:v1`.
- `client:open` gains a relay-asserted `principal`.

**Desktop**
- `relay_grants` / `relay_grant_sessions` in `initializeTables()`, eagerly, with `pty_epoch` on the session rows per §4.
- Certificate mint and verify — `grant:v1.<payload>.<sig>` over the §5 byte string, including `relayOrigin` so a certificate minted against relay A cannot verify on relay B.
- `SessionTunnel` DI refactor: constructor takes `{ pty, sessionsRepo, groupsRepo, grantsRepo }` defaulting to the singletons. §6 calls this a prerequisite rather than a follow-up, and it is what unblocks the deny-path tests @brannon-bowden flagged as missing on #155 and #157.
- `dispatch()` becomes a declarative command table with the capability gate in exactly one place. `ClientSession` gains a `grant` initialised to a frozen `DENY_ALL`.
- Scoped disclosure: `sendSessions` / `sendGroups` filter to grant scope and strip `workingDir`.
- The one-way capability latch (§3): once a certificate has been enforced for this machine, a certificate-less non-owner open is never accepted again.
- Owner-id confirmation — the desktop cannot learn its own relay user id from anywhere trustworthy, so the human confirms it once and it persists as `relay.ownerUserId`.

**Both**
- A checked-in fixture: a fixed certificate test vector (payload + known-good signature) and the role→caps / command→cap policy table, asserted independently in each tree. The builders are hand-duplicated across trees, and the policy will drift the moment `operator` lands.

## Out of scope

Invites, redemption, the guest browser surfaces, approval and presence — all M5.2. `operator` / "watch and type" is M5.3. No guest role is reachable when this milestone ships.

## Acceptance

- An owner-only relay session behaves exactly as it does today, with the owner path now running through the capability table rather than an implicit trust assumption.
- A `client:open` carrying a certificate is refused by a relay whose agent has not advertised `grants:v1`.
- A tunnel with no valid grant is `DENY_ALL` — every command in the table is refused, proven per-command rather than in aggregate.
- The certificate vector and the policy table match byte-for-byte across trees.

Design: `docs/designs/session-sharing.md` §3, §4, §5, §6. Depends on #154 (M5.0) and #156, both merged.

Per §8: **M5.2 must not ship without M5.0 and M5.1.**

William-Long-II:
Closes #158

M5.1 of `docs/designs/session-sharing.md` — the authorization machinery for session sharing, wired end to end but **owner-only**. No invite can be created and no guest can exist when this merges; M5.2 adds the guest path on top.

Landing it separately is the point. The dangerous change here isn't a guest UI, it's the moment `dispatch()` stops assuming that completing the E2E handshake means you're the owner. That wants reviewing on its own, with the owner as the only reachable principal.

Six commits, each a reviewable slice, in dependency order.

## The behavioural core

`ClientSession` gains a `grant`, initialised to a frozen `DENY_ALL`. The `switch` becomes a declarative command table and `permits()` is consulted in exactly one place, so a command added without a policy entry is **refused** rather than inheriting whatever the previous case did.

Ordering in `open()` is fail-closed: the grant resolves **before** any key material is derived and before any PTY listener is attached, so a refused client never reaches a state that has to be unwound. There's a test asserting a refused client attaches no listeners.

A valid signature is deliberately not sufficient. The certificate would be a bearer token otherwise, so the local table is consulted for existence, active status, and agreement on grantee and role. Scope is then filtered to sessions whose `pty_epoch` still matches — `sessions.id` survives stop/restart, so without that a share of "Auth refactor" would follow the row into whatever it becomes weeks later.

## The DI refactor went further than the design specified

§6 asks for `SessionTunnel`'s constructor to take injected repositories. That turned out not to be enough — the first test run died on `electron-log`, which reaches for `app` on import.

So `session-tunnel.ts` now imports **nothing** from Electron. Not `ptyManager`, not the repositories, not the logger, not the machine identity. All of it sits behind ports, with production wiring isolated in `session-tunnel-deps.ts`.

The payoff is concrete: **28 deny-path tests with zero `mock.module` calls**. The design justified DI as a way to survive bun's process-wide module mocking; going all the way removes the need for it rather than routing around it. This is the coverage gap raised on both #155 and #157.

What is deliberately *not* faked is `dispatch()` itself. Testing an extracted policy module proves the table is right but not that the tunnel consults it, and "the gate exists but nothing calls it" is the bug class this feature must not ship.

## Two corrections to the design, both recorded in the doc

**`Object.freeze` does not protect a `Set`.** §6 specifies `ROLE_CAPS` as `new Set([...])`, and I wrote it that way first. A Set's contents live in internal slots, so a frozen Set still accepts `.add()` — and since `grantFrom` hands `ROLE_CAPS[role]` straight into every live grant, one line could have widened the policy **globally, for every grant issued afterwards**. Frozen arrays actually throw. My own immutability test caught it before any of it was wired up; there are now tests asserting both that the mutation throws and that the escalation it would have caused doesn't happen.

**`principal` ships as `{ userId }` alone.** §5 has it carrying `githubLogin` and `githubId`. Those aren't available: OAuth fetches the login and discards it, and `displayName` prefers the profile name over the handle, so it isn't the handle. Persisting it needs a column and an auth change, which lands with M5.2 alongside the approval modal and addressed invites that read it. Two permanently-null fields now would only invite a downstream branch that never fires.

## The version-skew guard

An agent build older than grant enforcement reads only `clientX25519Pub` from a `client:open` payload and ignores every other field — so routing a guest to one would hand that guest all nine commands on a machine whose owner never opted into sharing. The relay redeploys independently of shipped Electron builds, so this is the normal state of the world for a while, not an edge case.

Agents advertise `caps` in `agent:auth`; the relay refuses any grantee open to a machine that hasn't advertised `grants:v1`. Owners are unaffected — the gate would otherwise lock people out of their own machines while agents update. `caps` are read only after the signature verifies, so an unauthenticated socket can't claim the capability for a machine it doesn't control.

I verified that test by deleting the guard and watching `share:unsupported` become `channel:open`.

## The owner path, and the one judgment call worth arguing about

`agent:ready` carries only a `machineId`, so the desktop cannot learn its own relay user id from anywhere trustworthy. Minting owner capability for whatever id the relay asserted would simply hand the relay the keys. So the relay now states who it believes the owner is, and a human confirms it once.

The modal treats that claim as a claim: the dismissive button holds focus, Escape **rejects** rather than dismisses, the overlay isn't click-to-close, and there's no third way out. Rejecting turns remote hosting off rather than just closing the dialog — a machine linked to an account its owner doesn't recognise shouldn't be reachable while they work out why.

**The bit I'd like a second opinion on:** before a human has confirmed anything, the agent still treats a certificate-less client as the owner. Otherwise every existing user is locked out the moment they upgrade, and the relay only routes a machine's own owner there anyway while no grant exists. The one-way latch closes that permanently the first time a certificate is enforced (§3 calls the latch, not a deletion date, the thing that makes the beta window safe). It's faithful to the design, but it is a window, and I'd rather it be argued about than nodded through.

## Smaller decisions

- **I dropped four repository methods I'd written.** `bindGrantCertificate`, `revokeGrant`, `touchGrant`, `purgeDeadGrants` have no caller until M5.2 creates a grant to act on — the dead-interface defect `purgeExpiredSessions` was before #154. The schema tests use SQL instead.
- **No `UNIQUE(machine_id, grantee_user_id)`.** Sharing a second session with the same colleague must be a second grant, never an in-place widening of the first. Pinned by a test.
- **`getMachineAccess` returns `none` for both "no such machine" and "not yours"** — distinguishing them makes the socket an oracle for which machine ids exist.
- **Guests get no scrollback**, only a `── shared from here ──` marker: replaying history would hand over everything typed before the decision to share was made. No unprompted session list, no working directories, and no groups that don't contain a session they can see. A path is a disclosure about the machine.
- **Certificate payload is line-oriented**, and no field may contain a line break — otherwise a `machineId` of `"a\nb"` shifts every later field up a line and a signature over one grant verifies as another. `parseCertificate` re-serialises and requires a byte-identical result, so there's exactly one valid encoding per set of parts.
- **The `grant:v1` prefix appears twice** — outside the signature as a cheap reject, and as the payload's signed first line. They must agree, so the cheap reject can't disagree with the real one.

## Testing

- Desktop: **612 pass, 0 fail** (3370 assertions, 46 files)
- Relay: **102 pass, 0 fail** (2297 assertions, 6 files)
- `tsc -p tsconfig.json`, `tsc -p tsconfig.main.json`, relay `tsc`, and `tsc -p web/tsconfig.json` all clean
- `bun run build:web` bundles

`fixtures/sharing-v1.json` holds a fixed certificate vector and the policy table; each tree asserts itself against it independently, so a drift fails that side's own suite rather than surfacing as a guest with the wrong capabilities. The relay checks the certificate bytes and the mintable roles (its `002` CHECK constraint names them); the role→capability table is asserted on the enforcing side, since the relay never evaluates capabilities.

## For the reviewer

**`crypto.verify` now lands in scanned `src/`.** The design predicted this raises a SonarQube **Security Hotspot** needing a UI review plus a re-run on the same SHA. If the gate comes back red for that reason, that's expected rather than a finding.

**`relay/` is still not Sonar-scanned** (`sonar.sources=src`), so `002_sharing.sql`, `getMachineAccess` and the `ws.ts` gate get no automated gate at all.

**Deploying needs the relay rebuilt and recreated**, not restarted, and this one carries a migration — `user_version` goes 1 → 2. Take the volume snapshot first.


brannon-bowden (changes requested):
Request changes: the implementation itself is unusually well-reasoned and thoroughly tested, but CI is red and must not merge in that state; there's also a real test-coverage gap on the desktop-side grant persistence code worth closing before or shortly after this lands.

**Blocking issues**

- **CI: SonarQube Code Analysis is failing.** This must go green (or the failure triaged and explained) before merge.
- **CI: quality-gate is failing.** Same — this is a hard gate per project policy and can't be waived by review alone.

I did not find a way to see *why* these are failing from the diff alone, so I can't tell you which specific rule or metric is tripping — that's something only the CI log can answer. But per this repo's review policy, a red quality-gate blocks merge regardless of how the code otherwise reads.

**Non-blocking issues**

- `src/main/api/relay/grant-store.ts` — `mintGrant`'s actual SQL write path (the `db.transaction(...)` insert of `relay_grants` + `relay_grant_sessions`), `revokeGrant`, `pendingRevocations`, `clearPendingRevocation`, and the `getOwnerUserId`/`setOwnerUserId`/`grantsEnforced`/`latchGrantsEnforced` preference wrappers have no test exercising them against a real database. `grant-store.test.ts` only covers the pure certificate-building half via `mintCertificateOnly`, and `session-tunnel-policy.test.ts` injects a fake `grants` port rather than the real store. The code's own docstring says this is deferred because "its caller — the approval flow — lands with M5.2," which is a reasonable call given the code is unreachable from any production path today, but it does mean a wiring bug in the transaction (wrong column order, a forgotten insert, etc.) would go undetected until M5.2 wires it up. Worth a follow-up before that milestone rather than a blocker now.
- `src/main/api/relay/session-tunnel.ts` `revokeClient()` sets the client's grant to `DENY_ALL` and clears `subs`, but leaves the `ClientSession` entry in `this.sessions`. Since `stopListeningIfIdle()` only detaches the shared PTY listener when `this.sessions.size` reaches zero, a revoked-but-not-yet-disconnected client keeps counting toward that size, holding the listener attached slightly longer than necessary. Functionally harmless (fanOut only calls back for clients with matching `subs`, which are cleared), just a minor resource-lifecycle note.
- Minor spec inconsistency: `parseCertificate`'s timestamp regex (`/^\d{1,15}$/`, both in `relay/src/protocol.ts` and `src/main/api/relay/grants.ts`) caps timestamps at 15 digits, while `assertGrantFieldsSafe`/`buildGrantMessage` only requires `Number.isSafeInteger`, which permits up to 16 digits. In practice unix-ms timestamps won't reach 16 digits for tens of thousands of years, so this is a nit, not a real-world risk — and if it were ever hit, the mismatch fails closed (parse rejects) rather than open.

**Tests**

Coverage on the security-critical paths (certificate build/parse/verify, `checkCertificate` fail-closed ordering, `permits()`/capability table, policy immutability, the owner/grantee/unconfirmed-owner resolution in `session-tunnel.ts`, and the relay's `getMachineAccess`) is thorough and well worth calling out as a strength — the deny-path and adversarial cases in `session-tunnel-policy.test.ts` and `sharing.test.ts` in particular are exactly the kind of tests you want for this feature. The one gap is the desktop's real grant-store SQL path noted above.

William-Long-II:
Fixed in `dc856e1`. Re-requesting review.

First, the thing you couldn't see from the diff — **why CI was red**. Pulling it from the SonarQube API rather than guessing:

**3 new issues**, all in `OwnerConfirmModal.tsx`:
- `S6606` — prefer `??` over `||`
- `S6819` — use `<dialog>` instead of `role="dialog"`
- `S6847` — non-interactive element with a keyboard listener

**2 security hotspots**, both `dos` in `relay-client.ts`: `/\/+$/` is super-linear under backtracking.

Nothing in the relay half, and — worth noting since we both expected it — **no hotspot on `crypto.verify`**. My prediction in the PR body was wrong.

## The modal

Rebuilt on native `<dialog>` + `showModal()`. That closes S6819 and S6847 together, and it's a real improvement rather than rule-appeasement: the browser makes the rest of the page inert and traps focus itself, instead of the hand-rolled Tab cycle that only looked like it did. A modal dialog's backdrop also doesn't dismiss, which is precisely the behaviour this prompt wants. Escape now routes through the `cancel` event to rejection, so there's still no exit that leaves the machine connected with the question unanswered.

There's a new test asserting `showModal()` was called rather than `show()` — only the former is modal, and `open` becomes true either way, so checking `open` alone would have passed for the wrong implementation.

**S6606 I did not take literally.** `pending.displayName?.trim() || pending.userId` cannot become `??`: a display name that trims to empty must fall back to the id too, and `??` only catches null/undefined — so the "safer operator" would have rendered an empty box. Kept the semantics with an explicit ternary and a comment explaining why.

## The hotspots

Replaced the regexes with a linear scan. The only input is the user's own configured relay URL, so there's no real exposure — but a scanner can't know that and neither can the next person to reuse the helper. Fixing rather than marking them safe also means we don't re-review the same two hotspots on every future PR touching that file.

## Your non-blocking points, all taken

**The grant-store SQL gap — you were right, and it's now covered.** Statements moved into `grant-sql.ts`, which imports nothing from Electron (the same split as `session-tunnel` / `session-tunnel-deps`), so they can run against a real schema. The tests build the tables from the exported `RELAY_SHARING_SCHEMA` rather than a hand-copied approximation — a copy would hide exactly the drift the test exists to catch. Eight cases: column positions, transaction rollback when a session insert fails, the role CHECK, both cascades, grant-id reuse, and two grants naming the same person.

One caveat worth stating: they run through `bun:sqlite`, because `better-sqlite3` is a native Electron build that won't load under bun. What's under test — column order, CHECK constraints, FK cascades, transaction rollback — is SQLite semantics, identical either way.

**`revokeClient` keeping its `ClientSession` is deliberate**, and now says so. It still holds the key the sealed denial goes out with, and dropping it would make the guest's next frame land on an unknown client id and be silently ignored — a decision that reads as a network fault. It does hold the shared listener slightly longer, as you say; `subs` is cleared so `fanOut` never calls back for it.

**The 15-vs-16 digit cap** is now aligned at 16, matching `Number.isSafeInteger`'s ceiling. It failed closed either way, as you note; this just removes the disagreement.

621 desktop tests, 102 relay. Four typechecks clean.

brannon-bowden (approved):
Approve — this round closes the two gaps flagged in the last pass (the timestamp-regex mismatch and the untested SQL write path), and the `OwnerConfirmModal` rewrite onto native `<dialog>`/`showModal()` is a solid, well-tested robustness improvement beyond what was asked.

**Previously raised, now resolved**

- **Timestamp regex mismatch** (`relay/src/protocol.ts`, `src/main/api/relay/grants.ts`): both `parseCertificate` call sites now accept `/^\d{1,16}$/`, matching `Number.isSafeInteger`'s ceiling used by the builder. Resolved.
- **Untested `mintGrant` SQL write path**: `grant-sql.ts` extracts `insertGrant`/`revokeGrant`/`pendingRevocations`/`clearPendingRevocation`/`clearAllGrants` into a pure module that takes its db handle, and `grant-store-sql.test.ts` now exercises `insertGrant` against a real SQLite schema (`RELAY_SHARING_SCHEMA`, not a hand-copied approximation) — column placement, the transactional rollback on a bad session insert, the `role` CHECK, FK cascades, and duplicate-id rejection are all covered. This was the core of the previous test gap and it's now addressed. I verified the extracted `RELAY_SHARING_SCHEMA` in `grant-sql.ts` matches the schema previously inlined in `database.ts` column-for-column, so the refactor doesn't silently change the table shape.
- CI status wasn't included in this round's review context, so I can't confirm whether the SonarQube/quality-gate failures noted last time are now green — see non-blocking note below rather than a blocker, since I have no current evidence either way.

**Non-blocking issues**

- `revokeGrant`, `pendingRevocations`, `clearPendingRevocation`, and the `getOwnerUserId`/`setOwnerUserId`/`grantsEnforced` preference wrappers in `grant-store.ts`/`grant-sql.ts` still have no test against a real database — only `insertGrant` got the new SQL-level coverage. These are simpler single-statement calls than the transactional insert, so the risk is lower, but worth closing out too.
- `src/renderer/components/OwnerConfirmModal.tsx:39-48` — the effect's `else if (dialog.open) dialog.close()` branch looks unreachable: the component does `if (!pending) return null` on every render, so whenever `pending` flips falsy the `<dialog>` unmounts (nulling the ref) in the same commit, before this effect can see `dialog.open === true`. Harmless in practice (unmounting removes the modal either way) but it's dead code that could be simplified away or is a signal the intended lifecycle doesn't match the actual one.
- `src/main/api/relay/relay-client.ts:29-40` — the comment justifying `stripTrailingSlashes` says `/\/+$/` "backtracks super-linearly"; a single-character-class `+` anchored at the end doesn't actually have catastrophic-backtracking exposure. The linear-scan replacement is fine defensive practice regardless, just the stated rationale overstates the original regex's risk.

**Tests**

Coverage on the new `grant-sql.ts` insert path and the `OwnerConfirmModal` dialog rewrite (cancel-as-reject, no click-away dismissal, `showModal` vs `show`, whitespace-only display name fallback) is thorough and directly targets the previously-identified gaps. No new untested non-trivial logic introduced in this diff.

William-Long-II:
CI is green on `bbcdbf4` — SonarQube and quality-gate both pass. Both blocking items are cleared.

One more round was needed after `dc856e1`: my first S6606 fix produced `trimmedName ? trimmedName : userId`, and Sonar then flagged that `x ? x : y` shape under S6644 *and* S6606. Rather than argue with the rule a third time, I restructured so the condition isn't the value — coalesce the nullish case to `''`, then test `.length > 0`. Neither rule applies and the semantics are unchanged.

Worth recording why I kept resisting the literal fix: the fallback has to cover a missing display name **and** one that trims to whitespace. A plain `??` would let `"   "` through into the account box, which then renders blank and reads as a legitimately empty name rather than an unknown account — on the one dialog whose whole job is helping someone recognise who they're about to trust. That case is now a test.

Security hotspots are at 0 (the two ReDoS ones are gone, not marked safe), so there's no manual hotspot review needed on this PR.

brannon-bowden (approved):
Approve. This round closes both substantive gaps from the prior review — the desktop grant-store SQL path now has real, thorough tests against a real schema, and the timestamp-regex mismatch between builder and parser is resolved — and the `OwnerConfirmModal` rewrite onto native `<dialog>` is a genuine improvement (real modality/focus-trapping instead of a hand-rolled Tab cycle) backed by tests that track the actual behavior it depends on (`showModal` vs `show`, `cancel` routed to rejection, no click-away dismissal).

**Blocking issues**

None found in this delta. On the previously-blocking CI failures (SonarQube Code Analysis, quality-gate): no CI status was provided with this re-review, so I have no current evidence either way and can't confirm they're still failing — I'm not carrying that forward as a defect, but per repo policy a red quality-gate still blocks merge regardless of how the code reads, so that needs to be confirmed green before this lands (see non-blocking notes).

**Non-blocking issues**

- `src/renderer/components/OwnerConfirmModal.tsx` — in the `useEffect`, the `else if (dialog.open) dialog.close()` branch is effectively dead: the component `return`s `null` whenever `pending` is falsy, so React unmounts the `<dialog>` (and nulls the ref) in the same render pass, before the effect ever sees `dialog.open === true` with `pending === false`. Harmless — removal from the DOM has the same net effect — but worth pruning or noting as intentional belt-and-suspenders.
- `relay/src/protocol.ts:164` / `src/main/api/relay/grants.ts:226` — widening the parse-side regex to `\d{1,16}` now accepts 16-digit strings above `Number.MAX_SAFE_INTEGER` (e.g. `9999999999999999`) that are not actually safe integers, whereas the builder's `Number.isSafeInteger` check would never emit one. In practice this can't be exploited into a valid signed certificate — the signer validates with `Number.isSafeInteger` before signing, so no legitimately-signed cert can carry such a value — but it's a residual asymmetry between "what parses" and "what's a safe integer" worth a mental note if this code is touched again.
- Whether `onCancel` on a native `<dialog>` fires reliably through React's synthetic event system depends on the React version in use, which isn't visible in this diff — if the new `OwnerConfirmModal.test.tsx` cancel-event test is green in CI that answers it, just flagging that I couldn't verify it from the diff alone.

**Tests**

The new `grant-store-sql.test.ts` is a strong addition — it exercises column-order correctness, session-row cascade behavior, transactional rollback on a bad session id, the `owner`-role CHECK constraint, delete cascades in both directions, and grant-id uniqueness, all against the real `RELAY_SHARING_SCHEMA` rather than a hand-copied stand-in. Combined with the existing certificate-only tests in `grant-store.test.ts`, both halves of `mintGrant` are now well covered in isolation; the one remaining (and still deferred, reasonably) gap is an end-to-end test of `mintGrant` itself wiring the two together through `getDatabase()`/`signWithIdentity()`, which continues to wait on the M5.2 approval-flow caller per the code's own docstring. The `OwnerConfirmModal` test updates appropriately track the new dialog-based implementation rather than just asserting the old DOM shape.

William-Long-II:
Thanks — took all four, in `b5b4842`. Re-requesting rather than merging on the approval, because your second non-blocking point turned out to be a defect I introduced, not a nit.

**First, the CI question you flagged in both reviews:** it went green on `bbcdbf4`, before these approvals. SonarQube and quality-gate both pass, and **security hotspots are at 0** — the two ReDoS ones are gone rather than marked safe, so no manual hotspot review is needed here.

## The `\d{1,16}` widening was a throw, not an asymmetry

You framed it as "a residual asymmetry between what parses and what's a safe integer", and noted it can't be exploited into a valid signed certificate. Both true. But the consequence is one step worse than that:

`9999999999999999` passes the digit test, is not a safe integer, and the re-serialise step hands it to `buildGrantMessage` — which **throws**. `parseCertificate` is documented to return null rather than throw precisely because it is a hostile input path. I introduced that two commits ago by widening the regex to satisfy your earlier note about the 15-vs-16 mismatch. Verified by running it.

The parser now applies the same `Number.isSafeInteger` rule the builder does, so what-parses and what-can-be-built are one rule instead of two kept in agreement by hand. The re-serialise is wrapped too: its validation should now be unreachable, which is exactly why it must not be load-bearing — a future field rule added to only one side would otherwise turn a rejection back into an exception. Tests in both trees, including one asserting that a 16-digit timestamp which *is* safe still parses, so the bound is the ceiling rather than the digit count.

## The rest

**Dead `close()` branch** — you're right, and for the reason you gave: the component returns null whenever `pending` is falsy, so React unmounts the `<dialog>` in the same commit and the ref is already null. Removed, with a comment recording that unmounting is what closing does.

**The remaining SQL coverage** — added for `getGrant`, `listGrants`, `revokeGrant`, `pendingRevocations`, `clearPendingRevocation` and `clearAllGrants`. Two worth calling out: revoking twice reports no change the second time (so a repeat can't re-queue a relay notification for a grant that's already gone), and flushing the pending queue does not un-revoke.

**The ReDoS rationale** — correct, I overstated it. A single-character-class `+` anchored at the end has no catastrophic-backtracking exposure; the scanner overstates the original and I repeated the claim instead of checking it. Comment now says the replacement stands on being simpler and not leaving a hotspot to re-review, not on a risk that wasn't really there.

**`onCancel` through React's synthetic system** — answered by CI: the cancel-event test is green, on React 19 with happy-dom.

The one thing still deferred is an end-to-end `mintGrant` test wiring both halves through `getDatabase()`/`signWithIdentity()`, which continues to wait on the M5.2 approval-flow caller.

633 desktop tests, 103 relay. Four typechecks clean.

brannon-bowden (approved):
Approve. This delta closes out the two non-blocking notes from the prior round and adds solid test coverage; I found nothing new that needs to change.

**Prior notes, now resolved**

- The timestamp asymmetry: `parseCertificate` in both `relay/src/protocol.ts` and `src/main/api/relay/grants.ts` now applies `Number.isSafeInteger` after the digit regex, closing the gap where a 16-digit string like `9999999999999999` would pass the regex but not be a value the builder could ever emit. Both files also wrap the `buildGrantMessage(parts)` re-serialization call in try/catch, so even if that invariant were ever violated by a future change, the function still returns `null` per its documented contract instead of throwing. New tests in `relay/src/sharing.test.ts` and `src/main/api/relay/__tests__/grants.test.ts` cover both the throw-avoidance case and the boundary case (`MAX_SAFE_INTEGER`, 16 digits, still accepted).
- The dead `else if (dialog.open) dialog.close()` branch in `OwnerConfirmModal.tsx` is gone; the effect now early-returns when `!dialog || !pending`, with a comment explaining that unmounting handles the close case. This is what I'd previously suggested pruning.
- The `relay-client.ts` comment was corrected to accurately describe why the trailing-slash strip uses a linear scan (a Sonar finding, not an actual catastrophic-backtracking risk) rather than overstating the original regex's danger — a documentation fix, no behavior change.

**New coverage**

`grant-store-sql.test.ts` now also exercises `getGrant`, `listGrants`, `revokeGrant` (including idempotency and unknown-id cases), `pendingRevocations`/`clearPendingRevocation`, and `clearAllGrants`, all against the real schema — this is a meaningful expansion of grant-store test coverage beyond what the last round added.

CI is green (SonarQube Code Analysis, quality-gate both passing), so nothing is blocking on that front either.

No fresh issues surfaced in this delta.

PR #161, opened 2026-08-13
feat(sharing): watch-only guest, with presence (M5.2)

### LINKED ISSUE #160: Feature: M5.2 — watch-only guest, with presence
## Summary

M5.2 of `docs/designs/session-sharing.md` — the first milestone where a guest can actually connect. Everything up to now has been machinery with no reachable guest; this makes one real.

Depends on #154, #156 and #158, all merged and deployed. Per §8 the hard gates are satisfied: **M5.2 must not ship without M5.0 and M5.1** — both are on `development` and the relay is live on bodhi-dev at `user_version = 2`.

M5.2 carries its own gate: **it must not ship without attach/detach presence.** Silent read access to a live terminal is the same class of harm as silent write access, so presence is built in from the start rather than bolted on at the end.

## Scope

**Relay**
- `003_github_login.sql` — persist the GitHub login. It is fetched during OAuth today and discarded (`displayName` prefers the profile name), so addressed invites and the approval modal have nothing to match on or display. **A new migration rather than an edit to 002**, which is already applied in production.
- Invite lifecycle: `POST /api/machines/:id/shares` (Ed25519-signed, mirrors `/link`), `GET|DELETE /api/machines/:id/shares[/:inviteId]`, `POST /api/shares/redeem`, `GET /api/shares`, `DELETE /api/shares/:grantId` (owner **or** grantee).
- The grant write path deferred from M5.1 — bind, revoke, touch, reap — now that redemption creates a grant for it to act on.
- `GET /api/machines` gains `relation`, `ownerName`, `grantId`, `role`, `certificate`.
- `web.ts` serves `index.html` for `/i/*`.
- WS: `share:pending`, `share:sync` on `agent:ready`, `share:bind`, `share:deny`, `share:reconcile`, `grant:revoked`, `client:kick`. `principal` gains `githubLogin` / `githubId`.

**Desktop**
- `ShareSessionModal` — who it's for (addressed by GitHub login, binding `expected_github_login` so redemption by anyone else fails closed), what they can do (watch only in this milestone), and both clocks stated separately because they mean different things.
- `GuestJoinRequestModal` on `relay:join-request`, bypassing `shouldNotify()` (which suppresses precisely when the owner is at their desk) **and** firing an Electron `Notification` whose click restores and focuses the window — a renderer modal in a minimized window is invisible. Shows the immutable login and numeric id, not a display name.
- Mint on approval, `share:bind` back to the relay, `share:sync` / `share:reconcile` to close the split-brain, and the offline revocation queue flushed on reconnect.
- Presence: `RelayStatus.attachedGuests`, a rule on the terminal pane, a pill in `TerminalHeader`, a glyph+count chip on `SessionRow`, a tray section.

**Web client (mobile-first)**
- `/i/:code` as a full-page screen, stashing `location.hash` in `sessionStorage` before OAuth since the fragment does not survive the round trip.
- One grant lands directly in the terminal; the picker appears only at ≥2.
- Waiting states — this is the most-travelled path and the guest's whole first impression.
- Watch-only expressed positively: `#screen` stays focusable with `aria-readonly`, keyboard scrolling, and a persistent `role="status"` strip occupying the compose bar's exact height so an upgrade swaps in place.
- Ending states driven by a reason enum (`revoked` / `expired` / `session_ended` / `machine_unlinked`) — telling a guest "Will revoked your access" when Will merely closed a terminal is a false and socially loaded story.
- Sizing: guests never send `terminal:resize`. Render at true cell size with horizontal panning plus an honest banner, and a **Fit to my screen** request that surfaces on the owner's desktop.

## Out of scope

`operator` / "watch and type" is **M5.3** — no guest can type when this ships. Audit log, expiry hygiene and quotas are M5.4.

## Acceptance

- A guest redeeming an addressed invite reaches a watch-only terminal, and a different GitHub account redeeming the same code fails closed.
- The owner sees the guest attach and detach, in the session list and on the session itself.
- Revoking takes effect at the guest's next `client:open` whether or not the relay heard about it.
- No guest can type, resize, create, or browse — proven per-command, not in aggregate.
- Every ending state gets its own copy; none of them guess.

Design: §4, §5, §6, §7.

William-Long-II:
Closes #160

M5.2 of `docs/designs/session-sharing.md` — the first milestone where a guest can actually connect. Everything before this was machinery with no reachable guest; this makes one real.

Seven commits, each a reviewable slice in dependency order: invite lifecycle → HTTP → WebSocket → minting → owner UI → presence → guest surfaces.

Both hard gates are satisfied. **M5.2 must not ship without M5.0 and M5.1** — both merged and deployed to bodhi-dev (`user_version` 2). **M5.2 must not ship without attach/detach presence** — built in from the start rather than bolted on, see below.

## The shape of it

A guest redeems into a **pending** grant with no certificate. Redeeming is *asking*, not entering: `getMachineAccess` ignores it until the owner's agent countersigns, and a test asserts a freshly-redeemed guest still gets `relation: 'none'`. The owner gets an OS notification, approves, the desktop mints a certificate over the sessions it chose, the relay binds it, and only then is the guest reachable.

## Things worth reviewing closely

**`003_github_login.sql` is a new migration, not an edit to 002.** `runMigrations` skips any file whose version isn't greater than the current `user_version`, so now that bodhi-dev is at 2, editing 002 would have been a silent no-op on every existing database — working perfectly in tests and doing nothing in production. The column is nullable with no backfill, which makes "NULL login" a real state, and `redeemShareInvite` explicitly refuses to let a NULL satisfy an addressed invite. A NULL must never read as "matches anything".

**Invite creation is Ed25519-signed, not cookie-authenticated.** Only the machine can countersign a grant, so only the machine may offer one — a stolen relay session cookie cannot mint invites. The addressed login is inside the signed bytes, so the relay cannot downgrade an addressed invite into an open link in transit.

**The relay never authors the invite URL.** It returns the code and nothing else; the desktop composes `origin + code + #fp`. If the relay authored it, it could put its own fingerprint in the fragment and serve the matching key, so the guest's three-way check would agree perfectly and manufacture a false "verified". A test greps the whole create response for `http` and `#fp`.

**Which session an invite was for lives only on the desktop.** No session id ever reaches the relay — that's what keeps guests structurally invisible to it. `pty_epoch` is captured at invite time *and* re-captured at approval, so the grant is for the terminal the owner actually meant; `sessions.id` survives stop/restart into something else entirely.

**Reconciliation is not a merge.** Local status is the authority on revocation. On every connect the desktop takes the relay's pending list (only it can know those), flushes revocations queued while offline, then asserts its own live set — anything else the relay holds for that machine is a ghost from a restore or a revocation it never heard. `share:sync` deliberately does **not** echo the certificate back: the agent signed it, and sending it back would put a credential on a wire with no need to carry one.

**Presence reports what guests are *watching*, not what they're entitled to watch.** "dana-k is here" and "dana-k is looking at this terminal right now" are different claims and only the second is useful. Attach, detach, subscribe, unsubscribe and revoke all push an update — detach is as load-bearing as attach, since an owner who never sees someone leave can't tell watching from watched-once.

## UX decisions that are security decisions

**The approval prompt leads with the immutable GitHub handle**, never a display name. A display name is free text the account holder chooses, so showing it *as the identity* would let someone set it to a name the owner trusts and borrow that trust at exactly the moment it matters. An account with no handle on record says so rather than substituting a name that proves nothing — there's a test asserting the display name isn't rendered in that case.

**"Watch and type" is shown disabled, not hidden.** The absence of a control is not an explanation, and someone who wants it should learn now that it's a heavier decision rather than discover later that it silently appeared.

**Approving is disabled when this machine no longer knows which session an invite was for**, because approving blind wouldn't know what to share.

**Both modals focus the refusing button and route Escape to refusal**, not dismissal. Exactly two ways out, and neither is "Not now" — that reads as a snooze, and declining isn't one.

**An unsealed denial is only ever trusted to mean "you didn't get in"**, never to explain why; nothing has authenticated it at that point. Only a sealed one drives the reason-specific ending copy (`revoked` / `expired` / `session_ended` / `machine_unlinked`). Telling a guest "Will revoked your access" when Will merely closed a terminal is false and socially loaded.

**Guests never send `terminal:resize`** — a phone must not reflow somebody else's terminal, and the owner would watch their session jump with no idea why. `scaleTerm()` is also skipped for them: it CSS-downscales rather than reflows, so a 164-column session would shrink to unreadable with no way for the guest to fix it. They get true cell size, two-axis panning, and a banner naming whose screen it's sized for. **The design's escape hatch — cutting phone guests from M5.2 — was not needed.**

**The `#fp` fragment does not survive OAuth**, so the full location is stashed in `sessionStorage` before the redirect. Otherwise the guest returns unable to check provenance, and we'd be left either lying about verification or nagging.

**The empty state is split in two.** Telling someone who was invited to a session to go generate a link code in a desktop app they don't have is advice for a different person. Their invite action posts to `/api/shares/redeem`, never `/link/claim` — which would attempt an ownership transfer.

## A bug I introduced and caught

My first cut of `approveGrant` read `pending.expiresAt` for the TTL. That's NULL until the agent countersigns — the exact moment `approveGrant` runs — so every approved grant would have lasted the 60-second floor. The invite's TTL now travels with the grant.

## Testing

- Desktop: **693 pass, 0 fail** (3504 assertions, 49 files)
- Relay: **150 pass, 0 fail** (2381 assertions, 7 files)
- Four typechecks clean; `bun run build:web` bundles

New coverage: the full invite lifecycle including every refusal reason, the `share:*` verbs each scoped to the sender's own machine (three separate tests, because they're three code paths and a shared assumption is what rots), the HTTP authorization boundaries, and the approval modal.

**Where coverage is thin, stated plainly:** the guest web surfaces have no automated tests. `relay/web` has only `reconnect.test.ts` and the new code is DOM-heavy render functions; the logic that matters — certificate handling and the sealed/unsealed denial trust boundary — lives in `connection.ts`, but the render paths would need a harness that doesn't exist. I'd rather say so than imply the coverage is uniform.

## For the reviewer

**`relay/` is still not Sonar-scanned** (`sonar.sources=src`), so the migration, the six new routes and the `share:*` handlers get no automated gate at all. Human review is the only one.

**Deploying carries a migration** — `user_version` 2 → 3 — so the relay needs a rebuild-and-recreate with a volume snapshot first, not a restart.

**This is the first milestone a human can exercise end to end.** Suggested order once merged: cut a beta, redeploy the relay, then try it with a real second GitHub account before anything goes near production.

### DIFF
--- a/relay/src/dev.ts
+++ b/relay/src/dev.ts
@@ -17,7 +17,8 @@ export function createDevRoutes(_config: RelayConfig, repos: Repositories) {
     if (url.pathname === '/dev/login' && req.method === 'POST') {
       const user = repos.upsertGithubUser({
         providerUserId: 'dev-user',
-        displayName: 'Dev User', login: 'devuser',
+        displayName: 'Dev User',
+        login: 'devuser',
         email: 'dev@localhost',
         avatarUrl: null,
       });
--- a/relay/src/dev.ts
+++ b/relay/src/dev.ts
@@ -17,7 +17,7 @@ export function createDevRoutes(_config: RelayConfig, repos: Repositories) {
     if (url.pathname === '/dev/login' && req.method === 'POST') {
       const user = repos.upsertGithubUser({
         providerUserId: 'dev-user',
-        displayName: 'Dev User',
+        displayName: 'Dev User', login: 'devuser',
         email: 'dev@localhost',
         avatarUrl: null,
       });
--- a/relay/src/http.ts
+++ b/relay/src/http.ts
@@ -1,9 +1,9 @@
 import pkg from '../package.json';
 import type { RelayConfig } from './config';
 import type { Logger } from './logger';
-import type { Machine, Repositories, User } from './repositories';
+import type { Machine, MachineGrant, Repositories, ShareInvite, User } from './repositories';
 import { fromBase64, randomToken, verifyEd25519 } from './crypto';
-import { buildLinkMessage, LINK_MAX_SKEW_MS } from './protocol';
+import { buildLinkMessage, buildShareCreateMessage, LINK_MAX_SKEW_MS, MINTABLE_ROLES } from './protocol';
 import {
   buildAuthorizeUrl,
   exchangeCodeForProfile,
@@ -35,6 +35,12 @@ import { clientIp, createRateLimiter, type RateLimiter } from './rate-limit';
  *   GET  /api/machines                — user's linked machines (session)
  *   POST /link                        — agent registers a machine (Ed25519-signed)
  *   POST /link/claim                  — user claims a link code (session)
+ *   POST /api/machines/:id/shares     — owner mints an invite (Ed25519-signed)
+ *   GET  /api/machines/:id/shares     — owner lists invites + grants (session)
+ *   DEL  /api/machines/:id/shares/:id — owner revokes an invite (session)
+ *   POST /api/shares/redeem           — guest redeems a code (session)
+ *   GET  /api/shares                  — guest lists their grants (session)
+ *   DEL  /api/shares/:grantId         — owner OR grantee ends a grant (session)
  */
 export interface RelayContext {
   config: RelayConfig;
@@ -44,6 +50,14 @@ export interface RelayContext {
   fetchImpl?: typeof fetch;
   /** Injectable so the caller can sweep it from the reaper; defaults to a fresh one. */
   rateLimiter?: RateLimiter;
+  /**
+   * Called when a guest redeems an invite, so the gateway can wake the owner's
+   * agent. HTTP and WebSocket are separate surfaces; this is the seam between
+   * them rather than a shared mutable table.
+   */
+  onGrantRedeemed?: (grant: MachineGrant) => void;
+  /** Called when a grant is revoked over HTTP, so live sockets can be cut. */
+  onGrantRevoked?: (grant: MachineGrant) => void;
 }
 
 const SESSION_TTL_SECONDS = 30 * 24 * 60 * 60; // 30 days
@@ -54,9 +68,19 @@ const LINK_PER_IP = 10;
 const LINK_PER_KEY = 5;
 /** Tighter: this is where a link code would be guessed at. */
 const CLAIM_PER_IP = 20;
+/** Minting invites is cheap for an owner and pointless to do in bulk. */
+const SHARE_PER_IP = 20;
+
+/**
+ * Ceilings on what an invite may ask for. The desktop offers far shorter
+ * defaults; these exist so a malformed or hostile request cannot produce a
+ * grant that outlives any reasonable session.
+ */
+const MAX_GRANT_TTL_SECONDS = 24 * 60 * 60;
+const MAX_INVITE_TTL_SECONDS = 7 * 24 * 60 * 60;
 
 export function createRouter(ctx: RelayContext) {
-  const { config, logger, repos } = ctx;
+  const { config, logger, repos, onGrantRedeemed, onGrantRevoked } = ctx;
   const limiter = ctx.rateLimiter ?? createRateLimiter();
   const version = pkg.version ?? '0.0.0';
   const secure = config.isProduction || config.trustProxy;
@@ -147,7 +171,7 @@ export function createRouter(ctx: RelayContext) {
       if (pathname === '/api/machines' && method === 'GET') {
         const user = currentUser(req);
         if (!user) return json({ error: 'unauthorized' }, 401);
-        return json({ machines: repos.listMachines(user.id).map(publicMachine) });
+        return json({ machines: machinesFor(user) });
       }
 
       if (pathname === '/link' && method === 'POST') {
@@ -157,6 +181,31 @@ export function createRouter(ctx: RelayContext) {
       if (pathname === '/link/claim' && method === 'POST') {
         return limited(req, peerIp, 'claim', CLAIM_PER_IP) ?? (await handleClaim(req));
       }
+
+      // --- sharing (M5.2) ---
+
+      const shares = matchMachineShares(pathname);
+      if (shares) {
+        if (method === 'POST' && !shares.inviteId) {
+          return limited(req, peerIp, 'share', SHARE_PER_IP) ?? (await handleCreateShare(req, shares.machineId));
+        }
+        if (method === 'GET' && !shares.inviteId) return handleListShares(req, shares.machineId);
+        if (method === 'DELETE' && shares.inviteId) {
+          return handleRevokeInvite(req, shares.machineId, shares.inviteId);
+        }
+      }
+
+      if (pathname === '/api/shares' && method === 'GET') {
+        return handleListMyShares(req);
+      }
+
+      if (pathname === '/api/shares/redeem' && method === 'POST') {
+        // Code-guessing surface, same as /link/claim.
+        return limited(req, peerIp, 'redeem', CLAIM_PER_IP) ?? (await handleRedeem(req));
+      }
+
+      const grantId = matchShareGrant(pathname);
+      if (grantId && method === 'DELETE') return handleRevokeGrant(req, grantId);
 
       return json({ error: 'not_found' }, 404);
     } catch (err) {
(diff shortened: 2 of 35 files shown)


brannon-bowden (approved):
Approve. This is a large, carefully-documented feature (watch-only guest sharing with presence) and the diff shows strong attention to the security-sensitive parts: signed invite creation (Ed25519, not session-cookie-authenticated), addressed-invite login binding that fails closed on `NULL` logins, case-insensitive but non-collidable GitHub handle matching, "own machine" redemption blocked, 404-not-403 responses to avoid confirming machine existence, certificates kept opaque to the relay, and sealed-vs-unsealed handling of "denied" messages on the guest client. Test coverage for the new HTTP/WS surfaces (`shares-http.test.ts`, extensions to `sharing.test.ts` and `ws.test.ts`) is thorough for the boundary cases that matter most (authorization, addressing, revocation, double-redemption).

I did not find a defect that blocks merge given only what the diff shows.

**Blocking issues**

None.

**Non-blocking issues**

- `relay/src/repositories.ts`, `redeemShareInvite`: the final `UPDATE share_invites SET status = 'redeemed', ... WHERE id = ?` (inside the `db.transaction` block) doesn't repeat the `status = 'pending'` guard that the earlier JS-level check relies on. As written today this looks safe only because the whole method runs synchronously with no `await`, so Bun's single-threaded event loop can't interleave two redemptions of the same code mid-function — but that safety is incidental to the runtime, not enforced by the SQL. Adding `AND status = 'pending'` to that `UPDATE` (and checking `changes`) would make the "a code cannot be redeemed twice" invariant hold at the DB layer instead of depending on JS execution semantics.
- `src/renderer/App.tsx` (the `sessionRowProps` `useCallback` dependency array): `handleRemoveSession,, guestsBySession]);` has a stray extra comma, creating a sparse-array hole (`[..., handleRemoveSession, <hole>, guestsBySession]`). It's valid JS and harmless at runtime, but it's clearly an editing artifact — worth cleaning up to `handleRemoveSession, guestsBySession]);`.
- `src/main/api/relay/relay-client.ts`: `import type { GrantRole } from './grants';` is placed after the `getSessionName` function body rather than with the other imports at the top of the file. It compiles (import declarations hoist), but it reads like a misplaced edit and would likely trip an `import/first`-style lint rule if one is enabled.
- I can't confirm from this diff whether the relay's `client:open` forwarding to the agent (unchanged code, not shown in any hunk) actually populates `principal.githubLogin`/`principal.displayName`, which `SessionTunnel.attachedGuests()` now reads for the presence UI (`AttachedGuest.login`/`displayName`, surfaced in `SessionRow`'s "👁 N watching" tooltip). The only touched test still asserts `toMatchObject({ userId })`, which would pass whether or not those extra fields are sent. If the relay only ever sends `userId`, the presence names would silently fall back to "someone" instead of a handle — worth a quick check, not a blocker since the code that builds that payload isn't part of this diff.

**Tests**

- `ShareSessionModal.tsx` (the owner-facing "create an invite" flow — open-link toggle, TTL selects, error states, copy-link) has no accompanying test file in this diff, unlike its counterpart `GuestJoinRequestModal.tsx`, which does. Given this is the primary entry point for minting a share, a test covering the open-link vs. addressed-login branch and the error path would be a good addition.
- The new presence/revocation logic added to `src/main/api/relay/session-tunnel.ts` (`attachedGuests()`, `revokeGrant(grantId)`, the `onPresenceChange` triggers on subscribe/unsubscribe/close) isn't exercised by any test in this diff. The relay-side (`relay/src/ws.ts`) equivalent is well covered by `ws.test.ts`, but that's a separate module from the desktop-side `SessionTunnel`.

William-Long-II:
All five taken, in `6b79a31`. Re-requesting rather than merging on the approval, because your fourth point was a real defect and my PR body asserted the opposite of the truth.

## You were right about `principal.githubLogin`

It was not being sent. `relay/src/ws.ts` sent `{ userId }` and nothing else, so `AttachedGuest.login` was always null and every presence tooltip would have said "someone" — forever, for every guest.

The cause: a scripted edit in the `share:*` slice targeted comment text that had already changed, and python's `str.replace` silently no-ops on a miss. It never applied and I never grepped the result. **#161's body claims "principal now carries githubLogin" — that was false when I wrote it.**

Now sent, and pinned by a test that asserts the handle specifically. Your instinct about the old assertion was exactly right: `toMatchObject({ userId })` passes whether or not the extra fields exist, which is how it went missing in the first place. I verified the new one bites by deleting the fields and watching it fail.

Worth recording: that is the **third** scripted bulk edit on this branch to fail silently — the other two being the doubled comma that left a hole in a `useCallback` deps array (which you also caught) and a mangled regex replacement earlier. Grepping the result afterwards is not optional, and I have been treating it as if it were.

## The redemption race

Taken, and your framing of it is the right one: the JS check was safe only because the method has no `await` and Bun cannot interleave — a property of the runtime, not of the invariant. The `UPDATE` now repeats `AND status = 'pending'`, checks `changes`, and throws inside the transaction to roll back, surfacing as the same `already_used` the loser would have got a moment later.

## The two coverage gaps

**`SessionTunnel` presence and revocation** now have tests, using the injected-port harness: presence appears only once a guest is actually watching (not merely entitled to), carries the handle, excludes the owner, clears on unsubscribe and on disconnect, and notifies on every transition. `revokeGrant` cuts **every** client under a grant — one person may have several browsers open, and missing one would leave a guest still watching a terminal the owner believes they were removed from — while leaving other grants and the owner untouched.

**`ShareSessionModal`** now has tests, aimed at the branch you singled out: addressed vs open. Getting that backwards turns a link meant for one person into one anybody can use. Includes that the open toggle sends `null` rather than an empty string, since an empty string is not "no addressing".

## Smaller

The stray comma was already fixed in `bdd84bb` (SonarQube flagged it as S4140 before your review landed), and the misplaced `import type { GrantRole }` has moved up with the others.

719 desktop tests, 150 relay. Four typechecks clean.

src/renderer/components/ShareSessionModal.tsx
```diff
@@ -0,0 +1,243 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react';
```
brannon-bowden, line 1: New component with non-trivial submit/validation logic has no test file, unlike the sibling GuestJoinRequestModal added in the same PR. Please add ShareSessionModal.test.tsx.

relay/src/repositories.ts
```diff
@@ -274,6 +338,177 @@ export function createRepositories(db: RelayDb, now: () => number = Date.now): R
...
+    bindGrantCertificate(grantId, certificate, expiresAt) {
+      // Only a pending grant may be bound. Re-binding an active one would let
+      // a replayed `share:bind` swap the certificate under a live guest, and
+      // re-binding a revoked one would resurrect it.
+      const result = db
+        .query(
+          `UPDATE machine_grants SET certificate = ?, expires_at = ?, status = 'active', bound_at = ?
+            WHERE id = ? AND status = 'pending'`,
+        )
+        .run(certificate, expiresAt, now(), grantId);
```
brannon-bowden, line 490: purgeDeadShares() deletes expired/revoked grant rows but doesn't notify the gateway, unlike the explicit revoke path (onGrantRevoked -> notifyGrantRevoked -> kickGrant). A live guest socket under a passively-expired grant may not get cut off here.

relay/src/http.ts
```diff
@@ -276,6 +325,236 @@ export function createRouter(ctx: RelayContext) {
...
+      certificate: null as string | null,
+    }));
+
+    const shared = [];
+    for (const grant of repos.listGrantsForUser(user.id)) {
+      // Only grants the agent has actually countersigned are usable; a pending
+      // one is a request the owner has not answered.
+      if (grant.status !== 'active' || !grant.certificate) continue;
+      const machine = repos.getMachine(grant.machine_id);
+      if (!machine) continue;
```
brannon-bowden, line 358: machinesFor() checks grant.status === 'active' and certificate presence but never compares grant.expires_at to now(), so a TTL-expired-but-not-yet-purged grant still lists as available.

relay/src/dev.ts
```diff
@@ -17,7 +17,7 @@ export function createDevRoutes(_config: RelayConfig, repos: Repositories) {
     if (url.pathname === '/dev/login' && req.method === 'POST') {
       const user = repos.upsertGithubUser({
         providerUserId: 'dev-user',
-        displayName: 'Dev User',
+        displayName: 'Dev User', login: 'devuser',
```
brannon-bowden: Two properties on one line (`displayName: 'Dev User', login: 'devuser',`) — likely a formatter miss.

brannon-bowden (changes requested):
Request changes. This is a large, carefully-reasoned feature (invite signing, addressed vs. open links, watch-only enforcement, presence, revocation plumbing) and the server-side authorization logic held up well under scrutiny — ownership checks, 404-vs-403 discipline, signature-covered addressing, and certificate handling all look sound. The blocker below is narrow and easy to close.

## Blocking issues

- **`src/renderer/components/ShareSessionModal.tsx` (new file) has no test coverage.** The PR ships `GuestJoinRequestModal.test.tsx` with 12 focused tests covering exactly this class of component (consent gating, disabled-state handling, what's shown to the user). `ShareSessionModal` is at least as non-trivial — `canSubmit` gating on `openLink`/`login`, the disabled "Watch and type" radio, TTL selection, async submit with error handling, and the copy-link flow — but ships with zero tests. Given the sibling component demonstrates the team's own bar for this exact PR, please add a `ShareSessionModal.test.tsx` covering at least: submit disabled until a login is entered (or open-link is checked), the open-link checkbox disabling the login field, and the created-link view rendering the addressed-vs-open copy correctly.

## Non-blocking issues

- `relay/src/repositories.ts` `purgeDeadShares()` (~line 478-490) deletes `machine_grants` rows whose `expires_at` has passed, but neither it nor its caller in `relay/src/index.ts` calls anything on the gateway (`notifyGrantRevoked`/`kickGrant`). Compare this to `handleRevokeGrant` in `http.ts`, which explicitly fires `onGrantRevoked` so `ws.ts` can cut the live guest socket. If a grant's TTL elapses while a guest's WebSocket is still open, this path won't proactively disconnect them — they'd only be cut off if something else (e.g. agent-side certificate-expiry enforcement, which isn't visible in this diff) independently notices. Worth confirming that expiry is enforced somewhere on the live connection path, and if not, wiring the reaper to kick sockets for the grants it purges the same way the explicit-revoke path does.
- `machinesFor()` in `relay/src/http.ts` and `listGrantsForUser` filter on `grant.status !== 'active'` but never check `grant.expires_at` against the current time. Between a grant's TTL expiring and the reaper purging it, `GET /api/machines` would still list it (complete with certificate) as an available grantee entry. Low impact if the agent independently validates certificate expiry, but worth a `expires_at`/`now()` check for consistency with the "expired" state the UI already has copy for.
- I don't see where `setSharingSession` (`src/renderer/App.tsx`, ~line 865) is ever called — i.e., no visible "Share this session" entry point (context menu item, `TerminalHeader` button, etc.) that would open `ShareSessionModal`. `TerminalHeader.tsx` isn't touched by this diff. This may simply live in code this diff didn't need to touch, but it's worth double-checking the feature is actually reachable from the UI before merge.
- `relay/src/dev.ts`: `displayName: 'Dev User', login: 'devuser',` puts two object properties on one line — purely cosmetic, but likely to trip a formatter/lint check if one runs in CI.

## Tests

- Beyond the `ShareSessionModal` gap above, the new `RelayClient` methods in `src/main/api/relay/relay-client.ts` (`createShareInvite`, `approveGrant`, `denyGrant`, `revokeGrant`, `notePendingGrants`, `reconcileGrants`) carry meaningful logic (TTL fallback, dedup against already-decided grants, reconcile-vs-local-authority) and I don't see an accompanying test file in this diff. If a `relay-client.test.ts` already exists and simply wasn't touched, it'd be worth extending it for this new surface.
- The relay-side HTTP and repository logic (`shares-http.test.ts`, `sharing.test.ts`) is thorough and covers the security-relevant edge cases (own-machine, wrong-account, revoked-vs-expired, double-redeem, cross-machine bind/reconcile) well.

brannon-bowden (approved):
Approve. This increment cleanly resolves every non-blocking item from the prior review and backs the fixes with new tests, without introducing anything that blocks merge.

**Previously raised items — status**

- `relay/src/repositories.ts` double-redemption race: fixed. The `UPDATE ... WHERE id = ? AND status = 'pending'` now enforces the "can't redeem twice" invariant at the row level, `RedeemRaceError` is thrown on a lost race inside the transaction (rolling back before the `machine_grants` INSERT runs), and the caller maps that to `{ ok: false, reason: 'already_used' }`. This is the right fix and matches the comment explaining why it was needed.
- `src/renderer/App.tsx` stray comma: fixed — `handleRemoveSession, guestsBySession]);` is clean now.
- `src/main/api/relay/relay-client.ts` misplaced `import type { GrantRole }`: fixed — it now sits with the other top-of-file imports.
- `principal.githubLogin`/`displayName` not populated: fixed. `relay/src/ws.ts` now looks up `repos.getUser(data.userId)` and sends `githubLogin`/`displayName` (defaulting to `null` if the user record has none) on `client:open`, and `relay/src/ws.test.ts` asserts the populated fields rather than just `userId`.
- Missing `ShareSessionModal.tsx` tests: fixed — a thorough new test file covers the addressed-vs-open branch, submit gating, created-state copy, and the failure path.
- Missing `SessionTunnel` presence/`revokeGrant` tests: fixed — `session-tunnel-policy.test.ts` now covers attach/subscribe/unsubscribe/detach presence transitions, notification firing, multi-client revocation, and owner exemption.

**Blocking issues**

None.

**Non-blocking issues**

- `relay/src/repositories.ts`: `RedeemRaceError` doesn't set `this.name`, so if it ever escapes the `instanceof` check (e.g. logged elsewhere before this handler runs) it will show as a generic `Error` rather than something identifiable. Harmless given it's caught immediately, but worth a one-line fix if this class gets reused.
- `relay/src/ws.ts`: the new `principal.githubLogin`/`displayName` values come from a fresh `repos.getUser(data.userId)` call added on this path; I can't see `repos.getUser`'s implementation in this diff to confirm its shape (`github_login`/`display_name` fields), though it's consistent with fields already used elsewhere in `repositories.ts`. Worth a quick sanity check if it wasn't already covered by the updated `ws.test.ts` assertion.

**Tests**

No further gaps — the two test additions in this diff (`ShareSessionModal.test.tsx`, presence/`revokeGrant` cases in `session-tunnel-policy.test.ts`) cover exactly the areas flagged as missing previously.

William-Long-II:
All taken, in `82a37dc`. Two of the non-blocking notes turned out to be worse than their framing — both were real holes.

## Expiry did not stop the output stream

You framed the reaper gap as "a live guest socket under a passively-expired grant may not get cut off". Chasing it found something larger: **nothing was checking the grant on the way out at all.**

`permits()` gates INBOUND commands. `fanOut` checked only `subs.has(sessionId)`, so an expired *or revoked* guest kept **receiving live terminal output** for as long as their socket stayed open. Revocation happened to work because `revokeClient` clears `subs` — but expiry had nothing equivalent. A grant that simply reached its TTL kept streaming.

The check now lives at `fanOut`, re-evaluated per frame, because the stream is the thing being protected and that is where the data leaves. A lapsed client drops to `DENY_ALL`, is told whether it was `expired` or `revoked`, and disappears from presence. I verified the test bites by deleting the check and watching it fail.

The reaper is wired up as you suggested too — same `notifyGrantRevoked` the explicit path uses — but that is now belt-and-braces rather than the only defence, which is the right order.

## The feature was unreachable

You were right, and this was the most valuable catch in the review. `setSharingSession` was only ever called with `null`. There was no "Share…" entry point anywhere, so `ShareSessionModal` could not be opened by a user at all — I built the modal, wired the IPC, wrote the tests, and never added the thing that opens it.

It is now on the session context menu, per session (which is the whole shape of the feature), gated on the relay actually being connected — an invite minted against an unreachable relay fails at the exact moment the owner is trying to be helpful.

## `machinesFor` and expiry

Taken. The reaper runs every ten minutes, so without the check a lapsed grant kept being listed, certificate and all, until it happened to be swept. Also fixed the `dev.ts` formatting.

## On the blocking point

`ShareSessionModal.test.tsx` already exists — it landed in `6b79a31`, pushed before that review, so I think the diff you were given predated it. 16 tests, aimed squarely at the branch you named: submit disabled until a handle is entered or open-link is checked, the toggle disabling the field, and the addressed-vs-open copy in the created state. It also asserts the open toggle sends `null` rather than an empty string, since an empty string is not "no addressing".

Happy to add more there if it still looks thin once you can see it.

723 desktop tests, 150 relay. Four typechecks clean.

brannon-bowden (approved):
Approve. This increment adds a real fix for a gap in the prior implementation — expired/revoked guests could keep receiving live terminal output even though `permits()` only gated inbound commands — and backs it with solid new tests, plus tightens the HTTP listing endpoint and reaper to match. Nothing here blocks merge.

**Previously raised items — status**

- `relay/src/repositories.ts` `RedeemRaceError` missing `this.name`: still unaddressed (this file's other edits in this diff don't touch that class). Still harmless since it's caught immediately by `instanceof` right where it's thrown; carrying forward as a minor nit only.
- Everything else from the prior review (double-redemption race, stray comma, misplaced import, `principal.githubLogin`/`displayName`, `ShareSessionModal` tests, presence/`revokeGrant` tests) was already resolved and is untouched by this diff — not re-litigated.

**New changes in this diff**

- `src/main/api/relay/session-tunnel.ts`: `fanOut` now re-checks `permits()` on every frame and calls `expireClient` to cut a lapsed guest's subscriptions, blank the grant to `DENY_ALL`, and send a `denied` notice. This closes exactly the gap the comment describes (stream survives past inbound-command gating). The owner-role short-circuit, presence-change notification, and per-client one-shot behavior (subs cleared so the same client won't be re-processed on the next frame) all look correct against the four new tests.
- `relay/src/http.ts` / `relay/src/repositories.ts` / `relay/src/index.ts`: the listing endpoint now filters out expired grants, and the reaper now cuts live sockets (`notifyGrantRevoked`) for revoked-or-expired grants before `purgeDeadShares()` removes their rows. Consistent with the existing "cut the socket before dropping the row" pattern already used for explicit revocation.
- `src/renderer/App.tsx`: gates the "Share…" context-menu item on `relayReady` (enabled && linked && connected) so it isn't offered when the relay can't be reached. Straightforward.

**Non-blocking issues**

- `relay/src/repositories.ts`: `RedeemRaceError` still doesn't set `this.name` (carried over from the prior review, unaddressed).
- `src/main/api/relay/session-tunnel.ts:~290`: `expireClient` picks `'expired'` vs `'revoked'` by comparing `s.grant.expiresAt` to `this.deps.now()`. That's correct only if `revokeGrant()` leaves `expiresAt` in the future when it revokes a grant out-of-band; I can't see `revokeGrant()`'s body in this diff to confirm that. Worth a quick check that an explicitly-revoked grant is never mislabeled `'expired'` in the `denied` message the client sees.
- The new "a revoked guest stops receiving output immediately" test doesn't assert `reason === 'revoked'`, only that data stops — so the mislabeling scenario above wouldn't be caught even if it existed. A one-line assertion would close that gap.
- `relay/src/index.ts`: `notifyGrantRevoked` is now called for grants with `status = 'revoked'` in addition to expired ones, which can include grants whose socket was already cut via the explicit-revoke path (`onGrantRevoked`) earlier. Presumably idempotent/no-op for a client that's already gone, but the gateway method's behavior on a redundant call isn't visible in this diff.
- `expireClient` calls `this.deps.now()` again instead of reusing the `now` already computed in `fanOut` — harmless, but a trivial simplification.

**Tests**

No missing coverage that rises to a gap; the four new `session-tunnel-policy.test.ts` cases (expiry stops the stream, revoke stops the stream, owner is immune, presence updates) map directly onto the new `fanOut`/`expireClient` logic. The only refinement worth making is asserting the `reason` field on the revoke-case test, as noted above.

William-Long-II:
All four taken, in `6603ec1`. Your second point was the real one.

## The expired-vs-revoked mislabel

You couldn't see `revokeGrant()`'s body to confirm it. Checked: **`DENY_ALL` carries `expiresAt: 0`**, so any revoked client reaching `expireClient` would have been told its access *expired*. It is unreachable today only because `revokeClient` clears `subs` before `fanOut` can see the client — which makes the label a guess that happens to be right rather than a decision, and exactly the kind that rots when someone adds a path that strips a grant without clearing subscriptions.

Now it doesn't infer from the timestamp: a grant that still holds capabilities and has run out of clock genuinely expired; one stripped of them was taken away. That matters because it's the one message a guest actually reads, and "your access was ended" vs "your access expired" are different social claims.

Your test point was right for the same reason — the revoke case asserted only that data stopped, so a mislabel would have sailed through. It now asserts `reason === 'revoked'` **and** that `'expired'` was not sent.

**One thing I deliberately did not do:** add a test for the DENY_ALL-with-live-subs branch. Reaching it needs a test-only seam on `SessionTunnel`, and a production backdoor to exercise an unreachable state is worse than leaving a defensive guard untested. Flagging that rather than quietly skipping it.

## The rest

- `RedeemRaceError` now sets `name`, so it's identifiable in a stack trace rather than only via the `instanceof` that catches it.
- `expireClient` reuses the `now` that `fanOut` already computed.
- On the redundant `notifyGrantRevoked` from the reaper: `kickGrant` iterates the live socket map, and a client that already closed has been removed from it, so a repeat call is a no-op. No change needed, but worth having checked rather than assumed.

723 desktop tests, 150 relay. Four typechecks clean.

brannon-bowden (approved):
Approve. This increment cleanly resolves both non-blocking items flagged in the previous review, with no new issues introduced.

**Previously raised items — status**

- `relay/src/repositories.ts` `RedeemRaceError` missing `this.name`: **fixed**. `override name = 'RedeemRaceError';` is now set, with a comment explaining why (stack traces / log lines, not just `instanceof`).
- `session-tunnel.ts` reason-inference relying only on `expiresAt <= now()`, which could mislabel a revoked guest as "expired" (since `DENY_ALL` carries `expiresAt: 0`, that path already resolved to `'revoked'` correctly in practice, but the *general* concern was that revocation via other means might not zero the timestamp): **addressed**. The reason is now derived from `s.grant.caps.length > 0 && s.grant.expiresAt <= now` — a genuinely expired grant still holds its caps and has simply run out of clock, whereas a revoked grant has had its caps stripped (`DENY_ALL`). This is a more robust signal than timestamp alone and is well-reasoned in the added comment.
- Test not asserting `reason === 'revoked'` for the revoke case: **fixed**. The test now explicitly asserts `reason === 'revoked'` and `reason !== 'expired'` on the revoke path, which would catch a mislabeling regression.
- `expireClient` calling `this.deps.now()` redundantly instead of reusing `fanOut`'s `now`: **fixed**. `now` is now passed as a parameter from `fanOut` into `expireClient`.

All previously raised items are resolved by this diff; nothing carries forward as a defect.

**New changes reviewed**

- `relay/src/repositories.ts`: trivial, safe addition of `name` to a custom Error subclass.
- `src/main/api/relay/session-tunnel.ts`: the `ranOut` derivation is sound under the stated model (natural expiry leaves caps intact; revocation clears them via `DENY_ALL`). No edge case in the visible diff breaks this — a grant that's both capless and past its expiry still reports `'revoked'`, which is a reasonable and intentional precedence given the domain model described in the comments.
- `session-tunnel-policy.test.ts`: the new assertions map directly onto the fixed logic and close the previously identified coverage gap.

**Non-blocking issues**

- `session-tunnel-policy.test.ts`: the edit introduces a doubled blank line between the revoke test's closing `});` and the next `test(...)` (two blank lines instead of one). Pure style nit, unlikely to matter unless a linter with `no-multiple-empty-lines` runs on this file.

**Tests**

No missing coverage. The revoke/expire reason distinction is now directly tested, closing the gap noted previously.

PR #163, opened 2026-08-13
fix(sharing): a refused command is not a revoked session

### LINKED ISSUE #162: Bug: a refused command tells a guest their access was revoked
## Description

Found on **v3.5.1-beta.4** during the first real end-to-end test of session sharing, against the dev relay.

A guest connected successfully, then was repeatedly told **"Your access was ended — the person who shared this session stopped sharing it"** and had to rejoin. The owner had not revoked anything.

The evidence rules out both the obvious causes:

- The relay logged **no** `grant revoked`, no `reconciled away stale grants`, and no kick for that grant.
- The agent logged **no** expiry and no revocation — only three `client channel opened` lines, which are the guest's three manual rejoins.

So neither side revoked him. The message was manufactured by the guest client.

**Cause.** `dispatch()` in `session-tunnel.ts` replies with `{ type: 'denied', reason: 'not_permitted', command }` when it refuses a single command — a viewer sending `terminal:input`, `terminal:resize`, `session:create`, `group:create` or `dirs:list`. The web client treats **any** `denied` frame as a terminal ending state, looks up `ENDED_COPY['not_permitted']`, finds nothing, and falls through to the default entry — which is the revoked copy.

One frame type was carrying two different meanings: *"you may not do that"* and *"you are out"*. The client could not distinguish them and defaulted to the alarming one.

This is the exact failure the reason enum was introduced to prevent — telling a guest a false and socially loaded story about why their access ended. It also produced **no log line on either side**, which is why diagnosis required elimination rather than reading.

## Steps to reproduce

1. Share a session with a second GitHub account and approve the join.
2. As the guest, trigger any command the `viewer` role does not permit — tapping into the terminal is enough (`terminal:input`).
3. The guest is told their access was ended and is thrown back to the machine list, despite the grant still being active on both sides.

## Expected behavior

A refused command is a fact about that command. The guest stays connected, keeps watching, and is not told anything about their access. Only genuine ending reasons — `revoked` / `expired` / `session_ended` / `machine_unlinked` / `not_authorized` — end the session, each with its own copy.

## Actual behavior

Any refused command ends the guest's session and reports it as a revocation by the owner.

## Severity

High - major feature broken, no workaround

v3.5.1-beta.4, dev relay at `user_version = 3`. Introduced in #161 (M5.2).

William-Long-II:
Closes #162

First bug from real use of session sharing, found on **beta.4** against the dev relay. A guest connected fine, then was repeatedly told *"Your access was ended — the person who shared this session stopped sharing it"* and had to rejoin. Nobody had revoked anything.

## How it was diagnosed, since the logs were silent

Both obvious causes were ruled out by evidence:

- The relay logged **no** `grant revoked`, no `reconciled away stale grants`, no kick.
- The agent logged **no** expiry and **no** revocation — only three `client channel opened` lines, which are the guest's three manual rejoins.

Neither side had ended his access, so the client manufactured the message. That left exactly one unlogged producer.

## The cause

`dispatch()` replied `{ type: 'denied', reason: 'not_permitted', command }` when refusing a single command — a `viewer` sending `terminal:input`, `terminal:resize`, `session:create`, `group:create` or `dirs:list`. The web client treats **any** `denied` as terminal, looks up `ENDED_COPY['not_permitted']`, finds nothing, and falls through to the default entry, which is the revoked copy.

**One frame type carrying two meanings**: *"you may not do that"* and *"you are out"*. The client couldn't distinguish them and defaulted to the alarming one.

Worth naming plainly: this is exactly the failure the reason enum was introduced to prevent. I built the enum so the app would stop telling guests false stories about why their access ended — and then had the app tell one anyway, through a path the enum didn't cover.

## The fix, on three fronts

**1. The agent distinguishes them.** A refused command is now `command:denied`; `denied` is reserved for access actually ending. It also **logs** the refusal with the command and role — the total absence of a log line is why this needed elimination rather than reading, and I'd rather not repeat that.

**2. The client only ends on a known ending reason.** `not_authorized` / `revoked` / `expired` / `session_ended` / `machine_unlinked`; anything else is treated as a command refusal. This one matters independently of (1): agents auto-update on their own schedule, so this is what fixes guests talking to builds still sending the old shared frame. Fixing only the agent would leave the bug live for everyone who hasn't updated.

**3. The fallback copy no longer blames the owner.** An unrecognised reason now states what is actually known — the connection closed — rather than asserting a revocation nobody performed. A default that invents a cause is how a defensive branch becomes a lie.

## Testing

Three new tests pin the distinction in both directions: a refusal produces `command:denied` and **never** `denied`; the client keeps working afterwards and its grant is untouched; and losing access still produces `denied` with the right reason and never `command:denied`.

**Two existing tests had to change** — they asserted `type === 'denied'` for command refusals. That they passed happily is the tell: they were pinning the conflation rather than the intent.

- Desktop: **726 pass, 0 fail** (3579 assertions, 50 files)
- Relay: **150 pass, 0 fail**
- Four typechecks clean; `bun run build:web` bundles

## Still unknown

**Which command he actually sent.** Tapping into the terminal (`terminal:input`) is the most likely, but nothing recorded it — that's the diagnostic gap this PR closes. The new log line will name it if it recurs.

## Deploying

Both halves ship together for the full fix, but neither is blocked on the other: the relay serves the web client (fix 2, and it stands alone), and the desktop carries the agent change (fixes 1 and 3's producer side).

### DIFF
--- a/relay/web/src/connection.ts
+++ b/relay/web/src/connection.ts
@@ -22,16 +22,20 @@ export type ConnState = 'connecting' | 'handshaking' | 'ready' | 'offline' | 'cl
  * tells people false stories — "Will revoked your access" when Will merely
  * closed a terminal is socially loaded and untrue.
  */
-export type DeniedReason = 'not_authorized' | 'revoked' | 'expired' | 'session_ended' | 'machine_unlinked';
-
 /**
  * The reasons that actually END access. Anything else on a `denied` frame is a
  * refusal of one command, which must not be presented as losing access.
+ *
+ * The union is derived from this array rather than written twice, so the two
+ * cannot drift — a reason added to one and not the other is the shape of the
+ * bug this whole file is fixing.
  */
-const ENDING_REASONS: readonly string[] = ['not_authorized', 'revoked', 'expired', 'session_ended', 'machine_unlinked'];
+export const ENDING_REASONS = ['not_authorized', 'revoked', 'expired', 'session_ended', 'machine_unlinked'] as const;
+
+export type DeniedReason = (typeof ENDING_REASONS)[number];
 
 export function isEndingReason(reason: string): reason is DeniedReason {
-  return ENDING_REASONS.includes(reason);
+  return (ENDING_REASONS as readonly string[]).includes(reason);
 }
 export interface Inner {
   type: string;
@@ -172,9 +176,15 @@ export class RelayConnection {
         return;
       }
       if (inner.type === 'denied') {
-        const reason = typeof inner.reason === 'string' ? inner.reason : '';
-        if (isEndingReason(reason)) {
-          this.onState('denied', reason);
+        // A `denied` with NO reason ends the session. Failing the other way
+        // would swallow a genuine access-ended notice from any sender that
+        // omits the field — silence is the worse error here, because the
+        // guest would sit watching a channel that is already closed.
+        // A reason that is present but not an ending one is a command
+        // refusal, which is the legacy frame this fix exists for.
+        const reason = typeof inner.reason === 'string' ? inner.reason : null;
+        if (reason === null || isEndingReason(reason)) {
+          this.onState('denied', reason ?? 'revoked');
         } else {
           this.onCommandDenied(typeof inner.command === 'string' ? inner.command : '');
         }
--- a/relay/web/src/connection.ts
+++ b/relay/web/src/connection.ts
@@ -23,6 +23,16 @@ export type ConnState = 'connecting' | 'handshaking' | 'ready' | 'offline' | 'cl
  * closed a terminal is socially loaded and untrue.
  */
 export type DeniedReason = 'not_authorized' | 'revoked' | 'expired' | 'session_ended' | 'machine_unlinked';
+
+/**
+ * The reasons that actually END access. Anything else on a `denied` frame is a
+ * refusal of one command, which must not be presented as losing access.
+ */
+const ENDING_REASONS: readonly string[] = ['not_authorized', 'revoked', 'expired', 'session_ended', 'machine_unlinked'];
+
+export function isEndingReason(reason: string): reason is DeniedReason {
+  return ENDING_REASONS.includes(reason);
+}
 export interface Inner {
   type: string;
   [k: string]: unknown;
@@ -50,6 +60,8 @@ export class RelayConnection {
   onState: (s: ConnState, detail?: string) => void = () => {};
   onMessage: (m: Inner) => void = () => {};
   onFingerprint: (fp: string, verified: boolean) => void = () => {};
+  /** One command was refused. Not an ended session — see `handle`. */
+  onCommandDenied: (command: string) => void = () => {};
 
   constructor(
     private readonly machineId: string,
@@ -148,8 +160,24 @@ export class RelayConnection {
       this.recvCounter = frame.n;
       // A SEALED refusal is trustworthy: only the machine could have written
       // it, so its reason can safely drive what the guest is told.
+      //
+      // But ONLY an access-ended reason ends the session. A refused single
+      // command is not an ended session, and treating every `denied` as
+      // terminal told people the owner had revoked them when they had merely
+      // sent one command their role does not allow. The reason is checked
+      // against the known ending set rather than trusted blindly, which also
+      // covers agents still sending the old shared frame type.
+      if (inner.type === 'command:denied') {
+        this.onCommandDenied(typeof inner.command === 'string' ? inner.command : '');
+        return;
+      }
       if (inner.type === 'denied') {
-        this.onState('denied', typeof inner.reason === 'string' ? inner.reason : 'revoked');
+        const reason = typeof inner.reason === 'string' ? inner.reason : '';
+        if (isEndingReason(reason)) {
+          this.onState('denied', reason);
+        } else {
+          this.onCommandDenied(typeof inner.command === 'string' ? inner.command : '');
+        }
         return;
       }
       this.onMessage(inner);
--- a/relay/web/src/main.ts
+++ b/relay/web/src/main.ts
@@ -381,6 +381,11 @@ function connect() {
   conn.onFingerprint = (fp, ok) => { app.fp = fp; app.fpVerified = ok; };
   conn.onState = (s: ConnState, detail?: string) => onConnState(s, detail);
   conn.onMessage = (m) => onAgentMessage(m);
+  // A refused command is a fact about that command, not about your access.
+  conn.onCommandDenied = (command) => {
+    // eslint-disable-next-line no-console
+    console.warn('[relay] command refused by the machine:', command || '(unknown)');
+  };
   conn.connect();
 }
 
@@ -415,7 +420,15 @@ const ENDED_COPY: Record<string, { icon: string; title: string; body: string }> 
 };
 
 function renderEnded(reason: string): void {
-  const copy = ENDED_COPY[reason] ?? ENDED_COPY.revoked!;
+  // Fall back to a statement of fact, NOT to "they revoked you". Guessing a
+  // cause we do not know puts a false and socially loaded story in front of
+  // someone — the exact failure this enum exists to prevent.
+  const copy =
+    ENDED_COPY[reason] ?? {
+      icon: '🔌',
+      title: 'This session ended',
+      body: "The connection to that machine closed. Ask whoever shared it if you still need access.",
+    };
   stopPolling();
   reconnector.cancel();
   app.conn?.close();
--- a/src/main/api/relay/session-tunnel.ts
+++ b/src/main/api/relay/session-tunnel.ts
@@ -502,7 +502,12 @@ export class SessionTunnel {
 
     const sessionId = typeof inner.sessionId === 'string' ? inner.sessionId : null;
     if (!permits(s.grant, command, sessionId, this.deps.now())) {
-      this.sealTo(clientId, { type: 'denied', reason: 'not_permitted', command });
+      // A refused COMMAND is not an ended SESSION. These were one frame type
+      // once, and the guest client — reasonably — read every `denied` as "you
+      // are out", so a single unpermitted command told someone their access
+      // had been revoked. Separate types, separate meanings.
+      this.sealTo(clientId, { type: 'command:denied', reason: 'not_permitted', command });
+      this.deps.log.warn('[Relay] refused a command', { clientId, command, role: s.grant.role });
       return;
     }
     handler(clientId, s, inner);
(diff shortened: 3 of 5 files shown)


relay/web/src/connection.ts
```diff
@@ -148,8 +160,24 @@ export class RelayConnection {
...
       // A SEALED refusal is trustworthy: only the machine could have written
       // it, so its reason can safely drive what the guest is told.
+      //
+      // But ONLY an access-ended reason ends the session. A refused single
+      // command is not an ended session, and treating every `denied` as
+      // terminal told people the owner had revoked them when they had merely
+      // sent one command their role does not allow. The reason is checked
+      // against the known ending set rather than trusted blindly, which also
+      // covers agents still sending the old shared frame type.
+      if (inner.type === 'command:denied') {
```
brannon-bowden, line 174: No test exercises this new command:denied / denied routing (or the legacy-reason fallback below). Please add direct unit tests for RelayConnection.handle() covering all three branches — this is the exact logic that caused the production incident.

relay/web/src/connection.ts
```diff
@@ -148,8 +160,24 @@ export class RelayConnection {
...
+        this.onCommandDenied(typeof inner.command === 'string' ? inner.command : '');
+        return;
+      }
       if (inner.type === 'denied') {
-        this.onState('denied', typeof inner.reason === 'string' ? inner.reason : 'revoked');
+        const reason = typeof inner.reason === 'string' ? inner.reason : '';
```
brannon-bowden: Fallback changed from defaulting missing/non-string reason to 'revoked' (ended) to defaulting to '' (treated as command denial). Confirm no legitimate 'denied' frame is ever sent without a reason field, or this would now be silently swallowed instead of ending the session.

relay/web/src/main.ts
```diff
@@ -381,6 +381,11 @@ function connect() {
   conn.onFingerprint = (fp, ok) => { app.fp = fp; app.fpVerified = ok; };
   conn.onState = (s: ConnState, detail?: string) => onConnState(s, detail);
   conn.onMessage = (m) => onAgentMessage(m);
+  // A refused command is a fact about that command, not about your access.
+  conn.onCommandDenied = (command) => {
+    // eslint-disable-next-line no-console
+    console.warn('[relay] command refused by the machine:', command || '(unknown)');
```
brannon-bowden, line 387: onCommandDenied only logs to the console — confirm there's no intended user-facing feedback for a refused command.

src/main/api/relay/session-tunnel.ts
```diff
@@ -502,7 +502,12 @@ export class SessionTunnel {
 
     const sessionId = typeof inner.sessionId === 'string' ? inner.sessionId : null;
     if (!permits(s.grant, command, sessionId, this.deps.now())) {
-      this.sealTo(clientId, { type: 'denied', reason: 'not_permitted', command });
+      // A refused COMMAND is not an ended SESSION. These were one frame type
+      // once, and the guest client — reasonably — read every `denied` as "you
+      // are out", so a single unpermitted command told someone their access
+      // had been revoked. Separate types, separate meanings.
+      this.sealTo(clientId, { type: 'command:denied', reason: 'not_permitted', command });
```
brannon-bowden, line 509: The 'reason: not_permitted' field is no longer consumed by the client (onCommandDenied only reads command). Consider dropping it or documenting why it's retained.

brannon-bowden (changes requested):
Request changes — the server-side fix and its test harness look solid, but the client-side logic that actually caused the production incident this PR is fixing (`relay/web/src/connection.ts`, `relay/web/src/main.ts`) ships with no test coverage at all.

## Blocking issues

- **No tests for the new client-side branching logic.** `relay/web/src/connection.ts:170-182` adds `isEndingReason`/`ENDING_REASONS` and the `command:denied` vs. `denied` split, including a legacy-compat fallback that reinterprets a `denied` frame with a non-ending reason as a command denial. This is exactly the code path that produced the "refused command shown as revoked" bug described in the PR. All the new/updated tests in `session-tunnel-policy.test.ts` exercise the *server* (`session-tunnel.ts`) side only — none of them invoke `RelayConnection.handle()` or assert on `onState`/`onCommandDenied` dispatch. `main.ts`'s new `onCommandDenied` handler and `renderEnded`'s new fallback copy (`main.ts:420-431`) are likewise untouched by any test in this diff. Please add direct unit tests for `connection.ts`'s frame-routing (`command:denied` → `onCommandDenied`; `denied` with an ending reason → `onState`; `denied` with a non-ending/legacy reason → `onCommandDenied`) before merging — this is precisely the logic that regressed in production once already.

## Non-blocking issues

- `relay/web/src/connection.ts:175` — the fallback for a `denied` frame changed: previously a missing/non-string `reason` defaulted to `'revoked'` (session ended); now it defaults to `''`, which `isEndingReason` rejects, routing it to `onCommandDenied` instead. If any code path (not visible in this diff) ever emits a terminal `denied` frame without a `reason` field, it would now be silently logged to the console instead of ending the session in the UI. Worth double-checking there's no such call site before relying on this.
- `relay/web/src/main.ts:387` — `onCommandDenied` only does a `console.warn`; there's no user-facing indication that a command was refused. That may be intentional for this first pass, but it's worth confirming that's the intended UX (a console-only signal is easy to miss for an end user).
- Rollout ordering: if the relay server and the `relay/web` client can be deployed independently, a client running the *old* `connection.ts` would receive a new `command:denied` frame from an upgraded server and fall through to `onMessage` (since it doesn't match either `'command:denied'` or `'denied'` in the old code). The PR explicitly handles old-agent→new-client compatibility but not old-client→new-server; confirm this isn't a real deployment scenario.

## Nits

- `session-tunnel.ts:509` — the `reason: 'not_permitted'` field on the new `command:denied` frame is no longer read by the client (`onCommandDenied` only uses `command`). Harmless, but consider dropping it or documenting why it's kept.
- `ENDING_REASONS` (`connection.ts`) duplicates the `DeniedReason` union as a literal string array; if the type ever changes the array won't be caught by the compiler. Deriving it from the type (or a `satisfies` check) would prevent drift.

## Tests

- Missing: direct unit tests for `connection.ts`'s `handle()` routing (`command:denied`, `denied` with ending reason, `denied` with legacy/non-ending reason) — see blocking issue above.
- Missing: a test for `main.ts`'s `renderEnded` fallback copy path (unknown/unmapped reason) and the `onCommandDenied` wiring.
- The additions to `session-tunnel-policy.test.ts` are otherwise thorough and clearly tied to the described production incident.

William-Long-II:
Blocking item closed in `7878eab`, along with three of the four other points. Re-requesting.

## The tests

You were right, and the gap was worse than "this PR lacks tests": `relay/web` had exactly one test file (`reconnect.test.ts`) and **nothing had ever touched `RelayConnection`**. The class that terminates every guest session was completely uncovered, which is why a human hitting it was the detection mechanism.

13 tests now drive the **real `handle()`** — through a genuine Ed25519-signed handshake and real AES-GCM sealed frames — rather than an extracted predicate. That distinction matters here: the bug was never in a predicate, it was in *which branch a frame reached*, so a test of `isEndingReason` in isolation would have passed throughout.

Covered: `command:denied` → `onCommandDenied` with no state change; `denied` + ending reason → `onState` with the reason preserved (parameterised over all five); the **legacy** `denied`/`not_permitted` → treated as a refusal; the connection still delivering messages afterwards; and the unsealed pre-key `denied` still meaning only `not_authorized`.

Verified by reverting the fix and watching the legacy case fail — which is precisely what the guest hit on beta.4.

## Your `reason`-missing point — taken, and it was a real regression

You flagged that a `denied` without a `reason` would now route to `onCommandDenied` instead of ending the session. Correct, and I'd introduced that: my first cut defaulted a missing reason to `''`, which `isEndingReason` rejects.

It now **fails toward telling them**: no reason ends the session; only a reason that is present *and* not an ending one is treated as a refusal. Silence is the worse error — a swallowed notice leaves someone watching a channel that is already closed. Tested.

## `ENDING_REASONS` drift — taken

`DeniedReason` is now derived from the array with `as const`, so the two cannot diverge. Two hand-maintained lists of the same thing is the shape of the bug this PR is fixing, so thank you for not letting it in.

## Old client → new server

Not a real scenario: **the relay serves the web client**, so they ship in the same container and update together. A stale cached bundle is the only way to get the mismatch, and it degrades safely — the old client falls through to `onMessage`, and `onAgentMessage` ignores unknown types. No crash, just no console line.

## `reason: 'not_permitted'` on the new frame

Kept deliberately. `onCommandDenied` doesn't read it today, but the agent logs it, and it's the field a future user-facing message would use. Happy to drop it if you'd rather not carry an unread field.

## Console-only feedback — your call

Deliberate for this pass, and I'd like your read. For a watch-only guest the UI offers no way to type, so a refusal means something unexpected happened rather than a user being told "no" — and the watch-only strip already states the rule permanently. A toast on every refused keystroke could be noisier than the situation warrants. But if you think silence is wrong, a transient line on the strip is a small change.

739 desktop tests, 163 relay (+13). Four typechecks clean, web bundle builds.

brannon-bowden (approved):
Approve — this update adds exactly the test coverage the prior review asked for, and it's substantial: `connection.test.ts` drives `RelayConnection.handle()` through a real handshake and real sealed frames, then asserts the `command:denied` vs. legacy `denied`/non-ending-reason vs. ending-reason branches all land where they should. That was the code path that caused the production incident, and it's no longer untested.

Specifically resolved from the last review:
- **Blocking: no client-side test coverage** — fixed. The new suite exercises `command:denied` → `onCommandDenied`, legacy `denied`/`not_permitted` → `onCommandDenied` (the exact incident frame), all five `ENDING_REASONS` → `onState('denied', reason)`, and a `denied` with no reason → `onState('denied', 'revoked')`. `isEndingReason` is also tested directly for accept/reject cases.
- **Non-blocking: `''` default for a reason-less `denied` frame silently dropping to `onCommandDenied`** — fixed. `connection.ts` now distinguishes "no reason" (`null`) from "reason is present but not ending" and treats a missing reason as session-ending (defaulting to `'revoked'`), matching the pre-existing behavior and covered by the new "a denied with NO reason ends the session" test.
- **Nit: `ENDING_REASONS`/`DeniedReason` duplication** — fixed. `DeniedReason` is now derived from `ENDING_REASONS` via `(typeof ENDING_REASONS)[number]`, so they can't drift.

## Suggestions (non-blocking)

- `main.ts`'s `onCommandDenied` handler and `renderEnded`'s fallback copy for unmapped reasons are still without direct test coverage — this diff doesn't touch `main.ts`. Given the root-cause routing logic in `connection.ts` is now well covered, I'm not treating this as blocking, but it's still worth adding since it's the last untested piece of this feature's UI wiring.
- Rollout ordering between relay server and `relay/web` client (an old client receiving a new `command:denied` frame it doesn't recognize) is unchanged by this diff — worth confirming that's not a real independent-deploy scenario, but not something this PR needs to solve.
- `session-tunnel.ts`'s `reason: 'not_permitted'` field on the `command:denied` frame still appears unread by the client (`onCommandDenied` only consumes `command`). Harmless as-is.

PR #166, opened 2026-08-14
Fix account switching losing the session, and show the live account

### LINKED ISSUE #164: Bug: switching a session's Claude account loses the conversation and orphans the pty
## What happens

Switching a session's Claude account is broken in three visible ways:

1. **Switching on a live session appears to do nothing.** The picked account never takes effect; the session keeps running (and billing) on the old one.
2. **Stopping the session, then switching, loses the conversation.** Resuming prints the CLI's "No session with that index" error.
3. **Resuming a session after a switch renders the TUI in roughly a quarter of the window.**

## Root cause 1 — pty lifecycle callbacks are keyed by session id, not by pty instance

`Terminal.tsx` restarts a session by cycling `isRunning` off, then back on after a fixed **100 ms** timer (`src/renderer/components/Terminal.tsx:200-216`). The cleanup calls `killSession`; `pty:kill` is fire-and-forget (`safeOn`, `src/main/index.ts:723`) even though `PtyManager.kill()` returns a promise nobody awaits.

Claude Code usually takes longer than 100 ms to die, so the **old** pty's `onExit` closure fires *after* the replacement pty was inserted under the same session id. That closure looks the session up by id alone (`this.sessions.get(id)`, `src/main/pty-manager.ts:296`), so it operates on the **new** session. Two branches, both destructive:

- **Non-zero exit** (typical for a killed CLI): the BDHLNDR-9 resume-failure fallback fires, because the new session has `resumeAttempted: true` and a `spawnedAt` inside the 5 s window. It calls `clearStoredClaudeSessionId(id)`, deletes the live session from the map, and spawns a third pty on a fresh conversation. **The stored conversation UUID is destroyed** — symptom 2.
- **Zero exit**: it falls through to `emit('exit')` + `this.sessions.delete(id)` (`src/main/pty-manager.ts:324-325`), **orphaning the brand-new pty**. `write`/`resize`/`onData` all guard on a map lookup, so every keystroke and every byte of output for the new process is silently dropped — symptom 1. It also tears down that session's chat parser (`src/main/index.ts:583-602`).

`onData` has the same id-only lookup, so a not-yet-dead old pty can interleave output into the new session's terminal.

## Root cause 2 — the silent resume-failure respawn spawns at 80x24

The fallback respawn calls `createSession`, which hardcodes `cols: 80, rows: 24` (`src/main/pty-manager.ts:246-249`), and deliberately hides the exit from the renderer ("the renderer never sees an 'exit' event for this path"). The renderer's xterm is still at full size and has no reason to re-send a resize, and `lastCols`/`lastRows` are not carried into the replacement. Claude's TUI therefore renders inside an 80x24 box in a full-size terminal — symptom 3.

## Root cause 3 — the transcript is carried at switch time, from a snapshot that can already be stale

`assignSessionAccount` copies the transcript into the new account's config dir at switch time (`src/main/account-switch.ts:36-50`). But a still-running Claude process keeps writing its transcript into the **old** account's dir, because `CLAUDE_CONFIG_DIR` was baked in at spawn (`buildAgentSpawn`, `src/main/pty-manager.ts:361-430`). After a raced live switch, the copy is stale or points at the wrong tree, the next `--resume` fails, and the fallback then wipes the UUID.

## Fix

1. Identity-guard the pty closures: `onExit`/`onData` must bail unless the entry still in the map is *this* pty (instance identity or a generation counter).
2. Make `pty:kill` a `handle` and await the kill before respawning, instead of the blind 100 ms timer.
3. Carry `lastCols`/`lastRows` into the resume-failure respawn so the silent replacement is not 80x24.
4. Carry the transcript lazily at launch time (in `buildAgentSpawn`, when the effective config dir differs from where the transcript actually lives) rather than at switch time, closing the stale-snapshot window.

## Acceptance

- Switching the account on a running session restarts it under the new account, keeps the conversation, and keeps the terminal at full size.
- Switching on a stopped session, then starting it, resumes the same conversation under the new account.
- Regression tests cover the late-exit race (both exit codes), the respawn size, and the lazy transcript carry.

### LINKED ISSUE #165: Feature: show the active Claude account on the session header and identify accounts by email
## Problem

Nothing in the UI says which Claude account a session is actually running under. The assignment lives in a context menu you have to re-open to read, and because `CLAUDE_CONFIG_DIR` is baked in at spawn (#164), the assignment in the database and the account the live pty is really using can disagree — with no way to see it. That disagreement is exactly what made #164 feel like "switching does nothing".

Accounts are also identified by a free-text label alone in most surfaces. Two accounts both labelled "Work" are indistinguishable, and the label says nothing about *which* Anthropic login it is.

## What to build

**1. Active-account indicator on the session header** (`src/renderer/components/TerminalHeader.tsx`)

- Show the account the session is running under: color swatch + label, with the email as the secondary identifier.
- Show the **live** account (what the running pty actually spawned with), not just the database assignment.
- When the assigned account differs from the live one — a switch that has not been applied yet — say so, and offer the restart that applies it.
- When no accounts are registered, render nothing (preserve today's pre-accounts look).

**2. Identify accounts by more than their label**

- Wherever an account is named (header indicator, the two account context menus in `src/renderer/App.tsx`, and the accounts panel), pair the label with its email and color swatch.
- `ClaudeAccount` already carries `email` and `color` (`src/shared/types.ts:259-273`), so this is presentation, not schema.

**3. Show usage in the accounts panel** (`src/renderer/components/ClaudeAccountsModal.tsx`)

- Mark which accounts are currently in use by running sessions, and how many.
- Keep the existing `default` tag distinct from "in use right now" — they answer different questions.

## Notes

- The live-account signal needs main to record the account each pty spawned under and expose it to the renderer; that plumbing lands with #164.
- Accessibility: the swatch is decorative — the account must be identifiable without relying on color alone.

## Acceptance

- The session header names the account the session is really on, with email and swatch.
- A pending (unapplied) switch is visibly distinct from an applied one, with a way to apply it.
- The accounts panel shows which accounts are in use and by how many sessions.
- Two accounts sharing a label are still tellable apart everywhere an account is named.

William-Long-II:
Closes #164
Closes #165

## The bug, in one sentence

A pty's `onData`/`onExit` close over the session id and re-look the session up when they fire, so a Claude Code that outlived the restart reported its death against its own replacement.

`Terminal.tsx` restarted by killing and respawning **100 ms** later, and `pty:kill` was fire-and-forget (`safeOn`) even though `PtyManager.kill()` returned a promise nobody awaited. When the old process died late, its exit closure found the *new* session under the same id and took one of two destructive branches:

- **non-zero exit** — the BDHLNDR-9 resume-failure fallback fired against the live replacement, cleared its stored conversation UUID and started the user over on a blank conversation (the "No session with that index" symptom);
- **zero exit** — it emitted `exit` and deleted the map entry, orphaning the brand-new pty. `write`/`resize`/`onData` all guard on a map lookup, so every keystroke and byte of output was silently dropped — which is why switching on a live session looked like it did nothing.

## What changed

| Fix | Where |
|---|---|
| Identity-guard the pty callbacks | `pty-manager.ts` — `isSupersededPty` |
| `pty:kill` settles on the real exit, and is awaited before respawn | `index.ts`, `preload.ts`, `Terminal.tsx` |
| POSIX `SIGKILL` escalation, so awaiting a kill means something | `pty-manager.ts` — `kill()` |
| Size carried into the silent resume-failure respawn | `pty-manager.ts` |
| Transcript carried at launch time, not switch time | `conversation-transcript.ts` (new), `account-switch.ts` |
| Conversation-id clear gated on the staging result | `pty-manager.ts` |

The guard is deliberately shaped as **"bail only if the map holds a DIFFERENT pty"**, never "bail unless the map holds this pty". `kill()` removes the entry *before* signalling the process, so the strict form would suppress the exit event for every deliberate kill and break `cancelLoginFlow`, the provider-install modal's completion detection, the ws-server `terminal:exit` fan-out and the relay tunnel. `settleKill` also runs *before* the superseded check, or a kill whose id was re-used would never resolve and the renderer would await it forever.

Instance identity rather than a generation counter: `spawnedAt` is `Date.now()` and collides for two spawns in the same millisecond, and a new counter would be a second, contradictable notion of the same fact.

### Adjacent defects found while proving the above

- Session idle/working timers outlived their own session and drove state changes — and notifications — for whatever took its place.
- `createSession` silently overwrote an incumbent map entry; the new guard would have turned that from loud corruption into a permanently invisible orphaned process, so it now throws.
- A stored `claude_session_id` reaching `buildAgentSpawn` from an imported group file (`group-import-export.ts` inserts it with no validation) was interpolated into a shell command line under a comment asserting UUIDs cannot contain metacharacters. Now validated at the point of use.
- `kill()` previously only ever resolved from its own 3 s timer, so every existing `await ptyManager.kill()` in provider-install and `killAll` was stalling a full three seconds.

## #165 — the indicator

The header names the account the pty **actually launched under**, not the database row — that distinction is the whole point, since the binding is fixed at spawn and the two can disagree. An unapplied switch says `Restart to use <name>` and offers the button that does it. The sidebar dot resolves through the same path, so the two surfaces in one window can no longer give contradictory answers.

The accounts panel marks what is in use and now names those sessions in the delete confirmation, since deleting removes the config directory out from under live ptys. New logins draw a colour from the sidebar palette — every account had been defaulting to the same `#888888`, so the visual half of the identifier conveyed nothing. Pre-existing accounts stay grey, which is why nothing is allowed to depend on the swatch alone.

Accessibility: email colour is a locally-defined `#9a9a9a` (5.4:1 on the header, 5.9:1 on the panel) because neither existing token cleared AA on both surfaces — and the refactor had regressed the panel's previously-passing email. A missing email now renders nothing rather than "Not yet logged in", which on a *running* session was both common on macOS and actively misleading. Accounts are identified as `Label (email)` wherever ambiguity is possible.

## Verification

`npx tsc --noEmit` clean · `build:main` and `build:renderer` clean · **813 pass / 0 fail** across 58 files (baseline before this work: 739 / 0).

The regression reviewer proved the new tests actually pin the fix rather than passing vacuously: reverting `isSupersededPty` to `return false` fails both late-exit tests, reverting the `onData` instance check fails the stale-output test, and restoring the 100 ms timer fails 3 of 5 restart tests.

## Deliberate scope boundaries

- **Context menus gain email but not a swatch.** `ContextMenu.tsx` renders string labels only; giving it rich labels is a separate change.
- **Silent spawns outside the renderer still start at 80x24.** `remote-sessions.ts:116` and `api/routes/sessions.ts:212,309` call `createSession` without a size. Symptom 3 is fixed for the resume-failure respawn; a session *started* remotely keeps the old stale-width behaviour.
- **`src/renderer/types/electron.d.ts` is deleted.** It was a stale duplicate declaring a second, conflicting `Window.electronAPI` global that `skipLibCheck: true` was hiding. Nothing references it.

## Known follow-ups (not in this PR)

- `ensureTranscriptInConfigDir` cannot distinguish "the transcript is nowhere" from "a config dir could not be read", so it cannot safely downgrade a provably-doomed `--resume` up front.
- No UI to reset a session's conversation id when the CLI genuinely cannot read a transcript it can see.
- The `kill()` force-timer paths (Windows `taskkill`, the new POSIX escalation, the `keepReaper` hand-off) are untested — they only fire after a real 3 s timeout, which needs the interval injectable.
- The accounts repository still has no tests: `pty-manager.test.ts` and `account-auth.test.ts` mock `../repositories/accounts` process-wide, so a real-repo test in the same run silently exercises their stub.
- `ptyManager.kill()` is still called unawaited at `account-auth.ts:222` and `api/routes/sessions.ts:270,336` (pre-existing).


src/main/repositories/accounts.ts
```diff
@@ -114,13 +114,28 @@ export function touchAccount(id: string): void {
...
 export function deleteAccount(id: string): void {
   const db = getDatabase();
   const tx = db.transaction(() => {
+    const wasDefault = db.prepare('SELECT is_default FROM claude_accounts WHERE id = ?').get(id) as
+      { is_default: number } | undefined;
     db.prepare('UPDATE sessions SET claude_account_id = NULL WHERE claude_account_id = ?').run(id);
     db.prepare('UPDATE groups SET claude_account_id = NULL WHERE claude_account_id = ?').run(id);
     db.prepare('DELETE FROM claude_accounts WHERE id = ?').run(id);
+    if (wasDefault?.is_default) {
+      db.prepare(`
```
brannon-bowden, line 134: New default-promotion behavior on deleteAccount has no test coverage. Add tests for: promoting the oldest survivor when the default is deleted, leaving default untouched when a non-default account is deleted, and deleting the last account without error.

brannon-bowden (changes requested):
Request changes: this is an unusually well-engineered PR — the pty-identity/kill-ordering rewrite, the launch-time transcript carry, and the live-vs-assigned account UI are all backed by extensive, well-targeted tests — but one piece of new, non-trivial database logic ships with no test coverage at all.

## Blocking issues

- **`src/main/repositories/accounts.ts:134`** — `deleteAccount` now promotes the oldest remaining account to default when the deleted account was the default (new `wasDefault` / promotion `UPDATE` block). This is exactly the kind of state-transition logic the rest of this PR is careful to test (compare with the exhaustive coverage of `ensureTranscriptInConfigDir` and the pty-identity guard), but no test file in the diff exercises it. Nothing here pins: that deleting the default with other accounts present promotes the oldest survivor, that deleting a non-default account leaves the default untouched, or that deleting the last remaining account doesn't throw when the promotion subquery returns no row. Given this directly feeds the "no default account" failure mode the docstring itself describes as confusing to users, please add a test (in whatever suite already covers `repositories/accounts.ts`, or a new one) before merge.

## Non-blocking issues

- **`src/renderer/components/AccountChip.tsx`** (`emailText = storedEmail`) — when an account's stored email has surrounding whitespace but non-empty content, the `.trim()` check passes but the untrimmed `storedEmail` (not the trimmed value) is what gets displayed. Minor cosmetic edge case.
- **`src/main/preload.ts` / `src/main/index.ts`** — `pty:kill` moved from a fire-and-forget `send`/`on` pair to an `invoke`/`handle` request-response pair. Every caller visible in this diff (`preload.ts`, `Terminal.tsx`) was updated consistently, but I can't see the rest of the codebase to confirm there's no other `ipcMain.on('pty:kill', ...)` listener or renderer-side `ipcRenderer.send('pty:kill', ...)` caller left stranded elsewhere (e.g. relay/API code). Worth a quick grep before merge, not a reason to block since nothing in the shown diff is inconsistent.

## Tests

- As noted above, `deleteAccount`'s default-promotion path has no test.
- `SessionRow.test.tsx` covers the "account present + pendingSwitch" and "account present + no pendingSwitch" cases, but not the case where the live account itself is null (deleted) while a switch is pending — the `role="img"`/`aria-label` branch in `SessionRow.tsx`'s fallback pending marker. Given this is precisely the "user deleted the account out from under a running session" scenario the PR calls out as the case that matters most elsewhere (e.g. `App.accounts.test.ts`), a corresponding `SessionRow` test would close the loop.
- Everything else — pty identity/supersession, `kill()` coalescing and force-escalation, the resume-failure/transcript-staged gating, `ensureTranscriptInConfigDir`'s freshness heuristics, and the renderer's live-vs-assigned account resolution — is thoroughly tested.

William-Long-II:
Thanks — the blocking catch was correct, and it turned out to be hiding something worse than a missing test.

## Blocking: `deleteAccount` default-promotion

Fixed in 3e5ab2b. But the reason it was untested matters: an earlier pass on this branch *did* try to test `repositories/accounts.ts`, hit bun's process-wide `mock.module` collision, and **deleted the tests rather than solve it** — verifying the promotion SQL against a throwaway harness instead. So the first job was undoing that shortcut.

**The collision was confirmed, not assumed.** With the new test file added and the existing mocks in place, the suite ran 819 pass / 21 fail while the same file passed 22/22 alone — and account ids accumulated across tests (`"a","a","old","middle","newest"...`), which identified the culprit exactly: the module-level `accounts: AccountRow[]` array inside `src/main/__tests__/account-auth.test.ts` was the subject under test, not the repository. Bun evaluates and runs each file in turn, so a mock registered by an earlier-sorting path (`src/main/__tests__/*` sorts before `src/main/repositories/__tests__/*`) is already installed by the time a later file imports the real module.

**Nothing mocks the accounts repository any more:**
- `pty-manager.test.ts` — the mock was dead scaffolding. All three call sites in `pty-manager.ts` (`resolveAccountForSession`, `touchAccount`, `getAllAccounts`) sit behind `provider.capabilities.accounts`, which is `false` for the passthrough `codex` provider those tests run. Removed; 23/23 still pass, no assertion touched.
- `account-auth.test.ts` — switched from the array fake to the **real** repository over a mocked `../database` backed by in-memory `bun:sqlite`. That's this repo's own precedent from the same directory: `account-switch.test.ts` already did exactly this.

**That swap defused one of its own assertions, which is worth flagging rather than burying.** With the real repo in play, `accounts.filter(isDefault).length === 1` is *guaranteed* by the demote-then-insert plus the partial unique index, so it would pass even with the atomicity bug present. Re-expressed as *which* account is default (the first caller's) and mutation-proved: inserting an `await` between the first-account check and `createAccount` in `account-auth.ts` turns it red. To be precise, since I'd rather not overstate it — the test as a whole would still have gone red via the neighbouring `seedCalls.length` assertion, so nothing was silently green; but that specific default assertion really had been defused, and the rewrite restored its discriminating power.

**22 new tests**, covering your three cases plus `createAccount`, `getAllAccounts`/`getDefaultAccount`, `updateAccount`, `setDefaultAccount` and `touchAccount`. Each was proved by mutation rather than assumed — the full battery is 13 mutations, all caught, each by exactly the test that claims that behaviour. The ones you named:

| Mutation | Test that goes red |
|---|---|
| promotion block removed | *deleting the default hands it to the oldest remaining account* |
| `ORDER BY created_at ASC` → `DESC` | same test — proves "oldest", not merely "somebody" |
| promote unconditionally | *deleting a non-default leaves the default where it was* |
| SELECT oldest into JS and deref `.id` | *deleting the last account leaves an empty table instead of throwing* — your third case exactly |

Isolation verified in three orderings (file alone, both formerly-colliding files first, mine first) plus the whole suite: **840 pass / 0 fail across 59 files**, `tsc --noEmit` clean.

## Non-blocking

**Untrimmed email** — you were right, and it reached the tooltip too, so the chip text and its `title` could disagree. Both now use the trimmed value, with the empty-string fallback semantics preserved (the `??`-vs-falsiness comment explains why: `parseAccountEmail` resolves its candidates with a `??` chain, so a credentials file with an empty email yields `''`, not `null`). Test covers a padded `'  a@b.com  '`.

**`SessionRow` null-live-account with pending switch** — added, asserting the accessible name rather than mere presence, since the point of that branch is staying identifiable when there's no account object to name.

**`pty:kill` grep** — done, whole tree. No stranded callers:

```
literal "pty:kill" — 5 hits, only 2 are code
  src/main/preload.ts:19    ipcRenderer.invoke('pty:kill', id)          <- the only sender
  src/main/index.ts:734     safeHandle('pty:kill', ...)                 <- the only listener
  (3 remaining hits are comments / a test error-message string)

ipcRenderer.send — 4 hits, none is pty:kill  (about:close, pty:write, pty:resize, pty:prime)
ipcMain.on / once — 2 hits, neither names pty:kill  (the generic safeOn wrapper, about:close)
killSession — 4 renderer call sites, all awaited except one deliberate
  fire-and-forget in an effect cleanup, which is `.catch()`ed and commented
```

## Two pre-existing issues found while doing this — not in scope, flagging for follow-up issues

1. `PtyManager.kill` is `async`, but three in-process callers invoke it without `await` inside a synchronous `try/catch` that therefore cannot observe the rejection — `account-auth.ts:241` (whose catch carries a now-dead "Pty may have already exited" comment) and `api/routes/sessions.ts:270,336`. A throw becomes an unhandled main-process rejection. Same class as #164 one layer up, but predates this branch.
2. `setDefaultAccount(id)` demotes the incumbent *unconditionally* and then promotes by id, so a stale id leaves **zero** defaults — the exact accountless state the `deleteAccount` promotion block was added to prevent, reachable when the renderer's account list is stale. Pinned as current behaviour by a test whose comment says it documents rather than endorses, so a future tightening fails loudly instead of silently.

Happy to open both as issues if you'd rather they not just live in this thread.

brannon-bowden (approved):
Approve — the one blocking gap from the previous review (untested `deleteAccount` default-promotion) is now closed, and the other notes have been addressed as well.

## Blocking issues

None. The prior blocking issue — `deleteAccount`'s default-promotion logic shipping without tests — is resolved by `src/main/repositories/__tests__/accounts.test.ts`, which pins exactly the state transitions I flagged: promoting the oldest survivor when the default is deleted (`deleting the default hands it to the oldest remaining account`), leaving the default untouched on a non-default delete, and not throwing when the last account is deleted (empty promotion subquery). It also covers `createAccount`'s demote-then-insert path and `setDefaultAccount`'s demote-then-promote path, both guarded against the partial unique index — solid coverage for this state machine.

## Non-blocking issues

- **`src/main/repositories/__tests__/accounts.test.ts` — `setDefaultAccount('ghost')`** — the test titled "pointing at an unknown id leaves the app with no default" documents that `setDefaultAccount` demotes the incumbent unconditionally, so a stale/unknown id strands the app with no default account — the same failure mode `deleteAccount`'s promotion was built to avoid. The test comment itself calls this out as "pinning current behaviour, not endorsing it." This isn't new code introduced by this diff, so it's not a blocker here, but since #165 is otherwise closing this exact class of bug, it may be worth a follow-up to guard `setDefaultAccount` the same way.
- The `pty:kill` invoke/handle consistency question from the previous review is untouched by this delta and remains unverifiable from the diff alone — not a reason to hold up this PR.

## Tests

Both test gaps from the previous review are now filled:
- `deleteAccount`'s promotion path, `createAccount`'s demotion path, and `setDefaultAccount` all have direct coverage in the new `accounts.test.ts`, run against a real in-memory `bun:sqlite` schema (including the partial unique index), which is a meaningfully stronger test than the hand-rolled array fake it replaces.
- `SessionRow.test.tsx` now covers the "live account deleted while a switch is pending" case (both a named target and a fall-back-to-default target), plus a test confirming the marker stays decorative when the dot is present — closing the accessibility-branch gap noted previously.
- `AccountChip.test.tsx` adds coverage for the whitespace-trimming fix, confirming both the rendered text and the tooltip agree.

No new non-trivial logic in this delta is left uncovered.

PR #172, opened 2026-08-19
Bug: a guest re-opening a shared session loses everything they had watched

### LINKED ISSUE #169: Bug: a guest re-opening a shared session loses everything they had watched
## What happens

A guest with a watch-only grant opens the shared session in the web client and sees live output from the moment they subscribe. If they navigate back to the session list — to look at one of their own sessions — and then re-open the shared session, **everything they had already watched is gone** and they start again from a blank screen.

Reported from testing on `bodhilander.bodhilabs.dev`: "me leaving the session after having joined to one of mine and then coming back resulted in again a wipe of all info from that point forward — this would be basically useless to anyone staying busy with their own work + monitoring a shared session."

## Why

Two independent clears stack up:

- `src/main/api/relay/session-tunnel.ts:543` — `handleSubscribe` sends `\x1b[2J\x1b[3J\x1b[H── shared from here ──` and returns early for **every** non-owner subscribe, not just the first one. The owner branch immediately below replays `pty.getSerializedBuffer(sessionId)`; guests never reach it.
- `relay/web/src/main.ts:716` — `openTerminal` calls `term.clear()` on every open, and the web client keeps no per-session buffer.

So re-entering a shared session re-runs both, and the guest's accumulated view is discarded.

## The constraint this has to respect

`docs/designs/session-sharing.md` lists "sharing scrollback history" as an explicit v1 non-goal, and the comment at `session-tunnel.ts:541` states the reason: a guest joins at the moment they were let in, not at whatever was on screen before the decision to share was made. Replaying `getSerializedBuffer` to guests would hand over pre-share scrollback and is not the fix.

## Proposed fix

Capture a **share transcript** per session, starting at the moment the grant is approved, and replay that (not full scrollback) on every guest subscribe.

- `PtyManager` already keeps a capped per-session `scrollbackBuffer` (`src/main/pty-manager.ts:49`, `:133`) and renders it through a headless xterm in `getSerializedBuffer` (`:1130`). Add a monotonic byte counter so a *mark* taken at share time stays meaningful after the cap trims, plus `getSerializedBufferSince(id, mark)`.
- Record the mark when a grant is minted (`relay-client.ts approveGrant` / `mintGrant`), alongside `ptyEpoch` — it belongs to the same moment of consent.
- `handleSubscribe`'s guest branch replays from the mark, chunked through `chunkText` like the owner path, keeping the "── shared from here ──" banner at the top of the replay.
- Bound the transcript so a long-running shared session cannot grow without limit.

This also fixes the adjacent cases: a browser refresh, a network reconnect, and a guest joining from a second device all currently start blank.

## Acceptance

- A guest leaves a shared session and returns: everything since the share was approved is still there.
- A guest never sees output produced before the grant was approved.
- Owner behaviour is unchanged.
- A shared session running for hours does not grow the agent's memory without bound.

William-Long-II:
## What this fixes

A guest watching a shared session lost everything they had seen the moment they navigated away and back. Reported from testing: *"me leaving the session after having joined to one of mine and then coming back resulted in again a wipe of all info from that point forward — this would be basically useless to anyone staying busy with their own work + monitoring a shared session."*

## Why it happened

`handleSubscribe` sent `\x1b[2J\x1b[3J\x1b[H── shared from here ──` and returned for **every** non-owner subscribe, not just the first. "No scrollback for guests" had been implemented as "nothing, on every attach", and those are not the same rule — the second one also throws away what the guest was already shown.

## The change

The line is drawn at the moment of consent instead of at each attach:

- `approveGrant` captures a **mark** per session alongside `ptyEpoch` — the same moment, for the same reason.
- Every guest subscribe replays `getSerializedBufferSince(sessionId, mark)` after the marker, chunked like the owner path.
- The mark counts characters the PTY has ever produced rather than indexing into the scrollback buffer, so it survives the 2 MB cap trimming. If the cap has already evicted it, the guest gets what survives — a subset of what they were entitled to, never a superset.
- A grant with no mark starts its window at first attach. A missing mark is never read as position zero; zero means "replay everything", which is the one thing a guest must never be sent.
- Marks live in memory on the tunnel keyed by `grantId`, dropped on revoke and on expiry. They index into a PTY instance's output and a grant is bound to that instance by `ptyEpoch`, so neither outlives the process — persisting one would create a number that survives the buffer it points into.

**What a guest can see is unchanged in extent**: everything from the moment they were let in, nothing from before it. It just survives them looking away — which also fixes a browser refresh, a reconnect, and joining from a second device, all of which started blank before.

## Tests

- `session-tunnel-policy.test.ts` — re-attach is replayed and pre-share output is not; a grant with no mark starts its window at first attach; the marker still leads every attach; revocation forgets the window.
- `pty-manager.test.ts` — replay from a mark, a mark at the current position replaying nothing, and a mark the cap has evicted yielding only what survives.
- Full desktop suite (848) and relay suite (163) green; `tsc -p tsconfig.main.json` clean.

## Manual verification wanted

Two testers, one shared session: watch, back out to your own session, produce output on the shared one, return — the intervening output should be there, and nothing from before the share.

Closes #169

### DIFF
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -38,7 +38,8 @@ the split §4 of the relay design already pre-committed to — "enforced by the 
 
 ## Non-goals (v1)
 
-Machine-wide grants · guest-initiated sharing · sharing scrollback history · guest-driven
+Machine-wide grants · guest-initiated sharing · sharing scrollback history *from before the
+share* (see §6.1) · guest-driven
 `terminal:resize` · `session:create` / `group:create` / `dirs:list` for any guest, ever ·
 browser-held guest device keys · session recording · federation of grants between relays.
 
@@ -339,8 +340,9 @@ Concretely in `session-tunnel.ts`:
   per repeat, and `PtyManager` never raises its default max of 10);
 - replace the three per-client listeners with **one tunnel-level set** fanning out over the
   client map;
-- gate `terminal:subscribe` on scope *before* `getSerializedBuffer`, and skip history entirely
-  for guests — seal `\x1b[2J\x1b[3J\x1b[H` plus `── shared from here ──`;
+- gate `terminal:subscribe` on scope *before* `getSerializedBuffer`, and skip pre-share
+  history for guests — seal `\x1b[2J\x1b[3J\x1b[H` plus `── shared from here ──`, then the
+  share window (§6.1);
 - filter `sendSessions`/`sendGroups` to scope and strip `workingDir`;
 - `revokeGrant()` → `DENY_ALL`, clear subs, detach, sealed `denied`, `client:kick`;
 - cache decrypted key material in memory — `deriveSharedSecret` and `signWithIdentity` each
@@ -355,6 +357,32 @@ in bun and will break `pty-manager.test.ts` in a load-order-dependent way. Refac
 defaulting to the singletons, matching the repo's existing injectable-database convention.
 Testing an extracted pure policy module proves the table is right but not that `dispatch()`
 consults it — which is the bug class this feature must not ship.
+
+### 6.1 The share window (#169)
+
+"No scrollback" was implemented as "nothing, on every attach", and those are not the same
+rule. A guest who left the shared session to look at one of their own and came back found it
+wiped, with no way to recover what they had already been shown — which makes watching a
+session useless to anyone who is also doing their own work, the exact person sharing is for.
+
+The line is drawn at the moment of consent, not at each attach:
+
+- `approveGrant` captures a **mark** per session alongside `ptyEpoch` — a count of characters
+  the PTY has ever produced, not an index into the scrollback buffer, so it stays resolvable
+  after the 2 MB cap trims.
+- Every guest `terminal:subscribe` replays `getSerializedBufferSince(sessionId, mark)` after
+  the `── shared from here ──` marker. Re-attaching, refreshing, reconnecting and joining
+  from a second device all become continuous.
+- A grant with **no** mark starts its window at first attach. A missing mark is never treated
+  as position zero: zero means "replay everything", which is precisely what a guest is not
+  entitled to.
+- Marks live in memory on the tunnel, keyed by `grantId`, and are dropped on revocation. They
+  index into a PTY instance's output and the grant is bound to that instance by `ptyEpoch`, so
+  neither outlives the process; persisting one would create a number that survives the buffer
+  it points into.
+
+What a guest can see is therefore unchanged in extent — everything from the moment they were
+let in, nothing before it — and changed only in that it survives them looking away.
 
 ## 7. UX
 
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -376,10 +376,16 @@ The line is drawn at the moment of consent, not at each attach:
 - A grant with **no** mark starts its window at first attach. A missing mark is never treated
   as position zero: zero means "replay everything", which is precisely what a guest is not
   entitled to.
-- Marks live in memory on the tunnel, keyed by `grantId`, and are dropped on revocation. They
-  index into a PTY instance's output and the grant is bound to that instance by `ptyEpoch`, so
-  neither outlives the process; persisting one would create a number that survives the buffer
-  it points into.
+- Marks live in memory on the tunnel, keyed by `grantId`. They index into a PTY instance's
+  output and the grant is bound to that instance by `ptyEpoch`, so neither outlives the
+  process; persisting one would create a number that survives the buffer it points into.
+- A window is dropped on revocation, and **swept on expiry** — on every write and every
+  `client:open`. Revocation reaches the map through `revokeGrant`, but a grant that is
+  approved, never connected to, and simply times out leaves no client behind to carry its
+  window out any other way.
+- A mark, once recorded for a session, is never overwritten. Recomputing it would move a
+  window **forward**, hiding output the guest was entitled to — the same failure as sending
+  nothing, arriving by a different route.
 
 What a guest can see is therefore unchanged in extent — everything from the moment they were
 let in, nothing before it — and changed only in that it survives them looking away.
--- a/src/main/__tests__/pty-manager.test.ts
+++ b/src/main/__tests__/pty-manager.test.ts
@@ -477,3 +477,56 @@ describe('getSerializedBuffer (rendered-text history)', () => {
     expect(await manager.getSerializedBuffer('nope')).toBe('');
   });
 });
+
+/**
+ * #169. A shared session replays the window since someone was let in — not the
+ * whole scrollback, which would hand over what was on screen before the
+ * decision to share, and not nothing, which wipes the guest's view every time
+ * they come back to it.
+ */
+describe('getSerializedBufferSince (the window a guest is entitled to)', () => {
+  test('replays from the mark onward and nothing before it', async () => {
+    const manager = new PtyManager();
+    const { ptyProc } = createAgentSession(manager);
+    ptyProc.dataCb!('SECRET BEFORE THE SHARE\r\n');
+
+    const mark = manager.scrollbackMark('session-1')!;
+    ptyProc.dataCb!('watched output\r\n');
+
+    const text = await manager.getSerializedBufferSince('session-1', mark);
+    expect(text).toContain('watched output');
+    expect(text).not.toContain('SECRET BEFORE THE SHARE');
+  });
+
+  test('a mark taken at the current position replays nothing', async () => {
+    const manager = new PtyManager();
+    const { ptyProc } = createAgentSession(manager);
+    ptyProc.dataCb!('already on screen\r\n');
+
+    const text = await manager.getSerializedBufferSince('session-1', manager.scrollbackMark('session-1')!);
+    expect(text).toBe('');
+  });
+
+  test('a mark the scrollback cap has evicted yields what survives, never more', async () => {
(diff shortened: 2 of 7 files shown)


brannon-bowden (changes requested):
Request changes. The share-window design itself is careful and the reasoning documented in `session-tunnel.ts` and `pty-manager.ts` holds up under scrutiny — the mark-vs-index distinction, the eviction math in `getSerializedBufferSince`, the mid-replay `s.subs`/session-identity re-checks, and the revocation cleanup are all sound and well tested. The blocker here is CI, not the logic.

## Blocking issues

- **CI is red: `SonarQube Code Analysis` and `quality-gate` are both failing.** Per this review's policy, no PR merges with failing required checks. Please fix whatever these checks are flagging (or share the failure output if it's a false positive) and get both green before merge. `GitGuardian Security Checks` is passing.

## Non-blocking issues

- `session-tunnel.ts`, in the endings-are-terminal block inside `revokeClient` (`if (s.grant.grantId) this.shareMarks.delete(s.grant.grantId);`): cleanup of `shareMarks` only happens via `revokeClient` (triggered when an active client hits expiry/revocation) or explicit `revokeGrant`. A grant that is approved, whose guest never reconnects, and that simply times out with no client attached to trigger the `ranOut` branch would leave its entry in `shareMarks` for the lifetime of the process. Given `noteShareMarks` is also called directly from `relay-client.ts` at approval time (independent of any connected client), this path seems reachable. Worth a look, though the per-entry footprint is small.
- `relay-client.ts:670-682` (`noteShareMarks(grantId, sessions.map(...))`) and `session-tunnel.ts`'s `noteShareMarks` (which unconditionally overwrites any existing mark per session): if the approval flow this hunk lives in can ever run a second time for the same `grantId` (e.g. extending an existing grant to more sessions, or re-approving), it would recompute *current* marks for every session in `sessions` and overwrite the ones already recorded — silently moving an already-connected guest's window forward and hiding content produced between the two approvals. I can't see the rest of `relay-client.ts` to know whether this code path is truly one-shot per grant; if it is, this is moot. Worth confirming.
- By design (per the updated `docs/designs/session-sharing.md` §6.1 and the "is replayed what it has already seen" test), every `terminal:subscribe` re-sends the *entire* window since the share began, not just what's new since the client's last attach. For a long-lived, heavily-reconnecting guest this means resending up to the full 2 MB scrollback on every reconnect. That looks intentional rather than accidental, but it's a real bandwidth/latency cost worth flagging if guests are expected to reconnect frequently (flaky mobile networks, etc.).

## Tests

The new coverage in `pty-manager.test.ts` and `session-tunnel-policy.test.ts` is solid for the core contract (mark-vs-index survives eviction, no-mark-means-first-attach, revoke wipes the window, marker always precedes replay). Two gaps I didn't see covered:
- A PTY restart between grant approval and guest attach (i.e., `scrollbackTotal` resets to 0 in a new instance while an old, larger `mark` is still recorded for that `sessionId`/grant). `getSerializedBufferSince`'s clamping means this fails safe (empty replay, no crash), but there's no test pinning that behavior, and no visibility in this diff into whether the existing `ptyEpoch`/`live`-session filtering already excludes restarted sessions from the grant before a stale mark could even be consulted.
- The re-approval/extend-grant scenario raised above — a test asserting that a second `noteShareMarks` call for an already-active guest's session does not regress their window would directly resolve that non-blocking question.

William-Long-II:
Thanks — all three non-blocking points were worth acting on. Pushed as 528a1ea.

**CI (blocking).** One new-code issue, not a false positive: `pty-manager.ts:1177` wanted an optional chain (`!session?.scrollbackBuffer` instead of `!session || !session.scrollbackBuffer`). Fixed. `GitGuardian` was already green.

**The `shareMarks` leak — confirmed, and fixed.** You had the path exactly right: `noteShareMarks` is called at approval independently of any client, so a grant that is approved, never connected to, and simply times out leaves nothing behind to trigger `revokeClient`'s cleanup. Each window now carries its grant's `expiresAt` and the map is swept on every write and every `client:open` — the two moments it is about to be used, rather than a timer. Covered by *"a window whose grant ran out is swept, not held for the life of the process"*.

**Re-approval overwriting a recorded mark — one-shot today, but fixed anyway.** `approveGrant` reads from `pendingGrants` and deletes the entry on success; `notePendingGrants` skips any grant already in the local table; and `insertGrant` would collide on the primary key. So the path is not currently reachable. It is still the wrong default: recomputing would move a window *forward* and hide output the guest was entitled to, which is the same failure as sending nothing, arriving from the other side. `noteShareMarks` is now first-write-wins — new sessions are added, recorded ones are left alone — with the test you asked for.

**Stale mark after a PTY restart.** Two layers, and you inferred both correctly. `grantFromCertificate` filters the grant's sessions by `ptyEpoch` and refuses the channel outright when none survive (pinned by the existing *"a valid grant whose sessions have all been restarted is refused"*), so a mark from a dead instance should never be consulted. `getSerializedBufferSince` clamps if one ever is. That clamp now has its own test rather than resting on inference.

**Full-window resend per attach — intentional, and load-bearing.** The web client calls `term.clear()` on every `openTerminal`, so a delta since the client's last attach would render as a blank screen with a tail stuck to it. Sending the window is what makes the reattach continuous. The cost is bounded by the 2 MB scrollback cap and is usually well under it: `getSerializedBufferSince` replays through a headless terminal and serializes *resolved text*, so the redraw traffic a TUI generates collapses rather than accumulating. Making this incremental means the client owning a per-session buffer and the agent tracking each client's position — worth doing if reconnect churn shows up as a real cost on mobile, but it is a different change from this one and I would rather not fold it in here.

brannon-bowden (approved):
Approve. This round resolves the CI blocker and both non-blocking concerns from the previous review, with tests added for each.

- **CI is now green** (`SonarQube Code Analysis`, `quality-gate`, `GitGuardian Security Checks` all passing) — the sole blocker from the prior review is resolved.
- **Timeout leak fixed**: `ShareWindow` now carries `expiresAt`, and `sweepShareWindows()` runs on every `client:open` and every `noteShareMarks` call (`session-tunnel.ts:383`, `session-tunnel.ts:600`), deleting entries whose grant has run out. This directly addresses the "approved-but-never-connected grant leaks forever" concern, and is covered by the new test `a window whose grant ran out is swept, not held for the life of the process`.
- **Re-approval window regression fixed**: `noteShareMarks` now applies "first write wins" per session (`session-tunnel.ts:602-609`: `if (!window.marks.has(sessionId)) window.marks.set(sessionId, mark)`), so a second call for an already-recorded session can no longer move the guest's window forward and hide content. Covered by the new test `a second approval does not move an already-recorded window forward`.
- **PTY-restart clamp test added**: `a mark from a previous PTY instance clamps to the end and replays nothing` pins the fail-safe-empty behavior in `getSerializedBufferSince` that was previously untested.

Both test gaps flagged in the prior review are now filled, and the fix for each underlying issue looks correct on inspection: `sweepShareWindows` deletes from the `Map` mid-iteration, which is safe per spec; the sweep runs before grant resolution on `client:open` so a just-expired window can't be read stale; and the "first write wins" logic in `noteShareMarks` is a minimal, targeted fix that matches the documented invariant.

## Non-blocking issues

- `session-tunnel.ts:601` — `ShareWindow.expiresAt` is fixed at the window's *first* creation (`this.shareMarks.get(grantId) ?? { expiresAt, marks: new Map() }`) and is never updated on subsequent `noteShareMarks` calls for the same `grantId` (e.g. the one triggered by `shareMark()` on detach). If a grant's expiry can ever be extended after initial approval, the window would still sweep on the original, earlier expiry. The code and docs assert approval is one-shot per `grantId` today, so this looks inert in practice — flagging only in case that invariant changes later.
- `relay-client.ts:706` — `grant.expiresAt ?? Date.now()` is documented as an unreachable fallback ("`insertGrant` always writes one"). I can't see `insertGrant`'s implementation to confirm that, but worth noting: if it were ever hit, the resulting window would carry an expiry of "now" and be swept on the very next sweep pass, i.e. it fails safe (window disappears) rather than fails open — so even in the unreachable case this isn't a correctness risk.
- The full-scrollback resend on every `terminal:subscribe` (not touched by this diff) remains a real bandwidth cost for frequently-reconnecting guests; still appears to be an intentional design tradeoff rather than an oversight.

## Tests

No further gaps — the two scenarios called out in the previous review (restarted-PTY mark clamping, re-approval not regressing an active window) are now both directly tested, plus the new sweep-on-expiry behavior.

William-Long-II:
Merged. Both remaining points, for the record:

**`ShareWindow.expiresAt` fixed at first creation.** Inert today and fails in the safe direction if it ever stops being. The only second caller for a live grant is `shareMark()`'s fallback, which passes `s.grant.expiresAt` — the same certificate's expiry, so there is nothing to drift. If a grant's expiry could later be extended, the window would sweep on the *earlier* time, and a swept window falls back to starting at the next attach: the guest loses replay they were entitled to, never gains replay they weren't. I left it alone rather than adding a max-of semantics for an operation ("extend a grant") that does not exist — inventing the rule before the feature would just be a second invariant to keep true.

**`grant.expiresAt ?? Date.now()`.** Confirmed unreachable: `insertGrant` (`grant-sql.ts`) takes `expiresAt` as a required argument and writes it in the same transaction as the grant row, and `mintGrant` computes it via `grantExpiryAt` before calling. The null in the row type is the window between redemption and countersigning, which is a state `mintGrant` has already left. Your read of the failure direction is right — a window stamped "now" is swept on the next pass, so it degrades to first-attach rather than persisting.

Thanks for both rounds — the timeout leak in particular was a real one that nothing in the test suite would have caught.

PR #173, opened 2026-08-19
Bug: an addressed share link rejects the correct GitHub account when the relay has no handle on file

### LINKED ISSUE #170: Bug: an addressed share link rejects the correct GitHub account when the relay has no handle on file
## What happens

An owner creates an addressed share link (`Who's it for? → @william-long-ii`). Opening that link in a browser signed in to the relay **as that exact GitHub account** is refused with "This link isn't for this account — It was addressed to a specific GitHub account. Sign in as that account…".

## Why

`relay/src/db/migrations/003_github_login.sql:20` added `users.github_login` with no backfill: *"existing users acquire it on their next sign-in."* Relay sessions live for 30 days (`relay/src/http.ts:63`), so any user whose cookie predates that deploy still has `github_login = NULL`.

`relay/src/repositories.ts:422-426` then fails closed:

```ts
if (row.expected_github_login) {
  const login = user.github_login?.trim().toLowerCase();
  if (!login || login !== row.expected_github_login) return { ok: false, reason: 'wrong_account' };
}
```

Failing closed on NULL is correct — a user who has not signed in since 003 must not silently satisfy an addressed invite. What is wrong is that the *outcome is indistinguishable from being the wrong person*, so the copy at `relay/web/src/main.ts:126` tells the right human that the link isn't theirs, and the only recovery (sign out, sign back in) is never mentioned.

## Fix

- Distinguish the cases in `redeemShareInvite`: return `login_unknown` when `expected_github_login` is set and the user's login is NULL, and keep `wrong_account` only for a genuine mismatch.
- Add `invite_login_unknown` copy to `REDEEM_COPY` that says the account's GitHub handle isn't on file yet and offers a sign-in-again action that returns to the invite.
- Consider refreshing `github_login` opportunistically rather than only on OAuth callback, so the window closes on its own.

## Open question worth confirming with data

`redeemShareInvite` checks `own_machine` **before** the login check (`repositories.ts:419`), so an owner clicking their own machine's link should see "That's your own machine", not the addressing error. Seeing the addressing error suggests the browser's relay user row is not the row the machine is linked to. `SELECT id, display_name, github_login FROM users;` plus the machine's `user_id` on the live relay would settle whether this is only the NULL-login path or a second, separate identity problem.

## Acceptance

- A user whose `github_login` is NULL gets an explanation that names the real cause and a way out.
- A genuinely wrong account still gets `wrong_account` and is still refused.
- No path lets a NULL login satisfy an addressed invite.

William-Long-II:
## What this fixes

An addressed share link (`Who's it for? → @william-long-ii`) refused the person it was addressed to, while they were signed in to the relay as exactly that GitHub account: *"This link isn't for this account — sign in as that account, or ask for a link addressed to you."*

## Why it happened

`003_github_login.sql` added `users.github_login` **nullable with no backfill** — "existing users acquire it on their next sign-in" — and relay sessions live 30 days (`http.ts:63`). So any account whose cookie predates that deploy has a NULL handle, and `redeemShareInvite` fails closed on NULL.

Failing closed is correct. Reporting it as `wrong_account` is not: it names the wrong cause, tells the right person the link isn't theirs, and points at the one recovery that doesn't work (signing in *as that account* — they already are) while hiding the one that does (signing in *again*).

## The change

- **Relay** — NULL handle on an addressed invite returns a new `login_unknown` reason. A genuine mismatch still returns `wrong_account`. Neither can be satisfied by a NULL.
- **Web** — `invite_login_unknown` copy explains the real cause and renders a *Sign in with GitHub* button, which stashes the invite through `INVITE_STASH` before redirecting so OAuth returns to the link with its `#fp=` fingerprint fragment intact, rather than to the home page. Re-auth upserts the profile, which is what puts the handle on file.
- **Observability** — refused redemptions are now logged with their reason. Previously a refusal was invisible from both ends: misleading copy for the guest, nothing at all for the owner.

A failed redemption leaves the invite `pending`, so the retry after signing in lands on a code that still works — covered by a test, because the alternative (a refusal that consumed the code) would make the fix unreachable.

## Tests

- `sharing.test.ts` — NULL handle reports `login_unknown`; the invite survives the refusal and the same user's retry succeeds after their handle is on file.
- `shares-http.test.ts` — the endpoint returns `409 invite_login_unknown` and fires no redemption callback.
- Relay suite (166) green; `tsc --noEmit` clean for both relay and web trees.

## Still worth confirming on the live relay

`redeemShareInvite` checks `own_machine` **before** the handle, so clicking your own machine's link as its owner shows "That's your own machine", not the addressing error. Seeing the addressing error means the browser's relay user row was not the row the machine is linked to. `SELECT id, display_name, github_login FROM users;` plus the machine's `user_id` would confirm whether this PR is the whole story or only part of it — and the new refusal log will say so directly on the next attempt.

Closes #170

### DIFF
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -373,6 +373,12 @@ roll-up, not the only route.
   `expected_github_login` so redemption by anyone else fails closed. A secondary *"Or make an
   open link"* covers the case where the owner doesn't know the handle. An addressed invite
   turns the approval modal into a confirmation instead of a stranger-check.
+
+  A NULL `users.github_login` — every account that has not signed in since migration 003 —
+  fails closed, and must, but it is **not** the same refusal as a wrong account (#170). It
+  gets its own reason, `login_unknown`, and copy that offers the one action that fixes it:
+  sign in again, which is what puts the handle on file. Telling the person the link *was*
+  addressed to that it "isn't for this account" is both wrong and a dead end.
 - **What they can do?**
   - **Watch** — *"They see this session's live output. They can't type."*
   - **Watch and type** — *"They can run any command you can, read any file you can, and spend
--- a/relay/src/http.ts
+++ b/relay/src/http.ts
@@ -484,6 +484,10 @@ export function createRouter(ctx: RelayContext) {
     const result = repos.redeemShareInvite(code, user, crypto.randomUUID());
     if (!result.ok) {
       const status = result.reason === 'not_found' ? 404 : 409;
+      // Logged because a refusal is the one outcome the guest cannot explain
+      // and the owner cannot see. `login_unknown` in particular looked exactly
+      // like a wrong account from the outside.
+      logger.info('share invite refused', { reason: result.reason });
       return json({ error: `invite_${result.reason}` }, status);
     }
 
--- a/relay/src/repositories.ts
+++ b/relay/src/repositories.ts
@@ -164,7 +164,18 @@ export interface CreateShareInviteInput {
 
 export type RedeemResult =
   | { ok: true; grant: MachineGrant; invite: ShareInvite }
-  | { ok: false; reason: 'not_found' | 'expired' | 'already_used' | 'revoked' | 'wrong_account' | 'own_machine' };
+  | {
+      ok: false;
+      reason:
+        | 'not_found'
+        | 'expired'
+        | 'already_used'
+        | 'revoked'
+        | 'wrong_account'
+        | 'own_machine'
+        /** Addressed invite, and this user's GitHub handle is not on file yet. */
+        | 'login_unknown';
+    };
 
 /** Thrown inside the redemption transaction to roll it back on a lost race. */
 class RedeemRaceError extends Error {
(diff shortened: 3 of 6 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped bug fix with logic, docs, and tests that reinforce each other.

**Blocking issues**

None.

**Non-blocking issues**

- `relay/src/repositories.ts:434-441` — The split of the NULL-login case into its own `login_unknown` reason is correct and the early `return` happens before `now()`/consumption logic, which the new "invite survives a login_unknown refusal" test (`relay/src/sharing.test.ts`) directly confirms — the invite isn't burned on a failed reauth-needed attempt. Good defensive test to have.
- `relay/src/http.ts:487-490` — `logger.info('share invite refused', { reason: result.reason })` only logs the reason enum, not user/session identifiers, so this looks safe from a PII standpoint. I can't see the top of the file to confirm `logger` is already imported here, but since only the reason (not full result or context) is used, this is a minor observability nicety rather than a risk.
- `relay/web/src/main.ts:186-198` — The reauth stash/redirect logic (rebuilding `/i/<code>` vs. using `pathname + hash`) is a bit fiddly; worth a comment sanity-check that `inviteCodeFromPath()`/`INVITE_STASH` continue to exist and behave the same way the pre-existing signed-out path expects, since this PR reuses that mechanism rather than defining it.

**Tests**

- Backend logic is thoroughly covered: `relay/src/shares-http.test.ts` and `relay/src/sharing.test.ts` both add cases for the new `login_unknown` reason, including the "old" NULL-login test being loosened to `ok: false` and a new precise assertion, plus a retry-after-reauth test proving the invite isn't consumed on refusal.
- The new front-end behavior in `relay/web/src/main.ts` (the `reauth` button, the `sessionStorage`/`INVITE_STASH` redirect logic, and the path-vs-code branching) has no accompanying test in this diff. If this project has any DOM/unit test harness for `main.ts` elsewhere, it'd be worth adding a case for the stash-and-redirect logic, since it's non-trivial branching (path-derived vs. code-derived stash value) that's easy to silently regress.

PR #174, opened 2026-08-19
Feature: an "until I revoke it" duration for shared session access

### LINKED ISSUE #171: Feature: an "until I revoke it" duration for shared session access
## What

Add an **Until I revoke it** choice to "Once they join, access lasts" in the share modal. Today the only options are 1, 4 and 8 hours (`src/renderer/components/ShareSessionModal.tsx:29`), and the relay refuses anything past 24 hours outright (`MAX_GRANT_TTL_SECONDS`, `relay/src/http.ts:79`).

Someone monitoring a shared session alongside their own work should not have their access die mid-afternoon, and re-issuing a link every few hours is friction with no security payoff — the owner already has revoke, and revocation is instant.

## Why this is safe here

The desktop, not the relay, is the authority on whether a grant is live: `session-tunnel.ts:437` refuses any certificate whose `grantId` is not `active` in the desktop's own table, and `session-tunnel.ts:445` additionally kills the grant when the shared session's PTY epoch changes — so a session restart already ends access regardless of clock. An unbounded certificate is not a bearer credential that outlives the owner's control.

## Shape

`grantTtlSeconds: 0` as the sentinel for "no expiry", carried through every place that already moves the TTL:

- `ShareSessionModal.tsx` — new option, and post-create copy that states plainly that access lasts until revoked.
- `buildShareCreateMessage` (`src/main/api/relay/grants.ts:92`) — the value is inside the signed message, so both trees must agree; `fixtures/sharing-v1.json` covers this and needs a case.
- `relay/src/http.ts:397` — `grantTtlSeconds <= 0` currently rejects; allow exactly `0`, keep rejecting negatives, keep the 24h cap for every non-zero value.
- `relay/src/repositories.ts` — `grant_ttl_seconds` carries through invite → grant unchanged.
- `mintGrant` (`grant-store.ts:114`) — a `0` TTL mints a far-future `expiresAt` rather than `now + 0`. Note `getMachineAccess` (`repositories.ts:339`) requires `expires_at IS NOT NULL AND expires_at > now`, and the sweep at `:512` deletes on `expires_at <= now`, so a far-future value behaves correctly in both without schema change.
- Settings' share list should render "Until revoked" rather than a date far in the future.

## Acceptance

- An "until revoked" grant still works the next day.
- Revoking it disconnects the guest immediately, as today.
- Restarting the shared session still ends the grant.
- A negative or over-cap TTL is still rejected by the relay.
- Both trees' `sharing-v1` fixture assertions cover the `0` case.

William-Long-II:
## What this adds

An **Until I revoke it** choice under "Once they join, access lasts". The options were 1, 4 and 8 hours, and the relay refused anything past 24 (`MAX_GRANT_TTL_SECONDS`). Anyone monitoring a shared session alongside their own work lost access mid-afternoon and had to be re-invited.

## Why it's safe to offer

The clock was never what bounded a share:

- the desktop's own table decides whether a grant is live, and `session-tunnel.ts:437` consults it on **every** `client:open`;
- scope is bound to the PTY **instance**, so restarting the shared session ends the grant whatever the certificate says;
- revocation reaches live sockets immediately (`revokeGrant` → `DENY_ALL` + sealed `denied`).

An unexpiring certificate is not an unrevocable one. The relay stores the owner's answer; it was never the authority on it.

## Shape

`grantTtlSeconds: 0` — a different answer to the question, not a longer one, which is why it isn't `3155760000`:

- **Modal** — new option, plus a line on the confirmation screen, because this is the one choice that keeps running after the owner has stopped thinking about it.
- **Certificate** — `grantExpiryAt(now, ttl)` maps 0 to `8_640_000_000_000_000`, the largest timestamp `Date` can represent. It's a safe integer (`assertGrantFieldsSafe` requires that) and every expiry check in both trees is a plain comparison, so nothing special-cases it.
- **Relay** — `grantTtlSeconds` must now be a whole number ≥ 0; 0 bypasses the cap, everything else is still capped at 24h. Negatives and fractions are newly rejected.
- **Approval** — `pending.grantTtlSeconds ?? DEFAULT` was already `??` rather than `||`, which is what keeps a chosen 0 from silently becoming four hours. Now commented as load-bearing.

## Correction to the issue

The issue said `fixtures/sharing-v1.json` covers the `share-create:v1` message and needed a case. It didn't cover it at all — the fixture only pinned the certificate and the policy table. That message is hand-duplicated in `relay/src/protocol.ts` and `src/main/api/relay/grants.ts`, is verified by signature, and a drift between the copies surfaces only as *"relay rejected the share request (401)"* for someone who did nothing wrong. So this PR **adds** a `shareCreate` vector, asserted independently by both trees, with `grantTtlSeconds: 0` in it — precisely the field one side might helpfully default away.

## Tests

- `ShareSessionModal.test.tsx` — the option exists, sends `0`, and the confirmation says so.
- `grants.test.ts` — `grantExpiryAt` for both branches; the never-expiry value is a safe integer and a valid `Date`; an unexpiring grant is still stopped by revocation. Plus the new cross-tree share-create assertions.
- `shares-http.test.ts` — 0 is accepted and stored as 0; negative and fractional TTLs are rejected; the ceiling still holds.
- Desktop suite (849) and relay suite (165) green; `tsc -p tsconfig.main.json` clean.

## Not included

Nothing renders a grant list in the desktop yet (`relayListShares` has no renderer consumer), so there is no "Until revoked" label to add to a settings screen. When that screen lands it must not format `8640000000000000` as a date.

Closes #171

### DIFF
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -382,6 +382,14 @@ roll-up, not the only route.
   *"This link works for 1 hour if nobody uses it. Once dana-k joins, they have access for 4
   hours."*
 
+  Access may also last **until the owner revokes it** (#171), carried as
+  `grantTtlSeconds: 0` — a different answer to the question, not a longer one. The clock was
+  never what bounded a share: the desktop's own table decides whether a grant is live and is
+  consulted at every `client:open`, the scope is bound to a PTY instance so a restart of the
+  shared session ends it regardless, and revocation reaches live sockets immediately. An
+  unexpiring certificate is not an unrevocable one, and a four-hour ceiling only meant
+  re-inviting the same colleague every afternoon.
+
 The created state shows the URL, a QR, the readable code, and `🔑 This link is the key. Send
 it over something private.`
 
--- a/fixtures/sharing-v1.json
+++ b/fixtures/sharing-v1.json
@@ -33,6 +33,25 @@
     "certificate": "grant:v1.Z3JhbnQ6djEKN2YzYTFjOTItNWI0ZS00ZDE4LTlhMjAtNmMxZThmMGIzZDQ3CmMyYjhhNGUwLTFkNzYtNGYzOS04ZTU1LTlhM2I3YzBkMmY2MQpodHRwczovL3JlbGF5LmV4YW1wbGUuY29tCmQ5MWU2YjczLTJhMDgtNGM1Zi1iMTM0LTdlNTBmOWE2Yzg4Mgp2aWV3ZXIKMTc2MDAwMDAwMDAwMAoxNzYwMDE0NDAwMDAw.xINoFOrJmE64iKEUDCsYEGL6JSpBOPOLl8KucZy8RFktvWqwxgcOmcn5IBXLgd-ejm6Aj44cMzuGkvNHiHZYAw"
   },
 
+  "shareCreate": {
+    "_note": [
+      "The message a desktop signs to create a share invite. Hand-duplicated in",
+      "src/main/api/relay/grants.ts and relay/src/protocol.ts, and a drift between them is",
+      "invisible until it surfaces as 'relay rejected the share request (401)'.",
+      "grantTtlSeconds is 0 on purpose: 0 means UNTIL REVOKED, and it is the value most",
+      "likely to be 'helpfully' defaulted away by one side and not the other."
+    ],
+    "parts": {
+      "machineId": "c2b8a4e0-1d76-4f39-8e55-9a3b7c0d2f61",
+      "expectedGithubLogin": "dana-k",
+      "role": "viewer",
+      "grantTtlSeconds": 0,
+      "inviteTtlSeconds": 3600,
+      "issuedAt": 1760000000000
+    },
+    "message": "share-create:v1\nc2b8a4e0-1d76-4f39-8e55-9a3b7c0d2f61\ndana-k\nviewer\n0\n3600\n1760000000000"
+  },
+
   "policy": {
     "_note": [
       "The certificate format will not drift; this will, the moment 'operator' becomes",
--- a/relay/src/http.ts
+++ b/relay/src/http.ts
@@ -76,6 +76,7 @@ const SHARE_PER_IP = 20;
  * defaults; these exist so a malformed or hostile request cannot produce a
  * grant that outlives any reasonable session.
  */
+/** Upper bound for a grant that expires at all; 0 means "until revoked". */
 const MAX_GRANT_TTL_SECONDS = 24 * 60 * 60;
 const MAX_INVITE_TTL_SECONDS = 7 * 24 * 60 * 60;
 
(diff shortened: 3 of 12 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped, thoroughly tested change, and the TTL-validation logic checks out under the edge cases I traced.

**What I verified**

- `relay/src/http.ts:398-400`: the new validation (`!Number.isInteger(grantTtlSeconds) || grantTtlSeconds < 0`, then a separate ceiling check) correctly admits `0` while still rejecting negatives, non-integers, `NaN`, `Infinity`, and stringified numbers (`Number.isInteger("0")` is `false`), and still enforces the 24h ceiling for every other value. Traced through `-1`, `1.5`, `0`, and values above `MAX_GRANT_TTL_SECONDS` — all resolve as the new tests expect.
- `src/main/api/relay/grants.ts`: `GRANT_NEVER_EXPIRES = 8_640_000_000_000_000` is the ECMAScript max representable date and is well under `Number.MAX_SAFE_INTEGER`, so it satisfies `Number.isSafeInteger` and stores cleanly in SQLite. `grantExpiryAt` centralizes the sentinel-to-expiry mapping, and `grant-store.ts:mintGrant` was updated to call it instead of hand-rolling `now + ttlSeconds * 1000` — the one call site the diff touches is consistent.
- `src/main/api/relay/relay-client.ts:672-680`: `pending.grantTtlSeconds ?? DEFAULT_GRANT_TTL_SECONDS` correctly preserves `0` (this already used `??`, not `||`, so no regression here — just a comment added).
- Test coverage is genuinely thorough: relay HTTP boundary tests (`shares-http.test.ts`), wire-format byte-for-byte tests against the fixture (`sharing.test.ts`, `grants.test.ts`), an expiry-computation unit test suite (`grantExpiryAt`), and a renderer test that drives the modal end-to-end and asserts both the submitted payload and the confirmation copy.

**Non-blocking issues**

1. `src/main/api/relay/grants.ts` / `grant-store.ts`: the safety argument for an unbounded grant rests entirely on revocation reaching live sessions immediately and PTY-restart invalidating scope — both are pre-existing mechanisms not shown in this diff, so I can't verify them here. Worth a sanity check that no other code path recomputes an expiry from `ttlSeconds` without going through `grantExpiryAt`, since a second call site would silently reintroduce the `now + 0*1000` bug for "until revoked" grants.
2. `src/renderer/components/ShareSessionModal.tsx:24-30`: "Until I revoke it" is offered as a plain dropdown option with the same weight as the timed ones. Given it's a meaningfully different risk profile (indefinite access vs. a few hours), consider whether it deserves slightly more visual distinction or a confirmation nudge — not required, just a UX suggestion.

**Tests**

Coverage looks solid across relay, main-process, and renderer layers for the new sentinel value; nothing missing that I can point to.

William-Long-II:
Merged. Both non-blocking points, answered rather than deferred:

**1. Second call site recomputing an expiry.** Checked — there is none. `grantExpiryAt` has exactly one production caller (`grant-store.ts:138`, inside `mintGrant`), and the only other `ttlSeconds * 1000` arithmetic in either tree is `repositories.ts:218` and `:249`, which compute expiries for relay **login sessions** and **link codes**. Neither takes a grant TTL, so neither can see the sentinel. Any future grant-expiry path has to go through the helper to get a certificate at all, since `mintGrant` is the only thing that signs one.

On the safety argument you couldn't verify from this diff, for the record: `session-tunnel.ts:437` refuses any certificate whose `grantId` is not `active` in the desktop's own table, checked on **every** `client:open` rather than at issue time; `:445` filters the grant's sessions by `ptyEpoch` and refuses the channel when none survive, so restarting the shared session ends the grant regardless of clock; and `revokeGrant` drops live clients to `DENY_ALL` with a sealed `denied` immediately. Those three are what make an unexpiring certificate a revocable one.

**2. "Until I revoke it" carrying the same visual weight as the timed options.** I left it as a peer option deliberately. The confirmation screen already states it separately — *"Access lasts until you revoke it — or until this session restarts"* — which is where the owner is actually about to send a link, rather than while they are still choosing.

Adding friction here would also misprice it relative to what is next to it: this is the watch-only role, revocable instantly, dead as soon as the session restarts, and it grants no input. The heavy-consent treatment is spent on **Watch and type**, which hands over the environment, and diluting that distinction with a nudge on the safe option makes the dangerous one look ordinary by comparison. If long-lived grants turn out to be the thing people forget, the honest fix is the grant list in settings — showing what is live and for how long — not a speed bump at creation.

PR #176, opened 2026-08-20
Web client has no way to sign out — or to see who you are signed in as

### LINKED ISSUE #175: Web client has no way to sign out — or to see who you are signed in as
## What's wrong

The relay web client has no account UI at all. There is no way to see who you are signed in as, and no way to sign out.

The server side already works — `POST /auth/logout` (`relay/src/http.ts:157`) deletes the session row and clears the cookie. Nothing in `relay/web/src/main.ts` ever calls it. The session-list header (`main.ts:289`) is brand + theme toggle; the only menu is the machine picker.

## Why it matters

Three concrete failures, in order of severity:

1. **Wrong-account recovery is impossible.** The `invite_wrong_account` redeem screen tells the guest "Sign in as that account" (`main.ts:133`) and then offers only "Go to my machines". There is no control anywhere in the app that lets them follow that instruction. The copy asks for something the UI cannot do.
2. **A borrowed or shared browser cannot be released.** Sessions last `SESSION_TTL_SECONDS`; short of clearing site cookies by hand, a signed-in user cannot end their own session.
3. **You cannot see which identity the relay has on file.** This is what made #170 hard to diagnose live: a user row with a NULL `github_login` looks identical to a correct one from the client. `publicUser` (`relay/src/http.ts:575`) doesn't return the handle, so the UI couldn't show it even if it wanted to.

## What to build

- **Account control in the session-list header** — avatar (or initial) button opening a sheet, matching the existing `openMachineMenu` / `openLinkMachine` pattern.
- **The sheet shows identity**: display name, and the GitHub handle *the relay actually has on file* — including an explicit note when it has none, since that is a real and previously invisible state.
- **Sign out**: `POST /auth/logout`, drop the connection, clear per-account local state (`bodhi.machineId`, the invite stash), return to the sign-in screen.
- **`invite_wrong_account` gets a working action** — a button that signs out and starts a fresh GitHub sign-in, keeping the invite stashed so the guest lands back on the link.
- **`publicUser` returns `githubLogin`** so the sheet can render it.

## Scope

Relay + relay web client only. No desktop changes, so this ships on a relay redeploy without a new beta build.

William-Long-II:
Found while testing the #170 fix live: the instruction "sign out and back in" turned out to be unfollowable, because the web client has no sign-out.

`POST /auth/logout` has always worked server-side (`relay/src/http.ts:157`) — it deletes the session row and clears the cookie. Nothing in the client ever called it, and nothing rendered your identity either.

## What that broke

- **`invite_wrong_account` is a dead end.** The screen tells the guest to "sign in as that account" and then offers only "Go to my machines". No control in the app could carry out its own instruction.
- **A borrowed browser can't be released.** Sessions run for `SESSION_TTL_SECONDS`; the only exit was clearing site cookies by hand.
- **The handle on file was invisible.** A user row with a NULL `github_login` looks identical to a correct one from the client. That is what made #170 hard to diagnose against a live relay — the account *was* right, the stored handle was missing, and nothing could show that.

## What's here

- **Account control** in the session-list header — avatar (initial as fallback) opening a sheet, following the existing `openMachineMenu` pattern: name, email, the handle the relay holds, and Sign out.
- **An explicit warning when there is no handle on file**, since that state silently breaks addressed invites and previously had no symptom the user could see.
- **Sign-out on the empty state too.** "Nothing here yet" is exactly what the wrong account looks like, so that screen needs its own way out — it has no header bar to hang the avatar on.
- **`invite_wrong_account` gets a working action** — sign out, then a fresh GitHub sign-in. Re-running OAuth alone would silently hand back the same account, so the session has to end first. The invite is re-stashed *after* the sign-out wipe so the pending link outlives the session that refused it.
- **`publicUser` returns `githubLogin`.**

Sign-out also clears `bodhi.machineId` and the invite stash — per-account state, not per-browser — and reloads rather than re-rendering, which drops the decrypted terminal buffers held in memory.

## Verification

`tsc` clean on both trees; 171 relay tests pass (168 + 3 new).

Exercised end-to-end against a local relay rather than only asserting in tests:

```
/api/me  → {"displayName":"Dev User", …, "githubLogin":"devuser"}
logout   → 204
/api/me with the same cookie → 401
```

That last line is the one that matters: the session dies server-side, so a cookie copied off the machine stops working too. Sign-out that only cleared the cookie would be decoration. I also nulled `github_login` in the local DB to confirm `/api/me` reports `githubLogin: null` — reproducing the exact state that caused #170 — so the new warning has something real to key off.

**Not visually verified.** The browser extension wasn't connected in this session, so the sheet's layout is typechecked and built but not rendered. Worth a look on the deployed relay.

## Scope

Relay + web client only. No desktop changes, so this ships on a relay redeploy — no new beta build needed.

Closes #175

### DIFF
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -413,6 +413,12 @@ roll-up, not the only route.
   gets its own reason, `login_unknown`, and copy that offers the one action that fixes it:
   sign in again, which is what puts the handle on file. Telling the person the link *was*
   addressed to that it "isn't for this account" is both wrong and a dead end.
+
+  A genuine `wrong_account` needs an escape too (#175). Re-running OAuth would hand back the
+  same session, so that refusal offers **sign out, then sign in** — and the invite is
+  re-stashed *after* the sign-out wipe so the pending link outlives the session it was
+  refused by. Without it the copy asks the guest to switch accounts in an app that, until
+  #175, had no sign-out control anywhere.
 - **What they can do?**
   - **Watch** — *"They see this session's live output. They can't type."*
   - **Watch and type** — *"They can run any command you can, read any file you can, and spend
--- a/relay/src/http.ts
+++ b/relay/src/http.ts
@@ -578,6 +578,11 @@ function publicUser(user: User) {
     displayName: user.display_name,
     email: user.primary_email,
     avatarUrl: user.avatar_url,
+    // Null is a real, reachable state: a row created before we recorded logins
+    // has none until the next sign-in backfills it. The client surfaces that
+    // rather than hiding it — it is the whole reason an addressed invite can
+    // refuse someone who IS the right person (#170).
+    githubLogin: user.github_login,
   };
 }
 
--- a/relay/src/m2.test.ts
+++ b/relay/src/m2.test.ts
@@ -295,6 +295,17 @@ describe('router: GitHub OAuth', () => {
   }
   const mockFetch = makeMockFetch();
 
+  /** Run the OAuth callback and return the issued session token. */
+  async function signIn(route: (req: Request) => Promise<Response> | Response): Promise<string> {
+    const res = await route(
+      new Request('http://relay.test/auth/github/callback?code=abc&state=xyz', {
+        headers: { cookie: 'bdl_oauth_state=xyz' },
+      }),
+    );
+    const session = res.headers.getSetCookie().find((c) => c.startsWith('bdl_session='))!;
+    return session.slice('bdl_session='.length).split(';')[0]!;
+  }
+
   test('login without OAuth config returns 503', async () => {
     const { config } = loadConfig({});
     const route = createRouter({ config, logger, repos: freshRepos() });
@@ -332,6 +343,42 @@ describe('router: GitHub OAuth', () => {
     const { user } = (await meRes.json()) as { user: { displayName: string; email: string } };
     expect(user.displayName).toBe('The Octocat');
     expect(user.email).toBe('octo@gh.com');
+  });
+
+  // The client shows this handle, and shows a warning when it is missing —
+  // an addressed invite matches on it, so an absent one is the difference
+  // between a link working and being refused (#170).
+  test('/api/me reports the GitHub handle the relay holds', async () => {
+    const { config } = loadConfig({ GITHUB_CLIENT_ID: 'id', GITHUB_CLIENT_SECRET: 'secret' });
+    const route = createRouter({ config, logger, repos: freshRepos(), fetchImpl: mockFetch });
+    const token = await signIn(route);
+
+    const meRes = await route(new Request('http://relay.test/api/me', { headers: { cookie: `bdl_session=${token}` } }));
+    const { user } = (await meRes.json()) as { user: { githubLogin: string | null } };
+    expect(user.githubLogin).toBe('octocat');
+  });
+
+  test('logout destroys the session, so the same cookie stops working', async () => {
+    const { config } = loadConfig({ GITHUB_CLIENT_ID: 'id', GITHUB_CLIENT_SECRET: 'secret' });
+    const route = createRouter({ config, logger, repos: freshRepos(), fetchImpl: mockFetch });
+    const token = await signIn(route);
+    const cookie = `bdl_session=${token}`;
+
+    const outRes = await route(new Request('http://relay.test/auth/logout', { method: 'POST', headers: { cookie } }));
+    expect(outRes.status).toBe(204);
+    expect(outRes.headers.get('set-cookie')).toContain('bdl_session=;');
+
+    // Clearing the cookie is only half of it. A cookie that was copied off the
+    // machine has to die server-side too, or "sign out" is decoration.
+    const after = await route(new Request('http://relay.test/api/me', { headers: { cookie } }));
+    expect(after.status).toBe(401);
+  });
+
+  test('logout without a session is not an error', async () => {
+    const { config } = loadConfig({ GITHUB_CLIENT_ID: 'id', GITHUB_CLIENT_SECRET: 'secret' });
+    const route = createRouter({ config, logger, repos: freshRepos() });
+    const res = await route(new Request('http://relay.test/auth/logout', { method: 'POST' }));
+    expect(res.status).toBe(204);
   });
 
   test('callback with mismatched state is rejected', async () => {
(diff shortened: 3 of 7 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped, carefully documented feature (client sign-out + account visibility) with server tests covering the new/exercised logout and `/api/me` behavior, and no blocking defects visible in the diff.

**Blocking issues**

None.

**Non-blocking issues**

- `relay/web/src/main.ts:92` (`app.user = ((await me.json().catch(() => ({}))) ...)`) — worth double-checking that the `me` response body hasn't already been consumed earlier in `boot()` before this call. If it has, `.json()` throws (body already read), which is swallowed by `.catch(() => ({}))` and silently leaves `app.user` as `null` — no crash, but the account UI would quietly never populate. Not visible in the shown hunk whether this is the first read of the body.
- `relay/web/src/main.ts` (`accountButton()`/`openAccount()`, e.g. `aria-label="Account — signed in as ${esc(name)}"` and `src="${esc(avatar)}"`) — these interpolate `esc()`-wrapped, GitHub-controlled values (`displayName`, `avatarUrl`) directly into HTML attributes. Worth confirming `esc()` escapes quote characters (not just `<`/`>`/`&`), since a GitHub display name containing a `"` could otherwise break out of the attribute. This is an existing helper reused here, not new code, so flagging as a question rather than a defect.
- `signOut()` (`relay/web/src/main.ts`) closes `app.conn` and disables the triggering button before awaiting `/auth/logout`; if that request hangs (no visible timeout in the `api()` helper from this diff), the UI sits in a disabled "Signing out…" state with no fallback. Can't confirm `api()`'s timeout behavior from this diff.

**Tests**

- Server-side coverage for the new `githubLogin` field and the `/auth/logout` behavior (session destruction, idempotency without a cookie) is solid.
- No test coverage is included for the new client-side ordering logic in `main.ts` — specifically `returnTo()` and the "re-stash the invite *after* the sign-out wipe" sequencing in `signOut()`, which the PR's own comments flag as a subtle, easy-to-get-backwards invariant (stash-then-wipe would silently drop the pending invite). If there's a harness for exercising `relay/web/src/main.ts` in isolation, a regression test for this ordering would be valuable; if there isn't one yet, at minimum it's worth being aware this path is only covered by manual/visual verification today.

William-Long-II:
Thanks — one of these was a real defect, so all three are addressed in a173eda.

**`esc()` in attribute context — confirmed, and fixed.** You were right to ask. `esc()` sets `textContent` and reads back `innerHTML`, which serializes a *text node*: per spec that escapes `&`, `<`, `>` and nbsp, and deliberately leaves quotes alone. Safe between tags, unsafe inside a quoted attribute.

The sharpest of the three I added is the avatar rather than the display name: `safeAvatar()` only checks the `https://` prefix, so `https://x" onerror="…` passes the guard and then breaks out of `src`. New `escAttr()` covers quotes, and I moved the one pre-existing case over with them — the session name in an aria-label at `main.ts:590`, which is authored on the owner's desktop.

**`/api/me` body — not consumed.** `me.ok` is a property, not a body read, and nothing else touches `me` between the fetch and `app.user`. That `.catch(() => ({}))` is there for a malformed body, not a double read.

**Hung logout — bounded at 5s.** Agreed, and `api()` has no timeout of its own. The alternative to a slow relay answering isn't waiting longer, it's a disabled button that never comes back; where we go next doesn't depend on the reply either way.

**Ordering test — there was no harness, so I built the smallest one that helps.** main.ts imports xterm and calls `boot()` at import time, so nothing in it is reachable from a test. The wipe now lives in `web/src/account.ts` with five tests.

Worth noting the mutation check, since a test of ordering is easy to write in a way that passes either way: inverting the two lines to stash-then-wipe still passes three of the five, and fails exactly the two that describe someone switching accounts to accept a link. I ran that edit to confirm the suite goes red rather than assuming it would.

176 relay tests pass; `tsc` clean on both trees.

brannon-bowden (approved):
Approve — this update addresses all three non-blocking concerns from the previous review and adds solid unit-test coverage for the trickiest piece of logic in the feature (the invite-stash-after-wipe ordering).

**What changed since the last review**

- `esc()`'s quote-escaping gap is fixed: a new `escAttr()` correctly layers `&quot;`/`&#39;` escaping on top of `esc()`'s existing `&`/`<`/`>` handling, and it's now consistently applied everywhere GitHub-controlled values (`displayName`, `avatarUrl`) are interpolated into HTML attributes (`main.ts:598`, `main.ts:998`, `main.ts:1001`, `main.ts:1019`).
- `signOut()`'s unbounded hang is fixed with a `Promise.race` against a 5s timeout (`main.ts:1056-1059`), so a slow `/auth/logout` no longer leaves the UI stuck on a disabled "Signing out…" button.
- `clearAccountState()` and its ordering invariant (invite re-stashed *after* the wipe, never before) are extracted to `account.ts` and now have direct unit tests (`account.test.ts`), including a test that specifically pins the invariant a naive stash-then-wipe implementation would violate.

- The double-read concern about `me.json()` at the old `main.ts:92` location (now shifted a few lines by this diff) was never confirmed or touched by these changes — still worth a quick check that the `/api/me` response body isn't consumed earlier in `boot()`, since a second `.json()` call would throw and be silently swallowed by `.catch(() => ({}))`.
- `main.ts` around the removed `INVITE_STASH` declaration (~line 123) now has two consecutive blank lines where the constant used to sit — harmless, but worth a quick cleanup pass.

William-Long-II:
Blank line cleaned up in the latest commit.

On the `me.json()` question — I did check it, and it is a single read. `me.ok` is a property access, not a body read, and nothing between the fetch and `app.user` touches `me`. The `.catch(() => ({}))` is there for a malformed body rather than a double read. (I posted this in the comment above; noting it here since the review read as still-open.)

brannon-bowden (approved):
Approve — this change is a trivial whitespace cleanup (removing an extra blank line left behind after the `INVITE_STASH` constant was removed), which is exactly the minor cleanup noted as a non-blocking suggestion in the previous review. No behavioral change, nothing new to flag, and CI is green (SonarQube, quality-gate, GitGuardian all passing).

The other outstanding suggestion from before — confirming `/api/me`'s response body isn't read twice via `me.json()` in `boot()` — remains unaddressed but is unaffected by this diff and was never blocking; still worth a quick check by the author when convenient.

Issue #183, opened 2026-08-24
Feature: Web push — a locked phone learns a session needs you


William-Long-II:
### User story / problem

As an owner away from my desk, I want my locked phone to get a push notification the moment a session enters `waiting` or `error`, so I can respond from anywhere without keeping the tab open.

The client's whole attention model ("needs you" sorting, count chips, banner) is driven by a 2.5s poll that dies with the tab. The plumbing already exists in pieces: `push_subscriptions` shipped in relay migration 001 (unused), `VAPID_SUBJECT` is parsed into relay config and read by nothing, and the desktop app has a complete VAPID + dispatcher web-push stack built for the legacy LAN PWA that can be ported.

### Proposed solution

- Relay: VAPID keypair in config, `GET /api/push/vapid-key`, subscribe/unsubscribe routes persisting to `push_subscriptions`, a dispatcher that sends on agent-forwarded attention events, 410-gone reaping, rate limits on the new routes.
- Client: subscription flow gated on the service worker from the installable-PWA issue, a per-machine or global notifications toggle, notification click opens the right machine (and session where the payload allows).
- Agent: on session state change to `waiting`/`error`, emit an attention event to the relay.
- **Zero-knowledge constraint (decide explicitly, record in the design):** session names and terminal content are E2E today — the relay must not learn them from push. Either (a) payloads carry only relay-visible metadata ("A session needs you on *machine-name*"), or (b) the agent performs the web-push `aes128gcm` encryption itself against the subscription keys and the relay blind-forwards to the push endpoint, keeping content E2E. Option (b) preserves the product's core promise; cost it before defaulting to (a).

### Acceptance criteria

- [ ] Tab closed, phone locked: a session entering `waiting` produces a push within seconds; tapping it opens the client on the right machine
- [ ] The relay never handles plaintext session names or terminal content in the push path; the chosen payload model is documented in `docs/designs/`
- [ ] Subscriptions can be turned on and off from the client; a revoked/expired browser subscription is reaped on 410
- [ ] `waiting` debounce respects the same event semantics the desktop notifier uses (no push storm from a flapping state)
- [ ] New relay routes are rate-limited and covered by tests

### Scope

Major feature

### Priority

Highest

Part of #177. Depends on the installable-PWA issue (service worker).

William-Long-II:
Carry-over notes from the #182 review (PR #188, approved) — these land here because #183 modifies the same worker:

[...]

Issue #177, opened 2026-08-24
Epic: Installable PWA + web push — the web client taps you on the shoulder


William-Long-II:
### Summary / goal

Make the relay web client an installable PWA with end-to-end web push, so attention events reach the phone when the tab is closed — and close the guest-UX and accessibility gaps the session-sharing design already specifies.

Today the client's "needs you" model (the organizing principle of its UI) is driven by a 2.5-second `sessions:list` poll that dies the moment the tab closes or the phone locks. There is no manifest, no service worker, no push — so the marquee promise of the relay ("know when Claude needs you, from anywhere") only holds while you're actively staring at the page.

### Why / business value

The web client is the surface testers judge; connectivity and E2E encryption are invisible to them. Every piece of this already exists somewhere: the desktop app has a working VAPID + dispatcher web-push stack (built for the legacy LAN PWA), the relay schema shipped `push_subscriptions` in migration 001, `VAPID_SUBJECT` is already parsed into relay config, and the legacy PWA has a reference service-worker + install-prompt flow. This is assembly, not invention.

### Child issues / scope

- [ ] #182 — Installable PWA: manifest, icons, service worker served from the relay origin
- [ ] #183 — Web push end to end: subscribe flow, relay routes + VAPID, agent-triggered pings on waiting/error (zero-knowledge payload model decided explicitly)
- [ ] #184 — Owner surfaces in the browser: list and revoke guests from a phone
- [ ] #185 — Guest arrival and sizing polish: single-grant lands in the terminal, guest-voice copy, role words, "Fit to my screen"
- [ ] #186 — Fingerprint provenance: wire the never-called check, 4-word phrase, honest provenance copy
- [ ] #187 — Sheet accessibility: focus traps, Escape, `aria-live`, retire `window.prompt()`

### Definition of done

A phone with the PWA installed and locked receives a push notification when a session enters `waiting` or `error`; tapping it opens that session. An owner can see and revoke guests from a browser. A guest with one grant lands in the terminal. The sheets pass a keyboard-only walkthrough. No terminal content ever appears in a push payload.

### Priority

Highest

PR #188, opened 2026-08-24
182 — Make the relay web client an installable PWA

### LINKED ISSUE #182: Feature: Make the relay web client an installable PWA
### User story / problem

As a Bodhilander user on my phone, I want the relay web client installed on my home screen, launching standalone with its own icon, so that reaching my sessions is one tap instead of browser → tab → bookmark — and so the client has the service worker that web push (the follow-on issue) requires.

Today the client has no manifest, no service worker, and no icons. `relay/src/web.ts` serves exactly five paths (`/`, `/i`, `/i/*`, `/app/main.js`, `/app/main.css`) — there is no route a manifest or root-scoped `sw.js` could even be served from. A cold load with no network shows the browser's offline page.

### Proposed solution

- `manifest.webmanifest` + icon set (192/512 + maskable + `apple-touch-icon`) served by the relay; `display: standalone`, theme/background colors matching the client's existing light/dark `theme-color` metas.
- A deliberately minimal root-scoped service worker: precache the app shell (`index.html`, `main.js`, `main.css`), network-first for everything else, and a branded offline fallback ("Can't reach the relay") instead of the dinosaur. No API caching — the terminal is live data; the SW exists for installability, the offline shell, and (next issue) push.
- Registration in `main.ts`; an update strategy that doesn't yank an open terminal session mid-use (decide and document skipWaiting/clients.claim behavior).
- New routes in `relay/src/web.ts` with correct content types; keep the path allow-list discipline that file already has.

### Acceptance criteria

- [ ] Chrome on Android offers install; iOS Add to Home Screen produces a standalone app with the right icon and theme colors in both light and dark
- [ ] A cold load with no network shows the branded offline page, not the browser error page
- [ ] A service-worker update activates without breaking a live terminal session; the chosen strategy is documented in the code
- [ ] Lighthouse installability checks pass against the deployed relay
- [ ] Web tests cover the new `web.ts` routes (the existing five-path tests extend rather than loosen)

### Scope

New feature

### Priority

Highest

Part of #177.

William-Long-II:
## Description
Makes the relay web client an installable PWA.

New:
- `relay/web/manifest.webmanifest` — standalone display, root scope, dark theme pair `#0c0f14`
- Icon set derived from `build/icon.png`: 192/512 plain + maskable variants composited onto `#0c0f14`, and a flattened 180×180 apple-touch-icon
- `relay/web/offline.html` — self-contained branded offline page (light + dark)
- `relay/web/sw.js` — hand-written, root-scoped service worker: precaches the shell, network-first, offline fallback for navigations only; never touches `/api`, `/auth`, `/ws`, non-GET, or cross-origin requests
- `relay/web/src/sw.test.ts`

Changed:
- `relay/src/web.ts` — eight new enumerated allow-listed routes, no directory serving
- `relay/src/relay.test.ts` — +8 route tests
- `relay/web/index.html` — manifest/icon/apple metas
- `relay/web/src/main.ts` — service worker registration

## Related issue
Closes #182 (child of epic #177)

## Type of change
- [x] New feature

## How was this tested?
- `bun test` in `relay/`: 192 pass, 0 fail, 2495 expects across 10 files (16 new tests)
- `typecheck` and `typecheck:web`: clean
- Live smoke: all 8 new routes return 200 with the correct MIME type; sibling paths 404
- Harness `verify.sh` at `feat/182-Bodhilander`: GREEN — with the caveat that the registry has `test`/`lint` as null for this repo, so those rows are SKIPs. The test evidence above is the owner's direct `bun test` run, not the harness.

Not verified: real-device home-screen install flows (Android/iOS), and push delivery — push handlers are #183 and not part of this change.

## For review attention
- SW update strategy is no-skipWaiting / no-claim: an update never swaps the controlling worker under a live terminal; it takes effect after the app fully closes once. Pinned by a behavioral test.
- The precached `/` is currently never served — offline navigations get `offline.html`. Kept for the push work in #183.
- The manifest carries the dark color pair only; per-scheme `theme-color` metas in `index.html` do the rest.
- `/sw.js` is served no-cache at root scope so push handlers (#183) can land without route changes.

## Checklist
- [x] Tests pass locally

## Gate results

**Gate 3 — reviewer: GREEN.** Operator path walked against a live relay on a scratch port: all 8 routes verified over real HTTP (status, MIME, cache-control); icons decoded pixel-level (opaque apple-touch-icon, correct maskable safe zone); Dockerfile ships the new assets unchanged. Mutation testing 5/5 killed — cross-origin guard, skipWaiting seizure, offline fallback key, sw.js MIME, allow-list widening — each mutation failed a test. Non-blocking notes:
- Cache-name bump discipline is undocumented in `sw.js` — hand-off note for #183, whose owner must bump `bodhi-shell-v1` when changing the worker.
- Manifest colors are dark-only — an inherent single-value manifest limitation; in-app chrome is covered by the dual `theme-color` metas.
- One residual AC needs a device: a manual phone pass against the deployed relay after merge (Chrome install prompt / iOS add-to-home-screen rendering).

**Gate 4 — verifier: VERIFIED.** Suite counts reproduced exactly (192 pass / 0 fail / 2495 expects, no flake); both typechecks clean with file coverage confirmed; icons byte-real at claimed dimensions; allow-list enumeration confirmed; zero deletions in test files; seams receipt on record. Caveats, stated honestly: `verify.sh` test/lint rows are registry SKIPs (the evidence is the direct runs), and web test files are executed by bun but excluded from the web typecheck (pre-existing convention).

CI: GitGuardian, SonarQube, and quality-gate all pass. `verify-merge-order.sh` released the draft itself (0 failures).

Order: Bodhilander


brannon-bowden (approved):
This is a clean, well-scoped PWA-installability change — approve. The service worker's install/activate/fetch logic is careful about scope (no `skipWaiting`/`clients.claim` so a live terminal session is never yanked out from under a page), the static-asset allow-list is exact-match (no directory traversal risk), and the test coverage (both the router-level asset tests and the mocked service-worker unit tests) is unusually thorough for a feature like this.

CI is green across all 3 checks (SonarQube, quality-gate, GitGuardian).

## Suggestions (non-blocking)

- `relay/web/sw.js` (the runtime cache-and-refresh branch, `fetch(req).then((res) => {... cache.put ...})`) caches the response unconditionally without checking `res.ok`. A transient 5xx/404 for a shell asset would get written into the cache and could be served back on a later offline fallback. Consider only caching on `res.ok`.
- `relay/web/offline.html:28` uses an inline `onclick="location.reload()"` handler. If the relay ever sets a `Content-Security-Policy` with a `script-src` that excludes `'unsafe-inline'` (not visible in this diff), the button would silently stop working. Worth confirming against whatever CSP policy (if any) applies to relay responses.
- `relay/web/src/main.ts:1154` swallows `serviceWorker.register()` failures with `.catch(() => {})` and no logging — reasonable given the "no update prompts" design goal, but it means a genuinely broken `sw.js` (e.g. a syntax error) fails with zero observability.
- The new tests confirm neighboring static paths (`/icons/icon-999.png`, `/manifest.json`, etc.) still 404, which is good, but there's no test exercising a dynamic SPA route like `/i/:code` after the `pwaAssets` lookup was spliced into the routing chain — the ordering in `web.ts` looks correct by inspection, but an explicit regression test would make that guarantee durable.

Issue #187, opened 2026-08-24
Task: Web client sheet accessibility — focus, Escape, live announcements


William-Long-II:
### What needs doing

An accessibility pass on the relay web client's six bottom sheets and its live-attention signals, bringing the dialogs up to the standard the watch-only work already set (that branch shipped real keyboard scrolling, `aria-readonly`, and focus-visible states — the sheets didn't get the same care).

- Focus management on all six sheets (fingerprint, create session, new group, account, machine switcher, link machine): move focus in on open, trap it while open, restore it on close.
- Escape closes any open sheet/scrim (today it's mouse or the Back gesture only).
- The app behind an open sheet is `inert`/`aria-hidden`.
- `aria-live` announcement for attention changes — the "N needs you" count currently updates silently.
- Replace the `window.prompt()` invite-code entry on the guest empty state with a proper labelled input.
- Keep the existing wins intact: 44×44 targets, no meaning carried by `--faint` text, `prefers-reduced-motion`.

### Context / why

The web client is the surface testers judge, and accessibility is a standing first-class requirement for it. The sheets have good ARIA attributes (`role="dialog"`, `aria-modal`, `aria-labelledby`) but none of the behavior those attributes promise, which is worse than absent for screen-reader users — the semantics claim a modal that doesn't act like one.

### Done when

- [ ] A keyboard-only walkthrough can open, operate, and close every sheet, ending focus where it started
- [ ] VoiceOver on iOS announces sheet open/close and attention-count changes
- [ ] No `window.prompt()` remains
- [ ] Escape dismisses every sheet
- [ ] Web tests cover the focus trap and restore behavior

### Priority

Medium

Part of #177.

William-Long-II:
Inventory update from #184 (PR #189): the client now has **eight** sheets, not six — #184 adds the owner "Sharing" sheet and the guest "Your access" sheet — plus two `window.confirm()` calls for destructive revoke/leave confirmation, which belong in this pass alongside the `window.prompt()` retirement. The sheets' close buttons also reuse the pre-existing 40×40 `.iconbtn` (below the 44×44 floor) — flagged by the #184 reviewer as this issue's territory.

PR #189, opened 2026-08-24
184 — Owners can see and revoke shares from the browser

### LINKED ISSUE #184: Feature: Owners can see and revoke shares from the browser
### User story / problem

As a machine owner on my phone, I want to see who has access to my sessions and revoke it, so that managing sharing doesn't require walking back to the desktop.

The server side already exists with no browser caller: `GET /api/machines/:id/shares` and `DELETE /api/shares/:grantId` (owner or grantee) are live routes, and the web client's entire API footprint today is config/me/machines/redeem/claim/logout/dev-login. An owner who shared a session from the desktop and then left is blind and powerless from the browser.

### Proposed solution

A "Sharing" sheet on the machine (owner relation only): each grant as a row — person (GitHub login), the role word (Watching / Watching and typing), expiry ("until you revoke it" for the 0-TTL grants), pending vs active — with a revoke action per row. Revocation copy stays honest about the disconnected-agent case the desktop UI already handles ("takes effect when this machine reconnects"). Guests get the mirror view for their own grants (`GET /api/shares`) with a leave action.

### Acceptance criteria

- [ ] An owner can list every grant and pending invite for their machine from the web client and revoke any of them
- [ ] A revoked live guest is disconnected and sees the revoked ending state (existing sealed-denied path); a revoke against an offline agent shows the queued-revocation copy
- [ ] A guest can see their own grants and walk away from one
- [ ] Role words only — no `viewer`/`operator`/`grantee` strings reach the UI
- [ ] Sheet meets the client's dialog conventions (labelled, keyboard reachable) and is covered by web tests

### Scope

Minor feature

### Priority

High

Part of #177.

William-Long-II:
## Description

Owners of the machine they are viewing get a **"Sharing" sheet** (button beside the machine pill) listing everything granted on that machine:

- rows for **active grants**, **pending requests**, and **invite links** — the person, a role word ("Watching" / "Watching and typing"), pending vs active, and expiry ("Until you revoke it" for the never-expires sentinel);
- actions per kind: **Revoke** (`DELETE /api/shares/:grantId`), **Refuse** for a pending request, **Cancel invite** (`DELETE /api/machines/:id/shares/:inviteId`);
- after-revoke copy distinguishes a connected machine from an offline one, driven by the `agent:offline` / `agent:ready` signals.

Guests get a **"Your access"** sheet (`GET /api/shares`) with a **Leave** action. Leaving the share you are connected through tears down the channel first, so the resulting kick cannot render a false "your access was ended" story.

**Fix commit `2cb6d24`** (landed after gate 3's first pass, in response to its one must-fix):

- Close-derived endings now emit a `CONNECTION_ENDED` sentinel rendered as **neutral, non-attributed copy** — the client never names a person for an ending it learned from an unsealed relay-authored close. `ENDED_COPY` is extracted to `relay/web/src/ended.ts` and pinned in both directions by `ended.test.ts`.
- Relay-side, a reconnecting revoked/expired guest is now closed with **their own grant's ending reason** instead of `'not your machine'`, ending the silent reconnect loop (`endedGrantReason` in `repositories.ts`, two new `ws.test.ts` tests). The reason literals are pinned in `ws.test.ts`.

Files:

- `relay/web/src/shares.ts` — new pure module: types, `roleWord`, `timeLeft`, row builders, copy — with `shares.test.ts` (26 tests, including a forbidden-vocabulary sweep over the user-facing copy)
- `relay/web/src/ended.ts` — new: `ENDED_COPY`, neutral close-derived ending copy — with `ended.test.ts` (4 tests)
- `relay/web/src/main.ts` — both sheets, the existing dialog/aria/`pushLayer` pattern, `machineOffline` tracking
- `relay/web/src/connection.ts` — a 4403 close carrying a known ending reason maps to the denied path (`connection.test.ts`: 16 tests)
- `relay/src/ws.ts`, `relay/src/repositories.ts` — reconnect close reasons for ended grants (`ws.test.ts`: 31 tests)
- `relay/web/src/styles.css` — token-only additions, 44px touch targets

**Merge note:** PR #188 (approved, unmerged) touches neighboring files in `relay/web`, and this branch was cut from `development` before it. A trivial merge-forward may be needed at merge time. Merges here are true merge commits — do not squash.

### For review attention

1. **The 4403 relay-authored close → ending-state mapping** in `connection.ts`. Rationale: an HTTP revoke makes the relay kick the socket before the agent can send the sealed denied frame; the mapping is scoped to known ending reasons, and other 4403 codes keep the reconnect behavior. Reviewed against the sealed-denied rationale in `docs/designs/session-sharing.md` §5 — see Gate results below.
2. **Revoking while the agent is offline leaves a desktop ghost** until reconnect reconciliation; the after-revoke copy says so explicitly.
3. `confirm()` is used for the destructive step, matching the existing client precedent.
4. Refusing a pending request leaves the waiting guest polling. That behavior exists on `development` today and this change does not alter it.
5. The sheet shows **no presence** (who is connected right now): the relay has no existing data path for it — the `ws.ts` clients map has no HTTP accessor — and this change does not invent a relay seam to add one. That decision is recorded on issue #184.

## Related issue
Closes #184

Initiative: 184 (single-repo; merge position 1 of 1).

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?

Verified (commands run in `relay/` at head `2cb6d24`):

- `bun test` — **211 pass / 0 fail (2512 expects, 11 files)**: `shares.test.ts` 26, `connection.test.ts` 16, `ended.test.ts` 4, `ws.test.ts` 31, plus the pre-existing suites
- typecheck — clean. Web test files are executed by bun but excluded from the web typecheck (pre-existing convention), so "typechecks clean" covers sources, not tests.
- `build:web` — bundles successfully
- harness `verify.sh Bodhilander=feat/184-Bodhilander` — **GREEN**. The registry test/lint rows are registry SKIPs for this repo, so the direct runs above are the test evidence, not the harness rows.

Not verified:

- No live end-to-end pass against a running relay + agent; the offline-agent, kick, and reconnect-close behaviors are covered at the unit level via the `agent:offline` / `agent:ready` and close-code paths, not against a live machine.

## Checklist
- [x] Tests pass locally
- [x] Docs updated where needed (no doc changes required)

## Gate results

**Gate 3 — reviewer: GREEN** (first pass RED, re-check of `2cb6d24` GREEN). The one must-fix: person-attributed copy was reachable from an unsealed relay-authored close — including an honest-path case where a guest's own Leave on one device was misattributed to the owner on their other device. The fix is complete and mutation-proven (reverting the `ws.ts` change fails exactly the two new tests), with no assertions weakened.

**Security review: no blocking findings.** The 4403 mapping is acceptable as scoped, with two standing conditions now encoded in the code itself: ending copy never interpolates a name, and the trusted-close set is pinned to `ENDING_REASONS`. Authorization, IDOR, CSRF, and escaping all clean. No existence oracle in `endedGrantReason`: it reads only the caller's own grant history, and a stranger's answer is byte-identical to a nonexistent machine's.

**Gate 4 — verifier: VERIFIED.** Receipts reproduced; wire/route shapes and reason literals hand-traced to server code. Non-blocking items recorded on issue #187 (sheet inventory, `confirm()` calls, 40×40 close buttons), plus one test gap noted there: a stranger-beside-a-revoked-guest fixture would pin the `grantee_user_id` filter in `endedGrantReason`.

CI: GitGuardian, SonarQube, and quality-gate all pass at `2cb6d24`. `verify-merge-order.sh` released the draft itself.

Order: Bodhilander

### DIFF
--- a/relay/src/repositories.ts
+++ b/relay/src/repositories.ts
@@ -108,6 +108,13 @@ export interface Repositories {
    */
   getMachineAccess(machineId: string, userId: string): MachineAccess;
 
+  /**
+   * Why a former guest no longer gets in: the ending their most recent grant
+   * met. Null for someone who never held one — a stranger must stay
+   * indistinguishable from one asking about a machine that does not exist.
+   */
+  endedGrantReason(machineId: string, userId: string): 'revoked' | 'expired' | null;
+
   // --- sharing (M5.2) ---
 
   createShareInvite(input: CreateShareInviteInput): { invite: ShareInvite; code: string };
@@ -354,6 +361,23 @@ export function createRepositories(db: RelayDb, now: () => number = Date.now): R
 
       if (!grant) return { relation: 'none' };
       return { relation: 'grantee', machine, grant };
+    },
+
+    endedGrantReason(machineId, userId) {
+      // The most recent grant tells the story; the reaper purging dead rows
+      // bounds how long it can be told, which is acceptable — after that the
+      // caller falls back to the stranger answer.
+      const grant = db
+        .query(
+          `SELECT * FROM machine_grants
+            WHERE machine_id = ? AND grantee_user_id = ?
+            ORDER BY created_at DESC LIMIT 1`,
+        )
+        .get(machineId, userId) as MachineGrant | null;
+      if (!grant) return null;
+      if (grant.status === 'revoked') return 'revoked';
+      if (grant.expires_at !== null && grant.expires_at <= now()) return 'expired';
+      return null;
     },
 
     createShareInvite(input) {
--- a/relay/src/ws.test.ts
+++ b/relay/src/ws.test.ts
@@ -48,8 +48,8 @@ function connect(url: string, headers?: Record<string, string>) {
     ws.onopen = () => res();
     ws.onerror = () => rej(new Error('ws error'));
   });
-  const closed = new Promise<{ code: number }>((res) => {
-    ws.addEventListener('close', (e) => res({ code: (e as CloseEvent).code }));
+  const closed = new Promise<{ code: number; reason: string }>((res) => {
+    ws.addEventListener('close', (e) => res({ code: (e as CloseEvent).code, reason: (e as CloseEvent).reason }));
   });
   const next = () =>
     new Promise<any>((res) => {
@@ -286,7 +286,11 @@ describe('client ↔ agent brokering (M3)', () => {
       const client = connect(`ws://127.0.0.1:${server.port}/ws/client`, { cookie: `bdl_session=${token}` });
       await client.opened;
       client.ws.send(JSON.stringify({ type: 'client:open', machineId }));
-      expect((await client.closed).code).toBe(4403);
+      const closed = await client.closed;
+      expect(closed.code).toBe(4403);
+      // A stranger gets the unrevealing answer — never a grant's ending,
+      // which would confirm a share existed on a machine that is not theirs.
+      expect(closed.reason).toBe('not your machine');
     } finally {
       server.stop(true);
     }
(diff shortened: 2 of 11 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped, thoughtfully tested change: the privacy-preserving `endedGrantReason` lookup, the client-side mapping of sealed 4403 reasons to a terminal "denied" state, and the new owner/guest sharing sheets are all backed by targeted unit tests, and the copy/logic extraction into `ended.ts`/`shares.ts` keeps the new business logic testable outside of `main.ts`'s import-time `boot()`.

## Issues
None blocking. CI is green across SonarQube, quality-gate, and GitGuardian.

## Suggestions

- `relay/web/src/shares.ts:55` — `UNTIL_REVOKED_HORIZON_MS` (1 year) is used to infer the "until revoked" sentinel from any expiry far enough out. It's a documented trade-off, but it's a magic-number heuristic that will silently misrender if certificate lifetimes are ever extended past a year without updating this constant. Consider having the server send an explicit `neverExpires`/sentinel flag instead of asking the client to guess from the timestamp's magnitude, to remove the coupling.
- `relay/web/src/main.ts` (`revokeOwnerRow`, `leaveShareRow`, `openSharing`, `openMyShares`) — the DOM wiring around confirm/delete/reload (including the `machineOffline` flag and the "leaving your own connected share" teardown-then-redirect path) isn't covered by any test, only the pure `shares.ts` helpers are. This mirrors the existing pattern in the file (logic extracted for testability, glue left untested), so it's not a regression, but it's worth a lightweight integration/DOM test given the non-trivial ordering (`stopPolling`/`reconnector.cancel`/`app.conn?.close()` before the DELETE, then either `location.href` or `connect()` depending on outcome).
- Not verifiable from this diff, so raised as a question rather than a blocker: `revokeOwnerRow` and `leaveShareRow` both call `DELETE /api/shares/:id` for grants. Please confirm the server-side handler enforces that only the machine owner (for a revoke) or the grantee themselves (for a leave) can delete a given share id — that authorization logic isn't part of this diff.
- `relay/src/repositories.ts`'s new `endedGrantReason` and the existing `getMachineAccess` are two separate queries against `machine_grants`; under concurrent revoke/expiry they could theoretically disagree by the time the close reason is computed. Worst case is a slightly wrong close-reason string (e.g. falling back to `'not your machine'`), not an access-control issue, so this is informational only.

William-Long-II:
Re the authorization question on `DELETE /api/shares/:grantId`: confirmed enforced server-side, independent of the client. `handleRevokeGrant` (`relay/src/http.ts:508-525`) requires a session (401), then computes owner-or-grantee from `machine.user_id` / `grant.grantee_user_id` server-side; any other caller gets the identical 404 a nonexistent grant produces, so there's no existence oracle. The invite route is owner-only with the same uniform 404, and its SQL is additionally machine-scoped (`WHERE id = ? AND machine_id = ?`, `repositories.ts:405-412`). Ids are `crypto.randomUUID()`. This was independently checked by a security review pass before the PR went ready — the client's `isGuest()` branching is display-only; every route re-checks.

PR #190, opened 2026-08-24
168 — setDefaultAccount with a stale id no longer demotes the incumbent to zero defaults

### LINKED ISSUE #168: Bug: setDefaultAccount with a stale id demotes the incumbent and promotes nobody, leaving zero defaults
Found while adding test coverage for `repositories/accounts.ts` in #166; predates that branch.

`setDefaultAccount(id)` (`src/main/repositories/accounts.ts:96-103`) demotes the incumbent **unconditionally**, then promotes by id:

```sql
UPDATE claude_accounts SET is_default = 0;          -- unconditional
UPDATE claude_accounts SET is_default = 1 WHERE id = ?;
```

If `id` matches no row, the second statement affects nothing and the app is left with **zero default accounts** — inside a transaction that commits happily.

That is the exact accountless state the `deleteAccount` promotion block added in #165 exists to prevent, and its docstring describes why it matters: every session with no override silently falls back to the legacy `~/.claude` login, and the session header can name neither a live account nor an assigned one. The user is told nothing; their account setup just quietly stops working.

**Reachable how:** the renderer holds a cached `claudeAccounts` list (`App.tsx`). Clicking "Make default" on a row whose account was deleted in another window — or after any staleness between that cache and the database — sends an id that no longer exists.

**Current status:** pinned as-is by the test `setDefaultAccount > pointing at an unknown id leaves the app with no default` in `src/main/repositories/__tests__/accounts.test.ts`. Its comment says explicitly that it documents rather than endorses the behaviour, so tightening this will fail that test loudly rather than changing behaviour silently.

**Suggested fix:** make the promotion conditional on the id existing — check first inside the transaction and no-op (or throw) when it does not, so the demote never lands without a matching promote. Then update that test to assert the new guarantee.

William-Long-II:
## Description
`setDefaultAccount` demoted the incumbent default unconditionally before promoting by id. When the id was stale — e.g. a renderer's cached list offering an account another window had already deleted — the transaction committed with **zero** defaults.

The transaction is now guarded on the target row existing, and the contract is `setDefaultAccount(id): boolean`:

- `true` — the row exists and is the default (idempotent re-default included)
- `false` — no row matched; nothing changed

No-op-with-signal was chosen over a throw to match this repository's pinned voice — `updateAccount`, `touchAccount` and `deleteAccount` all no-op on unknown ids — and because the modal's unconditional `refresh()` already reconciles the ghost row out of the list.

The full caller chain carries the new contract: the IPC handler returns the boolean (`src/main/index.ts`), and the preload + `electron.d.ts` signatures are now `Promise<boolean>`. The boolean is **available to** the renderer through the full IPC chain — its one caller (`ClaudeAccountsModal`) deliberately discards it and relies on `refresh()`; the renderer does not consume the value today.

Commit d209938 (a verifier finding): the panel test stub now answers `async () => true` to match the `Promise<boolean>` contract; a sweep confirmed no other stub anywhere misstates it.

Initiative 168.

### For review attention
1. The modal discarding the returned boolean is deliberate (see above), not an oversight.
2. One comment in a pinned `deleteAccount` test was reworded — it cited the now-fixed behavior as a route to the no-default state. Assertions are unchanged.

## Related issue
Closes #168

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
Evidence at head d209938:
- `bun test` (full) — 872 pass / 0 fail, 60 files (3902 expects)
- `bun test src/renderer` — 287 pass / 0 fail
- `bun test src/main/repositories` — 33 pass / 0 fail
- `tsc` — main and renderer both clean
- Harness `verify.sh` — GREEN (the registry test/lint rows are SKIPs for this repo; the evidence is the direct runs above)

**Not verified:** no manual multi-window run reproducing the original race (stale renderer cache offering a deleted account); the stale-id path is covered by the repository unit tests, not an end-to-end exercise.

## Checklist
- [x] Tests pass locally

## Gate results

**Gate 3 — reviewer: GREEN.** The only operator surface (the panel's "Make default") can no longer produce a click that appears to succeed while leaving zero defaults. Transaction guard confirmed atomic-in-effect (synchronous better-sqlite3, single-instance lock = one writer); mutation test: deleting the guard fails exactly the stale-id and empty-table tests. All other zero-default paths walked — creation, `deleteAccount` promotion, account-switch, resolver legacy fallback — none needed this guard. Non-blocking note: the ghost-click gives no message by design (`refresh()` removes the row); the boolean makes a toast a one-line change if ever wanted.

**Gate 4 — verifier: VERIFIED.** All counts reproduced exactly against a real in-memory DB with the single-default partial unique index; no global mocks in play; no pre-existing assertion weakened (the one changed assertion was the bug's own pin, replaced by the fix's pin). Commit d209938 closed the verifier's type-stale-mock finding.

CI: GitGuardian, SonarQube, and quality-gate all pass on d209938.

Order: Bodhilander

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -838,7 +838,7 @@ safeHandle('accounts:update', (
 });
 
 safeHandle('accounts:setDefault', (id: string) => {
-  accountsRepo.setDefaultAccount(id);
+  return accountsRepo.setDefaultAccount(id);
 });
 
 // Assigning an account is routed through account-switch (not db:sessions:update)
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -342,7 +342,7 @@ contextBridge.exposeInMainWorld('electronAPI', {
     ipcRenderer.invoke('accounts:delete', id),
   updateAccount: (id: string, updates: { label?: string; color?: string; email?: string | null }): Promise<void> =>
     ipcRenderer.invoke('accounts:update', id, updates),
-  setDefaultAccount: (id: string): Promise<void> =>
+  setDefaultAccount: (id: string): Promise<boolean> =>
     ipcRenderer.invoke('accounts:setDefault', id),
   assignAccountToSession: (sessionId: string, accountId: string | null): Promise<AccountSwitchResult> =>
     ipcRenderer.invoke('accounts:assignToSession', sessionId, accountId),
--- a/src/main/repositories/__tests__/accounts.test.ts
+++ b/src/main/repositories/__tests__/accounts.test.ts
@@ -217,32 +217,37 @@ describe('setDefaultAccount', () => {
     seedAccount('a', '2026-01-01T00:00:00.000Z', { isDefault: true });
     seedAccount('b', '2026-02-01T00:00:00.000Z');
 
-    accountsRepo.setDefaultAccount('b');
+    expect(accountsRepo.setDefaultAccount('b')).toBe(true);
 
     expect(defaultIds()).toEqual(['b']);
     expect(accountsRepo.getDefaultAccount()?.id).toBe('b');
   });
 
-  test('re-defaulting the current default is idempotent', () => {
+  test('re-defaulting the current default is idempotent, and still reports true', () => {
     seedAccount('a', '2026-01-01T00:00:00.000Z', { isDefault: true });
 
-    accountsRepo.setDefaultAccount('a');
+    expect(accountsRepo.setDefaultAccount('a')).toBe(true);
 
     expect(defaultIds()).toEqual(['a']);
   });
 
-  test('pointing at an unknown id leaves the app with no default', () => {
-    // Pinning current behaviour, not endorsing it: the demote runs
-    // unconditionally, so a stale id demotes the incumbent and promotes
-    // nobody. That is the same accountless state #165 just closed on the
-    // delete path. The window is narrow — the renderer only offers ids it read
-    // out of this table moments earlier — so this test exists to make the
-    // behaviour visible and to fail loudly if someone tightens it.
+  test('pointing at an unknown id leaves the incumbent default and reports false', () => {
+    // The renderer offers ids from a cached list, so it can ask for an account
+    // another window already deleted. The demote must not land without a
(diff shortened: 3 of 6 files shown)


brannon-bowden (approved):
This is a clean, well-tested fix: `setDefaultAccount` now checks the target row exists *before* demoting the incumbent, all inside the same `db.transaction`, so a stale id can no longer strand the app with zero defaults. The boolean return threads correctly through the IPC handler (`src/main/index.ts:841`), preload bridge, and `electron.d.ts` type, and the test suite covers the happy path, idempotent re-default, unknown id against a populated table, and unknown id against an empty table. CI is green (SonarQube, quality-gate, GitGuardian).

## Issues
- Minor: the diff updates the `ClaudeAccountsModal` test mock to resolve `setDefaultAccount` to `true`, but no renderer call site is included in this diff to show whether the component now does anything with that boolean. If the caller still fires-and-forgets, a stale-id click will silently no-op with no user feedback (the button just does nothing). Worth confirming the UI surfaces a message/toast when the promise resolves `false`, but since that code isn't in this diff I can't confirm it's missing — not blocking.

## Suggestions
- Non-blocking: consider a brief inline comment or follow-up ticket if the renderer doesn't yet handle the `false` case, so a stale-id click gives the user some indication (e.g., "account no longer exists, refresh") rather than a silent no-op.

William-Long-II:
Re the renderer question: confirmed by the gate-3 review (which walked the actual render path) — `ClaudeAccountsModal` deliberately discards the boolean and relies on its unconditional `refresh()` after the call. On a stale-id click the ghost row disappears from the list and the incumbent visibly keeps its default tag, so the click cannot appear to succeed and isn't a silent nothing — the list correcting itself is the feedback. There's no toast infrastructure in the renderer today; the boolean already reaches it end-to-end, so an explicit "that account no longer exists" message is a one-line addition whenever toast plumbing exists. Recorded rather than ticketed: it's noted in the PR body's gate results.

Issue #179, opened 2026-08-24
Epic: One remote path — finish the LAN rip-out, delete the legacy PWA


William-Long-II:
### Summary / goal

Finish the rip-out that `docs/designs/remote-hosting-relay.md` §7 ordered: one remote path (the relay), one web client (`relay/web`), no LAN-era attack surface. Roughly 25% of the §7 removals are done (mDNS, Tailscale UI); this epic completes the rest — almost all of it deletion.

### Why / business value

The legacy LAN HTTP/WS server still auto-starts on `0.0.0.0:8443` on every launch — untested (the entire server and its auth boundary have zero test coverage), legacy, and reachable from the local network — when the only thing that needs it is the Claude Code hook script posting to `127.0.0.1:8443`. And the repo builds and ships two complete mobile clients: the 6.2 KLOC LAN-era `src/pwa` (React, zero tests) and the strategic `relay/web`, sharing no code and no protocol. Every week both live, mobile investment is split or duplicated, and the build pays for both.

### Child issues / scope

- [ ] Bind the local server to `127.0.0.1` and delete the `localNetworkOnly` RFC1918/CGNAT allow-list — hooks keep working, the LAN attack surface dies. (Worth landing first, independently.)
- [ ] Delete on-LAN pairing: `pairing-manager.ts`, `routes/pairing.ts`, the QR flow, the Settings → Mobile tab, and its ~10 IPC handlers
- [ ] Delete `src/pwa`: the `build:pwa`/`dev:pwa` scripts, the `/m` static mount + SPA fallback in `http-server.ts`, and the `dist/pwa` asarUnpack rule. Decide the one porting candidate first: voice input (port to `relay/web` or file as its own issue)
- [ ] Dependency + residue sweep: `bonjour-service` (prod dep, zero imports), `react-is` (zero imports), `sodium-native` asarUnpack rules for a package that isn't installed, `electron-builder.yml` comments about tree-sitter/sqlite-vec, orphaned `CodeSearchModal.css` / `IndexStatus.css`, stale mDNS comments in `api/index.ts`, stale `dist/mcp-server` output
- [ ] Decide Microsoft Teams: the Graph OAuth backend initializes at startup with five IPC handlers while the Settings card hard-codes "Coming Soon". Finish it or delete it
- [ ] Narrow the web-push stack to its post-PWA consumer (relay push) once the epic for relay web push lands, or fold the port into that epic's work

### Definition of done

[...]

PR #191, opened 2026-08-24
167 — PtyManager.kill rejections no longer escape its fire-and-forget callers

### LINKED ISSUE #167: Bug: PtyManager.kill is awaited nowhere in three callers, so a rejection escapes as an unhandled rejection
Found while reviewing #166; predates that branch, so it was deliberately left out of scope.

`PtyManager.kill` is declared `async kill(id: string): Promise<void>` (`src/main/pty-manager.ts:981`), but three in-process callers invoke it **without `await`**, inside a synchronous `try/catch` that therefore cannot observe the rejection:

- `src/main/account-auth.ts:241` — its `catch` still carries the comment "Pty may have already exited", which describes a swallow that no longer happens
- `src/main/api/routes/sessions.ts:270`
- `src/main/api/routes/sessions.ts:336`

A throw from any of these becomes an unhandled main-process rejection rather than the intended swallow. It is the same class of defect as #164 one layer up: a promise-returning call treated as fire-and-forget.

#166 made `kill()` settle on the process's real exit and added POSIX `SIGKILL` escalation, so these calls now also return meaningfully later than they used to — which makes the missing `await` more visible, not less.

**Fix:** `await` each call (they are all already inside async functions — worth confirming) or attach an explicit `.catch()` with a comment saying what is being swallowed and why.

**Note for whoever picks this up:** the kill force-timer paths are currently untested, because they only fire after a real 3-second timeout. Testing them needs the interval injectable — worth doing in the same pass.

William-Long-II:
## Description
`PtyManager.kill()` returns a promise that three callers ignored, so a rejection escaped as an unhandled rejection. Rather than a blanket `await`, each call site got the handling its semantics call for:

- **`account-auth.ts` `cancelLoginFlow`** — handled with `.catch()` and a `log.warn`. This is the sync-IPC modal-close path; `kill()` on an absent session resolves cleanly, so a rejection here is a teardown glitch worth a log line, never a failed cancel.
- **`api/routes/sessions.ts` `POST /:id/stop`** — now awaited (handler made async). A rejection surfaces as the route's 500, and `success: true` now truthfully means the process is dead. The force path bounds the wait at ~3s.
- **`api/routes/sessions.ts` `DELETE /:id`** — awaited inside the existing `try`. A rejection aborts with the 500 envelope instead of half-succeeding, and retry is clean because `kill()` empties its session slot before anything can throw (pinned by a test).

Alongside: `pty-manager.ts` gained an injectable `killGraceMs` (constructor option, default 3000), which makes the force-timer paths testable.

**For review attention:**
1. `DELETE /:id` now blocks up to ~3s on a hung pty — truthful surfacing was chosen over the old instant-but-lying response.
2. `cancelLoginFlow` still `rmSync`s the account dir while the login pty may take up to 3s to die. That race predates this change and this PR does not alter it — flagged so the reviewer weighs it deliberately.
3. These routes belong to the legacy LAN API (epic #179). This fix keeps them honest for as long as they exist.

## Related issue
Closes #167

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?
Direct runs on the branch (`feat/167-Bodhilander`, worktree at `_wt-167-Bodhilander`):

- `pty-manager.test.ts` — **30 pass** (2 new: SIGKILL escalation resolves the waiter; a throwing graceful kill rejects the caller while the armed reaper still escalates and drains).
- `account-auth.test.ts` — **4 pass** (1 new: a rejecting `kill` is swallowed, rollback completes, no `unhandledRejection`).
- NEW `src/main/api/routes/__tests__/sessions.test.ts` — **4 pass** (stop gated on `kill` settling; stop/delete rejections → 500 with zero unhandled rejections; retry-after-failure completes the delete).
- Full suite: **878 pass / 0 fail** across 61 files.
- `tsc` (main) clean.
- Harness `verify.sh`: **GREEN**. Note: the registry's `test`/`lint` entries reported SKIP for this repo, so the evidence above is the direct runs, not the registry wrappers.

**Not verified:** no manual end-to-end run against a live hung pty; the ~3s force-timer behavior is exercised via the injectable `killGraceMs` in tests, not wall-clock. The `cancelLoginFlow` `rmSync`-vs-dying-pty race (item 2 above) has no test coverage in this PR.

## Checklist
- [x] Tests pass locally

## Gate results

**Gate 3 — reviewer: GREEN.** Completeness swept: six external `ptyManager.kill` call sites — the three fixed here, and three already handled (safeHandle-returned IPC, provider-install awaits, `killAll` in try/catch). Operator paths improved, not regressed: the PWA stop button now renders teardown failures in its existing error banner instead of a lying 200, and no consumer depends on the old instant response (the PWA's `apiFetch` has no timeout and shows a busy state). The benign already-exited path never rejects, so the cancel-login `log.warn` stays quiet. Mutation check: reverting the DELETE await kills its test. Discipline clean.

**Gate 4 — verifier: VERIFIED.** Every count reproduced first-try: scoped 38/0; full 878/0 vs 871/0 at base — exactly +7 tests / +1 file, purely additive. The new tests were also run against the OLD production code and fail there (the `unhandledRejection` listener captures the old escape at account-auth.ts:241; 3 of 4 route tests fail pre-fix), so their green is evidence, not decoration. Anchored claims re-grepped: `killGraceMs` defaults to 3000, the sole production constructor call stays bare, and the slot-emptied-before-throw retry safety is pinned. Stated not-verified: the win32 taskkill branch ran under no test (constant-for-constant change, tsc-covered, equally untested before this PR).

CI at `735e7de`: GitGuardian passes. SonarQube quality gate is red on a single condition — **0% Security Hotspots Reviewed on New Code** (needs 100%); no other condition failed and no new issues were reported. The likely subject is the pre-existing win32 `execFile('taskkill', ...)` block, whose surrounding lines changed in this PR. Hotspot review is a SonarQube-UI action; the `quality-gate` job re-runs green once it is done.

Order: Bodhilander

### DIFF
--- a/src/main/__tests__/account-auth.test.ts
+++ b/src/main/__tests__/account-auth.test.ts
@@ -111,7 +111,7 @@ function fakePtyManager(): PtyManager & { loginSessions: string[] } {
     },
     on() {},
     off() {},
-    kill() {},
+    kill: () => Promise.resolve(),
   };
   return stub as unknown as PtyManager & { loginSessions: string[] };
 }
@@ -193,5 +193,36 @@ describe('startLoginFlow legacy seeding', () => {
 
     accountAuth.cancelLoginFlow(pty, a.ptyId, false);
     accountAuth.cancelLoginFlow(pty, b.ptyId, false);
+  });
+});
+
+describe('cancelLoginFlow', () => {
+  test('a rejecting kill is swallowed and cleanup still runs to completion', async () => {
+    const pty = fakePtyManager();
+    let kills = 0;
+    pty.kill = () => {
+      kills++;
+      return Promise.reject(new Error('teardown glitched'));
+    };
+    const { ptyId } = await accountAuth.startLoginFlow(pty, null, 'Doomed');
+
+    const escaped: unknown[] = [];
+    const onUnhandled = (err: unknown) => { escaped.push(err); };
+    process.on('unhandledRejection', onUnhandled);
+    try {
+      accountAuth.cancelLoginFlow(pty, ptyId, true);
+      // Give a floating rejection an event-loop turn to surface.
+      await new Promise((r) => setTimeout(r, 10));
+    } finally {
+      process.off('unhandledRejection', onUnhandled);
+    }
+
+    expect(escaped).toEqual([]);
+    expect(kills).toBe(1);
+    // The abort rollback ran past the failed kill: the account row is gone.
+    expect(storedAccounts()).toEqual([]);
+    // And so is the flow itself — a second cancel is a no-op, not a re-kill.
+    accountAuth.cancelLoginFlow(pty, ptyId, true);
+    expect(kills).toBe(1);
   });
 });
--- a/src/main/__tests__/pty-manager.test.ts
+++ b/src/main/__tests__/pty-manager.test.ts
@@ -12,6 +12,9 @@
  *   the replacement bound to its id, kill() settles off the real exit and
  *   coalesces per id, and the silent resume-failure respawn inherits the last
  *   known size instead of dropping back to 80x24,
+ * - the kill force path, on an injectable grace: a pty trapping the graceful
+ *   signal is reaped by SIGKILL, and a graceful kill that throws rejects the
+ *   caller while the armed reaper still escalates,
  * - live-account publication (#165) for a provider without account support.
  *
  * Run with: bun test src/main/__tests__
@@ -28,11 +31,13 @@ interface FakePty {
   spawnOpts: { cols: number; rows: number };
   dataCb: ((data: string) => void) | null;
   exitCb: ((event: { exitCode: number }) => void) | null;
+  /** Signals delivered through the handle; undefined is the default kill(). */
+  killSignals: (string | undefined)[];
   onData(cb: (data: string) => void): void;
   onExit(cb: (event: { exitCode: number }) => void): void;
   write(data: string): void;
   resize(cols: number, rows: number): void;
-  kill(): void;
+  kill(signal?: string): void;
 }
 
 const spawned: FakePty[] = [];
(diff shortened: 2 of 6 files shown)


brannon-bowden (approved):
Approve — this closes a real gap (fire-and-forget `kill()` calls that could produce unhandled promise rejections) with logic changes that are matched by targeted regression tests, including tests that explicitly listen for `unhandledRejection` to prove the fix actually works.

**Blocking issues**

None.

**Non-blocking issues**

- `src/main/api/routes/sessions.ts:274` and `:337` — `DELETE /:id` and `POST /:id/stop` now `await ptyManager.kill(id)` before responding, whereas previously the response returned immediately. Since the SIGHUP→SIGKILL escalation path can take up to `killGraceMs` (default 3000ms) plus process teardown time, these endpoints can now take several seconds longer to respond in the escalation case. This looks intentional (comments and the new "response waits on kill()" test confirm the design), but it's a client-observable latency/contract change — worth confirming paired mobile clients have timeouts that comfortably exceed the worst-case grace period.
- `src/main/__tests__/pty-manager.test.ts:454` — the new `kill() force path` tests are gated to POSIX only (`posixTest`), which is reasonable since the escalation mechanism differs, but the Windows `taskkill` branch (also touched in this diff to parametrize the timeout via `killGraceMs`) has no equivalent test coverage of its own force-path behavior.
- `src/main/api/routes/__tests__/sessions.test.ts` shadows `ptyManager.getSession`/`kill` by assigning directly onto the singleton instance and restoring with `delete`. This only correctly restores the original behavior if those are regular prototype methods (not instance-bound arrow class fields) on `PtyManager`; the diff doesn't show enough of the class definition to confirm one way or the other, so flagging as worth a quick check rather than something to act on.

**Tests**

Coverage is strong and specifically targets the bug being fixed: `account-auth.test.ts` verifies a rejecting `kill()` no longer escapes as an unhandled rejection and that cleanup/idempotency still hold; the new `sessions.test.ts` verifies both routes turn a `kill()` rejection into a 500 instead of an unhandled rejection, and that a subsequent retry succeeds. No missing tests for the core behavior change; only the minor Windows-path gap noted above.

Issue #178, opened 2026-08-24
Epic: Ship what's built — stable promotion, CI test gate, signed builds


William-Long-II:
### Summary / goal

Get the month of shipped work into stable users' hands, and make quality enforcement real: promote development → production as 3.6.0, gate merges on tests in CI, sign Windows builds, make macOS notarization fail closed, and make the documentation truthful.

### Why / business value

Stable users have been on 3.5.0 since July 27 — the entire sharing feature (M5.0–M5.2), the account-switch fixes, and web sign-out all sit in `3.5.1-beta.6`. Meanwhile 60 test files (~9.7 KLOC) run only when someone remembers to type `bun test`: there is no `test` script in package.json, no CI workflow runs tests, and no coverage reaches Sonar — nothing blocks a merge or a release on tests passing. Windows installers ship unsigned (SmartScreen tax on every new user), and `scripts/notarize.js` silently skips when signing secrets are absent, so a mis-set secret would produce an unsigned macOS release with a green build.

### Child issues / scope

- [ ] Fix #167 (`PtyManager.kill` awaited nowhere in three callers — rejection escapes as an unhandled rejection)
- [ ] Fix #168 (`setDefaultAccount` with a stale id demotes the incumbent and promotes nobody)
- [ ] Add `"test": "bun test"` to package.json and a PR workflow that runs it; report coverage to Sonar (`sonar.javascript.lcov.reportPaths`); add `relay/` to the Sonar scan (the sharing design doc itself flags that the unscanned side attracts decisions)
- [ ] Windows code signing: procure a cert, wire `CSC_LINK`/`certificateFile` into the win matrix of `release.yml`
- [ ] Make notarization fail closed in CI: `notarize.js` hard-fails release builds when secrets are missing; verify the current secrets actually produce a notarized artifact (the README's `xattr -cr` note suggests this may never have been confirmed)
- [ ] README truth pass: drop sqlite-vec / tree-sitter / onnxruntime-node from the tech stack (removed in #146), drop the "HTTPS/TLS" claim for the plain-HTTP local server, drop mDNS/Tailscale-era instructions, reposition as a multi-provider AI-CLI session manager (7 providers) with Claude as flagship; update `relay/README.md` off "Milestone 1 skeleton" — it describes `/ws` as closing `1013 not implemented` while the service has full auth, linking, and sharing
- [ ] Promote development → production; on production, `npm version minor` → 3.6.0, push (no `--tags`)

### Definition of done

[...]

PR #194, opened 2026-08-24
[192] CI test gate — run the suite on every PR, coverage to Sonar, relay in scope

### LINKED ISSUE #192: Task: CI test gate — run the suite on every PR, coverage to Sonar, relay in scope
### What needs doing

Make the test suite enforceable: a `test` script, a CI workflow that runs it on every PR, coverage reported to SonarQube, and the relay brought under the Sonar scan.

- Add `"test": "bun test"` to package.json (there is none today — 61 test files run only when someone remembers the bare command).
- A GitHub Actions workflow (e.g. `test.yml`) running on `pull_request` and on push to `development`: install with bun, run the suite. Known risk to solve, not skip: `postinstall` runs `electron-rebuild` for better-sqlite3/node-pty — CI needs either a working headless install or a test path that avoids the electron rebuild (the suites use `bun:sqlite` and mock node-pty, so a `--ignore-scripts` install may suffice; prove it, don't assume it). The relay suite (`cd relay && bun test`) runs too — it has its own lockfile.
- Coverage: `bun test --coverage --coverage-reporter=lcov`, and wire `sonar.javascript.lcov.reportPaths` in sonar-project.properties (Sonar currently shows "no data about coverage").
- Add `relay/src` and `relay/web/src` to `sonar.sources` (currently `sonar.sources=src` — the sharing design doc itself flags that the unscanned side attracts decisions). Keep test files excluded per Sonar convention.
- Note for the human: making the new workflow a required status check is a repo-settings step; call it out in the PR body rather than assuming.

### Context / why

Tests exist and are good (the last four initiatives each shipped load-bearing suites) but nothing blocks a merge or a release on them passing — they are advisory. This is the largest process risk in the repo and a prerequisite for the 3.6.0 stable promotion (epic #178).

### Done when

- [ ] A PR with a deliberately failing test shows a red check (prove it on the branch, then revert the canary)
- [ ] `bun test` runs green in CI on both the app and relay suites
- [ ] SonarQube shows coverage data on the next PR
- [ ] relay/ code appears in the Sonar scan
- [ ] The PR body names the one manual step (marking the check required)

### Priority

High

Part of #178.

William-Long-II:
## Description
Initiative 192. Adds a CI test gate and wires coverage and the relay into Sonar:

- **New `.github/workflows/test.yml`** — runs on `pull_request` and on push to `development`. Pinned `checkout`/`setup-bun` action SHAs (matching the `release.yml` convention) and **`bun-version: 1.4.0` pinned** in both test.yml and sonarqube.yml (`release.yml` keeps floating per its own convention). Root install is `bun install --frozen-lockfile --ignore-scripts`, which skips the electron-rebuild postinstall — no Electron download or native toolchain on CI; the suites use `bun:sqlite` and mock `node-pty`. Plain frozen install in `relay/`, then both suites run with `--isolate`. PR-only concurrency cancellation, `contents: read`, 15-minute timeout.
- **Coverage to Sonar** — `bun test --coverage --coverage-reporter=lcov` produces `coverage/lcov.info` with repo-root-relative `SF` paths; a relay-cwd lcov would collide on `src/...` paths and was deliberately avoided. Generation steps added to `sonarqube.yml`'s workspace before the scan; `sonar.javascript.lcov.reportPaths=coverage/lcov.info`; `coverage/` gitignored.
- **Sonar scope** — sources are now `src,relay/src,relay/web/src`; exclusions add `**/*.test.ts,**/*.test.tsx,**/__tests__/**`. There were no test exclusions before, so test files were being scanned as main code.

### What the gate's own first run exposed
The first ubuntu run of the new workflow went red on **pre-existing order-dependent mock poisonings** (817 pass / 16 fail / 2 errors). Root cause, refined during the fix: **bun 1.4.0** — which the floating `setup-bun` resolved on CI — changed `bun:test` semantics: module mocks leak across files through one shared registry keyed by resolved path (1.3.x restored them between files), and the victim set varies with the repo's absolute path (38 failures in a Linux container at the CI path vs 16 + 2 on CI itself). Fixes on this branch:

- `bun test --isolate` everywhere a suite runs — the package.json script and both workflow steps, including the relay step (6ca167f).
- `bun-version: 1.4.0` pinned in test.yml and sonarqube.yml, so the semantics CI runs under are the semantics that were tested.
- Mock repairs (b9cd55f): `account-auth.test.ts` restores the real mcp-config/legacy-claude-seed modules in `afterAll` and its stub covers the full four-export shape; its pty-manager mock is deleted (`account-auth.ts` imports `PtyManager`/`BrowserWindow` as type-only); the engine/engine-ollama arena stubs gained a faithful `settleInterruptedResponses`; engine-ollama stubs key-vault itself instead of borrowing another file's mock.

## Related issue
Closes #192

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Hotfix
- [x] Refactor / chore
- [ ] Docs

## How was this tested?
**Local receipts (draft stage, b4cd2a7):** in a clean worktree with no `node_modules`, `bun install --frozen-lockfile --ignore-scripts` followed by the suites gave root **871 pass / 0 fail** (60 files, includes the relay's 9) and relay standalone **176 pass / 0 fail**. `coverage/lcov.info` confirmed with repo-root-relative `SF` paths. `actionlint` clean on both touched workflows. Harness `verify.sh` GREEN (registry test/lint entries SKIP for this repo — the direct runs are the receipts).

**CI proof chain (everything the draft listed as CI-only-provable is now proven):**

- **Canary red at ce54311** — [test run 32783011981](https://github.com/Software-Development-LLC/Bodhilander/actions/runs/32783011981) failed with exactly the one deliberately failing canary test, 871 passing beside it. The gate goes red on a failing test.
- **Byte-exact revert at bbbcd3e** — [test run 32783136578](https://github.com/Software-Development-LLC/Bodhilander/actions/runs/32783136578) green (871/0 app + 176/0 relay); [quality-gate run 32783136397](https://github.com/Software-Development-LLC/Bodhilander/actions/runs/32783136397) green with the scanner logging `Analysing [.../coverage/lcov.info]` — coverage ingested by Sonar.
- **Final head 6ca167f** — all four checks green: test, quality-gate, SonarQube, GitGuardian.

**Not verified / outstanding:** nothing in the code path. The one remaining step is manual and human-only (below).

## Checklist
- [x] Tests pass locally

## Gate results

**Gate 4 — verifier: VERIFIED at b4cd2a7.** Receipts reproduced; lcov shape confirmed; action pins byte-matched against the release.yml convention; exclusion patterns swept.

**Gate 3 — reviewer: first pass RED** — the CI-red finding (the order-dependent mock poisonings surfaced by the gate's own first ubuntu run). **Delta re-check at bbbcd3e GREEN:** mocks are faithful supersets and properly restored; the type-only import claim confirmed via clean `tsc`; `--isolate` proven non-masking on three configurations; zero changed `expect` lines; canary pair byte-exact. The isolate sweep at 6ca167f closes the reviewer's last carry-forward (the relay step now runs isolated too).

## Two warnings the team needs

1. The first post-merge `development` analysis pulls 17 relay files (both `crypto.ts`, `auth/github.ts`, `auth/cookies.ts`) into the Sonar new-code period and will likely raise Security Hotspots needing UI review plus a re-run — the PR #191 pattern.
2. 51 test files leave `sonar.sources`, so LOC drops and existing test-code issues auto-close on the next analysis. Deliberate, and stated here so nobody reads it as an accident.

## Manual repo-settings step (human)
Mark job `test` in workflow **"Tests"** as a required status check.

Order: Bodhilander

### DIFF
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -31,6 +31,25 @@ jobs:
           # Disabling shallow clone is recommended for improving relevancy of reporting.
           fetch-depth: 0
 
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+
+      # --ignore-scripts skips postinstall's electron-rebuild, which needs an
+      # Electron download and a native toolchain. The suites never load a
+      # native module: they use bun:sqlite and mock node-pty.
+      - name: Install dependencies
+        run: bun install --frozen-lockfile --ignore-scripts
+
+      - name: Install relay dependencies
+        run: bun install --frozen-lockfile
+        working-directory: relay
+
+      # One root run discovers relay's tests too, producing a single
+      # coverage/lcov.info whose paths are all repo-root-relative — the shape
+      # sonar.javascript.lcov.reportPaths expects.
+      - name: Generate test coverage
+        run: bun test --coverage --coverage-reporter=lcov
+
       - name: SonarQube Scan
         uses: sonarsource/sonarqube-scan-action@22918119ff8e1ca75a623e15c8296b6ea4fbe28f # v8.2.1
         env:
--- a/.github/workflows/sonarqube.yml
+++ b/.github/workflows/sonarqube.yml
@@ -33,6 +33,11 @@ jobs:
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          # Exact pin: bun:test module-mock semantics changed between minors
+          # (1.4.0 shares one mock registry across files unless --isolate), so
+          # a floating runtime can flip this gate without any code change.
+          bun-version: 1.4.0
 
       # --ignore-scripts skips postinstall's electron-rebuild, which needs an
       # Electron download and a native toolchain. The suites never load a
@@ -46,9 +51,10 @@ jobs:
 
       # One root run discovers relay's tests too, producing a single
       # coverage/lcov.info whose paths are all repo-root-relative — the shape
-      # sonar.javascript.lcov.reportPaths expects.
+      # sonar.javascript.lcov.reportPaths expects. --isolate keeps each file's
+      # module mocks to itself, same as the Tests workflow.
       - name: Generate test coverage
-        run: bun test --coverage --coverage-reporter=lcov
+        run: bun test --coverage --coverage-reporter=lcov --isolate
 
       - name: SonarQube Scan
         uses: sonarsource/sonarqube-scan-action@22918119ff8e1ca75a623e15c8296b6ea4fbe28f # v8.2.1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,8 @@ jobs:
         run: bun test --isolate
 
       # The relay deploys against its own lockfile (relay/bun.lock), so its
-      # suite must also pass standalone from relay/.
+      # suite must also pass standalone from relay/. --isolate for the same
+      # reason as above — the first relay mock.module must not change this.
       - name: Run relay suite
-        run: bun test
+        run: bun test --isolate
         working-directory: relay
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,6 +25,11 @@ jobs:
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          # Exact pin: bun:test module-mock semantics changed between minors
+          # (1.4.0 shares one mock registry across files unless --isolate), so
+          # a floating runtime can flip this gate without any code change.
+          bun-version: 1.4.0
 
       # --ignore-scripts skips postinstall's electron-rebuild, which needs an
       # Electron download and a native toolchain. The suites never load a
@@ -37,8 +42,10 @@ jobs:
         working-directory: relay
 
       # Discovers every *.test.ts(x) under the repo, relay/ included.
+      # --isolate gives each file its own module registry, so one file's
+      # mock.module can never become another file's subject.
       - name: Run app suite
-        run: bun test
+        run: bun test --isolate
 
       # The relay deploys against its own lockfile (relay/bun.lock), so its
       # suite must also pass standalone from relay/.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,47 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - development
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  contents: read
+
+# Cancel superseded runs only on pull requests; pushes to development run to
+# completion so the branch's check history stays gap-free.
+concurrency:
+  group: tests-${{ github.ref }}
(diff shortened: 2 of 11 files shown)


brannon-bowden (approved):
Approve — this is a well-documented CI infrastructure PR that adds a proper test gate and wires coverage into SonarQube, and the accompanying test-file changes correctly account for `--isolate` semantics.

## Blocking issues
None.

## Non-blocking issues

- **Relay tests run twice per PR** (`.github/workflows/test.yml:37-51`): the root "Run app suite" step (`bun test --isolate`) discovers and runs every `*.test.ts(x)` under the repo, including `relay/`, and then "Run relay suite" runs the same relay tests again standalone from `relay/`. The comments explain this is intentional (validate relay both within the monorepo run and standalone against its own lockfile), but it doubles relay test runtime on every PR. Worth confirming this trade-off is deliberate rather than incidental, and considering whether the root run should exclude `relay/**` to avoid the duplication if not.
- Related to the above: when the root run executes relay's test files, `process.cwd()` is the repo root, not `relay/`. If any relay test resolves fixtures or config via relative/cwd-based paths, behavior could differ between the two invocations. CI is currently green, so this isn't observable as a failure today, but it's worth double-checking if relay tests are ever extended to touch the filesystem.
- `src/main/account-auth.ts:18-24` changes `PtyManager` from a value import to `import type`, and the corresponding test file removes the `mock.module('../pty-manager', ...)` stub entirely on the premise that `PtyManager` is "only referenced in type positions" in `account-auth.ts`. That's consistent with what's shown in the diff, but the full body of `account-auth.ts` isn't visible here — if `PtyManager` is used as a value anywhere else in that file, this would be a compile break. Worth confirming a typecheck step (`tsc --noEmit`) actually runs against this file in CI, since `bun test` alone doesn't enforce type-only-import correctness.

## Tests

No new production logic was introduced (the `account-auth.ts` change is a mechanical type-only import refactor), so no new tests are required. The test-infrastructure changes themselves (restoring real `mcp-config`/`legacy-claude-seed` modules in `afterAll`, filling out mock export shapes, adding `key-vault` and `settleInterruptedResponses` stubs) are appropriately scoped to make the existing suites correct under `--isolate` and don't need their own tests.

## Nit

- The `bun-version: 1.4.0` setup block (with its multi-line rationale comment) is duplicated verbatim between `sonarqube.yml` and `test.yml`. Not required to fix now, but a reusable/composite action would keep the two workflows from drifting if the pin is ever bumped in only one place.

William-Long-II:
Answers to the three review questions, for the record:

- **Type-only import safety:** verified beyond the diff — both gate agents ran `tsc -p tsconfig.main.json --noEmit` against this branch (clean), and the gate-3 re-check specifically confirmed `PtyManager`/`BrowserWindow` appear only in type positions across the full body of `account-auth.ts`. Your underlying point stands though: **PR CI currently runs no typecheck step** — `test.yml` runs only `bun test`, and the typecheck implicit in `build:main` runs on release pushes, not PRs. Recorded on epic #178 as its own item (add a `tsc --noEmit` step for both tsconfigs to the Tests workflow) rather than amending this approved PR.
- **Relay tests running twice:** deliberate, per the step comments — the standalone run validates relay against its own lockfile (it has a separate bun.lock the root run never exercises). Cost is ~2s of relay suite. The cwd caveat is noted and currently moot: relay tests use in-memory DBs and no cwd-relative fixtures. Both facts worth rechecking if relay tests ever grow filesystem touchpoints.
- **Duplicated pin block:** carried to epic #178's notes as a composite-action cleanup candidate if the pin ever needs bumping.

PR #197, opened 2026-08-25
196 — Claude Accounts show logged in when the config dir says they are

### LINKED ISSUE #196: Bug: Claude Accounts all show 'not yet logged in' despite completed logins
### What happened

Settings → Claude Accounts lists every account as "not yet logged in", even though all three accounts were added through the login flow, completed it, and have been used for sessions since.

### Expected

An account that has completed login shows as logged in.

### Investigation notes (hypotheses to verify, not assume)

- The status is presumably derived from some credential artifact in the account's isolated `CLAUDE_CONFIG_DIR`. On macOS, Claude Code stores OAuth credentials in the system Keychain rather than (only) a `.credentials.json` in the config dir — a file-presence check would then read every healthy account as logged out. Verify what the derivation actually reads vs what a logged-in account's config dir actually contains on macOS.
- The accounts area moved recently: #166/#176 (session account switching, live account indicator) and #190 (default-account guard). Check whether the status field regressed alongside, or whether it was always wrong and the modal only recently started surfacing it.
- Check both layers: the main-process derivation and what `ClaudeAccountsModal` renders from it.

### Done when

- [ ] Accounts that have completed login display as logged in on macOS (and the derivation is correct per-platform, not just patched for one)
- [ ] A test pins the status derivation against the real credential layout(s)

William-Long-II:
## Description

Settings → Claude Accounts reported every account as "not yet logged in" on macOS, including three accounts that had completed login and been used for sessions since.

**Root cause.** The status was never derived from a credential artifact at all. It came from `acc.email`, and `email` was only ever written by `parseAccountEmail()` parsing `<configDir>/.credentials.json` at login time. `AccountChip` rendered "Not yet logged in" whenever email was blank. macOS never writes that file — Claude Code puts the tokens in the Keychain — so the column stayed NULL and every healthy account read as logged out.

**Ground truth on the reporter's machine.** All three account config dirs are heavily used. None contains `.credentials.json` or `auth.json`. All three carry `.claude/.claude.json` with a populated `oauthAccount` block holding a distinct address, which confirms both the hypothesis and that per-config-dir isolation is working. (Directory listings and JSON key names only — no file contents and no secrets were read or recorded.)

**When it broke: never worked on macOS.** `parseAccountEmail` and the label both landed 2026-04-17 with the original accounts feature, so this is not a regression from the recent accounts work. #165 narrowed where the label appears, which is why it showed precisely in Settings → Claude Accounts.

**The fix.** A new `src/main/account-identity.ts` probes the account's config dir in order:

1. `<configDir>/.claude.json` → `oauthAccount` — primary, and correct on every platform. It requires an object carrying `emailAddress` or `accountUuid`, because the file itself exists from first run and so its presence proves nothing; the block is the signal.
2. `.credentials.json` / `auth.json` presence — the fallback for platforms that write a plaintext token store.
3. The recorded row address — and only where the config dir could not be read at all.

`accounts:list` maps through `withAccountIdentity`, so one boundary feeds every renderer surface.

**`loggedIn` is a tri-state, not a boolean.** `undefined` means "not consulted, or consulted and unreadable" — it is never "logged out". A `.claude.json` that exists but will not parse yields UNKNOWN rather than logged-out. That matters concretely: Claude Code rewrites that file roughly once a minute, so a torn read is a real event, and under a two-state flag it would have reproduced the reported bug intermittently and at random. Both surfaces test `=== false`, so a surface that consulted nothing cannot accuse an account of being logged out — the exact shape this bug took.

**The address fallback is narrow on purpose.** A recorded address stands in for the probe *only* where the directory could not be read:

```ts
let loggedIn = identity.loggedIn;
if (loggedIn === undefined && storedEmail !== null) loggedIn = true;
```

A readable directory showing no login is authoritative. So `/logout`, and a config dir that has been moved or is on an unmounted volume, now surface as `false` instead of being masked by a stale address.

**The completion event carries `verified`.** The field runs main → preload → `electron.d.ts` → renderer, so the login overlay words itself from what main actually resolved rather than asserting "this account is signed in" on the strength of a button press. Unverified now reads "Recorded — but no login was found in this account's config directory yet".

**The panel pairs address with status.** Rows carry the account's address *and* an explicit `not signed in` tag beside the existing `default` / `in use` tags. The old email-over-label precedence was a safety net for a login state that could not be trusted; now that the state is authoritative and every login records an address, that same rule would have hidden every logout rather than protecting against a false one. `noEmailLabel` was deleted from `AccountChip` entirely, because it existed only to render a missing address as a login status — the exact conflation this fix removes.

**The login watcher** is keyed to `LOGIN_ARTIFACTS` and gated on the same resolver, so a macOS login auto-detects instead of waiting on the manual confirm button.

**Why probe rather than persist our own marker.** A `loggedInAt` marker would not fix any existing account without probing anyway; it records a past event rather than current state; on macOS it would be pure self-report; and it misses logins performed outside the app.

**Keychain deliberately not consulted.** Reading it requires a prompt-triggering `security` call, and its item is not demonstrably keyed per `CLAUDE_CONFIG_DIR`, whereas `.claude.json` is.

## Related issue

Closes #196

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?

**Automated, at head `855bf17`.**

| suite | tests |
| --- | --- |
| `account-identity.test.ts` (new) | 29 |
| `account-auth.test.ts` | 10 |
| `preload.test.ts` (new) | 4 |
| `ClaudeAccountsModal.test.tsx` | 23 |
| `AccountChip.test.tsx` | 12 |
| `App.accounts.test.ts` | 18 |

`account-identity.test.ts` uses zero mocks — it builds real fixture directories laid out per platform.

```
full root suite:  987 pass / 0 fail across 66 files
tsc (main):       clean
tsc (renderer):   clean
```

**CI at `855bf17`: all four checks green** — SonarQube Quality Gate, `quality-gate`, `test`, GitGuardian.

**Sonar's one new issue was resolved without obeying it.** It flagged the blank-address collapse in `AccountChip`. The suggested `??` would have been wrong: `''` is not nullish, so nullish coalescing preserves it and hands the chip an empty address that identifies nothing. It was rewritten instead as the explicit truthiness guard the file already uses for its colour swatch two lines below, which is both correct and consistent with its neighbour. A new test pins that an empty-string email renders no address at all.

**Harness.** `scripts/verify.sh Bodhilander=feat/196-Bodhilander`, run from the copy pinned in `team.yaml` (`.../bodhi-marketplace/bodhi/2.18.4`):

```
16 file(s) changed
21% of added lines are comment (172/832)
running: bun test --isolate src/main/__tests__/account-auth.test.ts …
specs:   19 enumerated (never a glob)

UNVERIFIED   coverage
PASS         tests
SKIP         lint
PASS         seams
```

It enumerates the specs reachable from the diff — **19** here, never a glob — and runs them under `bun test --isolate`. Tests **PASS**, seams **PASS**. It is RED on exactly one row, `coverage`, for `src/main/index.ts`; both gates examined that row specifically and accepted the justification recorded under Gate results.

**`lint` SKIP is verified-absent, not a detection failure.** This repo has no `lint` script in `package.json` and no eslint, biome or oxlint configuration anywhere in the tree. There is no lint step to run, so `null` is the correct record rather than a missed one.

The suite and typecheck evidence above came from direct runs, and the suite is independently enforced by the `test` check on this PR.

**Not verified.** No manual run against a live macOS install was performed on this branch. The macOS behaviour is pinned by fixture directories reproducing the reporter's layout and by the verifier's read of the reporter's three real config dirs, not by an end-to-end login through the UI. The Windows and Linux paths are covered by fixtures rather than by a real login on those platforms. `src/main/index.ts` carries no direct coverage — see the coverage note under Gate results.

## For review attention

1. `oauthAccount` is a profile cache that most likely survives `/logout`, so the honest reading of the flag is "has completed a login" rather than "holds a valid token" — a revoked or expired token still reads as logged in. This is deliberate: the alternative is prompting the user for Keychain access on every list.
2. Tier 1 couples us to Claude Code's on-disk shape. If `.claude.json` or `oauthAccount` is restructured upstream, only accounts whose directory is unreadable fall through to the recorded address.
3. `confirmLoginMacOS` keeps its name while now serving as a general manual override. Renaming it ripples through IPC, preload, types and renderer, so the name was left as-is and is flagged here rather than changed silently.
4. The `.claude.json` watcher parses on each write during the login window. That is bounded by the window itself and by the completed guard.

## Gate results
- **Gate 3 (adversarial reviewer), first pass: RED** — two ways the UI could still state something false to the user. Addressed, and the re-check came back **GREEN**.
- **Final re-check: GREEN.** The banner mutation that survived the earlier pass is now killed. The reviewer ran a full five-state render matrix over the new status tag, killed six mutations on it, measured its contrast at **7.44:1 (AAA)**, confirmed it is distinguishable from its sibling tags without relying on colour, and confirmed **zero `noEmailLabel` remnants** anywhere in `src/`.
- **Gate 4 (verifier): VERIFIED.** Test counts reproduced independently against a real in-memory database, with no weakened assertions. The verifier also confirmed directly that the reporter's three real accounts resolve as logged in, from artifacts the old code could never have read.
- **One equivalent-mutant pair, analysed and accepted.** Two individually correct guards mutually mask each other's substitution, so neither mutation alone is caught. Removing **both** fails the two tests that assert the invariant. That is redundancy between the guards, not a gap in coverage.
- **Coverage.** `src/main/index.ts` remains the single unverified file, for the stated reasons: 1486 lines, 91 IPC registrations, and import-time side effects pulling in Electron, the database and `node-pty`. Moving the identity overlay into the accounts repository to clear the gate was considered and **rejected** — it would put filesystem I/O into a database repository that is called on every PTY spawn.

## Checklist
- [x] Tests pass locally (987 pass / 0 fail across 66 files; both typechecks clean; all four CI checks green)
- [x] Docs updated where needed — no user-facing or contributor doc describes login-state derivation, so none needed changing

Order: Bodhilander

### DIFF
--- a/src/main/__tests__/account-auth.test.ts
+++ b/src/main/__tests__/account-auth.test.ts
@@ -222,6 +222,24 @@ async function waitFor(predicate: () => boolean, timeoutMs = 5000): Promise<void
   throw new Error('condition was never met');
 }
 
+type MainWindow = Parameters<typeof accountAuth.confirmLoginMacOS>[0];
+
+interface SentEvent {
+  channel: string;
+  data: { accountId: string; email: string | null; verified: boolean };
+}
+
+/** Captures what the renderer is told, which is the claim the overlay words. */
+function fakeWindow(): { sent: SentEvent[]; win: MainWindow } {
+  const sent: SentEvent[] = [];
+  const win = {
+    webContents: {
+      send: (channel: string, data: SentEvent['data']) => { sent.push({ channel, data }); },
+    },
+  };
+  return { sent, win: win as unknown as MainWindow };
+}
+
 /**
  * Which writes into the config dir mean "a login landed". The watch used to
  * answer that with the presence of a token file, which a platform holding its
@@ -283,6 +301,39 @@ describe('login detection', () => {
 
     expect(storedEmail(account.id)).toBeNull();
     expect(storedAccounts()).toEqual([{ id: account.id, isDefault: true }]);
+    accountAuth.cancelLoginFlow(pty, ptyId, false);
+  });
+
+  // The button can be pressed before OAuth finishes. Main writes no address in
+  // that case, so it must not let the overlay say the account is signed in
+  // while the panel behind it reads the same empty dir and says otherwise.
+  test('a confirmation main could not corroborate is reported as unverified', async () => {
+    const pty = fakePtyManager();
+    const { win, sent } = fakeWindow();
+    const { ptyId } = await accountAuth.startLoginFlow(pty, win, 'Work');
+
+    accountAuth.confirmLoginMacOS(win, ptyId);
+
+    const completed = sent.filter((e) => e.channel === 'accounts:login-completed');
+    expect(completed).toHaveLength(1);
+    expect(completed[0].data.verified).toBe(false);
+    expect(completed[0].data.email).toBeNull();
+    accountAuth.cancelLoginFlow(pty, ptyId, false);
+  });
+
+  test('a login main found on disk is reported verified, with its email', async () => {
+    const pty = fakePtyManager();
+    const { win, sent } = fakeWindow();
+    const { account, ptyId } = await accountAuth.startLoginFlow(pty, win, 'Work');
+
+    writeConfigFile(account.configDir, '.claude.json', {
+      oauthAccount: { accountUuid: 'u-1', emailAddress: 'will@acme.test' },
+    });
+    accountAuth.confirmLoginMacOS(win, ptyId);
+
+    const completed = sent.filter((e) => e.channel === 'accounts:login-completed');
+    expect(completed[0].data.verified).toBe(true);
+    expect(completed[0].data.email).toBe('will@acme.test');
     accountAuth.cancelLoginFlow(pty, ptyId, false);
   });
 });
--- a/src/main/__tests__/account-auth.test.ts
+++ b/src/main/__tests__/account-auth.test.ts
@@ -202,6 +202,91 @@ describe('startLoginFlow legacy seeding', () => {
   });
 });
 
+function storedEmail(id: string): string | null {
+  const row = db.prepare('SELECT email FROM claude_accounts WHERE id = ?').get(id) as
+    { email: string | null } | undefined;
+  return row?.email ?? null;
+}
+
+function writeConfigFile(configDir: string, name: string, contents: unknown): void {
+  fs.writeFileSync(path.join(configDir, name), JSON.stringify(contents));
+}
+
+/** fs.watch delivers asynchronously, so the assertion has to outlast it. */
+async function waitFor(predicate: () => boolean, timeoutMs = 5000): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (Date.now() < deadline) {
+    if (predicate()) return;
+    await new Promise((r) => setTimeout(r, 10));
+  }
+  throw new Error('condition was never met');
+}
+
+/**
+ * Which writes into the config dir mean "a login landed". The watch used to
+ * answer that with the presence of a token file, which a platform holding its
+ * tokens in a keyring never writes.
+ */
+describe('login detection', () => {
+  test('a login that writes only a profile completes the flow and takes its email', async () => {
+    const pty = fakePtyManager();
+    const { account, ptyId } = await accountAuth.startLoginFlow(pty, null, 'Work');
+
+    // What Claude Code leaves in a dir it has started in but nobody has logged
+    // into: the file is there, the profile block is not.
+    writeConfigFile(account.configDir, '.claude.json', { userID: 'u1', projects: {} });
+    await new Promise((r) => setTimeout(r, 150));
+    expect(storedEmail(account.id)).toBeNull();
+
+    writeConfigFile(account.configDir, '.claude.json', {
+      userID: 'u1',
+      oauthAccount: { accountUuid: 'u-1', emailAddress: 'will@acme.test' },
+    });
+    await waitFor(() => storedEmail(account.id) === 'will@acme.test');
+
+    // And it got there without the artifact the old check demanded.
+    expect(fs.existsSync(path.join(account.configDir, '.credentials.json'))).toBe(false);
+
+    accountAuth.cancelLoginFlow(pty, ptyId, false);
+  });
+
+  test('a token file appearing still completes the flow', async () => {
+    const pty = fakePtyManager();
+    const { account, ptyId } = await accountAuth.startLoginFlow(pty, null, 'Work');
+
+    writeConfigFile(account.configDir, '.credentials.json', {
+      claudeAiOauth: { accessToken: 'not-a-real-token', subscriptionEmail: 'will@linux.test' },
+    });
+    await waitFor(() => storedEmail(account.id) === 'will@linux.test');
+
+    accountAuth.cancelLoginFlow(pty, ptyId, false);
+  });
+
+  test('the manual confirmation records the email the watch would have', async () => {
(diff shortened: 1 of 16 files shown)


brannon-bowden (approved):
This PR replaces the "credentials file exists" heuristic with a proper `account-identity` resolver that reads `.claude.json`'s `oauthAccount` block (which macOS logins populate even though they never write a token file), while still honoring the plaintext token files on Linux/Windows. The tri-state result (`true`/`false`/`undefined`) is used carefully throughout: the fs.watch handler only completes a flow on a positive `loggedIn: true`, `withAccountIdentity` only lets a stored email stand in when the dir is genuinely unreadable, and the renderer surfaces a distinct "unverified" state when the user presses the manual macOS confirm button before OAuth actually finished. The three-way UI split (`accountMenuLabel`, `AccountChip`, `signed-out-tag` in the accounts panel) is consistent with the resolved `loggedIn` value, and the test suite (`account-identity.test.ts`, updated `account-auth.test.ts`, `preload.test.ts`, component tests) exercises the tricky cases well — torn reads mid-rewrite, org logins with no email, stale stored emails being overridden by a readable dir, unverified manual confirmations, etc. CI is green across all four checks.

I traced the branch logic in `account-identity.ts` (`probeProfile`/`probeTokenFile`/`resolveAccountIdentity`/`withAccountIdentity`) against every listed test case and didn't find a case where the implementation and the asserted expectation diverge.

## Suggestions (non-blocking)

- `src/main/account-identity.ts:82-98` — `probeProfile` distinguishes a present-but-unparseable `.claude.json` as `'unreadable'` (deferring judgment), but `probeTokenFile` treats a present-but-unparseable credentials file as an immediate `{ loggedIn: true, email: null }` rather than `'unreadable'`. That's evidently deliberate ("only writing it creates it," per the accompanying test), but it means genuine on-disk corruption of a credentials file is indistinguishable from a torn write-in-progress and will permanently read as logged-in with no email. Worth double-checking this is the intended tradeoff, or leaving a short note next to `probeTokenFile` explaining why it doesn't need the same "unreadable" deferral that `probeProfile` gets.
- `src/main/account-auth.ts` — worth a quick check that `confirmLoginMacOS`'s existing body (not shown in this diff) still guards on `flow.completed` the same way the watcher does, now that `handleLoginCompleted` can be invoked from two independent paths with different `known` values — just to make sure a watcher-completed flow can't be re-resolved (and its email potentially nulled out) by a late button press.

William-Long-II:
Both suggestions checked before merge:

**Re-invocation after a watcher-completed flow: already guarded.** `confirmLoginMacOS` opens with `if (!flow || flow.completed) return;` (`account-auth.ts:203`) and the watcher sets `flow.completed = true` at `:139` before calling `handleLoginCompleted`. So a late button press on an already-completed flow returns early and never re-resolves — the email cannot be nulled out by the second path. The `known` parameter only narrows this further: it is supplied solely from the watcher, immediately after an `if (!identity.loggedIn) return;`, so it is always a freshly-read positive and is never stored across ticks.

**`probeTokenFile` not deferring on an unparseable file: deliberate, and worth the note you suggested.** The asymmetry is that the two artifacts prove different things. `.claude.json` is written from first run and rewritten constantly, so its *existence* proves nothing and only a parsed `oauthAccount` is evidence — which is why a torn read there must defer rather than accuse. A credentials file is only ever created by a completed OAuth login, so its *presence* is the evidence and its contents are a bonus; corruption doesn't retract the fact that a login happened. You're right that this makes on-disk corruption indistinguishable from a torn write, and that the reasoning should live next to the code rather than only in a test name — I'll have that comment added.

Issue #180, opened 2026-08-24
Epic: Finish sharing — M5.3 watch-and-type and M5.4 hardening


William-Long-II:
### Summary / goal

Complete the session-sharing design (`docs/designs/session-sharing.md` §8): M5.3 watch-and-type behind high-friction consent with typing attribution and the pause control, and M5.4's audit log, expiry hygiene, quotas, and a11y sweep.

### Why / business value

Watch-only (M5.2) shipped well and has been polished across four follow-up PRs, but the design's back half is unbuilt — and it left a trap: the `operator` role is accepted by the relay (`MINTABLE_ROLES`, SQL CHECKs), enforced by the agent (`ROLE_CAPS.operator` includes `input`), and offered by no UI. `ShareSessionModal.tsx` hardcodes `role: 'viewer'`, and the web client ignores `Machine.role` entirely — `applyWatchOnly()` strips the compose bar from every guest, so an operator grant minted today would be *allowed* to type and given no way to do it. Sharing is the feature no first-party client offers; finishing it completes the moat.

### Child issues / scope

- [ ] M5.3 consent flow: watch-and-type behind type-the-machine-name confirmation with honest copy ("they can run any command you can, read any file you can, and spend your Claude account") — per the design, this is machine trust wearing session-scope clothing and the copy must say so
- [ ] M5.3 typing attribution (hard gate: no typing without it) and the re-mint path
- [ ] Pause control: the `client:kick` mechanism exists relay-side with a comment saying it's for pause; build the owner UI (pause from the presence pill), blank the guest's rendered screen, keep grant/socket/certificate alive
- [ ] Role-aware web client: read the `role` field the API already returns; operator guests keep compose bar + key toolbar; viewer copy stays as-is
- [ ] M5.4 desktop-authoritative audit log (grep for "audit" in src/ returns zero today — relay comments assert the desktop is the audit trail; make that true)
- [ ] M5.4 expiry hygiene: T-5-minute warning to both sides, an Extend action
- [ ] M5.4 quotas: invites per machine, active grants per machine (overlaps the hosted-relay epic — dedupe at planning)
- [ ] M5.4 a11y sweep of all sharing surfaces (overlaps the PWA epic's sheet work — dedupe at planning)
[...]

PR #198, opened 2026-08-25
185 — Guest arrival and sizing polish: land in the terminal, honest copy, Fit to my screen

### LINKED ISSUE #185: Feature: Guest arrival and sizing polish — land in the terminal, honest copy, Fit to my screen
### User story / problem

As an invited guest, I want to land in the session that was shared with me, read honest copy about what's happening, and be able to read a wide desktop terminal — the arrival-and-reading polish the session-sharing design (§7, Guest) specifies that didn't land with M5.2.

Concretely, four shipped gaps:
1. A guest with exactly one grant still gets the session list and machine pill — the design (and an in-file comment in `main.ts`) says they land directly in the terminal.
2. The offline state uses owner vocabulary for guests ("This machine is offline" instead of "Will's machine is offline right now. We'll ask again as soon as it's back.").
3. The two user-visible role words (Watching / Watching and typing) are not rendered anywhere.
4. "Fit to my screen" doesn't exist — panning shipped, the request half didn't, so a phone guest on a 164-column session pans forever with no way to ask for a resize.

### Proposed solution

- Single-grant guests render the terminal directly; the machine picker appears only at ≥2 grants, sectioned by person ("SHARED WITH ME — Will's laptop").
- Guest-voice connection/offline/waiting copy keyed off the existing relation field.
- Role word rendered in the terminal subtitle ("Shared by Will · watching").
- "Fit to my screen" button beside the wide-terminal banner → a resize request over the tunnel → owner desktop surfaces a one-tap prompt `[ Resize once ] [ Keep my size ]` (renderer + `session-tunnel.ts` message). Guests still never send `terminal:resize` directly.

### Acceptance criteria

- [ ] One grant → terminal, no picker; two grants → sectioned picker
- [ ] Offline/waiting/ended copy in guest voice, naming the owner, per the design's §7 strings
- [ ] "Fit to my screen" round trip works end to end; declining leaves the guest exactly as before; the owner prompt names the requested size
- [ ] Role words only, everywhere a role is shown
- [ ] Web tests cover the single-grant branch and the resize-request message; desktop tests cover the prompt

### Scope

Minor feature

### Priority

High

Part of #177.

William-Long-II:
## Description

The four arrival-and-reading gaps the session-sharing design (§7, Guest) specifies and that didn't land with M5.2. A guest now arrives somewhere, is told the truth about what's happening in their own voice, and can ask a wide desktop terminal to fit their screen.

### 1. Single-grant arrival — new `relay/web/src/arrival.ts`

One machine plus a relation of grantee means no machine pill, no picker: the one session the grant covers opens as soon as the list arrives. The decision is made **once**, and the guard lives in `autoOpenSessionId()` where it is tested. The list polls every 2.5s, and re-deciding on each poll would drag a guest who tapped Back straight back into the terminal.

A single grant covering two or more sessions still shows the list — deliberate, because there is a real choice to make. At two or more grants the picker returns, sectioned "My machines" / "Shared with me", with rows labelled by person, and the sheet titled "Shared with you" when the person owns nothing themselves. Section headings render only when there is more than one section to separate, symmetric across owner and guest.

The `＋` new-session button is removed for guests: `create` belongs to no guest role, so for a guest it could only ever produce a refusal. "Link a machine" moved into the account sheet so it stays reachable when the machine pill is suppressed.

### 2. Guest-voice copy — new `relay/web/src/guest-copy.ts`

"Will's machine is offline right now. We'll ask again as soon as it's back." and "Waiting for Will to let you in…".

The owner's name is not in the redeem response, so it comes from `/api/shares`, which lists pending grants with `ownerName`, and repaints only if the waiting screen is still the thing on screen. Where no name is known the fallback reads "Their machine is offline right now…", capitalised, with a test pinning the exact string.

A guest who lands directly in a terminal never sees the session list — which was the only place connection state was ever stated — so the same copy is repeated on a `#connStrip` above their screen.

### 3. Role words

`guestSubtitle()` builds on the existing `roleWord()` in `shares.ts`: "Shared by Will · watching", "Shared by Will · watching and typing". Both `openTerminal` and `updateTermHeader` render through it, so a poll cannot overwrite the subtitle with a raw role. Tests assert that `viewer`, `operator` and `grant` never reach the string.

### 4. Fit to my screen

The guest's wide-terminal banner gains a 44×44 button. Asking flips it to copy that promises nothing and reverts after 45s, so an ask that goes unanswered can simply be repeated:

> Asked Will to resize. They may not be at their desk — nothing changes unless they say yes, so keep reading meanwhile.

Two paths end in silence by design — a requested size below the floor, and a second guest asking while a prompt already stands — and that copy makes silence an expected outcome rather than a mystery.

The owner's terminal shows `👀 @dana-k asked to fit this session to their screen (80×24) — yours is 164×48.` with `[ Resize once ] [ Keep my size ]`. Declining sends nothing at all, so a refusal reaches the guest as silence followed by the 45s revert.

**An accepted fit is held.** `Terminal.tsx` carries `heldFitRef`; `handleResize` early-returns while a granted fit stands, so focus, window-resize and session-activation can no longer re-measure and push the desktop grid back over the guest. The hold clears exactly two ways: the owner clicking "Resume desktop size" — now genuinely the take-back — or the desktop size arriving over `pty:resized`. Accepting also sets `mobileSize`, so a held fit always implies the banner is on screen and the take-back is always reachable.

### The message it travels as

`terminal:resize-request {sessionId, cols, rows}`, client→agent, in the existing `terminal:*` vocabulary.

It requires only the `view` capability — asking is not resizing. Viewer and operator capabilities are unchanged, and `resize` stays owner-only. The tunnel:

- clamps through the existing `sanitizeSize`
- requires a live subscription to that session, because asking about a session you cannot see would leak that it exists
- **throttles per grant, not per socket** — a tunnel-level map keyed `grantId ?? userId ?? clientId`, swept so it cannot grow without bound. Six channels open under one certificate raise one prompt, not six.
- forwards through a new `onResizeRequest` dep, and never calls `pty.resize` itself

`shouldPrompt` refuses anything below 10×2, so the wire floor cannot become a button. The read/click race between two arriving requests is closed first-wins, via `setResizeRequest((pending) => pending ?? request)`.

Added to `fixtures/sharing-v1.json` commandCaps with a note recording that needing `resize` here would itself be the finding. The message, the per-grant throttle and the held fit are written into `docs/designs/session-sharing.md` §7.

### Two fixes that reach beyond guests

**A reconnected client re-subscribes.** `readyCommands(activeSessionId)` now returns groups → sessions → `terminal:subscribe` when a terminal is open. This was a pre-existing owner bug too; landing guests directly in a terminal made it the default path and so made it visible.

**Owner replays clear the screen first.** `\x1b[2J\x1b[3J\x1b[H` is emitted before the owner's chunks, through a single `CLEAR_SCREEN` constant both branches share, which kills a duplicate-scrollback regression the re-subscribe fix introduced. The invariant belongs with the sender: the frame is what claims "here is the whole buffer", and `readyCommands` deliberately knows nothing about the terminal widget.

## Related issue

Closes #185

Part of #177.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [x] Docs

## How was this tested?

```
$ bun test --isolate                 # repo root
1007 pass, 0 fail (68 files)

$ cd relay && bun test --isolate
268 pass, 0 fail (14 files)

$ cd relay && bun run typecheck      # relay server
$ cd relay && bun run typecheck:web  # relay web
$ bun run build:main                 # tsc -p tsconfig.main.json
$ tsc -p tsconfig.json --noEmit      # renderer + shared
all four clean, no diagnostics

$ cd relay && bun run build:web
bundles
```

Harness `verify.sh` is green on tests, seams and the comment-wall check.

Two test doubles were themselves wrong, and were part of why the held-fit defect stayed invisible: `FitAddon.fit` now re-measures against a fixed container, and `FakeTerm.resize` actually moves the grid. Before that, a double reported a resize it had not performed.

### What was not verified

**Coverage is RED in `verify.sh`, across four files — not three.**

Three are pure wiring: `relay-client.ts`, `src/main/index.ts` and `src/main/preload.ts`. Each received between 3 and 8 lines with no branch in any of them — construct the dep, forward the IPC, expose the bridge. The repo has no spec for any of the three today, because covering them requires `mock.module('electron')`, which is exactly what the design doc's dependency-injection argument exists to avoid. All behaviour behind that wiring is tested on both sides of it.

The fourth is different, and is a **false green rather than a stated exemption**. `relay/web/src/main.ts` is reported as covered and is not. The checker's `importing_specs` builds its needle from the file stem — here `main`, which is not in `GENERIC_STEMS` — and then matches `/main['"./]`. The literal `src/main/` in an import satisfies that pattern, so **29 specs across the repo match, of which none import this file**; an account-auth test under `src/main/` is credited as covering a browser entry point it has never loaded. Verified at `54e15dc`: no spec imports `relay/web/src/main.ts`, and `requestFit`, `maybeLandInTerminal` and `updateWideBanner` are module-private and referenced by zero specs.

The logic those three call was extracted precisely so it could be tested — `arrival.ts` and `guest-copy.ts` are covered directly — but the wiring inside `main.ts` that calls them is executed by no spec, and the checker will not tell you so.

Also not verified: no manual end-to-end run of the fit request against a real phone and a real desktop. The round trip is covered by tests on each side and at the tunnel, not by one live exercise of the whole path.

## Gate results

**Gate 4 — verifier: VERIFIED.** Every count above was reproduced independently. The tunnel tests drive the production dispatch path rather than a re-implementation, and pin the deny cases. The new capability entry is held by a pre-existing exact-equality assertion against the policy fixture, so the code and the fixture cannot drift apart without a test failing.

**Gate 3 — reviewer: first pass RED**, three majors, each proven by running a probe rather than by reasoning about the code:

1. An accepted fit was undone by a single window-focus event.
2. The throttle was per socket, so six tabs raised six prompts.
3. A reconnected guest never re-subscribed, while the page still read as connected.

**Delta re-check: GREEN.** Both probes were re-run byte-identical and both flipped — the fit held at 80 cols, and six sockets raised one prompt. Four mutations were killed, including the arrival guard that survived the first round. The hold was verified as an invariant across all five `heldFitRef` call sites, so an owner cannot be stranded at a phone-sized terminal.

## For review attention

**(a) Picker sectioning.** The design writes `SHARED WITH ME — Will's laptop`. That was read as "section header plus row", so it is implemented as one "Shared with me" section with person-labelled rows sorted by owner, rather than one section per person. If per-person headings were what was meant, that is a one-line change in `machineSections()`.

**(b) The owner prompt renders on the terminal it is about.** An owner sitting in Analytics or in another session will not see it. This is deliberate: the prompt decides nothing about access, so unlike a join request it does not bypass `shouldNotify()` and does not raise an OS notification. The guest's 45s revert is what keeps that honest — an unseen ask expires and can be asked again. The banner is an `<output>` element rather than `role="status"`.

**(c) "Resize once" now means "until you take it back", not "until the next repaint".** This is the one review finding that changed the shipped behaviour rather than the code behind it, so the button's label is worth a second opinion: it still reads `Resize once`, while the hold persists until the owner clicks "Resume desktop size" or a desktop size arrives over `pty:resized`.

**(d) Test doubles.** `Terminal.restart.test.tsx`'s `electronAPI` stub was extended with `onRelayResizeRequest` — a faithful superset. The new `Terminal.resizeRequest.test.tsx` mocks xterm the same way that file does. Both rely on `--isolate`.

## Merge note

Initiative #183 (web push) is concurrently editing `relay/web/src/main.ts` and `sw.js`. Whichever of the two lands second will need a merge-forward before it goes in. This PR's changes to `main.ts` are substantial, so that merge wants a real read rather than an automatic resolution.

## Checklist
- [x] Tests pass locally

Order: Bodhilander

### DIFF
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -499,9 +499,15 @@ If neither lands, cut phone guests from M5.2 rather than ship an illegible prima
 
 The request travels as `terminal:resize-request {sessionId, cols, rows}`, which requires only
 `view`: it resizes nothing, so a guest role never needs `resize`. The tunnel clamps the size,
-scopes it to a live subscription, throttles it (each one interrupts the owner) and forwards it
-to the renderer, which performs the resize itself on accept. Declining sends nothing back, so
-a refusal reaches the guest as silence and their view is left exactly as it was.
+scopes it to a live subscription, throttles it **per grant** — a guest may open as many sockets
+as they like, so a per-socket throttle is one refresh away from free — and forwards it to the
+renderer, which performs the resize itself on accept. Declining sends nothing back, so a
+refusal reaches the guest as silence and their view is left exactly as it was.
+
+An accepted fit is **held** against that window's own re-fits until the owner takes it back
+through the existing `Resume desktop size` banner. Without the hold the next focus or window
+resize re-measures an unchanged container and pushes the desktop grid back — undoing the grant
+silently, seconds after it was given, with the guest panning again and nobody told.
 
 **Ending states get distinct copy**, driven by a reason enum the agent already knows —
 `revoked` / `expired` / `session_ended` / `machine_unlinked`. Telling a guest *"Will revoked
--- a/docs/designs/session-sharing.md
+++ b/docs/designs/session-sharing.md
@@ -497,6 +497,12 @@ size with horizontal panning (`overflow-x:auto`, two-axis drag) plus an honest b
 request that surfaces on the owner's desktop as a one-tap `[ Resize once ] [ Keep my size ]`.
 If neither lands, cut phone guests from M5.2 rather than ship an illegible primary surface.
 
+The request travels as `terminal:resize-request {sessionId, cols, rows}`, which requires only
+`view`: it resizes nothing, so a guest role never needs `resize`. The tunnel clamps the size,
+scopes it to a live subscription, throttles it (each one interrupts the owner) and forwards it
+to the renderer, which performs the resize itself on accept. Declining sends nothing back, so
+a refusal reaches the guest as silence and their view is left exactly as it was.
+
 **Ending states get distinct copy**, driven by a reason enum the agent already knows —
 `revoked` / `expired` / `session_ended` / `machine_unlinked`. Telling a guest *"Will revoked
 your access"* when Will merely closed a terminal is a false, socially loaded story.
--- a/fixtures/sharing-v1.json
+++ b/fixtures/sharing-v1.json
@@ -58,7 +58,10 @@
       "reachable. 'owner' is NOT a certificate role -- there is deliberately no persisted",
       "owner certificate (§3), so it never appears in mintableRoles.",
       "'create' and 'browse' belong to no guest role, which is what keeps session:create,",
-      "group:create and dirs:list unreachable for any guest, ever (a v1 non-goal)."
+      "group:create and dirs:list unreachable for any guest, ever (a v1 non-goal).",
+      "'terminal:resize-request' requires only 'view' because it resizes nothing: the guest",
+      "asks, and the owner's desktop decides. If it ever needs 'resize', a phone has been",
+      "given the ability to reflow somebody else's terminal and that is the finding."
     ],
     "caps": ["view", "list", "input", "resize", "create", "browse"],
     "roleCaps": {
@@ -74,6 +77,7 @@
       "terminal:unsubscribe": "view",
       "terminal:input": "input",
       "terminal:resize": "resize",
+      "terminal:resize-request": "view",
       "session:create": "create",
       "group:create": "create",
       "dirs:list": "browse"
--- a/relay/web/src/arrival.test.ts
+++ b/relay/web/src/arrival.test.ts
@@ -10,6 +10,7 @@ import {
   machineMenuTitle,
   machineSections,
   planArrival,
+  showSectionTitles,
   type ArrivalMachine,
 } from './arrival';
 
@@ -108,7 +109,13 @@ describe('machineSections', () => {
 
 describe('machineMenuTitle', () => {
   test('a guest never reads the word "machines" as though they owned any', () => {
-    expect(machineMenuTitle([shared(), shared({ id: 'g2' })])).toBe('Shared with you');
+    expect(machineMenuTitle([shared(), shared({ id: 'g2' })])).toBe('Shared with me');
+  });
+
+  test('the sheet and the section inside it say the same thing, in one voice', () => {
+    // Two names for one list on one screen reads as two different lists.
+    const machines = [shared(), shared({ id: 'g2' })];
+    expect(machineSections(machines)[0]!.title).toBe(machineMenuTitle(machines));
   });
 
   test('anyone with one of their own gets the owner heading', () => {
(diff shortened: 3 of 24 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped, well-tested change with no CI failures and no defect I can point to as merge-blocking from the diff.

## Blocking issues
None.

## Non-blocking issues

- `src/main/api/relay/session-tunnel.ts` (`handleResizeRequest`): the per-asker throttle key falls back through `s.grant.grantId ?? s.principal?.userId ?? clientId`. If a connected session ever has neither a grant id nor a principal user id (I can't see how `s.grant`/`s.principal` are populated elsewhere in this file), the key degenerates to `clientId`, which is fresh per socket and would make the "one asker, one prompt per interval" guarantee (the thing six of the new tests specifically pin down) bypassable by reconnecting. Worth double-checking that grant/principal are always populated for any client that can reach this handler — if they are, this is a dead fallback and no action is needed.
- Client-side cooldown (`FIT_ASK_MS = 45_000` in `relay/web/src/main.ts`) is more conservative than the server-side throttle (`RESIZE_ASK_INTERVAL_MS = 10_000` in `session-tunnel.ts`). Not a bug — a reload resets the client timer but the server still silently drops the repeat — but it means a “successfully asked” UI state can outlive what the server would actually still be blocking, which is a slightly confusing asymmetry if you ever want the two to visibly agree.
- Throttled resize-requests are dropped with only a `log.warn` and no signal back to the caller (by design — matches the “declining is silence” philosophy elsewhere in this feature), so there's no way to distinguish "throttled" from "owner declined" from "owner never saw it" in metrics/logs beyond that one warn line. If this path ever needs on-call visibility (e.g. to tell abuse from a slow owner), it has none today.

## Tests

Coverage on the new logic is thorough — `arrival.ts`, `guest-copy.ts`, `reconnect.ts`, `resizeRequestPrompt.ts`, the tunnel's resize-request handling (throttling, clamping, scoping, multi-socket/multi-grant isolation), and the Terminal.tsx accept/decline/hold-until-resumed flow all have direct unit tests, including the specific regressions they're guarding against (duplicate scrollback replay, silent grid snap-back on focus). The only gaps are the thin IPC wiring in `src/main/index.ts` (`relay:resize-request` forward) and `relay-client.ts` (`emit('resize-request', ...)`), which are essentially one-line pass-throughs and reasonable to leave uncovered given the layers on either side are tested.

William-Long-II:
Checked the throttle-key fallback before merging — it is a dead branch for any client that could abuse it.

Every guest reaches this handler through `grantFromCertificate`, which refuses outright without a principal: `if (!machineId || !principal?.userId) return null;` (`session-tunnel.ts:506`). So a certificate-bearing client always has **both** a `grantId` and a `principal.userId`, and `asker` resolves to the grant id — the reconnect-to-reset bypass isn't reachable from the guest population the throttle exists to bound.

The only paths that can leave both unset are the owner branches (`resolveGrant`, `:482` and `:500`), and an owner holds `resize` directly — they never need to ask, so spamming the request path would only prompt themselves. Hence `clientId` as the last fallback: it keeps the key total without inventing a shared bucket, and nothing that can exploit it can reach it.

On the 45s client cooldown vs the 10s server interval: deliberate in that direction — the client's timer is the longer one, so a reload can only ever make the UI *more* conservative than the server, never claim an ask went through that the server would still refuse. Your point that the two never visibly agree is fair; that's a copy question rather than a logic one, and the asked-state string was just reworded to stop implying delivery at all.

On observability for the throttled path: agreed there's nothing but the one `log.warn` today. Recording it against the sharing epic (#180) rather than widening this PR — that epic already owns the audit-log work, which is where a throttle/decline/never-saw-it distinction belongs.

Issue #200, opened 2026-08-25
Feature: Transfer bundle — export and restore everything that can move


William-Long-II:
### User story / problem

As someone setting up a new machine, I want to take my whole Bodhilander state with me in one file, so that I can restore groups, sessions, history and settings instead of rebuilding them by hand.

Today `group-import-export.ts` carries group and session rows and stops there. It writes each session's `claudeSessionId` into the export, but the `.jsonl` transcript that `--resume` actually reads never travels — so the import produces sessions that look restored and cannot resume. That gap is the reason this phase exists.

This phase is deliberately relay-free. It ships as a backup/restore feature on its own merits, and Phase 2 (#201) reuses the same archive as its payload.

### Proposed solution

A single `.bodhilander-bundle` archive (compressed; format version in a manifest, the way the existing portable JSON already versions itself):

- **manifest** — format version, source app version, source platform, source `userData` root, export timestamp, and the working-directory roots observed across all groups and sessions
- **data** — the portable tables serialized as JSON: `groups`, `sessions`, `session_events`, `chat_events`, `arena_runs`, `arena_responses`, portable `preferences`, `claude_accounts` metadata
- **transcripts** — each account's `projects/**/*.jsonl`, keyed by account id rather than by absolute path

Export extends the existing exporter rather than replacing it; it must keep reading v1 files so an older export still imports.

Import runs in a transaction against the live database and performs three jobs beyond inserting rows:

1. **Rewrite absolute paths.** `claude_accounts.config_dir` points into the source machine's `userData`; it is rewritten to the destination's account root before insert.
2. **Remap working directories.** The manifest's observed roots are presented to the user for mapping (`/Users/will/Work/Repos` → `D:\Projects`). Unmapped roots are allowed — they produce "needs relink" sessions, not failures.
3. **Land the transcripts** under each restored account's config dir, reusing the copy mechanics in `conversation-transcript.ts` / `legacy-claude-seed.ts` rather than a second implementation.

The exclusion list from the epic is enforced at export, not import — a secret that cannot be decrypted on arrival should never be written to the archive in the first place.

### Acceptance criteria

- [ ] Export writes a single archive containing manifest, portable tables, and per-account transcripts
- [ ] `safeStorage` blobs (provider API keys, `teamsTokens`, relay keypair) are absent from the archive; a test asserts their absence rather than trusting the filter
[...]

PR #203, opened 2026-08-25
183 — Web push: the agent seals, the relay forwards, a locked phone learns a session needs you

### LINKED ISSUE #183: Feature: Web push — a locked phone learns a session needs you

William-Long-II:
## Description

A locked phone now learns that a session needs you. When a session enters `waiting` or `error`, the agent raises a Web Push notification that reads

> **deploy-prod** · Waiting for your input

and the relay, which carried that notification end to end, holds no key for that text.

### The zero-knowledge decision — option (b), and it is the headline

The issue asked for this to be decided explicitly rather than defaulted. It is decided, it is **option (b)**, and it is recorded in `docs/designs/remote-hosting-relay.md` §10.

Web Push has two separable halves, and the split falls out of who holds what:

- **Identification** — the VAPID `ES256` JWT that tells the push service who is asking. That is the **relay's** job: the browser subscribed with the relay's application-server key, so only the relay can sign for it.
- **Confidentiality** — the RFC 8291 `aes128gcm` content encoding. That is the **agent's** job, because the payload names a session, and session names are end-to-end today. The agent seals; the relay blind-forwards an opaque body to the endpoint.

The issue asked for (b) to be *costed* before defaulting to (a). It costs one ECDH, two HKDF derivations and one GCM record — about 90 lines of `node:crypto` in `src/main/api/relay/push-seal.ts`, no new dependency. And decisively: **RFC 8291 §5 publishes a complete worked example**, so `push-seal.test.ts` reproduces the RFC's own vectors **byte for byte**. That is verified interop against a published answer, not a round-trip that a matched pair of mistakes on both ends would happily agree on.

The desktop's existing web-push stack was read and **deliberately not reused**. Two reasons, both structural: it needs the subscription endpoint, which the agent is never told (withholding it is the point); and its VAPID details are module-global state shared with the LAN dispatcher living in the same process, so a second consumer would be mutating the first one's configuration.

### What this explicitly does NOT claim

The relay holds `p256dh` and `auth` — the browser hands them over at subscribe time, and there is no way for it not to. **A malicious relay could therefore forge a notification.** It cannot read one.

Option (b) buys confidentiality against the relay *as it actually runs*: its logs, its database, its operators, its backups. It does not buy unforgeability. §10 says exactly this, in those terms, rather than letting "end-to-end" do unearned work.

### Relay

- Three rate-limited HTTP routes for the VAPID public key and subscribe/unsubscribe, persisting to the `push_subscriptions` table that shipped unused in migration 001.
- Two agent-socket messages: `push:sync` sends the subscription keys **down** to the agent with the endpoint deliberately withheld, and `push:send` carries sealed bodies **up**.
- VAPID identity resolves env → kv → minted-and-persisted, so a self-host deployment that sets nothing still comes up with a stable keypair.
- `410 Gone` from a push service reaps the subscription row.
- A per-machine rate limit on the WS path, so the socket is not a way around the HTTP limits.
- `isAllowedPushEndpoint` is enforced at **subscribe and again at send** — the stored URL is re-checked at use, not trusted because it passed once.

### Agent

- `push-seal.ts` — pure crypto, no I/O.
- `push-attention.ts` — `planAttentionPush` holds the **entire** trigger decision: guards, debounce, fan-out. That is what makes it testable without Electron, without a socket and without a PTY.
- `CAP_PUSH_V1` capability gate.
- An **8-line call site** in `handleStateChange`, sitting beside the existing LAN dispatch.
- Debounce mirrors the LAN dispatcher exactly: keyed per `(sessionId, state)`, 30s, bounded map, survives reconnects, cleared on `disable()`. The acceptance criterion asked for the same event semantics; it is the same semantics because it is the same shape.

### Client

- `push.ts` with dependencies injected — no globals, which is why it has 28 tests that run in plain node.
- A notifications switch in the account sheet.
- Service worker `push` and `notificationclick` handlers.
- `?m=` deep link consumed **before** `boot()`, so tapping a notification lands on the right machine rather than on the machine list.

### Carry-overs from #182

All five landed here:

- Cache bumped `bodhi-shell-v1` → `bodhi-shell-v2`.
- Shell caching gated on `res.ok`, with a test.
- **Verified there is no CSP anywhere in the relay** — then moved the offline-page reload from an `onclick` attribute into a script block anyway. A block can be permitted by nonce or hash; an attribute needs `'unsafe-inline'` *plus* `'unsafe-hashes'`. This makes a future CSP cheap instead of blocked.
- `serviceWorker.register()` failure now `console.warn`s instead of failing mute.
- `relay/src/web.test.ts` pins that `/i/:code` still reaches the shell after the static lookup.

## Related issue

Closes #183

Part of #177.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Hotfix
- [ ] Refactor / chore
- [ ] Docs

## How was this tested?

**172 tests added.**

| Suite | Added |
| --- | --- |
| `relay/src/push-http.test.ts` | 29 |
| `relay/src/push-vapid.test.ts` | 32 |
| `relay/src/push-ws.test.ts` | 13 |
| `relay/src/web.test.ts` | 17 |
| `relay/web/src/push.test.ts` | 28 |
| `src/main/api/relay/__tests__/push-seal.test.ts` | 15 |
| `src/main/api/relay/__tests__/push-attention.test.ts` | 18 |
| `relay/web/src/sw.test.ts` | 8 → 20 |
| `relay/src/relay.test.ts` | 18 → 26 |

Suite results:

```
root:            1102 pass / 0 fail  (71 files)
relay standalone: 366 pass / 0 fail  (17 files)
```

Three typechecks clean, `build:web` clean.

The 15 `push-seal` tests include the RFC 8291 §5 worked example reproduced byte for byte — same salt, same server key, same ciphertext.

### Two findings the tests surfaced

Stating these because they are the suite doing its job, not incidental cleanup:

1. **`enablePush` was leaning on browser behaviour when permission is `denied`.** It now short-circuits explicitly instead of asking and hoping.
2. **A config warning had quietly broken `loadConfig`'s "a complete production config is warning-free" contract.** It would have cried wolf on every boot of the documented self-host setup. The warning moved into `push/vapid.ts`, where it fires once — when a keypair is actually minted.

## Coverage — stated under its own heading, and not presented as covered

`verify.sh` is **PASS** on comment-wall, tests and seams, and **RED on coverage**. It names four changed files that no spec imports. Taking them in turn:

**One is a checker artifact.** `relay/src/config.ts` — `config` is in the checker's `GENERIC_STEMS`, so its needle collapses to `src/config`, which no relative `./config` import can ever match. The file **is** genuinely covered by `relay.test.ts`, and 8 tests were added there for the new VAPID parsing, which really did have none.

**Three are entrypoints and glue, and are genuinely not covered:**

- `relay/src/index.ts`
- `src/main/index.ts`
- `src/main/api/relay/relay-client.ts`

Covering `relay-client.ts` would take roughly **11 process-wide module mocks** for code that is pure wiring — and process-wide mocks in this suite leak into files that test the same modules for real. Extracting `planAttentionPush`, which is fully tested, was the alternative chosen instead: the decision-making moved out of the untestable file rather than the untestable file being wrapped in mocks. The wiring that remains is uncovered and that is the honest state of it.

## Expected Sonar hotspots

Naming these up front so nobody is surprised by the count. Everything landed in one push, so PR analysis sees the change whole.

- `src/main/api/relay/push-seal.ts` — `createECDH`, `hkdfSync`, `createCipheriv`.
- `relay/src/push/vapid.ts` — key generation and JWT signing.
- `relay/src/push/send.ts` — outbound `fetch` to a stored URL. This is the SSRF flag; mitigated by `isAllowedPushEndpoint` at both subscribe and send, and argued in §10.
- `relay/web/src/push.ts` — base64 / `atob` handling.
- Possibly `relay/web/sw.js` — the `renotify` / `showNotification` surface.

## Merge note

Concurrent initiative #185 (#198) also touches `relay/web/src/main.ts` and `relay/web/sw.js`. Changes here were confined to the account-sheet region and the bottom of each file to keep the two apart, but whichever lands second may need a merge-forward.

## For review attention

Five things the owner wants a human decision on rather than a silent default:

1. **DNS-rebinding residue.** A hostname that *resolves* to a private address still passes the allow-list. §10 records this as "do not treat the relay's outbound egress as trusted." Is that acceptable for the hosted deployment?
2. **Forgeability.** Closing it needs the agent to sign the payload with ed25519 and the service worker to verify. Deliberately not done — is that the right call?
3. **`CAP_PUSH_V1` gating means an un-updated desktop silently sends no push.** Honest, but silent. Should it surface in the client instead?
4. **Guests get no push by design** — fan-out goes to the machine owner only. That is a sharing-policy question the owner did not feel entitled to decide.
5. **`push_subscriptions` has no reaper for rows whose owner never returns.** They die only on a `410`. Is endpoint-liveness-on-410 sufficient, or should the table age rows out?

## Checklist
- [x] Tests pass locally


William-Long-II:
Brought this up to date with `development` — it was 74 behind and had been sitting since 25 Aug. `test` is green again and the PR is `MERGEABLE`. Two blockers remain, and **both are dashboard actions I can't take**.

## What the merge involved

74 commits, including the whole machine-handoff epic (#200/#201/#202). Five conflicts, every one where web push and the handoff had independently added to the same file.

Four were additive and resolved by keeping both sides. Two needed hands:

- **`relay-client.ts`** — both sides import from `./grants`, so concatenating would have declared `CAP_GRANTS_V1` and `buildShareCreateMessage` twice.
- **`http.ts`'s fourth hunk is not additive**, which cost me a first attempt. Our side ends by closing `createRouter` and then adds two top-level helpers; the incoming side is handoff handlers that belong *inside* that closure. Concatenated, the handlers landed in the body of `isAuthSecret` and lost `repos`, `config` and `limiter`. Spliced at the brace instead.

**1700 pass / 1 fail** under bun 1.4.0 — the failure is `account-auth`'s `a token file appearing still completes the flow` timing out at 15s under full-suite load, which is the #228 flake, present on `development` too and green in isolation. 111 spec files run, up from 103, so the web-push suites are live. No unhandled errors, so nothing is silently contributing zero tests. Both typechecks clean.

## Blocker 1 — GitGuardian, and I believe it is a false positive

One finding: *Generic High Entropy Secret* in `src/main/api/relay/__tests__/push-seal.test.ts`, in the `RFC8291` block.

Those values are the **published test vector from RFC 8291 §5, "Push Message Encryption Example"** — plaintext `"When I grow up, I want to be a watermelon"`, and the receiver/sender keypairs, auth secret and salt printed alongside it in the RFC. They are in the standard precisely so implementations can check themselves against a known answer, and nothing anywhere is protected by them.

They are also the most valuable thing in that file. The spec's own header says it:

> The first test is the one that matters: it reproduces the worked example in RFC 8291 §5 byte for byte. […] The RFC vector is an answer nobody here chose, so agreeing with it is evidence a browser can read what we send.

Deleting or obfuscating them to satisfy the scanner would trade the one test that can catch a matching pair of mistakes for a green check. **The right resolution is to mark incident 36589257 as a false positive in the GitGuardian dashboard**, which needs someone with access. A repo-side `.gitguardian.yaml` would not help — the check is the GitGuardian app's platform scan, configured there rather than here.

## Blocker 2 — Sonar security hotspots

```
Quality Gate failed
- 0.00% Security Hotspots Reviewed on New Code
```

New issues are now at **zero** — the merge cleared whatever the pre-merge run was flagging. The only condition left is hotspot review, which is a person marking them in the SonarQube UI; the list is not in the check run, the annotations, or the bot comment, and `api/hotspots/search` needs a token. Same hand-back as #234.

Given what this PR does — VAPID keys, ECDH, HKDF, a server that makes outbound requests to push endpoints — hotspots are expected here and reviewing them is genuine work, not a rubber stamp.

## After those

Still a draft, and `REVIEW_REQUIRED`. Once the two dashboards are cleared it needs taking out of draft and a review.

Issue #199, opened 2026-08-25
Epic: Machine handoff — sign in on a new machine and keep going


William-Long-II:
### Summary / goal

Get a new machine to the state the old one was in: sign in, restore, and pick every session back up mid-conversation. Not file-level sync between live machines — a one-directional handoff that ends with the new machine being the one you work on.

The user-visible promise is narrow and testable: **every session that was resumable on the old machine is resumable on the new one.** Terminal processes do not move; conversations do.

### Why / business value

Today a new machine is a full manual rebuild. `group-import-export.ts` moves group and session rows and nothing else — no conversation transcripts, no preferences, no accounts, no analytics history. The `claude_session_id` it carries is a promise the import cannot keep, because the `.jsonl` transcript that `--resume` actually reads stays behind on the old machine. So the one thing a user most wants preserved is precisely the thing that silently does not come across.

Everything needed to fix that already exists in pieces:

- `src/main/group-import-export.ts` — versioned portable JSON of groups + sessions
- `src/main/conversation-transcript.ts` `carryTranscript()` — copies transcripts between config dirs, with path-safety checks
- `src/main/legacy-claude-seed.ts` — bulk-copies a whole `projects/` tree into a fresh account dir
- the relay's `users` → `machines` model, where a second machine under one GitHub identity is already schema-legal
- the relay's existing posture of storing ciphertext it cannot read

This epic assembles those into one flow rather than inventing a new subsystem.

### Child issues / scope

- [x] #200 — Phase 1, transfer bundle: archive format, export, import, path remapping. Standalone value as a backup/restore feature; no relay dependency.
- [x] #201 — Phase 2, relay handoff: the new machine signs in, sees the old machine's bundle, restores. Client-side encryption; the relay stores an opaque blob.
- [x] #202 — Phase 3, arrival polish: repo-root auto-detection for remapping, inline account re-auth, an honest report of what came back resumable.

### What moves, and what does not

Decided up front so it is chosen rather than discovered.

**Moves:** `groups`, `sessions`, `session_events`, `chat_events`, `arena_runs`, `arena_responses`, portable `preferences`, `claude_accounts` metadata, and the per-account `projects/**/*.jsonl` transcripts.

**Deliberately excluded:**

- `push_subscriptions` — endpoints bound to the old machine's pairings.
- The relay sharing tables — `machine_grants.certificate` is signed by the old machine's key and is meaningless under a new identity.
[...]

PR #205, opened 2026-08-25
200 — Transfer bundle: export and restore everything that can move

### LINKED ISSUE #200: Feature: Transfer bundle — export and restore everything that can move

William-Long-II:
Phase 1 of the machine-handoff epic #199.

## What this delivers

One `.bodhilander-bundle` archive that carries a machine's state to a new machine: groups, sessions, history, portable settings, account metadata, and — the part that was missing — the `.jsonl` transcripts that `--resume` actually reads.

The gap this closes: `group-import-export.ts` already wrote each session's `claudeSessionId` into its export, but the transcript stayed behind on the source machine. The import produced sessions that looked restored and could not resume.

This phase is relay-free and stands on its own as backup/restore.

### Format — `src/main/transfer/bundle-format.ts`

A versioned container (`BUNDLE_FORMAT_VERSION = 2`, magic `BDHLBNDL`, extension `.bodhilander-bundle`) holding:

- **manifest** — format version, source app and version, source platform, source `userData` root, export timestamp, the distinct working-directory roots observed across every group and session, and per-table counts
- **`data/tables.json`** — `groups`, `sessions`, `session_events`, `chat_events`, `arena_runs`, `arena_responses`, portable `preferences`, `claude_accounts` metadata
- **`transcripts/<accountId>/…`** — each account's `projects/**/*.jsonl`, keyed by account id rather than by the source machine's absolute path

The v1 portable JSON shapes are now defined once here and consumed by `group-import-export.ts`, so the two formats cannot drift apart.

### Export — `src/main/transfer/bundle-export.ts`

Reached through the existing `exportGroupsAndSessions` entry point rather than a second command. The user chooses which format to write; "Groups & sessions only" produces the same v1 JSON as before.

Exclusion is enforced at export, not import, so a secret that could not be decrypted on arrival is never written to the archive in the first place. `isPortablePreferenceKey` refuses the `providerApiKey.`, `providerApiKeyUse.` and `relay.` prefixes, and the keys `teamsTokens`, `windowBounds`, `customShellPath`, `preferredEditor`, the four `sound*CustomPath` keys, and the two legacy cleanup flags.

Archive size is computed and shown to the user before the file is written (`sizeLabel`, via `formatBytes`).

### Import — `src/main/transfer/bundle-import.ts`

One transaction against the live database, doing three things beyond inserting rows:

1. **Absolute paths are rewritten.** `claude_accounts.config_dir` points into the source machine's `userData`; it is rewritten to the destination's account root before insert.
2. **Working directories are remapped.** `src/main/transfer/working-dirs.ts` collects the distinct roots for the manifest and applies the user's mapping per root. It parses both platforms' paths itself rather than using `path`, whose separator is fixed to the host — a bundle written on macOS is imported on Windows and back.
3. **Transcripts are landed** under each restored account's rewritten config dir, reusing `carryTranscript()` and the `legacy-claude-seed.ts` copy mechanics rather than a second implementation.

Unmapped roots are allowed and produce needs-relink sessions, not failures. Transcript entries whose path escapes their account prefix are skipped and logged.

### Needs-relink — `src/shared/types.ts`, `SessionRow.tsx`, `App.tsx`

A restored session whose mapped directory is not present here gets `state: 'needs-relink'` (`NEEDS_RELINK_STATE`). It never reaches the `pty-manager` spawn path, which throws on a missing `cwd` and surfaces as a spawn failure with nothing to act on. In the sidebar its status pill becomes a button titled with the missing folder; clicking it opens a directory picker and returns the session to `idle`. `updateSession` in `src/main/repositories/sessions.ts` gained a `workingDir` field so the relink can persist.

## What was verified

Harness verification at `feat/200-Bodhilander` (head `26d3079`), plugin copy `bodhi/2.18.4`:

```
$ scripts/verify.sh Bodhilander=feat/200-Bodhilander --initiative initiatives/200
verify Bodhilander at feat/200-Bodhilander
  worktree /Users/will/Work/Repos/_wt-200-Bodhilander  (HEAD confirmed == feat/200-Bodhilander)
  base origin/development
  19 file(s) changed
  8% of added lines are comment (169/2253)
  specs:   23 enumerated (never a glob)

  PASS         tests
  SKIP         lint
  PASS         seams
```

The transfer specs and the suites the change touches:

```
$ bun test --isolate src/main/transfer/__tests__/ \
    src/main/__tests__/group-import-export.test.ts \
    src/renderer/components/__tests__/SessionRow.test.tsx

 101 pass
 0 fail
 250 expect() calls
Ran 101 tests across 7 files. [1.75s]
```

Covered by those specs: the archive round-trips; `safeStorage` blobs, `push_subscriptions`, the relay sharing tables and machine-local preferences are asserted **absent** from a built archive rather than assumed filtered; `config_dir` rewriting; root collection and remapping across both platforms' path shapes; the needs-relink state and its sidebar control; transaction rollback leaving the existing database untouched; re-import not duplicating groups or sessions; a v1 file from the current exporter still importing; and `resume-end-to-end.test.ts` following a restored session to a readable transcript at the rewritten path rather than asserting a row exists.

## What was NOT verified

- **Lint did not run** — `verify.sh` reported `SKIP` for it, not `PASS`. Nothing here has been linted.
- **No type-check was run** on this branch independently of the test run.
- **Sonar has not reported.** That is the reason this PR is a draft: analysis cannot start until the PR exists.
- **No manual end-to-end run** of a real export on one machine and import on another. The resume path is covered by an automated test against a fixture database, not by a human moving a bundle between two machines.
- **Archive size behaviour is untested at realistic volume.** The size label is asserted in a spec against small fixtures; no bundle built from a large transcript corpus was measured.

## One defect found while opening this PR

`src/main/transfer/working-dirs.ts` contains a **literal NUL byte** at line 81, used as a separator in a composite map key:

```ts
const key = `${parsed.prefix}\x00${parsed.absolute}`;
```

That is a raw `0x00` in the file, not the two-character escape. The file is otherwise valid UTF-8, TypeScript parses it, and the tests pass — but git classifies the whole file as binary, so it appears in this diff as `Bin 0 -> 4503 bytes`. Consequences: 4.5 KB of new logic cannot be read in the GitHub diff view, Sonar is likely to skip the file, and every future change to it will be undiffable too.

Writing the separator as the escape sequence `\u0000` instead of a raw byte restores the file to text with no behaviour change. Flagged rather than fixed here — this dispatch does not edit source.

## Review state

Gates 3 (review) and 4 (verification) have **not** reported. This PR is deliberately a draft so that Sonar analysis runs while they do. Their verdicts and evidence will be carried into this body before it leaves draft, and reviewers are added at that point.

Closes #200


src/main/transfer/bundle-import.ts
```diff
@@ -0,0 +1,464 @@
...
+    VALUES (?, ?, ?, ?, ?, 0, ?, ?)
+  `);
+
+  let inserted = 0;
+  for (const account of tables.accounts) {
+    if (existing.has(account.id)) continue;
+    insert.run(
+      account.id,
+      account.label,
+      destinationConfigDir(options.accountsRoot, account.id),
```
brannon-bowden, line 335: account.id from the imported bundle is used unsanitized to build destinationConfigDir(accountsRoot, account.id); a crafted id with '../' segments escapes accountsRoot. Downstream, registerRestoredAccountHooks() writes settings.json into this path for every account, turning this into an arbitrary file write via a malicious .bodhilander-bundle. Validate account.id (e.g. reuse the isSafeSegment check used for transcript paths) before using it here.

brannon-bowden (changes requested):
Request changes — this is a well-tested feature (the working-dir remapping, transcript path-safety, and preference-exclusion tests are genuinely thorough), but the account-restore path in the bundle importer has an unsanitized path-traversal that becomes an arbitrary file write once hooks are re-registered, and that needs to be closed before this ships.

## Blocking issues

- **`src/main/transfer/bundle-import.ts` (`restoreAccounts`, ~line 298-318, and `destinationConfigDir`, ~line 349-351): account IDs from the imported bundle are never validated before being used to build a filesystem path.**
  `tables.accounts[].id` comes straight from parsed, attacker-controllable JSON (the `TABLES_ENTRY` payload has no schema/format validation — `parseBundle` only checks that `groups`/`sessions` are arrays for the v1 shape). That `account.id` is used unmodified at `destinationConfigDir(options.accountsRoot, account.id)` → `path.join(accountsRoot, accountId, '.claude')` and stored as `claude_accounts.config_dir`.
  Compare this to the transcript path, which is careful: `stageTranscripts`/`isSafeSegment` in the same file reject any path segment containing `/`, `\`, `.` or `..` before it's used to build a filesystem path. No equivalent check exists for `account.id`.
  The impact is concrete, not theoretical: after `restoreTransferBundle` commits, `group-import-export.ts`'s `registerRestoredAccountHooks()` (~line 236-243) calls `registerHooks(account.configDir)` for *every* account in the DB, including the malicious one just inserted. Per this PR's own test (`a restored account gets its hooks registered…`), `registerHooks` writes a `settings.json` containing the hook wiring into that directory. A crafted `account.id` such as `"../../../../.claude"` (or any relative traversal appropriate to the destination machine) resolves outside `accountsRoot` and lets an imported `.bodhilander-bundle` write/overwrite `settings.json` anywhere the app process can reach — including the user's real `~/.claude/settings.json`, which is exactly the file that wires up hook execution for the actual Claude Code CLI.
  Since this feature's whole premise is a portable file meant to be moved between machines (and plausibly shared/handed to someone), this is a realistic attack: "import my bundle" from an untrusted source becomes an arbitrary-file-write / hook-injection primitive.
  Fix: validate `account.id` the same way transcript path segments are validated (e.g. reuse `isSafeSegment`, or require it to match the UUID shape the app actually generates) before it's used in `destinationConfigDir`/inserted into the DB, and reject or drop accounts that fail.

## Non-blocking issues

- **`src/main/transfer/bundle-format.ts` (`decodeBundle`) — `zlib.gunzipSync` is called on each entry with no output-size ceiling.** `buildTransferBundle` bounds the *export* side via `MAX_TRANSCRIPT_BYTES`, but on *import* a crafted bundle with a small, highly-compressed entry could expand to consume large amounts of memory before any size check runs. Worth capping decompressed size per entry (or overall) given the bundle is untrusted input at import time.
- **`src/main/transfer/bundle-format.ts` (`isPortablePreferenceKey`) — the export filter is a denylist** (`EXCLUDED_PREFERENCE_PREFIXES` / `EXCLUDED_PREFERENCE_KEYS`) rather than an allowlist. It's well covered by tests today, but any future preference key that happens to hold a secret will be exported by default unless someone remembers to add it here. An allowlist of known-portable keys (or at least a naming convention like `secret.*`/`local.*` enforced elsewhere) would fail safer as the preferences surface grows.

## Tests

- No test exercises a hostile/malformed `account.id` (or `group.id`/`session.id`) in the tables payload — the transcript-path-traversal test (`an entry naming a path outside its account is refused`) has no analogue for account restoration. Once the blocking issue above is fixed, a test asserting that a traversal-shaped `account.id` is rejected (or safely neutralized) rather than reaching `config_dir`/`registerHooks` would pin the fix.
- No test covers a decompressed-entry size ceiling on import, if one is added per the non-blocking note above.

William-Long-II:
Thanks — the blocking finding was real and the exploit chain checked out end to end. Confirmed before fixing: `account.id` reached `path.join(accountsRoot, id, ".claude")` with no validation, that path was stored as `claude_accounts.config_dir`, and `registerRestoredAccountHooks()` then called `registerHooks(configDir)` for every account in the DB, which lands in `writeClaudeSettings` — so a crafted id wrote `settings.json`, the file that wires up hook execution, at a location the bundle author picked.

## Blocking — fixed

`account.id` is now validated before it can reach a path, against the same rule conversation ids already follow in `conversation-transcript.ts` (`/^[A-Za-z0-9][A-Za-z0-9._-]*$/`). That accepts the UUIDs the app actually mints and refuses separators, `..`, a leading dot, drive letters and non-strings.

Two deliberate choices:

- **The check runs up front in `restoreRows`,** before a single row is written, rather than at the point of use partway through. The guard is then obviously total and does not depend on the evaluation order of the counts object literal.
- **A bad id fails the whole import** rather than dropping that account. The restore is one transaction, a legitimate bundle never trips this, and dropping would leave groups and sessions pointing at an account that no longer exists.

I also traced the transcript path, since it builds a config dir from a bundle-chosen key too: `landTranscripts` skips any `accountKey` not in `idsIn(db, "claude_accounts")`, which is read back after the transaction commits — so it can only reach ids that passed the guard. No change needed there.

## Non-blocking — fixed

**Decompression ceiling.** `decodeBundle` now passes `maxOutputLength` to `gunzipSync`, defaulting to `MAX_ENTRY_OUTPUT_BYTES` (1 GiB, matched to the export-side transcript budget so nothing this app writes can trip it), and translates `ERR_BUFFER_TOO_LARGE` into a readable message. `decodeBundle` takes an optional `maxEntryBytes` so the limit itself is testable without allocating a gigabyte.

## Non-blocking — not taken

**Denylist → allowlist for preferences.** I agree with the direction and left it. Flipping it now would silently drop legitimate settings on every existing bundle, and the failure mode it guards against is a future key that does not exist yet — that is a change worth making deliberately against the whole preferences surface rather than inside this PR. Happy to file it against the epic if you want it tracked.

## Tests

- `an account id shaped like a traversal is refused, and nothing is written` — asserts the throw, that no directory appears outside `accountsRoot`, and that `claude_accounts` is left empty.
- `a separator in an account id is refused too` — covers `a/b`, `a\b`, `..`, `.hidden` and `""`.
- `the uuid the app actually mints is still accepted` — pins that the guard is not over-tight, and that `config_dir` lands where it should.
- Three tests on the decompression ceiling: refused over the cap, whole under it, and the shipped 1 GiB default passing a 5 MB entry.

I checked the traversal tests are not vacuous: with the guard commented out both go red, and green again with it restored.

Full suite 1263 pass / 0 fail, relay 268 pass / 0 fail, and all four checks are green on `9c0f18e`.

One thing unrelated to this PR, flagged for visibility: the suite has intermittent load flake on `development`'s code — `session-tunnel-policy`, `account-auth` login detection, and `POST /sessions/:id/stop`. That last one lives in a file this PR touches, so I checked it specifically: the failing test pre-exists on `development` unchanged, and the file passes 5/5 in isolation. It only fails in the full 83-file concurrent run under load. Worth its own issue rather than something to chase here.

brannon-bowden (approved):
Approve — the account-restore path traversal I previously blocked on is now closed, with tests that pin the fix.

The blocking issue from the last review was that `account.id` from an imported bundle flowed unsanitized into `destinationConfigDir(accountsRoot, account.id)`, letting a crafted bundle write `settings.json` (via `registerHooks`) outside `accountsRoot` once hooks are registered post-restore. This diff adds `assertRestorableAccountId` (`src/main/transfer/bundle-import.ts`) using `PATH_SAFE_ACCOUNT_ID = /^[A-Za-z0-9][A-Za-z0-9._-]*$/`, which:

- Rejects any path separator (`/`, `\`) and drive-letter colons — neither character is in the allowed set.
- Rejects a leading `.`, which blocks both `..` and dotfile-shaped ids (`.hidden`) outright, since the first character must be alphanumeric.
- Rejects the empty string, since the regex requires at least one leading alphanumeric character.
- Still accepts the UUID shape the app actually mints.

The check is applied both inside `restoreAccounts` (right before insert) and upfront in `restoreRows` (before any row is written), which is a deliberate belt-and-suspenders choice per the inline comment — fail the whole transaction before any partial writes rather than at point-of-use. That's sound given the restore is meant to be all-or-nothing.

Test coverage for this is solid: `bundle-import.test.ts` adds cases for a `../../evil`-style id, a battery of other hostile shapes (`a/b`, `a\b`, `..`, `.hidden`, `''`), and confirms the real UUID shape still round-trips and lands at the expected `config_dir`. Each hostile case asserts both the rejection message and that nothing was written to disk or the destination DB.

The other non-blocking note from last time — unbounded `gunzipSync` output on import — is also addressed: `decodeBundle` now takes a `maxEntryBytes` ceiling (default 1 GiB, matched to the export-side transcript budget), passes it as `maxOutputLength` to `gunzipSync`, and translates the resulting error into a clear message. `bundle-format.test.ts` covers both a bomb tripping a tight ceiling and one passing under a looser one, plus a sanity check that the shipped default doesn't clip legitimate payloads.

I don't see new issues introduced by this delta. CI is green across all four checks (SonarQube, quality-gate, test, GitGuardian).

## Suggestions (non-blocking, carried over)
- `bundle-format.ts` (`isPortablePreferenceKey`) still filters exported preferences via a denylist rather than an allowlist. Still well-tested today, but worth converting to an allowlist (or an enforced naming convention) as the preferences surface grows, so a future secret-holding key doesn't get exported by default.

PR #208, opened 2026-08-25
207 — Account failover: when one account runs dry, the work moves and the conversation follows

### LINKED ISSUE #207: Feature: Fail over to another Claude account when one hits its usage limit
### User story / problem

As someone running several sessions against a Claude Max subscription, I want work to continue on another registered account when the one I'm on runs out of quota, so that a five-hour limit doesn't stop every session at once and force me to reassign them by hand.

Multi-account support (BDHLNDR-31) already gives each account an isolated `CLAUDE_CONFIG_DIR`, and #164/#165 already carry a conversation across accounts and report which ptys are billing which account. What's missing is the reaction: when the CLI announces a usage limit, nothing notices, nothing moves, and every session on that account sits blocked until a person intervenes.

### Proposed solution

Detect the limit in the pty output scanner (the same place `waitingPatterns` is read), record it against the account, and move that account's live sessions to the next one in an explicit failover order.

- **Detect** — provider-declared `usageLimitPatterns`, with the reset time parsed out of the CLI's own message: the `|<epoch>` suffix headless mode prints, `resets Nov 12 at 9am`, `reset at 3pm`, `in 42 minutes`. No reset named falls back to a five-hour window (Claude's rolling window).
- **Mark** — `claude_accounts` gains `limited_until` / `limited_at`. Expiry is decided at read time; nothing sweeps the column.
- **Move** — every *live* session on the spent account, not just the one that noticed, since the limit is the account's. Target is the next healthy account by a user-set `fallback_rank`.
- **Return** — once the limit lifts, sessions go back to their original account, but only while idle or stopped, so a switch back never interrupts a turn in progress.

Both halves are switchable in Settings → Claude Accounts, defaulting on.

### Acceptance criteria

- [ ] A usage-limit announcement in a session's output marks that account limited until the reset time named in the message
- [ ] An announcement with no reset time falls back to a bounded default window rather than an optimistic guess
- [ ] Every live session on the spent account is moved, not only the session that reported the limit
- [ ] The replacement is chosen by an explicit, user-reorderable failover order, skipping accounts already cooling down
- [ ] A moved session resumes its conversation — verified against the transcript-carry path, not by asserting the row changed
- [ ] With no other account available, the user is told so explicitly rather than the event passing silently
- [ ] A session that inherited its account from its group returns to *inheriting*, not pinned to what that inheritance resolved to
- [ ] Two hops (A limited, then B limited) still return the session to A, not B
- [ ] A session the user reassigned by hand is never moved back by a pending fail-back
- [ ] Fail-back only fires while a session is idle or stopped
- [ ] Detection does not fire on output that merely quotes the trigger phrase (source, diffs, grep hits)

### Scope

Major feature

### Priority

High

William-Long-II:
Closes #207.

Multi-account support (BDHLNDR-31) gave each account an isolated `CLAUDE_CONFIG_DIR`; #164 and #165 made a conversation follow a session across accounts and made the UI honest about which account a pty is actually billing. What was missing was the reaction. The CLI would print its usage limit, every session on that account would sit blocked, and moving them was manual, one session at a time.

## How it works

**Detect** — `usage-limit.ts`. There is no structured channel for this: the CLI announces a limit by printing it, so this rides the existing output scanner next to `waitingPatterns`. The reset time is read out of the message itself — the `|<epoch>` suffix headless mode prints, `resets Nov 12 at 9am`, `reset at 3pm`, `in 42 minutes`. A message naming no reset falls back to a five-hour window (Claude's rolling window) rather than an optimistic guess: an account handed back early is a second failover and a second restart for every session on it.

**Mark** — `claude_accounts.limited_until` / `limited_at`. Nothing sweeps the column; expiry is decided at read time, so every reader agrees on it. A second, vaguer report can't shorten a precise cooldown.

**Move** — `account-failover.ts`, policy only. It never spawns or kills a pty — a switch reaches the CLI as `CLAUDE_CONFIG_DIR`, fixed at spawn, so applying one takes a respawn and only the renderer owns those. Every *live* session on the spent account moves, not just the one that noticed; the limit is the account's, so the others are already dead in the water. The target is the next healthy account in a user-set order.

**Return** — the point of a secondary account is to absorb an outage, not to become the new home. Sessions go back once the limit lifts, but only while idle or stopped, so a switch back can never interrupt a turn in progress.

## The part that had to be right

Failover routes through `account-switch`, so the transcript follows. Combined with the launch-time carry from #164 — which searches every config dir and picks newest-and-not-smaller — and the renderer's existing kill-before-spawn ordering, the respawn's `--resume` finds the conversation instead of starting over. If a resume still fails, `transcriptStaged === 'carried'` means the conversation id is kept and the exit surfaces; it cannot silently start you over.

Three row-level cases the bookkeeping exists for, each covered by a test:

- a session that **inherited** its account from its group returns to *inheriting*, not pinned to whatever that resolved to that day
- **two hops** (A limited, then B limited) still go home to A, not to the equally-spent B in between
- a session the user **reassigned by hand** is never moved back by a decision the app made hours earlier — and one with a switch already pending is restarted rather than re-pointed, so their choice is applied instead of overwritten

## The risk this design carries

Detection is text-matching, and terminal output is an open channel — an agent reading these very files puts the trigger phrase on screen. It is bounded on three sides: the patterns are narrow (`limit` and `rate` alone appear constantly and are deliberately absent), matches on lines shaped like code, comments, diffs or grep hits are rejected, and a pty reports at most once. What that misses is bounded downstream: failover never targets an account already cooling, and the whole thing is switchable.

I could not verify against a real limit event. If your Claude Code build words its message differently, detection won't fire — the log line `[Accounts] Session … reported a usage limit` is the thing to watch. It can be exercised without waiting for a real limit by echoing the phrase into a session.

## Known limitation

A **group** still points at the spent account. Sessions get a per-session override; rewriting the group's assignment would affect sessions that were never running. A new session in that group starts on the spent account and fails over on its own within a turn — self-correcting, but not free.

## UI

Settings → Claude Accounts gains the failover order (↑/↓, numbered), a `limited · back 9:30 PM` badge with a **Clear limit** escape hatch for when the reset time was a guess, and two toggles — switching accounts, and moving back — both defaulting on and separately switchable, because wanting the first without the second is a reasonable position. An in-window notice reports each switch and stays until dismissed; the desktop notification does not suppress on a focused window, since this happened without being asked for.

## Verification

- `bun test` — 906 pass, 0 fail (35 new: 18 policy, 12 detection, 5 notice copy)
- `tsc` clean on both projects; `bun run build` green
- New columns are additive with `PRAGMA table_info` guards; with no ranks set, `accounts:list` returns byte-for-byte the previous ordering, so nothing moves for an estate whose owner never touches the order

### DIFF
--- a/src/main/account-failover.ts
+++ b/src/main/account-failover.ts
@@ -1,0 +1,321 @@
+import log from 'electron-log';
+
+import {
+  AccountFailoverEvent,
+  ClaudeAccount,
+  LiveAccountBindings,
+  SessionState,
+} from '../shared/types';
+import { resolveAccountForSession } from './account-resolver';
+import { assignSessionAccount } from './account-switch';
+import * as accountsRepo from './repositories/accounts';
+import { getPreference } from './repositories/preferences';
+import * as sessionsRepo from './repositories/sessions';
+
+/**
+ * What happens when an account runs out of quota mid-session.
+ *
+ * The mechanics of moving a session between accounts already exist — the user
+ * can do it by hand, and account-switch.ts carries the conversation across and
+ * reports which ptys need respawning. This module is only the policy on top:
+ * which account is spent, which one takes the work, and when the session is
+ * allowed to go back.
+ *
+ * Two things it deliberately does NOT do. It never spawns or kills a pty —
+ * a switch takes effect on respawn, and the renderer owns that (CLAUDE_CONFIG_DIR
+ * is fixed at spawn). And it never decides that output means a limit; it is
+ * handed that verdict by the pty scanner, so the text-matching risk stays in
+ * one place.
+ */
+
+/** Preference keys. Both default to on — absent means enabled. */
+const FAILOVER_PREF = 'accountFailoverEnabled';
+const FAILBACK_PREF = 'accountFailbackEnabled';
+
+export function isFailoverEnabled(): boolean {
+  return getPreference(FAILOVER_PREF) !== 'false';
+}
+
+export function isFailbackEnabled(): boolean {
+  return getPreference(FAILBACK_PREF) !== 'false';
+}
+
+export interface UsageLimitReport {
+  /** Session whose pty printed the announcement. */
+  sessionId: string;
+  /** Account that pty was BILLING — its live binding, not its assignment. */
+  accountId: string | null;
+  /** Reset time parsed from the message, or null when it named none. */
+  resetAt: Date | null;
+  /** Which accounts running ptys are currently bound to. */
+  liveAccounts: LiveAccountBindings;
+}
+
+/**
+ * The next account willing to take work, in the user's failover order.
+ *
+ * Exported because the accounts panel shows the same answer ("Work is next in
+ * line") and must not compute it a second way.
+ */
+export function nextHealthyAccount(excludeId: string | null, now: Date = new Date()): ClaudeAccount | null {
+  return accountsRepo.getAccountsInFallbackOrder().find(
+    account => account.id !== excludeId && accountsRepo.isAccountHealthy(account, now)
+  ) ?? null;
+}
+
+/**
+ * Record a usage limit and move the account's live sessions off it.
+ *
+ * Returns null only when there is nothing this could possibly be about: no
+ * account was bound (the pre-accounts ~/.claude setup, where there is no
+ * second account to move to), or the bound account has since been deleted.
+ * Every other outcome — including "nowhere to go" and "you switched this off" —
+ * comes back as an event, because each of those is something the user needs
+ * told rather than a no-op to swallow.
+ */
+export function handleUsageLimit(report: UsageLimitReport): AccountFailoverEvent | null {
+  if (!report.accountId) return null;
+
+  const from = accountsRepo.getAccount(report.accountId);
+  if (!from) return null;
+
+  // Bookkeeping happens even when failover is switched off. Knowing an account
+  // is spent until 9pm is worth having on its own — the panel shows it, and it
+  // stops that account being picked as somebody else's failover target.
+  const resetAt = report.resetAt ?? new Date(Date.now() + accountsRepo.DEFAULT_COOLDOWN_MS);
+  accountsRepo.markAccountLimited(from.id, resetAt);
+  log.info(
+    `[Failover] ${from.label} is out of quota until ${resetAt.toISOString()}` +
+    `${report.resetAt ? ' (from the CLI message)' : ' (default window — the message named no reset)'}`
+  );
+
+  if (!isFailoverEnabled()) {
+    return { reason: 'limit', from, to: null, sessionIds: [], resetAt, blocked: 'disabled' };
+  }
+
+  const to = nextHealthyAccount(from.id);
+  if (!to) {
+    log.warn(`[Failover] ${from.label} is spent and no other account is available.`);
+    return { reason: 'limit', from, to: null, sessionIds: [], resetAt, blocked: 'no-healthy-account' };
+  }
+
+  // Every pty billing the spent account, not just the one that noticed. The
+  // limit is the account's, so the others are already dead in the water; making
+  // each wait to hit the wall itself would restart them one at a time, minutes
+  // apart, for no gain.
+  const stranded = new Set(
+    Object.entries(report.liveAccounts)
+      .filter(([, binding]) => binding?.accountId === from.id)
+      .map(([sessionId]) => sessionId)
+  );
+  // The reporter belongs in the set even if its binding says otherwise — it is
+  // the one session we have direct evidence about.
+  stranded.add(report.sessionId);
+
+  const sessionIds: string[] = [];
+  for (const sessionId of stranded) {
+    if (moveSession(sessionId, from, to)) sessionIds.push(sessionId);
+  }
+
+  log.info(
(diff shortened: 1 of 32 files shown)


src/main/account-failover.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  const sessionIds: string[] = [];
+  for (const sessionId of stranded) {
+    if (moveSession(sessionId, from, to)) sessionIds.push(sessionId);
+  }
+
+  log.info(
+    `[Failover] Moved ${sessionIds.length} session(s) from ${from.label} to ${to.label}.`
+  );
+  return { reason: 'limit', from, to, sessionIds, resetAt };
```
brannon-bowden, line 133: If moveSession throws for one stranded session, the whole handleUsageLimit call throws and publishFailover is never invoked for any of the sessions moved before the throw — they end up silently reassigned with no UI refresh/notification. Consider isolating per-session failures.

brannon-bowden (approved):
Approve — this is a large, carefully designed feature and it's backed by an unusually thorough test suite (`account-failover.test.ts`, `usage-limit.test.ts`) that exercises the tricky invariants directly (second-hop failover keeping the original home, not overriding a manual switch, NULL-vs-inherited semantics, cooldown extension on a vaguer second report, unranked-sorts-last, deleted-account cleanup). I traced the core state machine (`handleUsageLimit` → `moveSession` → `failBackSession` → `failbackCandidates`) against the tests and didn't find a case where it produces wrong data. CI is green.

## Blocking issues
None.

## Non-blocking issues

- **`src/main/account-failover.ts` (the `for (const sessionId of stranded)` loop in `handleUsageLimit`)**: if `moveSession` throws partway through (e.g. `assignSessionAccount` fails for one of several stranded sessions), the exception propagates out of `handleUsageLimit` entirely. The account is already marked limited and any sessions moved *before* the throw stay moved in the DB, but the caller in `index.ts` catches the exception and never calls `publishFailover` — so the renderer never gets `sessions:refresh`/`accounts:failover`, and no notification is shown, even for the sessions that did move successfully. Consider catching per-session inside the loop (or building the event from whatever succeeded) so a single bad session doesn't silence the whole event.
- **`src/main/usage-limit.ts`** doc comment: "failover refuses to fire twice for the same account inside a short window" — I don't see logic implementing an account-level debounce anywhere in `account-failover.ts`; the only debounce is `usageLimitReported` on `PtySession`, which is per-pty, not per-account, and resets on every respawn. Two ptys on the same account (or the same pty after being respawned onto a new account that also mis-fires) can both trigger a real `markAccountLimited`/session-move. It's not catastrophic (moves are idempotent per the `moveSession` early-return), but the comment overstates a protection that isn't there and could mislead a future maintainer relying on it.
- **`src/main/repositories/sessions.ts`** `getSession()` does `getAllSessions().find(...)`, i.e. a full table load for a single-row lookup. It's called on essentially every failover/failback/sweep path. Given the stated "tens of rows" scale it's not a functional problem, but a `WHERE id = ?` query would be the more honest implementation here.
- **`src/shared/types.ts`**: `AccountHealth` is added but never referenced anywhere in this diff — looks like leftover/unused scaffolding.

## Tests

- `describeFailover` (the desktop-notification copy generator in `account-failover.ts`) has no direct unit test, unlike its renderer counterpart `FailoverNotice`, even though it has the same "must never claim a move that didn't happen" branches (`disabled`, `no-healthy-account`, `failback`).
- The renderer-side decision of *which* moved sessions actually get restarted (`respawnable()` in `App.tsx`, and the `onAccountFailover` effect that calls it) has no test coverage — a regression there would either restart a session the user explicitly stopped or fail to restart one that needs it.
- No test covers the partial-failure path in `handleUsageLimit`'s stranded-session loop discussed above.

William-Long-II:
Thanks — all four non-blocking findings were correct, and all four are fixed in 4796af1. Approval is untouched (`dismiss_stale_reviews` is off on `development`), so this is additive rather than a re-request.

**The stranded-session loop was a real defect**, not just a tidy-up. You were right about the consequence and it's worse than "no notification": `handleUsageLimit`'s caller catches, so a throw meant `publishFailover` was never reached for *any* of them — sessions already reassigned stayed reassigned in the DB, the renderer never got `sessions:refresh`, **no pty was respawned**, and the user was told nothing, while those sessions kept billing the account that had just run dry. Now caught per session. The regression test fails without the catch and passes with it — I verified by reverting the fix rather than assuming.

**The comment overstating the debounce** was the finding I'm most glad you caught. There is no per-account debounce; the only guard is `usageLimitReported`, per-pty and reset on respawn, exactly as you said. The comment now states what is actually true and says explicitly that no debounce exists, so nobody builds on a protection that isn't there.

**`getSession`** hits the primary key now, with the row mapping shared with `getAllSessions` so the two can't drift.

**`AccountHealth`** removed — it was scaffolding I never used.

All three test gaps closed: `describeFailover`'s copy branches (the same "must never claim a move that didn't happen" cases `FailoverNotice` already pins), `respawnable()`'s decision about which moved sessions get a pty back, and the partial-failure path.

One correction to something I said in the PR description, now that I've run it against latest `development`: the two local full-suite failures (`preload.test.ts` + `account-auth.test.ts` colliding through the `electron` module stub) are present on base `development` too — 1059/2/2 there versus 1104/1/1 here — and both pass in isolation. CI's `test` job is green either way, so nothing in this PR touches them.

PR #210, opened 2026-08-25
209 — Detect the wording a real weekly limit actually produces

### LINKED ISSUE #209: Bug: usage-limit detection misses the wording a real weekly limit produces
### What happened

A real weekly limit fired today and produced:

```
You've hit your weekly limit · resets Aug 26 at 2pm (America/New_York)
```

`detectUsageLimit()` returns **null** for that string, so #207's failover never fires. Confirmed by running the merged code against the message directly, not by reading the patterns.

### Why

Every pattern shipped in #207 required the word **"reached"**. The real message says **"hit"**. It also joins the reset to the limit with a bullet rather than a sentence, so neither `limit reached` nor `Your limit will reset at` appears anywhere in it either.

Reset *parsing* was never the problem — `parseResetAt` reads `Aug 26 at 2pm` correctly. Detection is the whole failure.

This is the risk called out in #207's description ("if your Claude Code build words its message differently, detection won't fire"), now observed rather than hypothesised. The gap was one verb, which is why guessing the wording could not have closed it.

### Impact

Silent. No error, no log line, no notification — sessions simply sit on an exhausted account exactly as they did before the feature existed. Two live sessions were in this state when it was found.

### Fix

Broaden the verb (`hit` / `reached` / `used up`), make the qualifier optional (`You've hit your limit`), and add the status-line form (`weekly limit · resets …`). The real message is pinned verbatim as a regression test, along with the wrapped form an agent reports it through.

The quoting guard still refuses all of: the new patterns quoted as source, a comment containing the phrase, a diff line, and a string literal — verified, since broadening detection is exactly where a false-positive hole would open.

### Acceptance criteria

- [ ] The verbatim message above is detected, with its reset parsed to Aug 26 14:00
- [ ] The same message wrapped in an agent error line is detected
- [ ] `You've hit your limit` (no qualifier) is detected
- [ ] Quoted/source/diff/comment forms of the new patterns are still refused

William-Long-II:
Closes #209. Follow-up to #208.

A real weekly limit fired today:

```
You've hit your weekly limit · resets Aug 26 at 2pm (America/New_York)
```

`detectUsageLimit()` returned **null** for it. #207's failover never fires for this message — silently, with no error and no log line. Two live sessions were sitting on the exhausted account when it was found.

## The gap was one verb

Every pattern in #207 required **"reached"**. The message says **"hit"**. It also joins the reset to the limit with a bullet rather than a sentence, so neither `limit reached` nor `Your limit will reset at` appears in it either.

I confirmed this by running the merged code against the string rather than by reading the regexes — and the same way after fixing it. Reset parsing was never at fault: `parseResetAt` read `Aug 26 at 2pm` correctly the whole time.

This is the risk the #208 description called out as unverifiable, now observed. Worth being blunt about the lesson: those patterns were written from what the message plausibly says rather than what it does say, and a one-word difference is not something more careful paraphrasing would have caught. The real message is pinned verbatim now, because only a literal would have failed.

## Changes

- verb broadened — `hit` / `reached` / `used up`
- qualifier made optional — `You've hit your limit` carries no adjective
- status-line form added — `weekly limit · resets Aug 26 at 2pm`

## False positives

Broadening detection is exactly where a hole opens, so the quoting guard is re-checked against the new patterns quoted as source, as a comment, as a diff line, and as a string literal. All still refused.

## Verification

- The verbatim message detected, reset parsed to Aug 26 14:00
- The same message wrapped in an agent's error line detected
- `bun test` — 1106 pass, only the two pre-existing `preload`/`account-auth` collisions that are present on base `development` as well
- `tsc` clean

### DIFF
--- a/src/main/__tests__/usage-limit.test.ts
+++ b/src/main/__tests__/usage-limit.test.ts
@@ -65,6 +65,27 @@ describe('detectUsageLimit', () => {
     expect(detectUsageLimit('5-hour limit · resets 9pm', undefined, NOW)).not.toBeNull();
   });
 
+  /**
+   * The accepted false-positive surface, made visible.
+   *
+   * "You've hit your <something> limit" is a sentence other services say too,
+   * and this terminal shows their errors as readily as Claude's. A false
+   * positive is not inert — it restarts live sessions and benches a healthy
+   * account for hours — so the qualifier is an enumerated set of the windows
+   * Anthropic actually meters on, not any word at all.
+   */
+  test('does not treat another service\'s limit as a Claude usage limit', () => {
+    const foreign = [
+      "You've hit your credit limit",
+      "You've reached your spending limit",
+      "You have reached your character limit",
+      "You've hit your context limit",
+    ];
+    for (const line of foreign) {
+      expect(detectUsageLimit(line, undefined, NOW), line).toBeNull();
+    }
+  });
+
   test('ignores ordinary agent output', () => {
     const lines = [
       'Rate limiting the relay to 100 requests per minute',
--- a/src/main/__tests__/usage-limit.test.ts
+++ b/src/main/__tests__/usage-limit.test.ts
@@ -32,6 +32,39 @@ describe('detectUsageLimit', () => {
     }
   });
 
+  /**
+   * The message a real weekly limit produced, verbatim, on 2026-08-25.
+   *
+   * The first version of these patterns did not match it. Every one of them
+   * required the word "reached"; this says "hit", and joins the reset to the
+   * limit with a bullet rather than a sentence, so neither "limit reached" nor
+   * "will reset at" appears anywhere in it. Detection returned null and
+   * failover silently never fired — the exact failure the feature exists to
+   * prevent, arrived at by guessing the wording instead of observing it.
+   *
+   * It is pinned here as a literal because a paraphrase would not have caught
+   * it: the difference between the guess and the truth was one verb.
+   */
+  test('catches the weekly-limit message an actual run produced', () => {
+    const real = "You've hit your weekly limit · resets Aug 26 at 2pm (America/New_York)";
+    const hit = detectUsageLimit(real, undefined, NOW);
+    expect(hit).not.toBeNull();
+    expect(hit!.resetAt?.getMonth()).toBe(7); // August
+    expect(hit!.resetAt?.getDate()).toBe(26);
+    expect(hit!.resetAt?.getHours()).toBe(14);
+  });
+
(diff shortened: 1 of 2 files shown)


brannon-bowden (approved):
This PR is a well-targeted, well-tested fix: it broadens the usage-limit detection regexes to match the actual wording production traffic produced ("hit" instead of only "reached", the bullet-joined status-line form), and it pins the real failing message as a literal test fixture along with a wrapped/embedded variant and the existing "don't match quoted/commented occurrences" regression tests. CI is green across all four checks (SonarQube, GitGuardian, test, quality-gate).

Issues
- `src/main/usage-limit.ts:23` — the qualifier group was widened from an enumerated list (`usage|weekly|5-hour`) to any single word (`[a-z0-9-]+`). This is what makes "You've hit your limit" and "You've hit your weekly limit" both match with one pattern, which is the point of the change, but it also means unrelated phrases like "You've hit your credit limit" or "You've reached your daily limit" (in agent output quoting some other API's error) would now match as a Claude usage limit. Worth confirming this trade-off is acceptable given the cost of a false positive (unwanted failover) vs. the cost of a false negative (silent non-failover) that motivated this PR — not blocking, since biasing toward catching real limits is a reasonable and likely intentional choice here.

Suggestions (non-blocking)
- The comments in `usage-limit.ts` and the test file explaining *why* each pattern exists (with the real-world message that broke the old one) are genuinely useful for future maintainers — good practice to keep doing when patching detection heuristics like this.
- Consider adding one negative test for the widened qualifier (e.g., a plausible unrelated "reached your X limit" phrase) to document the accepted false-positive surface, since that trade-off isn't otherwise visible from the tests.

William-Long-II:
Good catch on the qualifier — I've narrowed it rather than documenting the trade-off, in 526ca8e. Approval is untouched (`dismiss_stale_reviews` is off).

`[a-z0-9-]+` bought the bare `You've hit your limit` case at the price of every other service's sentence, and this terminal shows their errors as readily as Claude's. Since a false positive isn't inert here — it restarts live sessions and benches a healthy account for hours — the qualifier is now the enumerated set of windows Anthropic actually meters on (`usage|weekly|monthly|daily|session|hourly|5-hour`), plus the no-qualifier form.

That still matches everything this PR set out to catch, and a wording it misses now has to be a *new* one, which I'd rather bet on than assume no other API ever says "limit" in the second person.

Your second suggestion is in too: the accepted surface is visible in the tests instead of implied by the regex — `credit limit`, `spending limit`, `character limit`, `context limit` pinned as non-matches.

PR #212, opened 2026-08-25
211 — Drop the Jira key-prefix workflow

### LINKED ISSUE #211: Chore: drop the Jira key-prefix workflow — we track work in GitHub issues
### Problem

`.github/workflows/jira-key-prefix.yml` runs on every `issues: opened` event and **fails every time**. It has failed on every issue opened today (#207, #209, and others). We no longer use Jira — work is tracked in GitHub issues — so the workflow is prefixing issue titles with a key for a system nobody reads.

It is pure noise: a red X on every new issue, for a job whose success would be undesirable anyway.

### Change

- Delete `.github/workflows/jira-key-prefix.yml`
- `.github/dependabot.yml` names that workflow in the comment explaining the `org-standards` ignore — the ignore stays (it still covers `pr-title`), the comment drops the dead name
- `.github/ISSUE_TEMPLATE/feature.yml` described its Scope dropdown as mapping to "the Jira Major/Minor/New Feature distinction" — same options, no longer explained by reference to a system we left

### Noted, not changed

`.github/workflows/pr-title.yml` — the other `org-standards` reusable workflow — **also fails on every PR**, at 0s, and has done so on every recent PR (#185, #196, #200, #210), not just mine. Same 0s signature as the Jira one, which suggests a shared cause in how the `@v1` reusable workflow resolves rather than anything about titles.

That is out of scope here: it is not Jira, it may be something we want fixed rather than deleted, and it is failing for everyone rather than as a result of anything in this change. Filing this note so it is written down somewhere.

William-Long-II:
Closes #211.

`.github/workflows/jira-key-prefix.yml` runs on every `issues: opened` event and fails every time — it failed on both issues I opened today (#207, #209) and on others before that. We track work in GitHub issues now, so it is prefixing titles with a key for a system nobody reads: a red X on every new issue, for a job whose *success* would be undesirable anyway.

Two collateral references go with it, since both describe things that stop being true when the file is deleted:

- `dependabot.yml` names the workflow in the comment explaining the `org-standards` ignore. **The ignore stays** — it still covers `pr-title`, and removing it would break the github-actions job with `git_dependencies_not_reachable`. Only the dead name goes.
- `ISSUE_TEMPLATE/feature.yml` described its Scope dropdown as mapping to "the Jira Major/Minor/New Feature distinction". Same options, minus the explanation-by-reference to a system we left.

## Noted, not changed

`pr-title.yml` — the other `org-standards` reusable workflow — **also fails on every PR**, at 0s, and has on every recent one (#185, #196, #200, #210), not just mine. Same 0s no-jobs signature as the Jira workflow, which points at how the `@v1` reusable workflow resolves rather than at anything about titles.

I've left it alone deliberately: it isn't Jira, it's failing for everyone rather than because of anything here, and it may be something you want repaired rather than removed. Written down in #211 so it isn't lost.

No source changes; nothing to test beyond CI continuing to pass.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yml
+++ b/.github/ISSUE_TEMPLATE/feature.yml
@@ -36,7 +36,7 @@ body:
     id: scope
     attributes:
       label: Scope
-      description: Maps to the Jira Major/Minor/New Feature distinction.
+      description: How large a change this is.
       options:
         - Major feature
         - Minor feature
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,8 +12,8 @@ updates:
         patterns:
           - '*'
     ignore:
-      # Private org repo hosting the reusable workflows (jira-key-prefix,
-      # pr-title). Dependabot has no access to it — without this ignore the
+      # Private org repo hosting the reusable workflows (pr-title).
+      # Dependabot has no access to it — without this ignore the
       # whole github-actions job fails with git_dependencies_not_reachable.
       # Versioning of those workflows is managed upstream in org-standards.
       - dependency-name: Software-Development-LLC/org-standards
--- a/.github/workflows/jira-key-prefix.yml
+++ b/.github/workflows/jira-key-prefix.yml
@@ -1,9 +1,0 @@
-name: jira-key-prefix
-on:
-  issues:
-    types: [opened]
-jobs:
-  prefix:
-    uses: Software-Development-LLC/org-standards/.github/workflows/jira-key-prefix.yml@v1
-    with:
-      key: B


brannon-bowden (approved):
Approve — this is a clean, well-scoped removal of the Jira key-prefix workflow with consistent follow-up edits.

**Blocking issues**

None.

**Non-blocking issues**

- `.github/dependabot.yml:12-14` — The updated comment now only mentions `pr-title` as justification for ignoring `Software-Development-LLC/org-standards`, and the ignore rule itself still stands, which is correct since that workflow still consumes the reusable workflow. Just flagging as a nit that if `pr-title` is ever also dropped in the future, this ignore block should be revisited/removed — not an issue now.
- Worth a quick sanity check (not blockable from this diff) that no other docs (e.g., CONTRIBUTING.md, README) still reference the `B` Jira key prefix or instruct contributors to expect auto-prefixed issue titles, since those aren't visible in this diff.

**Tests**

None needed — this is a workflow/config deletion and doc-text change with no executable logic to cover.

Issue #213, opened 2026-08-25
Bug: a group account switch commits before the prompt, and 'Use default account' sits next to the account it replaces


William-Long-II:
### What happened

A group was moved off an exhausted account onto a healthy one. Minutes later it was back on the exhausted one, with nothing in the UI reporting the change.

Reconstructed from the log and the on-disk transcripts:

```
18:10Z, 18:12Z   groups.claude_account_id = Brannon   (read directly from the DB)
14:13:39 local   [Accounts] Conversation c98f6d1d-… is readable from the new account
14:13:39 local   Work/…/c98f6d1d-….jsonl  mtime — the carry landed in Work
18:14Z, 18:15Z   groups.claude_account_id = NULL      → falls back to default → Work
```

No `[GroupsAPI]` line in that window, so this was not a mobile/web PATCH — it was a local `assignGroupAccount(groupId, null)`, i.e. the **Use default account** menu item.

### Why it is easy to hit

`handleGroupContextMenu` builds:

```
─────────────────────────
✓  Use default account      ← writes NULL
   Work                     ← listed first: accounts:list orders is_default DESC
   Brannon
   Brannon (GoBodhi)
─────────────────────────
```

The item that clears the assignment is directly above the list of accounts, one row from the account you are reaching for. When the default *is* the account you are trying to escape — which is the whole reason you are in this menu during a usage limit — a one-row miss silently puts every session in the group back on it.

### The part that makes it stick

`handleAssignGroupAccount` writes the assignment first, then raises "N running sessions are still on the previous account. Restart them now?". The change is already committed when that dialog appears; **Not now** declines only the restart. The code says so explicitly, and it is the correct behaviour for a deliberate switch — but read as a confirmation dialog it looks like the change is still pending, so dismissing it feels like cancelling.

Net effect: the group is switched, the prompt is dismissed, nothing on screen changes, and every session without its own override moves to the new account at its next respawn — which may be hours later, under a limit, with no memory of the click that caused it.

### Suggested fixes

- Separate **Use default account** from the account list, or name the account it resolves to (`Use default account (Work)`) so the destination is visible before the click
- Make the restart prompt honest about what has already happened — it is "apply now or on next restart", not "confirm this change"
- Consider an undo path, or at minimum surface the resulting assignment somewhere non-modal after the switch

### Related

Sibling to the silent-no-op case: choosing the account a session already effectively uses returns `affectedSessionIds: []` and reports nothing at all.

PR #216, opened 2026-08-25
215 — Stop the test suite writing into the installed app's log

### LINKED ISSUE #215: Bug: the test suite writes into the installed app's log file and corrupts it
### What happened

Running `bun test` writes into `~/Library/Logs/bodhilander/main.log` — the same file a running Bodhilander is appending to.

electron-log's file transport resolves a default path per platform (`~/Library/Logs/<appName>/main.log` on macOS). Any module under test that imports `electron-log` — `account-auth`, `account-switch`, `pty-manager`, `conversation-transcript`, and more — therefore logs into the live app's file. Two processes append at independent offsets, so the entries interleave and overwrite each other.

### Why it matters

This was found while reading that log to diagnose a real account-switch bug (#213). The file contained fixture accounts from `account-auth.test.ts` — `label="Work"`, `label="Doomed"`, `will@acme.test`, `will@linux.test`, `Failed to kill login pty … teardown glitched` — timestamped 17:33 but sitting *before* the app's own 14:11 entries, with the app's real lines partly overwritten.

Sixty-odd fake account logins in the log, out of order, on a machine where someone is trying to work out why their account switch reverted. Tests corrupting the evidence you debug with is worse than tests being noisy — and the fixtures are convincing enough to send someone chasing a login flow that never happened.

### Fix

Disable the file transport in `test-setup/happydom.ts`, which `bunfig.toml` preloads for every test file, so it covers the whole suite at one point:

```ts
log.transports.file.level = false;
```

Disabling rather than redirecting to a temp path: nothing asserts on log *files*, so redirecting would only move the litter. Console output is untouched, so a test that needs to see a log line still can.

### Verified

- `bun test src/main/__tests__/account-auth.test.ts` — previously the biggest writer; log size now unchanged
- full suite (1105 tests) — log delta **0 bytes**

### Note

Existing pollution in the file is historical and not cleaned up by this; it just stops growing.

William-Long-II:
Closes #215.

`bun test` was writing into `~/Library/Logs/bodhilander/main.log` — the file a running Bodhilander appends to. electron-log's file transport resolves a default per-platform path, so every module under test that imports it logged into the live app's file, at an independent offset. The two interleave and overwrite.

## How it was found

Reading that log to diagnose #213 — a real account switch that reverted. The file contained fixture accounts from `account-auth.test.ts` (`label="Work"`, `label="Doomed"`, `will@acme.test`, `teardown glitched`) timestamped 17:33 but sitting *before* the app's own 14:11 entries, with real lines partly overwritten.

Sixty fake account logins in the record someone is using to work out why their account switch reverted. That is worse than noise: the fixtures are plausible enough to send you chasing a login flow that never happened.

## Fix

One line in `test-setup/happydom.ts`:

```ts
log.transports.file.level = false;
```

`bunfig.toml` preloads that file for every test in the repo, so this covers the whole suite — including suites not yet written — at a single point.

**Disabled rather than redirected to a temp path.** Nothing asserts on log *files*, so redirecting would only relocate the litter. Console output is untouched, so a test that needs to see a log line still can.

## Verified

| | |
|---|---|
| `account-auth.test.ts` alone (previously the biggest writer) | log size unchanged |
| full suite, 1105 tests | log delta **0 bytes** |

Existing pollution is historical and not cleaned up here; it just stops growing.

### DIFF
--- a/test-setup/happydom.ts
+++ b/test-setup/happydom.ts
@@ -11,6 +11,30 @@
  * globals back. Component tests only need document/window/HTMLElement.
  */
 import { GlobalRegistrator } from '@happy-dom/global-registrator';
+import log from 'electron-log';
+
+/**
+ * Stop the suite writing into the installed app's log file.
+ *
+ * electron-log's file transport resolves a default path per platform — on macOS
+ * `~/Library/Logs/<appName>/main.log`, which is the same file a running
+ * Bodhilander is appending to. Any module under test that imports electron-log
+ * therefore logs into it, at an offset of its own, while the app writes at its
+ * own: the two interleave and overwrite, and the result is a log that is not a
+ * record of either.
+ *
+ * That is not a tidiness problem. It was found while reading that log to
+ * diagnose a real account-switch bug (#213), where the file contained fixture
+ * accounts from `account-auth.test.ts` — labels "Work" and "Doomed",
+ * will@acme.test, "teardown glitched" — interleaved with, and in places on top
+ * of, the app's own entries. Tests corrupting the evidence you debug with is
+ * worse than tests being noisy.
+ *
+ * Disabling the transport rather than redirecting it: nothing asserts on log
+ * FILES, so a temp path would only move the litter somewhere else. Console
+ * output is untouched, so a test that needs to see a log line still can.
+ */
+log.transports.file.level = false;
 
 // Capture Bun's natives before happy-dom replaces them.
 // Network/crypto only. DOM-side globals (Event, EventTarget, MessageEvent, …)


brannon-bowden (approved):
Approve — this is a narrow, well-justified change to test setup that disables electron-log's file transport so test runs stop interleaving into the installed app's log file, and CI is green.

## Issues
None blocking.

## Suggestions
- `test-setup/happydom.ts:15` (new `log.transports.file.level = false`) — this only takes effect if this setup file runs and configures the transport *before* any test module imports/configures `electron-log` itself. If another setup file or a module under test also sets `transports.file.level` (e.g. re-enabling it for a specific test), this could be silently overridden depending on load order. Worth a quick sanity check that this file is guaranteed to load first, since the whole point of the change is to prevent exactly that race with the file transport.
- No test asserts the new behavior (e.g. `expect(log.transports.file.level).toBe(false)`), which would guard against a future refactor accidentally dropping the line. Given this is test infrastructure disabling itself, a minimal assertion is optional but would make the invariant explicit rather than relying on the comment alone.

William-Long-II:
Both suggestions were worth raising, and chasing the first turned up that my original verification was weaker than I claimed.

**On load order.** I had asserted "0 bytes across the suite" without ever running the control — that number alone can't distinguish "the fix worked" from "nothing wrote this time". Properly controlled now, on this branch, same test, back to back:

```
WITH fix:    delta 0 bytes
WITHOUT fix: delta 3331 bytes
```

So it holds behaviourally. Nothing else re-enables the transport from a test path either — `index.ts:79` and `auto-updater.ts:9` both set `file.level = 'info'` at module scope, but no test imports either, directly or transitively.

**On the assertion test — I tried it, and it can't be written the obvious way.** `expect(log.transports.file.level).toBe(false)` fails even though the transport is demonstrably off, because a test file importing `electron-log` gets a different module instance than the preload does (in the full-suite run the test-side object has no `transports` property at all). So the assertion would fail while the invariant holds — a guard that fires on correct code is worse than none.

I'd rather leave that gap visible than paper over it with a test asserting the wrong object. If it's worth closing properly, the honest version is a behavioural check — stat the resolved log path around a run that logs — which is a bigger piece of test infrastructure than this one-line change warrants. Happy to do it as a follow-up if you think it's worth it.

PR #218, opened 2026-08-25
217 — Drop pr-title: a public repo cannot call an internal repo's workflow

### LINKED ISSUE #217: Chore: drop pr-title — a public repo cannot call an internal repo's reusable workflow
### Problem

`.github/workflows/pr-title.yml` fails on **every** pull request, at 0s, with zero jobs and the message "This run likely failed because of a workflow file issue". It has failed on every recent PR — #185, #196, #200, #208, #210, #216 — not just one author's.

### Cause

```
Bodhilander    visibility: public
org-standards  visibility: internal (private: true)
```

A public repository cannot consume a reusable workflow from a private or internal one. `org-standards` has `access_level: organization`, which grants org repos access, but that does not extend to public callers — GitHub cannot resolve the workflow, so the run fails before scheduling any job.

Ruled out on the way to that, each with evidence:

- inputs match the callee's contract (`key`, required, string — caller passes `key: B`)
- `default_workflow_permissions` on this repo is `write`, so it is not permission intersection
- `org-standards` has a real `v1` tag, byte-identical to its default branch
- both caller and callee YAML parse cleanly

### Consequence

The workflow has **never worked in this repo**. Its job is to stamp titles as `[B-<issue#>][<Type>] <title>`; every PR title here is `<issue#> — <title>`, written by hand. The only thing it has ever produced is a red X.

### Change

Delete `.github/workflows/pr-title.yml`, and the `org-standards` ignore in `dependabot.yml` that existed only to stop Dependabot choking on the now-unused reusable-workflow reference.

### Alternatives not taken

- **Inline the ~65 lines into this repo** — viable if the `[B-<issue>]` stamping is wanted. It is not today; the existing convention works and is applied by hand.
- **Make `org-standards` public** — fixes it for every consumer at once, but publishes org tooling. Not this repo's call to make.

William-Long-II:
Closes #217. Companion to #212.

`pr-title.yml` has failed on **every** PR in this repo — #185, #196, #200, #208, #210, #216 — at 0s, with zero jobs and "This run likely failed because of a workflow file issue". It is not about titles:

```
Bodhilander    public
org-standards  internal (private: true)
```

A public repository cannot consume a reusable workflow from a private or internal one. `org-standards` sets `access_level: organization`, which grants org repos access but does not extend to public callers — GitHub can't resolve the workflow, so the run dies before scheduling a job.

## Ruled out first

Each against evidence, not inspection:

| Hypothesis | Finding |
|---|---|
| input contract mismatch | callee wants `key` (required, string); caller passes `key: B` ✓ |
| permission intersection | `default_workflow_permissions` on this repo is `write` ✓ |
| bad or missing `v1` tag | tag exists, byte-identical to default branch ✓ |
| malformed YAML | both caller and callee parse ✓ |

## Consequence

It has **never worked here**. Its job is to stamp `[B-<issue#>][<Type>] <title>`; every title in this repo is `<issue#> — <title>`, written by hand. The only output it has ever produced is a red X on the checks list.

The `org-standards` ignore in `dependabot.yml` goes with it — it existed solely to stop Dependabot choking on the reusable-workflow references, and there are none left.

## Alternatives not taken

- **Inline the ~65 lines** — the right move if `[B-<issue>]` stamping is wanted. It isn't today.
- **Make `org-standards` public** — fixes every consumer at once, but publishes org tooling, and that isn't this repo's call.

Both stay available; this just stops a check that cannot pass from failing on every PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,10 +11,4 @@ updates:
       actions:
         patterns:
           - '*'
-    ignore:
-      # Private org repo hosting the reusable workflows (pr-title).
-      # Dependabot has no access to it — without this ignore the
-      # whole github-actions job fails with git_dependencies_not_reachable.
-      # Versioning of those workflows is managed upstream in org-standards.
-      - dependency-name: Software-Development-LLC/org-standards
 
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,9 +1,0 @@
-name: pr-title
-on:
-  pull_request:
-    types: [opened, edited]
-jobs:
-  title:
-    uses: Software-Development-LLC/org-standards/.github/workflows/pr-title.yml@v1
-    with:
-      key: B


brannon-bowden (approved):
Approve — this is a clean, well-scoped removal of a workflow that could never have succeeded from a public repository.

## Blocking issues
None. The change is a straightforward deletion of `.github/workflows/pr-title.yml` and its associated dependabot ignore entry, with no impact on application code, schemas, or public APIs.

## Non-blocking issues
- **Branch protection dependency**: if the repo's branch protection rules list the `title` job (from the now-deleted `pr-title` workflow) as a required status check, PRs may get stuck waiting for a check that will never run again. This isn't visible in the diff — worth a quick check in the repo's branch protection settings before/after merge, but it's not something to hold up this PR for.
- The commit message/PR title explains the "why" (public repo can't call a private org's reusable workflow) clearly — good practice, and the dependabot ignore removal is a correct, consistent follow-up since that entry existed solely to work around the now-removed workflow.

## Tests
Not applicable — this is a CI/workflow configuration removal with no testable logic.

William-Long-II:
Checked the branch-protection point before merging rather than after: `development` has no `required_status_checks` configured, and the only ruleset is "Copilot Auto Review" (`copilot_code_review`). No `title` context is required anywhere, so nothing will wedge waiting on the deleted workflow. Good thing to raise — that failure mode would have been silent and would have hit every future PR.

PR #220, opened 2026-08-25
219 — Detect quota rejections from the transcript, not the screen

### LINKED ISSUE #219: Bug: usage-limit detection marks healthy accounts limited, and resuming re-arms it
### What happens

After updating to 3.5.1-beta.7, the app marks accounts limited that are not. Observed on a live machine: three accounts marked within nine seconds, two of them demonstrably healthy. Clearing them is not durable — a session that fails back to its original account re-triggers immediately.

### Mechanism

`checkUsageLimit` matches `CLAUDE_USAGE_LIMIT_PATTERNS` against `PtySession.outputBuffer`. That buffer is not text. It is the TUI's repaint stream with ANSI stripped, so cursor-positioning becomes nothing and words from different screen regions concatenate. Captured from the log, verbatim:

```
⏺Agent"Ownissue206LinuxCIflake"failed:AgentterminatedearlyduetoanAPIerror:You'vehityourweeklylimit·resets…
'') || '~'Lines794/856 fallback when themeasurementis0(ahiddenelement)—??keepsthe0…
lyingthebehaviour-preservingfixes:Ran 2 shell commands ⏺Backgroundcommand"Watchthereleasebuild"completed…
```

Two of those contain no limit message at all. They matched on prose *about* limits — which is the text most likely to contain the trigger words.

The quoting guard was supposed to catch this and could not: it tests the line around the match, and the TUI emits almost no newlines, so `lineAround()` returns the whole buffer. All 12 captured matches had a "line" of 725–1877 characters. A real limit message is ~71. The guard was matching line-shaped patterns against a 1200-character blob.

**Resuming re-arms it.** `--resume` replays the conversation into the terminal, so any session that has ever discussed a usage limit re-renders the trigger text on every restart. That is the loop: mark → clear → fail back → re-render → mark again.

### The design error

Marking happens **even when failover is switched off** — a deliberate choice, on the reasoning that a cooldown is worth recording regardless. It means the kill switch does not stop this. In this failure mode that reasoning is exactly backwards.

### Why the signal is unsound, not just under-guarded

A length guard would reject every one of the 12 false positives. It would also reject genuine limit messages, because in the TUI those appear inside the same repaint blobs. Terminal output cannot distinguish the CLI announcing a limit from the CLI rendering a conversation about one. No pattern refinement fixes that; the input is the problem.

### Fix in this issue

Stop the harm, do not paper over the mechanism:

- [ ] The failover preference gates **detection**, not just the move — nothing is marked while it is off
- [ ] The preference defaults to **off**, so an update cannot re-enter this state unasked
- [ ] A structural guard rejects matches whose containing line is implausibly long for a message, so an opted-in user cannot hit the soup case
- [ ] The 12 captured strings are pinned as regression fixtures, verbatim

Detection is deliberately conservative afterwards — closer to inert in the TUI than wrong. A sound replacement signal is a separate piece of work.

### Follow-up

Redesign detection against the conversation transcript, which records the CLI's own messages structurally, rather than against rendered output.

William-Long-II:
Closes #219. Regression from #207, shipped in 3.5.1-beta.7.

Three accounts marked limited within nine seconds on a live machine, two of them healthy. Clearing didn't hold. **86 such detections in one log.**

## Why patterns could never work

`checkUsageLimit` matched prose against `PtySession.outputBuffer`. That buffer is not text — it's the TUI's repaint stream with ANSI stripped, so cursor positioning becomes nothing and words from unrelated screen regions fuse:

```
⏺Agent"Ownissue206LinuxCIflake"failed:…You'vehityourweeklylimit·resetsAug26at2pm…
'') || '~'Lines794/856 fallback when themeasurementis0(ahiddenelement)…
lyingthebehaviour-preservingfixes:Ran 2 shell commands ⏺Backgroundcommand…
```

**Two contain no limit message at all** — they matched on prose *about* limits, the text most likely to carry the trigger words. The quoting guard never had a chance: a TUI emits almost no newlines, so `lineAround()` returned the whole buffer. Every captured match had a "line" of **725–1877 characters** against a real message of ~70.

And `--resume` replays the conversation, so any session that ever discussed a limit re-armed it on restart: mark → clear → fail back → re-render → mark again.

No pattern refinement fixes this. **Terminal output cannot separate a CLI announcing a limit from a CLI rendering a discussion of one.** The input is wrong, so the signal changes rather than the patterns.

## The signal that is actually sound

The agent already records a refusal structurally, in the conversation it writes to disk:

```json
{ "type": "assistant", "isApiErrorMessage": true, "apiErrorStatus": 429,
  "quotaLimits": { "status": "rejected", "resetsAt": 1787767200,
                   "rateLimitType": "seven_day" } }
```

That is the CLI asserting a fact about itself. It cannot be produced by rendering text; it carries the **exact reset as an epoch**, so nothing is parsed out of a sentence; and `rateLimitType` names the window, so the copy stops guessing. The five-hour default survives only as a fallback that shouldn't fire.

Output is still the *trigger* — new bytes prompt a look, throttled to once per three seconds — but no longer the *evidence*.

Two conditions stop a replayed transcript re-arming anything: an entry older than the pty is history, and a reset already past describes a limit that has lifted. `status: 'rejected'` is required rather than merely present — the same shape reports quota running low, and being warned is not being refused.

## Removed

The prose patterns, the `usageLimitPatterns` provider hook that carried them, and their tests are **deleted**, not left as a weaker second detector. Failover returns to **on by default**, since what it acts on is now a fact rather than a guess.

## Verification

Fixtures are real lines lifted from transcripts on a machine that hit genuine limits:

- every captured rejection recognised; every other captured entry rejected
- a turn whose text is the limit message **and** every old pattern's trigger → **null**
- a connection error carrying the same `isApiErrorMessage` flag → null
- `allowed_warning` → null; already-passed reset → null; entry older than the pty → null
- truncated first line (tail reads start mid-line) → still finds the rejection
- path-traversal conversation id → null
- `bun test` — 1102 pass, only the two pre-existing `preload`/`account-auth` collisions present on base `development`

I shipped #207 saying detection was the part I couldn't verify, and guarded the direction I expected to fail — a missed detection. It failed the other way.


brannon-bowden (approved):
This PR is well-scoped and consistently threads its change through implementation, types, UI copy, and tests — approving.

The core change flips `accountFailoverEnabled` from opt-out (`!== 'false'`) to opt-in (`=== 'true'`) and, more importantly, moves the `isFailoverEnabled()` gate in `handleUsageLimit` (src/main/account-failover.ts:87) so that `markAccountLimited` is never called when the feature is off — previously an account could be marked "limited" purely from detection even with failover disabled, which is the false-positive harm the PR title targets. The `'disabled'` branch of `AccountFailoverEvent.blocked` is removed cleanly from the type, `describeFailover`, and `FailoverNotice`, with matching test removals. The new `MAX_MESSAGE_LINE` guard in `usage-limit.ts` addresses a concrete observed failure mode (repainted TUI screens collapsing into one long unbroken run) and is exercised with real captured false-positive blobs plus a positive-control test confirming genuine single-line messages still match. The `ClaudeAccountsModal` checkbox default and copy are updated consistently with the new opt-in default.

## Suggestions (non-blocking)

- `src/main/__tests__/usage-limit.test.ts`: `expect(detectUsageLimit(blob, ...), blob.slice(0, 70)).toBeNull()` passes a second argument to `expect()` as if it were a custom failure message — most Jest-compatible `expect` implementations don't support that signature, so the label likely won't surface on failure (the assertion itself is still correct). Worth double-checking against the test runner in use, or switching to a loop with a `try/describe.each`-style per-blob test name for a message that reliably shows.
- `REPAINT_BLOBS` (src/main/__tests__/usage-limit.test.ts) has one entry indented with 4 spaces while the rest use 2 — cosmetic only.
- Worth a quick sanity check (not a blocker): `isFailbackEnabled()` still defaults to opt-out (`!== 'false'`) while `isFailoverEnabled()` now defaults to opt-in-required. Since failback presumably only matters once failover has fired, this is likely inert while failover is off by default, but worth confirming that's the intended asymmetry going forward.

William-Long-II:
Re-requesting review — the approval on this PR is against `6c028d8`, and the approach changed completely after it. `dismiss_stale_reviews` is off on `development`, so GitHub still shows it green and mergeable; please don't take that as covering what's here now.

**What you approved (`6c028d8`)** kept the output-scanning detector and added guards: reject a match whose containing line is implausibly long, gate detection behind the preference, default the preference off.

**What's here now** throws that away. The premise was wrong — no guard makes rendered terminal output a sound signal, because it cannot separate a CLI *announcing* a limit from a CLI *rendering a conversation about one*.

Three commits since:

- `b5add9c` — detection moved to the conversation transcript, where the agent records a refusal structurally (`isApiErrorMessage` + `quotaLimits.status === 'rejected'` + `resetsAt` as an epoch). The prose patterns, the `usageLimitPatterns` provider hook and their tests are **deleted**; `quota-limit.ts` is new. Failover returns to **on by default**, since it now acts on a fact.
- `9cb5276` — one Sonar finding.
- `2cb62cf` — a **one-time data migration** clearing every account cooldown, and removal of `DEFAULT_COOLDOWN_MS`.

The two things I'd most want a second pair of eyes on:

1. **The migration writes to `claude_accounts`.** It clears all cooldowns, not just provably-bad ones, on the reasoning that cooldowns only ever existed under the broken detector and a genuinely spent account is re-marked at its next refusal. It is guarded by a preference key rather than a schema check. If it ever re-fired it would erase a legitimate cooldown — `does not run a second time` is the test for that, and it is the one I would attack first.

2. **`resetAt` is now non-nullable** on `UsageLimitReport`, which is what makes the deleted fallback unreachable-by-type rather than unreachable-by-comment. Worth checking I haven't made a null reachable from some path I did not consider.

brannon-bowden (approved):
This increment replaces the regex/screen-scraping usage-limit detector with one that reads the CLI's own structured transcript entries (`quota-limit.ts`), and consistently threads that change through `pty-manager.ts`, `account-failover.ts`, provider types, the migration that clears stale cooldowns, and a thorough new test suite (including a verbatim-fixture round-trip test and a path-traversal guard test) — approving.

The reversal of `isFailoverEnabled()` back to opt-out (`!== 'false'`, src/main/account-failover.ts:37) — the exact opposite of what the prior review praised as opt-in — is justified by the underlying change: the false-positive risk that motivated the opt-in default was "prose that discusses a limit reads the same as an announcement," and the new detector requires a structured `isApiErrorMessage: true` + `apiErrorStatus: 429` + `quotaLimits.status === 'rejected'` entry, which prose cannot produce by construction. The UI copy, default state, and tests in `ClaudeAccountsModal.tsx` and `account-failover.test.ts` were all updated to match, and this also resolves the failover/failback default asymmetry flagged in the previous review, since both are now opt-out. The one-time `clearCooldownsFromOutputScanning` migration (src/main/database.ts) is scoped tightly (two columns, one preference-guarded run) and has direct tests for idempotency and blast radius.

- `clearCooldownsFromOutputScanning(database)` is invoked from inside `initializeArenaTables` (src/main/database.ts:576-604), which is a confusing home for an accounts-table migration unrelated to the arena feature — worth moving to its own call site in the init sequence for readability.
- `quota-limit.ts`'s `parseEntry` reads `Date.now()` directly for the expiry check rather than accepting an injectable `now`, unlike the old `usage-limit.ts` (which took `now: Date` for deterministic tests). Tests work around this with `soon()`, but it makes the module inherently time-coupled rather than pure — a minor testability regression, not a correctness issue given the multi-hour windows involved.

PR #222, opened 2026-08-25
221 — Clear the Sonar findings in the relay web client

### LINKED ISSUE #221: Chore: clear the Sonar findings in the relay web client
### Problem

The `development` quality gate is red, and has been on every push for days. Three conditions fail; `new_violations: 54 > 0` is the substantive one. Twenty of those 54 sit in a single file, `relay/web/src/main.ts` — 37% of the total.

They are mechanical: `??` over `||`, nested ternaries, unbraced `if`s with a trailing statement, unnecessary assertions, a nested template literal.

### Not all of them should be fixed

Three of the five `||` → `??` suggestions in that file would introduce bugs, because the falsy fallback is deliberate:

- `getBoundingClientRect().width || scrollWidth || 1` — width is **0** for a hidden element; `??` keeps the 0 and scales the terminal to nothing
- `.xterm-viewport?.clientHeight || screenEl.clientHeight` — clientHeight is **0** for a hidden element and the result is divided by `term.rows`; `??` divides by zero
- `(… workingDir …) || '~'` — an empty `workingDir` is the **inherit sentinel** across this codebase, so `''` must fall through

Those are marked **Accepted** in SonarQube with that rationale rather than "fixed".

A fourth looked like the same case and is not: `sel.parentId || sel.id`. `repositories/groups.ts` maps `parentId: row.parent_id || null` before it leaves the desktop, and the wire type is `string | null`, so `''` is unreachable and `??` is exactly equivalent. Fixed, not accepted.

### Changes

- **S2681** ×5 — brace the inner `if`s. None was a bug today; each is the shape that becomes one on the next edit, since the trailing `return`/assignment only *looks* conditional.
- **S3358** ×5 — extract nested ternaries. The largest is the session-state class chain, which turns out to be `BADGE[s.state].cls` — the map already existed two lines above, with identical values for all five states.
- **S6606** ×2 fixed (`rows[0] ?? null`, `parentId ?? id`), ×3 accepted as above.
- **S6582**, **S4325** ×2, **S4624** — optional chain, redundant assertions, nested template literal.
- **S3776** — `renderSessions` cognitive complexity, expected to fall out of the ternary extraction rather than needing its own refactor.

### Acceptance criteria

- [ ] `relay/web/src/main.ts` reports zero unresolved Sonar issues
- [ ] The three accepted findings carry a written rationale in SonarQube
- [ ] `bun run typecheck:web` and `bun run build:web` pass; the relay suite stays green
- [ ] No behaviour change — the state-class lookup is verified identical for every `SessionState`

### Remaining after this

~30 findings across 11 other files, plus `new_software_quality_maintainability_issues: 51 > 50`. Tracked separately; this file is the third that clears in one place.

William-Long-II:
Closes #221.

Twenty of the 54 violations reddening the `development` gate sit in this one file — 37% of the total, in 17 changed lines.

## The part that isn't mechanical

**Three of the five `||` → `??` suggestions would introduce bugs.** The falsy fallback is deliberate:

| line | why `??` breaks it |
|---|---|
| 794 | `getBoundingClientRect().width` is **0** for a hidden element; `??` keeps the 0 and scales the terminal to nothing |
| 856 | `clientHeight` is **0** for a hidden element, and the result is **divided by `term.rows`** |
| 1076 | an empty `workingDir` is the **inherit sentinel** used across this codebase, so `''` must fall through to `'~'` |

Those are marked **Accepted** in SonarQube with that rationale written on each, rather than "fixed".

**A fourth looked identical and wasn't.** `sel.parentId || sel.id` is safe to convert: `repositories/groups.ts` maps `parentId: row.parent_id || null` before it leaves the desktop and the wire type is `string | null`, so `''` never arrives. Fixed, not accepted. The difference between that case and the three above is invisible at the call site, which is the argument for checking each rather than applying the rule.

## The rest

- **S2681 ×5** — unbraced `if` with another statement after it on the same line. **None is a bug today**; in each the trailing `return`/assignment is meant to be unconditional and merely reads as if it isn't. That's the shape that becomes a bug on the next edit.
- **S3358 ×5** — nested ternaries extracted. The largest was the session-state class chain, which turned out to be `BADGE[s.state].cls` — the map already existed two lines above. Verified state by state, not by eye:

  ```
  waiting→wait  working→work  idle→idle  error→err  stopped→stop
  identical for every state
  ```
- **S6582**, **S4325 ×2**, **S4624** — optional chain, redundant assertions, nested template literal.
- **S3776** — `renderSessions` complexity, expected to fall out of the ternary extraction.

## Verification

- `bun run typecheck:web` and `bun run build:web` pass
- relay suite: 268 pass; root suite unchanged at 1109 pass (two pre-existing failures)
- no behaviour change intended anywhere in this diff

## Remaining after this

~30 findings across 11 other files, plus `new_software_quality_maintainability_issues: 51 > 50` (one over). The gate won't be green until those land too.

### DIFF
--- a/relay/web/src/main.ts
+++ b/relay/web/src/main.ts
@@ -1027,7 +1027,7 @@ function buildCreateTree() {
   let pick: HTMLElement | null = null;
   if (pendingGroupName) { pick = rows.find((r) => app.groups.find((g) => g.id === r.dataset.gid)?.name === pendingGroupName) || null; pendingGroupName = ''; }
   if (!pick && selGroupId) pick = rows.find((r) => r.dataset.gid === selGroupId) || null;
-  if (!pick) pick = rows[0] ?? null;
+  pick ??= rows[0] ?? null;
   if (pick) selectGroup(pick, pick.dataset.gid!); else { selGroupId = ''; const go = $<HTMLButtonElement>('#crGo'); if (go) go.disabled = true; }
 }
 function addGroupRow(tree: HTMLElement, g: RGroup, isSub: boolean, parent?: RGroup) {
--- a/relay/web/src/main.ts
+++ b/relay/web/src/main.ts
@@ -468,8 +468,7 @@ function applyWatchOnly(): void {
 
   // The existing scroll path is a touch-only touchmove handler, so without
   // this a keyboard user cannot move through the scrollback at all.
-  screen.addEventListener('keydown', (e) => {
-    const ev = e as KeyboardEvent;
+  screen.addEventListener('keydown', (ev) => {
     if (!term) return;
     const page = Math.max(1, term.rows - 1);
     if (ev.key === 'PageUp') { term.scrollLines(-page); ev.preventDefault(); }
@@ -646,7 +645,7 @@ function onConnState(s: ConnState, detail?: string) {
 }
 
 function onAgentMessage(m: Inner) {
-  if (m.type === 'groups') { app.groups = (m.groups as RGroup[]) || []; renderSessions(); if ($('#gtree')) buildCreateTree(); return; }
+  if (m.type === 'groups') { app.groups = (m.groups as RGroup[]) || []; renderSessions(); if ($('#gtree')) { buildCreateTree(); } return; }
   if (m.type === 'dirs') { onDirs?.({ path: String(m.path), entries: (m.entries as string[]) || [] }); return; }
   if (m.type === 'sessions') {
     const list = (m.sessions as RSession[]) || [];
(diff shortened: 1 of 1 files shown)


brannon-bowden (approved):
Approve — this is a low-risk, mechanical Sonar-lint cleanup (removing redundant casts, adding braces to single-statement `if`s, optional chaining/nullish coalescing, de-duplicating a computed string) with no new features and essentially no behavioral surface to test. All four CI checks (SonarQube, quality-gate, test, GitGuardian) are green.

**Issues**

None blocking.

- `relay/web/src/main.ts:726` — the CSS class computation was changed from an inline `work/wait/err/idle/stop` ternary chain to `b.cls` (`b = BADGE[s.state]`). The `BADGE` table definition isn't in this diff, so I can't confirm its `cls` values line up 1:1 with the five strings the old ternary produced. TypeScript will catch a missing field but not a wrong string value, so a quick visual pass over each session state (working/waiting/error/idle/stopped) is worth doing before merge to rule out a silent CSS-class regression.
- `relay/web/src/main.ts:1049` — `sel.parentId || sel.id` was changed to `sel.parentId ?? sel.id`. This is only equivalent to the original if `parentId` can never be an empty string (only `null`/`undefined`); if it can be `''`, `ngParent` would now resolve to `''` instead of falling back to `sel.id`. Worth a quick confirmation that `parentId`'s type/usage rules out `''`, since the rest of the file wasn't shown here.

Issue #224, opened 2026-08-25
Chore: split the three relay protocol dispatchers Sonar flags for complexity


William-Long-II:
### Problem

Three functions exceed the cognitive-complexity threshold of 20:

| function | complexity | size |
|---|---|---|
| `relay/src/ws.ts` `handleAgent()` | 60 | 131 lines |
| `relay/src/http.ts` `route()` | 47 | 98 lines |
| `relay/web/src/connection.ts` `handle()` | 24 | 77 lines |

All three are message dispatchers: a long chain deciding what an incoming frame or request means and routing it.

### Why they were not split in the Sonar sweep

A fourth one was — `ws.ts` `close()`, complexity 21, two independent branches that separated cleanly. Even that one surfaced something: the `ws.data.role` check was narrowing `SocketData` to the agent or client shape, and the narrowing did not survive being moved into a helper. It had to be passed down explicitly. A compiler caught that immediately.

The remaining three are the case where nothing catches it. They dispatch live traffic between the desktop agent, the relay and the web client, and splitting a 60-branch dispatcher means deciding which state each extracted piece still needs — the same class of mistake as the narrowing above, but in code where the failure shows up as a dropped frame rather than a type error.

That is worth doing deliberately, with the protocol in front of you. It is not worth doing at the end of a mechanical cleanup to move a metric.

The three findings are marked **Accepted** in SonarQube pointing at this issue. That records a deferral, not a refutation — the complexity is real.

### Acceptance criteria

- [ ] Each dispatcher is under the threshold
- [ ] Every message type still round-trips: agent→relay→client and back
- [ ] The three Accepted findings are reopened and resolved as fixed

Issue #223, opened 2026-08-25
Chore: migrate off recharts Cell before Recharts 4.0 removes it


William-Long-II:
### Problem

`AnalyticsPanel.tsx` colours three charts with `<Cell>` children (lines 201, 231, 333). Recharts deprecated the component:

> This component is now deprecated and will be removed in Recharts 4.0.
> Please use the `shape` prop or `content` prop on the respective chart components instead.

### Why it is not being done in the Sonar sweep

It is a real API migration, not a lint fix. The three call sites assign per-slice and per-bar colours — a pie keyed by session state, a bar chart cycling a palette, a second pie. Moving them to `shape`/`content` changes how each element renders, and whether it still looks right is a question a test cannot answer. Doing it at the tail end of a mechanical cleanup, unverified, is how a chart quietly starts drawing the wrong colours.

The four findings are marked **Accepted** in SonarQube pointing at this issue. That records a deferral, not a refutation: unlike the `||`/`??` findings accepted alongside them, Sonar is **right** here and the work is real.

### Acceptance criteria

- [ ] The three charts render identically — verified by eye against the current build, not only by tests passing
- [ ] `Cell` is gone from the imports
- [ ] The four Accepted findings are reopened and resolved as fixed

### Deadline

Recharts 4.0. There is no urgency before then, but the deprecation will become a build error rather than a warning.

PR #226, opened 2026-08-25
225 — Clear the remaining Sonar findings, and take the gate to zero

### LINKED ISSUE #225: Chore: clear the remaining Sonar findings outside the relay web client
### Problem

After #221 cleared `relay/web/src/main.ts`, 34 findings remain across 20 files, and the `development` gate is still red on `new_violations > 0`.

### Fixed

- **`db.exec` → `db.run`** ×4 (`relay/src/db/index.ts`) — the `exec` overload is deprecated in bun:sqlite; `run` is the current one.
- **`React.FormEvent` → `React.SyntheticEvent`** ×3 — the React types deprecate it with the reason *"FormEvent doesn't actually exist"*.
- **`data.toString()` on a websocket payload** ×3 — `RawData` is `Buffer | ArrayBuffer | Buffer[]` and only the first stringifies to the message; an ArrayBuffer yields `'[object ArrayBuffer]'` and a fragment array comma-joins its halves. A shared `messageText()` handles all three. This one was a latent bug, not a style nit.
- **promise where void was expected** ×2 — Electron's `click` and a hook's `refresh: () => void`.
- **unnecessary assertions** ×2 — including `child.stdout!`, which the compiler confirms was never needed.
- **`Promise<unknown | null>`** — collapses to `unknown`; the null was never expressible.
- **`await` on a non-promise**, **nested ternary**, **indexed loop over a `Uint8Array`**, **duplicate `*` selector**.
- **`ws.ts` `close()`** — complexity 21, two independent branches, split cleanly. The split immediately lost a type narrowing (`ws.data.role` narrows `SocketData`, and that does not survive a call), which is why the data is now passed down explicitly.

### Accepted as wrong

Five more `||` → `??` suggestions, all deliberate empty-string-means-absent:

```ts
githubClientId:   env.GITHUB_CLIENT_ID || null,             // GITHUB_CLIENT_ID= is not a client id
allowedGithubOrg: env.ALLOWED_GITHUB_ORG?.trim() || null,   // whitespace-only is absent
vapidSubject:     env.VAPID_SUBJECT ?? 'mailto:...',        // ?? where empty IS meaningful
```

`config.ts` uses both operators three lines apart, chosen per case. That is the codebase answering the rule.

Running total: **8 of the ~25 `||` findings across this sweep would have introduced bugs** if applied as suggested.

### Accepted as deferred

Seven, each with a tracking issue and a rationale saying Sonar is right:

- **#223** — recharts `Cell` ×4. A real API migration whose correctness is visual.
- **#224** — three protocol dispatchers, complexity 24–60. Splitting live relay routing deserves its own change.

### Acceptance criteria

- [ ] `new_violations` reaches 0 on the branch gate
- [ ] Every accepted finding carries a written reason distinguishing *wrong* from *deferred*
- [ ] All three typechecks, both builds and both suites stay green

William-Long-II:
Closes #225. Completes the sweep started in #221.

## One latent bug

Three sites called `data.toString()` on a websocket payload. `RawData` is `Buffer | ArrayBuffer | Buffer[]`, and **only the first stringifies to the message** — an ArrayBuffer yields `'[object ArrayBuffer]'`, a fragment array comma-joins its halves. Which arrives depends on `binaryType` and on whether the frame was fragmented, so the common idiom is right for the case that usually happens rather than for the type. A shared `messageText()` handles all three.

## One lesson, from the one complexity finding I took

`ws.ts` `close()` (complexity 21) had two independent branches and split cleanly — except the `ws.data.role` check was **narrowing** `SocketData` to the agent or client shape, and narrowing doesn't survive a call. The compiler caught it instantly, and the data is now passed down explicitly.

That is exactly why the other three dispatchers (complexity 24–60) are **not** split here. They carry live traffic between agent, relay and client, and the same mistake there surfaces as a dropped frame, not a type error.

## Also fixed

`db.exec` → `db.run` ×4 (deprecated overload) · `React.FormEvent` → `SyntheticEvent` ×3 (the types deprecate it with *"FormEvent doesn't actually exist"*) · two promise-where-void handlers · two unnecessary assertions, including a `child.stdout!` the compiler confirms was never needed · `Promise<unknown | null>` collapsed to what it always was · an `await` on a non-promise · a nested ternary · an indexed loop over an iterable · a duplicated `*` selector.

## Accepted as *wrong* (5)

Every remaining `||` → `??`. All are deliberate empty-string-means-absent:

```ts
githubClientId:   env.GITHUB_CLIENT_ID || null,            // GITHUB_CLIENT_ID= is not a client id
allowedGithubOrg: env.ALLOWED_GITHUB_ORG?.trim() || null,  // whitespace-only is absent
vapidSubject:     env.VAPID_SUBJECT ?? 'mailto:...',       // ?? where empty IS meaningful
```

`config.ts` uses both operators three lines apart, chosen per case — the codebase answering the rule.

**Across this whole sweep, 8 of ~25 `||` findings would have introduced bugs** if applied as suggested. One in three. That is the argument for reading each rather than bulk-applying, and it's why a fourth site that *looked* identical (`sel.parentId || sel.id`) was fixed rather than accepted after checking that `repositories/groups.ts` normalises it to null upstream.

## Accepted as *deferred* (7)

Distinguished deliberately from the above — here **Sonar is right** and the work is real:

- **#223** — recharts `Cell` ×4. A genuine API migration whose correctness is visual, not testable.
- **#224** — the three protocol dispatchers.

Each carries that reasoning on the finding itself, so nobody later reads an Accept as "we decided this was nonsense".

## On `FormEvent` → `SyntheticEvent`

Not because the handler is shared. The React types deprecate `FormEvent` with the note *"FormEvent doesn't actually exist"* — there is no such DOM event, and `SyntheticEvent` is the honest supertype. Only `preventDefault()` is used at all three sites.

## Verification

- `tsc` clean on both root projects, `relay` and `relay/web`
- `build:web` builds; relay suite 268 pass; root suite 1109 pass (two pre-existing `preload`/`account-auth` collisions, present on base `development`)
- no behaviour change intended outside the websocket-payload fix, which is a correction

### DIFF
--- a/relay/src/db/index.ts
+++ b/relay/src/db/index.ts
@@ -16,8 +16,8 @@ export function openDb(dbPath: string): RelayDb {
     fs.mkdirSync(path.dirname(path.resolve(dbPath)), { recursive: true });
   }
   const db = new Database(dbPath, { create: true });
-  db.exec('PRAGMA journal_mode = WAL;');
-  db.exec('PRAGMA foreign_keys = ON;');
+  db.run('PRAGMA journal_mode = WAL;');
+  db.run('PRAGMA foreign_keys = ON;');
   runMigrations(db);
   return db;
 }
@@ -45,10 +45,10 @@ export function runMigrations(db: RelayDb): void {
 
     const sql = fs.readFileSync(path.join(dir, file), 'utf8');
     db.transaction(() => {
-      db.exec(sql);
+      db.run(sql);
       // user_version does not accept bound parameters; `version` is a validated
       // integer parsed from the filename above, so interpolation is safe here.
-      db.exec(`PRAGMA user_version = ${version};`);
+      db.run(`PRAGMA user_version = ${version};`);
     })();
   }
 }
--- a/src/main/api/ws-message.ts
+++ b/src/main/api/ws-message.ts
@@ -1,0 +1,17 @@
+import type { RawData } from 'ws';
+
+/**
+ * A websocket payload as text.
+ *
+ * `RawData` is `Buffer | ArrayBuffer | Buffer[]`, and only the first of those
+ * has a `toString()` that yields the message: an ArrayBuffer stringifies to
+ * '[object ArrayBuffer]' and a fragment array to comma-joined halves. Which
+ * arrives depends on `binaryType` and on whether the frame was fragmented, so
+ * the common `data.toString()` is right for the case that usually happens
+ * rather than for the type.
+ */
+export function messageText(data: RawData): string {
+  if (Buffer.isBuffer(data)) return data.toString('utf8');
+  if (Array.isArray(data)) return Buffer.concat(data).toString('utf8');
+  return Buffer.from(data).toString('utf8');
+}
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -969,9 +969,9 @@ const App: React.FC = () => {
     }
   }, [sessions, activeSessionId, setActiveSessionId]);
 
-  const handleCloseSession = useCallback(async () => {
+  const handleCloseSession = useCallback(() => {
     if (activeSessionId) {
-      await handleRemoveSession(activeSessionId);
+      handleRemoveSession(activeSessionId);
     }
   }, [activeSessionId, handleRemoveSession]);
 
--- a/src/renderer/store/analytics.ts
+++ b/src/renderer/store/analytics.ts
@@ -59,8 +59,8 @@ export function useAnalytics(timeRange: TimeRange): UseAnalyticsResult {
   }, [timeRange]);
 
   useEffect(() => {
-    loadStats();
+    void loadStats();
   }, [loadStats]);
 
-  return { globalStats, recentEvents, loading, error, refresh: loadStats };
+  return { globalStats, recentEvents, loading, error, refresh: () => { void loadStats(); } };
 }
--- a/relay/src/http.ts
+++ b/relay/src/http.ts
@@ -424,7 +424,7 @@ export function createRouter(ctx: RelayContext) {
 
     const { invite, code } = repos.createShareInvite({
       machineId,
-      expectedGithubLogin: (expectedGithubLogin as string | null) ?? null,
+      expectedGithubLogin: expectedGithubLogin ?? null,
       role: role as 'viewer' | 'operator',
       label: (label as string | undefined) ?? null,
       grantTtlSeconds,
@@ -597,7 +597,7 @@ function publicMachine(m: Machine) {
   };
 }
 
-async function readJson(req: Request): Promise<unknown | null> {
+async function readJson(req: Request): Promise<unknown> {
   try {
     return await req.json();
   } catch {
--- a/relay/src/ws.ts
+++ b/relay/src/ws.ts
@@ -122,31 +122,38 @@ export function createGateway(ctx: WsGatewayContext) {
     notifyGrantRevoked,
 
     close(ws: ServerWebSocket<SocketData>) {
-      if (ws.data.role === 'agent') {
-        const { machineId, authTimer } = ws.data;
-        if (authTimer) {
-          clearTimeout(authTimer);
-          ws.data.authTimer = null;
-        }
-        if (machineId && agents.get(machineId) === ws) {
-          agents.delete(machineId);
-          logger.info('agent offline', { machineId });
-          // Tell any clients bound to this machine that their agent dropped.
-          for (const client of clients.values()) {
-            if ((client.data as ClientSocketData).machineId === machineId) send(client, { type: 'agent:offline' });
-          }
-        }
-      } else {
-        const { clientId, machineId } = ws.data;
-        if (clients.get(clientId) === ws) clients.delete(clientId);
-        // Tell the agent this client went away so it can tear down the session.
-        if (machineId) {
-          const agent = agents.get(machineId);
-          if (agent) send(agent, { type: 'client:closed', clientId });
-        }
-      }
+      // Narrow here and hand the result down: the role check is what tells the
+      // compiler which shape ws.data has, and it does not survive a call.
+      const data = ws.data;
+      if (data.role === 'agent') closeAgent(ws, data);
+      else closeClient(ws, data);
     },
   };
+
+  function closeAgent(ws: ServerWebSocket<SocketData>, data: AgentSocketData): void {
+    const { machineId, authTimer } = data;
+    if (authTimer) {
+      clearTimeout(authTimer);
+      data.authTimer = null;
+    }
+    if (!machineId || agents.get(machineId) !== ws) return;
+
+    agents.delete(machineId);
+    logger.info('agent offline', { machineId });
+    // Tell any clients bound to this machine that their agent dropped.
+    for (const client of clients.values()) {
+      if ((client.data as ClientSocketData).machineId === machineId) send(client, { type: 'agent:offline' });
+    }
+  }
+
+  function closeClient(ws: ServerWebSocket<SocketData>, data: ClientSocketData): void {
+    const { clientId, machineId } = data;
+    if (clients.get(clientId) === ws) clients.delete(clientId);
+    // Tell the agent this client went away so it can tear down the session.
+    if (!machineId) return;
+    const agent = agents.get(machineId);
+    if (agent) send(agent, { type: 'client:closed', clientId });
+  }
 
   async function handleAgent(ws: ServerWebSocket<SocketData>, data: AgentSocketData, msg: Record<string, unknown>) {
     if (!data.authed) {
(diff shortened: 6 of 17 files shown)


src/main/api/ws-message.ts
```diff
@@ -0,0 +1,17 @@
...
+ * A websocket payload as text.
+ *
+ * `RawData` is `Buffer | ArrayBuffer | Buffer[]`, and only the first of those
+ * has a `toString()` that yields the message: an ArrayBuffer stringifies to
+ * '[object ArrayBuffer]' and a fragment array to comma-joined halves. Which
+ * arrives depends on `binaryType` and on whether the frame was fragmented, so
+ * the common `data.toString()` is right for the case that usually happens
+ * rather than for the type.
+ */
+export function messageText(data: RawData): string {
```
brannon-bowden, line 13: New branching logic for Buffer/Buffer[]/ArrayBuffer WebSocket payloads, used on the pairing-auth and message-parsing path, has no unit test coverage in this PR. Please add tests for each branch.

src/renderer/store/analytics.ts
```diff
@@ -59,8 +59,8 @@ export function useAnalytics(timeRange: TimeRange): UseAnalyticsResult {
   }, [timeRange]);
 
   useEffect(() => {
-    loadStats();
+    void loadStats();
   }, [loadStats]);
 
-  return { globalStats, recentEvents, loading, error, refresh: loadStats };
+  return { globalStats, recentEvents, loading, error, refresh: () => { void loadStats(); } };
```
brannon-bowden, line 65: refresh now returns void instead of a Promise. Please confirm no caller awaits or chains off useAnalytics().refresh().

relay/src/db/index.ts
```diff
@@ -45,10 +45,10 @@ export function runMigrations(db: RelayDb): void {
 
     const sql = fs.readFileSync(path.join(dir, file), 'utf8');
     db.transaction(() => {
-      db.exec(sql);
+      db.run(sql);
```
brannon-bowden, line 48: db.exec -> db.run for migration SQL that may contain multiple statements. Please confirm db.run supports multi-statement execution the same way db.exec did.

src/renderer/App.tsx
```diff
@@ -969,9 +969,9 @@ const App: React.FC = () => {
     }
   }, [sessions, activeSessionId, setActiveSessionId]);
 
-  const handleCloseSession = useCallback(async () => {
+  const handleCloseSession = useCallback(() => {
     if (activeSessionId) {
-      await handleRemoveSession(activeSessionId);
+      handleRemoveSession(activeSessionId);
```
brannon-bowden, line 974: handleRemoveSession is no longer awaited and the wrapper is no longer async, so a rejection here becomes an unhandled promise rejection instead of propagating.

brannon-bowden (changes requested):
Request changes — this is a well-scoped Sonar-cleanup PR and almost all of the individual edits are safe, mechanical simplifications, but one of them (`messageText`) is genuinely new parsing logic on a security-relevant path and ships with zero test coverage.

## Blocking issues

- **`src/main/api/ws-message.ts:13`** — `messageText()` is new, non-trivial logic: it branches on `Buffer`, `Buffer[]` (fragmented frames), and `ArrayBuffer` to correctly stringify a WebSocket payload, replacing the previous (buggy) `data.toString()` calls in `relay-client.ts` and `ws-server.ts`. It's used to decode both the pairing auth handshake and post-auth JSON messages — i.e. untrusted network input on a security-sensitive path. This PR touches no test files, so none of the three branches (especially the fragmented-array concat and the ArrayBuffer fallback, which are the cases the old code got wrong) are exercised by anything added here. Please add a unit test for `messageText()` covering a `Buffer`, a `Buffer[]`, and an `ArrayBuffer` input before merge.

## Non-blocking issues

- **`src/renderer/store/analytics.ts:65`** — `refresh` in the hook's return value changes from `refresh: loadStats` (a function returning `Promise<void>`) to `refresh: () => { void loadStats(); }` (a function returning `void`). This is a shape change to what `useAnalytics()` hands back. If any consumer does `await refresh()` or chains off its return value, that now silently no-ops instead of waiting for the reload. Worth a quick check of call sites before merge; not blocking since I can't see the consumers from this diff.
- **`relay/src/db/index.ts:48`** — `db.exec(sql)` → `db.run(sql)` for migration files, which can contain multiple semicolon-separated statements. If `db.run` in this Bun/sqlite binding only executes a single statement (unlike `exec`), multi-statement migrations would silently only apply the first statement or throw. Worth confirming `db.run` has the same multi-statement semantics as `db.exec` here.
- **`src/renderer/App.tsx:974`** — `handleCloseSession` no longer awaits `handleRemoveSession`, and it's not `async` anymore, so a rejection from `handleRemoveSession` becomes an unhandled promise rejection instead of propagating to the caller. If `handleRemoveSession` can throw, worth adding a `.catch()` or logging here.
- **`src/renderer/components/ClaudeAccountsModal.tsx:432`, `NamePromptModal.tsx:76`, `ShareSessionModal.tsx:86`** — widening the submit-handler parameter from `React.FormEvent` to `React.SyntheticEvent` loses some type specificity; harmless since only `preventDefault()` is used, but worth a one-line note in the PR description on why (e.g. handler now shared between form submit and button click) if that's the intent.

## Tests

- No test files are included in this PR. The one change that actually warrants coverage is `messageText()` (see blocking issue above) — everything else here is a like-for-like refactor/type cleanup with no new branching behavior.

William-Long-II:
Blocking issue fixed in `acf4282`, and I checked all three non-blocking ones rather than waving them off — one of them was the right question to ask.

## Blocking — `messageText()` coverage

Fair catch, and the framing is right: it decodes the pairing handshake and every post-auth frame, so it reads untrusted network input, and I shipped new parsing logic on that path with nothing exercising it. Five tests now cover Buffer, `Buffer[]`, and ArrayBuffer.

Two of them assert the **old** behaviour inline rather than only the new:

```ts
expect([whole.subarray(0, cut), whole.subarray(cut)].toString()).not.toBe(FRAME);
expect(String(ab)).toBe('[object ArrayBuffer]');
```

So they demonstrate the defect, not just cover the fix. And note *why* those two shapes had no coverage before: nothing in ordinary local use produces a fragmented frame or an ArrayBuffer, which is how this survived.

There's a fifth that pins the implementation rather than the result — a multi-byte character split across the fragment seam. It only survives because the bytes are concatenated *before* decoding; a per-fragment decode would pass every other test in the file.

## Non-blocking — `db.run` multi-statement semantics

**The right question, and I had not checked.** Verified empirically rather than by reading docs:

```
run   tables: ["a","b"] rows in b: 1
exec  tables: ["a","b"] rows in b: 1
```

Identical for `CREATE TABLE a(x); CREATE TABLE b(y); INSERT INTO b(y) VALUES (1);`. Both statements execute under both. Migrations are safe.

## Non-blocking — `refresh` shape change

Checked the consumers: `AnalyticsPanel.tsx:420` and `:454`, both `onClick={refresh}`. No `await refresh()`, no chaining. `onClick` wants `void`, so returning void is the more correct shape — which is what Sonar was pointing at.

## Non-blocking — `handleCloseSession` no longer awaits

`handleRemoveSession` is `useCallback((id: string) => {…})` — not `async`, and its only `return` is a bare early exit. There is no promise to reject, which is what S4123 was reporting. No `.catch()` to add.

## Non-blocking — `FormEvent` → `SyntheticEvent`

Not a shared handler; the reason is narrower. The React types deprecate `FormEvent` with the note *"FormEvent doesn't actually exist"* — there is no such DOM event, and `SyntheticEvent` is the honest supertype. Only `preventDefault()` is used at all three sites. I'll add that to the PR description.

brannon-bowden (approved):
Approve — the previously-blocking gap is closed: this delta adds `src/main/api/__tests__/ws-message.test.ts`, which exercises `messageText()` across exactly the cases that mattered (a plain `Buffer`, a fragmented `Buffer[]` rejoined without an inserted separator, an `ArrayBuffer`, a multi-byte character split across a fragment boundary, and empty-payload input for both `Buffer` and array forms). That covers the security-relevant parsing path this PR introduced, so I'm lifting the prior `request-changes`. CI is green across all four checks.

## Suggestions (non-blocking, carried over from prior review — not touched by this delta, so unverified against current code)

- `src/renderer/store/analytics.ts:65` — `refresh` in the hook's return shape reportedly changed from returning a `Promise<void>` to a void-returning wrapper; worth confirming no consumer does `await refresh()`.
- `relay/src/db/index.ts:48` — `db.exec(sql)` → `db.run(sql)` for migration files that may contain multiple semicolon-separated statements; worth confirming `db.run` has the same multi-statement semantics.
- `src/renderer/App.tsx:974` — `handleCloseSession` reportedly no longer awaits/propagates a rejection from `handleRemoveSession`; worth a `.catch()` if that call can throw.
- `ClaudeAccountsModal.tsx:432`, `NamePromptModal.tsx:76`, `ShareSessionModal.tsx:86` — parameter type widened from `React.FormEvent` to `React.SyntheticEvent`; harmless if only `preventDefault()` is used, just worth a one-line rationale in the PR description.

None of these were touched by the changes since the last review, so treat them as outstanding notes rather than newly-verified findings.

Issue #214, opened 2026-08-25
Bug: an account switch that changes nothing reports nothing — the click looks broken


William-Long-II:
### What happened

Right-clicking a session and picking an account produced no visible response at all. The click had registered and the database was written; the *effective* account simply hadn't changed, so nothing restarted and nothing was said.

Observed state afterwards:

```
"Session 1"  session override: Brannon
             group account   : Brannon
             => effective    : Brannon   (unchanged)
```

The session had been inheriting Brannon from its group. Picking Brannon explicitly wrote a session-level override of the same account, so:

```js
const before = resolveAccountForSession(sessionId);   // Brannon
sessionsRepo.updateSession(sessionId, { claudeAccountId });
const after  = resolveAccountForSession(sessionId);   // Brannon
if ((before?.id ?? null) === (after?.id ?? null)) return { affectedSessionIds: [] };
```

`affectedSessionIds: []` → `restartSessions([])` → no pty replaced, no banner, no toast, no log line the user can see. Indistinguishable from a menu that didn't register the click.

### Why it matters more than it looks

Two reasons this is worse than a cosmetic gap.

**It happens exactly when the user is under pressure.** The reason to open that menu is usually that the current account has hit a limit. Getting silence back invites the reasonable conclusion that the feature is broken, and the next move is to try something else — which is how the group ends up mis-clicked onto the default (#213).

**It quietly changes the session's semantics.** The click converts an *inherited* assignment into an explicit override pinned to the same account. Nothing says so, and the consequence surfaces later: changing the group no longer moves that session, because its own override now wins. A user who "did nothing" has in fact opted that session out of group-level control permanently.

**#164 and #165 went to real lengths to stop the UI claiming a switch that had not taken effect.** This is the same class of problem from the other side: rather than saying something untrue, it says nothing at all, and the user has to guess.

### Suggested fix

Report the outcome in all three cases, not just the one that restarts something:

- **moved and restarted** — already covered by the restart itself
- **already on that account** — say so, and do not write a pointless override (or say that an override was set and what it means)
- **assignment recorded, applies on next start** — for a stopped session

Cheapest honest version: when `affectedSessionIds` is empty, tell the user why it was empty.

### Related

- #213 — the group-menu counterpart, where a switch commits before the prompt that appears to confirm it

PR #231, opened 2026-08-26
213/214 — An account switch that says what it did, and a default that names where it goes

William-Long-II:
Fixes the two bugs filed out of the failover work — the ones behind "I changed the group from my exhausted work account to Brannon" and finding it back on the exhausted account minutes later, with nothing on screen having said so.

They're one incident from two sides, so they're one PR.

## #214 — a switch that restarts nothing reports nothing

`affectedSessionIds` answers exactly one question — which ptys to replace — and answers it with `[]` in two very different situations: nothing *needed* to move, and nothing *could*. Both reach the user as an unchanged screen, indistinguishable from a menu that never registered the click.

`AccountSwitchResult` now carries an `outcome`: the account the target resolves to, the sessions that stayed put, and which of those were pinned to their own account. The renderer turns that into one line in the window.

The case that speaks for itself stays quiet — a live session about to visibly restart doesn't need a notice narrating it.

The case most worth naming is the one in the issue: picking the account a session **already inherits** writes an override, which silently opts that session out of group-level control. Nothing said so, and the consequence surfaced hours later when a group switch mysteriously left it behind. It now says so.

## #213 — the default that didn't name itself

`Use default account` sat one row above the account list and named no destination. During a usage limit the default is usually the account being fled, so a one-row miss put the group straight back onto it.

- the item now names where it resolves to — `Use default account (Work)`
- a separator keeps it off the account list, so a slip lands on nothing
- the same separator goes into the session menu, where `Inherit account from group` has the same shape
- the restart prompt no longer reads as a confirmation. The assignment is already written when it appears and `Not now` declines only the restart, so it now says so outright rather than leaving "dismiss = cancel" as the natural reading

## Shared shell

`FailoverNotice`'s markup moves to a `Notice` component. An automatic failover and a manual switch are the same event to the user — only the instigator differs — and they shouldn't drift into looking like different kinds of thing.

## What I did not do

The issues also suggest an undo path (#213) and *not* writing a pointless override (#214). I left both alone deliberately:

- **undo** is a real feature with its own design questions, not a bug fix
- **suppressing the override write** would invent new semantics — explicitly picking an account is a reasonable thing to mean "pin here". The bug was the silence, not the write, so the write stays and is now reported

Both are judgement calls rather than omissions; say the word if you'd rather have either.

## Testing

`bun test --isolate` — **1288 pass, 0 fail**. New coverage:

- `accountSwitchReport.test.ts` (9) — every reporting branch, including the two that must stay silent
- `account-switch.test.ts` (+5) — the outcome fields, incl. a group switch separating "already there" from "pinned elsewhere", and an assignment to a deleted account falling through to the default
- `App.accounts.test.ts` (+3) — `defaultAccountMenuLabel`, incl. the no-default-set case
- `AccountSwitchNotice.test.tsx` (3) — the sentence reads as one announcement with the account named inside it

One of my own assertions caught a plural bug while writing them (`2 sessions have its own account`); fixed the string, kept the test.

### DIFF
--- a/src/main/__tests__/account-switch.test.ts
+++ b/src/main/__tests__/account-switch.test.ts
@@ -352,6 +352,17 @@ describe('switch outcome', () => {
     expect(result.affectedSessionIds).toEqual(['s1']);
   });
 
+  test('names no account rather than undefined when no default is configured', () => {
+    // AccountChip renders `account: ClaudeAccount | null`; an undefined
+    // leaking through the default lookup would be a different thing entirely.
+    addGroup('g1');
+    addSession('s1', 'g1');
+
+    const result = accountSwitch.assignGroupAccount('g1', null);
+
+    expect(result.outcome.account).toBeNull();
+  });
+
   test('an assignment to a deleted account falls through to the default', () => {
     addAccount('acct-a', true);
     addGroup('g1');
--- a/src/main/__tests__/account-switch.test.ts
+++ b/src/main/__tests__/account-switch.test.ts
@@ -287,3 +287,78 @@ describe('assignGroupAccount', () => {
     expect(result.affectedSessionIds).toEqual([]);
   });
 });
+
+/**
+ * The outcome (#214) exists so a switch that restarts nothing can still be
+ * explained. `affectedSessionIds: []` is returned both when nothing needed to
+ * move and when nothing could, and the renderer cannot tell those apart — nor
+ * could the user, which is the bug.
+ */
+describe('switch outcome', () => {
+  test('a session pick that changes nothing still names the account and the session', () => {
+    addAccount('acct-a', true);
+    addGroup('g1', 'acct-a');
+    addSession('s1', 'g1');
+
+    // Picking the account the session already inherits: no restart, but the
+    // row is now an override and the session has left group control.
+    const result = accountSwitch.assignSessionAccount('s1', 'acct-a');
+
+    expect(result.affectedSessionIds).toEqual([]);
+    expect(result.outcome.account?.id).toBe('acct-a');
+    expect(result.outcome.unchangedSessionIds).toEqual(['s1']);
+  });
+
+  test('a session that does move reports no unchanged sessions', () => {
+    addAccount('acct-a', true);
+    addAccount('acct-b');
+    addGroup('g1');
+    addSession('s1', 'g1');
+
+    const result = accountSwitch.assignSessionAccount('s1', 'acct-b');
+
+    expect(result.outcome.account?.id).toBe('acct-b');
+    expect(result.outcome.unchangedSessionIds).toEqual([]);
+  });
+
+  test('a group switch separates sessions it could not move from ones already there', () => {
+    addAccount('acct-a', true);
+    addAccount('acct-b');
+    addGroup('g1');
+    addSession('inherits', 'g1');
+    addSession('pinned', 'g1', { accountId: 'acct-a' });
+    addSession('already', 'g1', { accountId: 'acct-b' });
+
+    const result = accountSwitch.assignGroupAccount('g1', 'acct-b');
+
+    expect(result.affectedSessionIds).toEqual(['inherits']);
+    // Both stayed put, but for different reasons — and only one of them is
+    // something the user can act on from the group menu.
+    expect(result.outcome.unchangedSessionIds.sort()).toEqual(['already', 'pinned']);
+    expect(result.outcome.overriddenSessionIds.sort()).toEqual(['already', 'pinned']);
+  });
+
+  test('clearing a group assignment names the default account it falls back to', () => {
+    addAccount('acct-a', true);
+    addAccount('acct-b');
+    addGroup('g1', 'acct-b');
+    addSession('s1', 'g1');
+
+    const result = accountSwitch.assignGroupAccount('g1', null);
+
+    // #213: this is the item sitting one row above the account list, so what
+    // it resolves to has to be knowable rather than inferred.
+    expect(result.outcome.account?.id).toBe('acct-a');
+    expect(result.affectedSessionIds).toEqual(['s1']);
+  });
+
+  test('an assignment to a deleted account falls through to the default', () => {
+    addAccount('acct-a', true);
+    addGroup('g1');
+    addSession('s1', 'g1');
+
+    const result = accountSwitch.assignGroupAccount('g1', 'gone');
+
+    expect(result.outcome.account?.id).toBe('acct-a');
+  });
+});
--- a/src/main/account-switch.ts
+++ b/src/main/account-switch.ts
@@ -4,6 +4,7 @@ import { AccountSwitchResult, ClaudeAccount } from '../shared/types';
 import { getDatabase } from './database';
 import { resolveAccountForSession } from './account-resolver';
 import { carryTranscript, legacyClaudeConfigDir } from './conversation-transcript';
+import * as accountsRepo from './repositories/accounts';
 import * as groupsRepo from './repositories/groups';
 import * as sessionsRepo from './repositories/sessions';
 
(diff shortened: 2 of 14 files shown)


brannon-bowden (changes requested):
Request changes — the reporting logic has a solid design and good test coverage for most branches, but one branch in `reportGroupSwitch` silently drops the exact information this PR exists to surface, in a scenario the PR's own fixtures already construct.

## Blocking issues

**`reportGroupSwitch` loses pinned/unchanged info whenever any session also moved** — `src/renderer/accountSwitchReport.ts:98-107`

```ts
if (result.affectedSessionIds.length > 0) {
  return {
    tone: 'info',
    prefix,
    account,
    suffix: `. ${count(result.affectedSessionIds.length)} in it `
      + `${plural(result.affectedSessionIds.length, 'is', 'are')} stopped and will use it `
      + 'when you start ' + plural(result.affectedSessionIds.length, 'it', 'them') + '.',
  };
}
```

This branch returns as soon as `affectedSessionIds` is non-empty, before the `unchangedSessionIds`/`overriddenSessionIds` handling below it ever runs. So for a group where some sessions move and others are pinned to their own account, the notice only mentions the moved sessions and says nothing about the pinned ones being left behind.

This is directly reproducible with the PR's own fixture from `account-switch.test.ts` ("a group switch separates sessions it could not move from ones already there"): `inherits` moves, `pinned` and `already` do not. Feed that `AccountSwitchResult` into `reportGroupSwitch` and the resulting notice reports only "1 session in it is stopped…", with no mention that two other sessions in the same group didn't move because they're pinned to their own accounts. That's exactly the "switch that appears to do nothing to some sessions" ambiguity #213/#214 set out to fix — reintroduced for the mixed case. It also means a live-restart in this shape (`liveAffected.length > 0`, so `reportGroupSwitch` returns `null` and the confirm-dialog message in `App.tsx` is shown instead) never mentions the pinned sessions either, since that dialog message has the same gap.

Fix: when `result.affectedSessionIds.length > 0` and `unchangedSessionIds.length > 0`, the suffix needs to combine both the "N stopped, will use it on restart" fact and the `describeStuck(already, pinned)` fact, rather than returning early.

## Non-blocking issues

- `src/renderer/accountSwitchReport.ts:62-69` — when a session is unchanged and `pickedAccountId` is non-null, the message always reads "it is now pinned to that account…" even if the session was already explicitly pinned to that exact account before this click (i.e. the user re-picked the same account). Not false, but slightly overstates a "new" consequence that in that specific case isn't new. Minor wording nit, not worth blocking on.
- `src/main/account-switch.ts:117-120` (`groupAccount`) — falls back to `accountsRepo.getDefaultAccount()` when no account is picked or the picked id doesn't resolve. If a workspace somehow has no default account configured, this could yield `undefined` rather than `null`, which wouldn't match the declared `ClaudeAccount | null` return type consumed by `AccountChip`. Worth a quick check against `getDefaultAccount`'s actual signature, but I can't confirm this from the diff alone.
- The `AccountSwitchResult` shape gained a required `outcome` field (`src/shared/types.ts`). Worth double-checking that any other construction sites of this type (IPC handlers/preload marshalling not shown in this diff) either pass the value straight through or were updated — the diff only shows `account-switch.ts` producing it and the renderer consuming it.

## Tests

The new `accountSwitchReport.test.ts` covers `reportGroupSwitch`'s branches individually (affected-only, unchanged-only, all-pinned) but never combines "some affected + some unchanged/pinned," which is precisely the branch with the bug above. Given `account-switch.test.ts` already builds that exact mixed-group fixture at the data layer, a matching case at the reporting layer would have caught this.

William-Long-II:
Fixed the blocking issue, and it was a real gap — thanks for constructing the repro from the data-layer fixture, that's exactly what made it obvious.

## Blocking — `reportGroupSwitch` dropped everything after the first applicable fact

You were right about the cause and right that my own tests should have caught it: I covered each branch in isolation and never combined them, which is the one combination that fails.

Rather than special-case the mixed shape, the function now **composes clauses** instead of picking one — a group switch genuinely can do three things at once, and the early return meant only the first ever got said:

```
“Api Service” now uses Work. 1 session in it is stopped and will use it when you
start it; 1 session was already on it and 1 has its own account, so they did not move.
```

It now returns `null` in exactly one case: the restart prompt is showing *and* there is nothing it could have missed.

**That also closes the second half of your point** — the live case, where the prompt is shown instead. I fixed it in the notice rather than the dialog string: the prompt names only what it's about to restart, so anything it omits belongs in the persistent notice, and duplicating the logic into a modal message would just be a second place to get it wrong. The prompt handles the movers; the notice handles what's left:

```
“Api Service” now uses Work. 1 session has its own account, so it did not move.
```

Tests added at the reporting layer for both mixed shapes — movers-stopped and movers-live — including an assertion that the live one does *not* repeat what the prompt already says.

## Non-blocking

**Wording nit — accepted.** Now reads "it **is** pinned to that account", which is true whether or not the pin is new.

**`getDefaultAccount` returning `undefined`** — checked, it can't:

```ts
export function getDefaultAccount(): ClaudeAccount | null {
  const row = db.prepare('SELECT * FROM claude_accounts WHERE is_default = 1 LIMIT 1').get();
  return row ? mapRow(row) : null;
}
```

Fair thing to flag from the diff alone, though, so I added a test for the workspace-with-no-default case rather than leave it resting on my reading.

**Other construction sites** — also checked: `account-switch.ts` is the only producer of `AccountSwitchResult` (`preload.ts` and `electron.d.ts` only marshal it through), and `tsconfig.json` covers `src/**/*`, so a missed site would have failed the typecheck rather than shipped.

## Verification

`bun test --isolate` — **1290 pass, 0 fail**. I also printed the rendered sentence for all eight outcome shapes and read them, which is how two clunky ones got caught that the assertions were happy with (`1 session was already on it and 1 session has its own account` → `…and 1 has its own account`).

brannon-bowden (approved):
Approve — this update fixes the blocking bug from the previous round and adds fixture-level test coverage for exactly the scenario that exposed it.

The core issue previously flagged — `reportGroupSwitch` returning early on `affectedSessionIds.length > 0` and silently dropping pinned/unchanged info in mixed groups — is resolved. The rewrite in `src/renderer/accountSwitchReport.ts:80-135` now builds a `clauses` array combining the "stopped movers" fact and the `describeStuck` fact instead of picking one branch, and only returns `null` when the restart prompt is showing *and* there's genuinely nothing else to report (`ctx.liveAffected.length > 0 && clauses.length === 0`). I traced both new tests (`accountSwitchReport.test.ts`: "a mixed group reports the sessions that moved AND the ones left behind" and "still reports sessions left behind when a restart prompt is covering the movers") against the new logic and both produce the expected suffixes — the mixed-group case (moved + pinned) now surfaces both facts in one sentence, and the case where the restart prompt already covers the movers correctly omits "stopped" while still surfacing the pinned sessions. This is precisely the gap the earlier review called out, now closed with a matching fixture.

The two non-blocking notes from last round are also addressed: the "it is now pinned" wording was softened to "it is pinned" (`accountSwitchReport.ts:67`), and `account-switch.test.ts` gained a test asserting `assignGroupAccount` returns `null` rather than `undefined` when no default account is configured — confirmed passing per the green `test` CI check.

## Suggestions (non-blocking)

- `src/renderer/accountSwitchReport.ts:127-129` — when a mover-only group has all movers live (`stoppedMovers === 0`) but also has pinned/unchanged sessions, `tone` is `'info'` even though the only visible clause is about sessions that *didn't* move. Not incorrect, just worth a glance if `'muted'`/`'info'` are meant to signal "did something" vs. "just confirming" — minor enough not to block.
- The open question about other `AccountSwitchResult` construction sites (IPC/preload marshalling) from the prior review remains unconfirmed either way in this diff — worth a quick self-check by the author, but there's no new evidence here to treat it as a defect.

CI is fully green (SonarQube, quality-gate, test, GitGuardian) with no failing checks to flag.

Issue #206, opened 2026-08-25
Bug: Terminal.resizeRequest test fails on Linux CI, reddening development's Sonar gate


William-Long-II:
### What happened

`src/renderer/components/__tests__/Terminal.resizeRequest.test.tsx:155` fails on the Linux CI runner:

```
- Expected  - 1      []
+ Received  + 3      [ "164x48" ]
```

A resize reached the PTY that the assertion expects not to. Observed in the SonarQube workflow's coverage run; the Tests workflow passed on the same commit, so it is load- or timing-sensitive rather than deterministic.

### Where it bites

- The **SonarQube workflow on `development` is red**, so the branch quality gate has no clean run.
- It also fails the `quality-gate` check on open PRs (seen on #203), where it reads as that PR's fault. It is not — the test arrived with #185 and the file is unchanged on the branches it fails.

### What is already known (verify, don't re-derive)

- Not reproducible on macOS: 6/6 isolated runs pass, and 3/3 full-suite runs under `--coverage --coverage-reporter=lcov` pass (1007/1007).
- The two workflows differ in that the Sonar job runs the suite with coverage instrumentation, which changes timing.
- The assertion sits in the held-fit work: after an accepted "Fit to my screen", `handleResize` must early-return while `heldFitRef` is set, so focus/resize/session-activation cannot push the desktop grid back. Receiving `164x48` means a re-fit ran anyway — so either the hold was not yet set when the event fired, or an event source the test does not control (ResizeObserver, a queued focus) fired after the assertion window.
- `Terminal.resizeRequest.test.tsx` stubs `getBoundingClientRect`, makes `FitAddon.fit` re-measure a fixed container and `FakeTerm.resize` genuinely move the grid — deliberately, so the doubles can see the bug the hold prevents. That faithfulness is worth keeping; the fix should not go back to no-op doubles.

### Done when

- [ ] The failure's mechanism is identified (not "added a wait until it stopped failing")
- [ ] The test is deterministic under coverage instrumentation and under load
- [ ] The doubles stay faithful — `fit()` still re-measures and `resize()` still moves the grid
- [ ] `development`'s SonarQube workflow runs green
- [ ] Sibling timing-sensitive assertions in the same file and in `Terminal.restart.test.tsx` are checked for the same pattern

William-Long-II:
The blocking half of this is no longer true, but the flake itself is unproven either way — so this is a re-triage, not a close.

## What the issue claims

[...]

PR #232, opened 2026-08-26
228/204 — Wait on the condition, not the clock, in the load-sensitive suites

### LINKED ISSUE #204: Bug: connection.test.ts flakes on cold full-suite runs, and the CI gate now depends on it
### What happened

`relay/web/src/connection.test.ts:226` failed once on a cold full-suite run (`expected length 1, received 0`, after `await Bun.sleep(20)`), then passed on two subsequent runs and 5/5 in isolation. Observed during gate verification of an unrelated branch; `connection.ts`, `crypto.ts` and the spec itself were byte-identical to `development` at the time, so the flake is pre-existing rather than introduced.

### Why it matters now

Until recently the suite ran only when someone remembered to. It is now a required-shaped CI gate on every PR (see the Tests workflow), so a latent timing flake becomes a spurious red check that blocks merges and — worse — trains people to re-run rather than read a red gate. That erosion is exactly what the gate was added to prevent.

### Investigation notes (verify, don't assume)

- The assertion sits after a fixed `Bun.sleep(20)`. A fixed sleep waiting on an async settle is the usual shape of this failure; prefer waiting on the condition (poll with a deadline, or expose a settle promise) over lengthening the sleep.
- It reproduced only under a cold, fully parallel run — load-dependent. Adding test files to the same run plausibly raises the odds, so it will surface more often as the suite grows.
- Check the sibling timing-sensitive specs in the same file and in `reconnect.test.ts` for the same pattern rather than fixing this one line in isolation.

### Done when

- [ ] The assertion no longer depends on a fixed sleep
- [ ] The full suite passes repeatedly from cold (10 consecutive runs is a reasonable bar)
- [ ] Any sibling spec sharing the same fixed-sleep pattern is fixed or explicitly cleared

### LINKED ISSUE #228: Test suite flakes under load in the full-repo run
Surfaced while landing #205 (#200). Not caused by that PR — every affected test lives in code it does not touch, or pre-exists unchanged on `development`.

## Symptom

`bun test --isolate` from the repo root (the CI "Run app suite" step, which sweeps `relay/` too — 83 files) intermittently fails a handful of tests. Roughly **3 of 7 full runs** had at least one failure; the rest were clean at 1263 pass / 0 fail. **Every affected test passes in isolation**, and the specific one checked hardest passed 5/5 runs on its own.

Failures are not a stable set. Observed across runs:

| Test | File |
|---|---|
| `agent WebSocket handshake > answering the challenge disarms the reaper` | `relay/src/ws.test.ts` |
| `the handshake still works > a verified proof reaches ready and carries the certificate` | `relay/web/src/connection.test.ts` |
| `losing access still ends the session > reason expired / reason session_ended / a denied with NO reason` | `relay/web/src/connection.test.ts` |
| `a refused command must not end the session > command:denied routes to onCommandDenied` | `relay/web/src/connection.test.ts` |
| `login detection > a token file appearing still completes the flow` | `src/main/__tests__/account-auth.test.ts` |
| `POST /sessions/:id/stop > the response waits on kill() and reports success only once it settles` | `src/main/api/routes/__tests__/sessions.test.ts` |

Errors are timing-shaped rather than assertion-shaped: `error: ws error`, and `condition was never met` after a 5002ms wait against a 5000ms timeout. Slow runs correlate with failures — a clean run took ~30s, a failing one ~56s on the same tree.

## Why this is worth fixing rather than tolerating

CI has passed on every run so far, so this currently reads as green. But the failure mode is load-dependent, and a busier runner will start failing builds for reasons unrelated to the change under test — which trains everyone to re-run rather than read the failure. It also cost real time during #205: the first full run after a merge showed 4 failures that took investigation to separate from genuine regressions (two of which, that run, were genuine).

## Notes for whoever picks this up

- `POST /sessions/:id/stop` deserves a look first even though it is the least likely culprit — it lives in a file #205 edited, so rule it in or out cleanly. The failing test itself pre-exists on `development` unchanged, and the file passes in isolation, so the evidence points away from it.
- The relay WebSocket tests bind real ports and run concurrently with 80+ other files under the root sweep. Port contention or a shared reaper timer is a likely mechanism.
- `account-auth`s 5s ceiling against a 5002ms observation is a timeout that is simply too tight for a loaded machine, and may just need raising or a fake clock.
- `--isolate` gives each file its own module registry, so this is not mock bleed between files; it is resource or wall-clock contention.

## Acceptance

- 10 consecutive full-repo `bun test --isolate` runs with 0 failures on an otherwise loaded machine.
- No test depends on a real wall-clock deadline it can miss under load.

William-Long-II:
Closes #204
Closes #228

## Reproduced first

The issues describe flakes that never appear on an idle machine, so the first job was a harness that fails on demand:

| condition | result |
|---|---|
| idle, `bun test --isolate` ×5 | **5/5 clean**, ~29s per run |
| all 16 cores saturated, ×3 | **3/3 failed**, ~50s per run |

Each loaded failure hit one of the tests #228 names — `POST /sessions/:id/stop`, `losing access still ends the session`, `the handshake still works`. That harness is what makes the fix checkable rather than hopeful.

## Four wall-clock dependencies, each replaced with the thing it was waiting for

**`relay/web/src/connection.test.ts`** — seven `Bun.sleep(20)` calls waiting on a WebCrypto unseal or a state push. Enough on an idle machine, not under an 83-file parallel run — exactly why these passed 5/5 in isolation and failed in CI (#204). Each now polls the specific condition it was sleeping for, with a generous backstop deadline. A genuine hang still fails on the assertion that follows, which says far more than a bare timeout.

**`src/main/api/routes/__tests__/sessions.test.ts`** — a 25ms sleep waiting for an in-flight `fetch` to reach the route. Under load it hadn't arrived and the assertion read an empty array. Now waits for the kill to actually be requested; the real assertion — that the response has *not* settled — is untouched.

**`src/main/__tests__/account-auth.test.ts`** — `waitFor`'s 5000ms ceiling was also bun's default per-test timeout, so the two raced and the test could die either way (`condition was never met` at 5002ms against a 5000ms limit). The ceiling now sits under a raised per-test timeout, so a real failure reports *what it was waiting on*.

**`relay/src/ws.test.ts`** — the reaper's 50ms auth window had to contain a connect, a challenge round-trip and an Ed25519 signature. Loaded, the reaper fired before the auth meant to disarm it and the test died with `ws error`. The window is now about the reaper rather than about how busy the machine is.

## Deliberately left alone

The 10ms and 150ms grace periods before **negative** assertions in those same files. Under load those can only make a test weaker, never redder — they are not flake sources, and tightening them is a different piece of work with a different justification.

## Verification — #228's own acceptance bar

> 10 consecutive full-repo `bun test --isolate` runs with 0 failures on an otherwise loaded machine.

**10/10, 1291 pass / 0 fail**, every core saturated throughout, at 124–128s per run — roughly 4× the idle baseline and well past the ~50s runs that were failing before the change.

## Not addressed

**#206** (`Terminal.resizeRequest` under coverage instrumentation on Linux) is a distinct mechanism on a platform I can't reproduce on here, and it stays open. Worth noting its premise is now stale in one respect: it says the failure reddens `development`'s Sonar gate, and that gate is currently green at `new_violations: 0`.

## A note on this branch

The commit on it carries another session's message and trailers — a second Claude Code session working in this same checkout committed my uncommitted edits before I did. The code in it is the work described above (verified: the identifiers are ones coined here), but that commit message lacks the `Closes` trailers and the evidence, which is why both are stated in full here. I did not rewrite the shared branch's history to fix the message, since another session may be working on it.

### DIFF
--- a/relay/src/ws.test.ts
+++ b/relay/src/ws.test.ts
@@ -10,6 +10,15 @@ import { toArrayBuffer } from './crypto';
 const logger = createLogger('error');
 const b64 = (b: Uint8Array) => Buffer.from(b).toString('base64');
 
+/**
+ * Auth window for the reaper tests. Large enough that a loaded machine can
+ * finish a real handshake inside it, since a window the handshake cannot meet
+ * tests the machine's speed rather than the reaper.
+ */
+const REAPER_WINDOW_MS = 500;
+/** How far past the window a disarmed socket must survive to prove the point. */
+const REAPER_MARGIN_MS = 300;
+
 function startServer(repos: Repositories, authTimeoutMs?: number) {
   const gateway = createGateway({ repos, logger, authTimeoutMs });
   return Bun.serve({
@@ -213,12 +222,14 @@ describe('agent WebSocket handshake', () => {
   });
 
   test('answering the challenge disarms the reaper', async () => {
-    // The mirror of the test above: with a 50ms timeout, an authenticated
-    // socket that stays quiet well past it must still be alive. Without the
-    // clearTimeout in the auth path this fails.
+    // The mirror of the test above: an authenticated socket that stays quiet
+    // well past the timeout must still be alive. Without the clearTimeout in
+    // the auth path this fails. The window has to outlast the handshake it is
+    // disarmed by — a connect, a challenge round-trip and a signature — or the
+    // reaper fires first and the test measures the machine, not the reaper.
     const repos = freshRepos();
     const { pub, sign, machineId } = await registerMachine(repos);
-    const server = startServer(repos, 50);
+    const server = startServer(repos, REAPER_WINDOW_MS);
     try {
       const c = connect(`ws://127.0.0.1:${server.port}/ws`);
       await c.opened;
@@ -229,7 +240,7 @@ describe('agent WebSocket handshake', () => {
 
       // Prove liveness after the window rather than sleeping blind: a ping
       // round-trip that completes can only happen on an open socket.
-      await Bun.sleep(120);
+      await Bun.sleep(REAPER_WINDOW_MS + REAPER_MARGIN_MS);
       c.ws.send(JSON.stringify({ type: 'ping' }));
       expect(await c.nextOfType('pong')).toMatchObject({ type: 'pong' });
     } finally {
--- a/relay/web/src/connection.test.ts
+++ b/relay/web/src/connection.test.ts
@@ -37,6 +37,18 @@ function ab(u: Uint8Array): ArrayBuffer {
   return b;
 }
 
+/**
+ * Wait until `predicate` holds rather than for a fixed number of milliseconds.
+ * Every assertion below waits on an async settle — a WebCrypto unseal, a state
+ * push. The deadline is a backstop, not a wait: it returns as soon as the
+ * condition holds, and missing it fails on the assertion that follows, which
+ * says far more about what went wrong than a bare timeout would.
+ */
+async function until(predicate: () => boolean, timeoutMs = 5000): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (!predicate() && Date.now() < deadline) await Bun.sleep(1);
+}
+
 /** A stand-in for the browser WebSocket the connection opens. */
 class FakeSocket {
   static last: FakeSocket | null = null;
@@ -129,8 +141,8 @@ async function connected(): Promise<Session> {
     agentX25519Pub: agentXPubB64,
     signature: toB64(sig),
   });
-  // Let the async handshake settle.
-  await Bun.sleep(20);
+  // Let the async handshake settle — it is done when the state lands.
+  await until(() => states.some((x) => x.state === 'ready'));
 
   // Mirror the client's key derivation from the agent side.
   const clientPub = await subtle.importKey('raw', ab(fromB64(clientPubB64)), algo({ name: 'X25519' }), false, []);
@@ -184,7 +196,7 @@ describe('a refused command must not end the session', () => {
     const before = s.states.length;
 
     s.socket.deliver(await s.seal(0, { type: 'command:denied', reason: 'not_permitted', command: 'terminal:input' }));
-    await Bun.sleep(20);
+    await until(() => s.refusals.length > 0);
 
     expect(s.refusals).toEqual(['terminal:input']);
     // No new state at all — the session is untouched.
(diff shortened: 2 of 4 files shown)


William-Long-II:
Stronger evidence than the original verification, arrived at by accident.

I tried to reproduce #206 on macOS by combining the two conditions its notes had only tried separately — coverage instrumentation **and** full CPU saturation. Partway through, the shared working tree was switched off this branch by another session running in the same checkout, so the run executed against `development` instead. That turned into an unintentional but clean A/B:

| branch | condition | result |
|---|---|---|
| `development` (unfixed) | `--coverage` + 16 cores saturated | **4 failures** per run, ×2 runs |
| `fix/228-Bodhilander` (this PR) | identical | **0 failures**, 1291 pass, ×2 runs |

The `development` failures were the expected set — `the handshake still works`, `command:denied routes to onCommandDenied`, `losing access still ends the session` (several reasons), `POST /sessions/:id/stop`.

This is a better result than the original verification. Coverage instrumentation is the harsher condition — it's what the SonarQube workflow runs, and it produced **4 failures per run** where load alone produced 1. The fix holds under it.

I've since moved to an isolated `git worktree` so a concurrent session can't swap the branch mid-run again.

## Two things this turned up that are not for this PR

**`ChatParser real-corpus regression (BDHLNDR-72)`** hit bun's 5000ms per-test ceiling (`5016.05ms`) on one unfixed run. It's the same class of problem as the four fixed here — a real wall-clock limit that a loaded machine can miss — but it's not in #228's list and I haven't characterised it. Worth its own issue rather than being swept in here silently.

**#206 still did not reproduce**, even under coverage + saturation on macOS. That's now a stronger negative than the issue's existing "3/3 full-suite runs under coverage pass" note, and it points harder at something Linux-specific. Recording it so whoever picks up #206 doesn't repeat this attempt.

brannon-bowden (approved):
Approve — this is a well-scoped test-hardening change that replaces fixed-duration sleeps with condition-based polling, which should reduce flakiness under load without altering any production code paths.

## Blocking issues

None.

## Non-blocking issues

- **`src/main/__tests__/account-auth.test.ts:227`** — the default `timeoutMs` for `waitFor` was raised from 5000ms to `WATCH_TIMEOUT_MS` (15000ms), and only the two tests explicitly touched in this diff got a matching bump to their per-test timeout (`WATCH_TEST_TIMEOUT_MS`). If any other test in this file calls `waitFor()` with the default timeout and still relies on the framework's original default test timeout, it could now hit the test-runner's timeout before `waitFor`'s internal deadline elapses — reintroducing the bare-timeout failure this PR is trying to eliminate. Worth double-checking the rest of the file for other `waitFor` call sites that might need the same per-test timeout bump. (Not verifiable from this diff alone, so treated as a question rather than a blocker.)
- **`src/main/api/routes/__tests__/sessions.test.ts:81-87`** — this `until` helper is byte-for-byte identical to the one added in `relay/web/src/connection.test.ts:40-49`. Worth pulling into a shared test-utils module rather than maintaining two copies.

## Tests

No new logic was introduced — these are test-infrastructure changes to existing tests, so no additional coverage is needed. The rationale comments explaining *why* each wait now targets a condition instead of a duration are a nice touch for future maintainers.

PR #234, opened 2026-08-26
201 — Relay-carried machine handoff: prepare on the old machine, restore on the new

### LINKED ISSUE #201: Feature: Relay handoff — sign in on the new machine and restore
### User story / problem

As someone who just installed Bodhilander on a new machine, I want to sign in and find my previous machine's state offered to me, so that restoring is something I accept rather than a file I have to remember to carry.

Phase 1 (#200) makes the state portable. This phase removes the USB stick.

### Proposed solution

Ride the identity model the relay already has. `users` (GitHub OAuth) → `machines` (Ed25519/X25519, linked by `POST /link` plus a code) already permits a second machine under one identity — no new account concept is required.

**On the old machine:** an explicit "prepare handoff" action builds the Phase 1 bundle, encrypts it client-side, and uploads the ciphertext to the relay under the user's id. The user is shown a recovery phrase; the key is derived from it and never leaves the machine.

**On the new machine:** after linking, if a bundle exists for the user, offer "restore from `<machine name>`". The user enters the phrase, the bundle is pulled and decrypted locally, and Phase 1's importer runs — same code path, same remapping step.

**Relay side:** stores an opaque blob plus routing metadata. It must not be able to read the bundle, consistent with how it already handles terminal channels. One bundle per user at a time, TTL'd, size-capped, and deleted after a successful restore acknowledgement.

The new machine generates its own relay identity as it always has. The bundle carries no keypair, so there is nothing to accidentally reuse.

### Acceptance criteria

- [ ] "Prepare handoff" builds a Phase 1 bundle, encrypts it client-side, uploads ciphertext, and displays a recovery phrase
- [ ] The relay never receives the key or the plaintext; a test asserts the stored blob is not decryptable with anything the relay holds
- [ ] Bundle storage is TTL'd, size-capped, and one-per-user; an existing bundle is replaced rather than accumulated
- [ ] Upload and download routes are rate-limited and quota'd, consistent with the hardening tracked in #181
- [ ] After linking, a new machine with a bundle available offers to restore from it, named by the source machine
- [ ] A wrong recovery phrase fails clearly and does not consume or damage the stored bundle
- [ ] Restore invokes the Phase 1 importer unchanged, including the working-directory remapping step
- [ ] The bundle is deleted from the relay after the new machine acknowledges a successful restore
- [ ] The new machine holds a freshly generated relay identity; the old machine's identity is untouched and still linked
- [ ] Declining the restore offer is durable — it is not re-offered on every launch

### Scope

Major feature

### Priority

Medium

William-Long-II:
Closes #201. Phase 2 of #199 — Phase 1 (#200) made a machine's state portable as a file; this removes the USB stick.

**Ready for review.** All four checks are green, including the Sonar quality gate at 0 new issues and 0 hotspots. This body was written when the PR opened as a draft; the Verification section below has been rewritten against the current head, and [a comment](https://github.com/Software-Development-LLC/Bodhilander/pull/234#issuecomment-5431378380) records everything that changed after the draft — two CI failures neither of which reproduced on a Mac, and the Sonar pass.

## What it does

**Old machine.** "Move to Another Machine" builds the Phase 1 bundle, seals it locally, uploads only the ciphertext to the user's slot on the relay, and shows a recovery phrase once. One confirmation dialog, naming the size and the counts, before anything leaves the machine.

**New machine.** After linking, a waiting bundle is offered by the name of the machine that prepared it. The phrase opens it locally and `restoreTransferBundle` runs — the same importer the file picker drives, reached through the same working-directory remapping prompt.

**The relay is a courier.** It stores bytes plus routing metadata. Nothing in its database opens them, which `handoff-http.test.ts` asserts directly rather than by inspection.

## Relay (`relay/`)

- **`db/migrations/004_handoff.sql`** — `handoff_bundles`. `user_id` is `UNIQUE`, so one-bundle-per-account is a schema guarantee rather than a rule the upload path has to remember: an upload is an upsert on that key. Foreign keys cascade from both `users` and `machines`, so unlinking the source machine withdraws the offer. New file rather than an edit to `003`, because `runMigrations` skips any version not greater than the current `user_version` and editing an applied migration is a silent no-op on every existing database.
- **`repositories.ts`** — `putHandoffBundle` (upsert; the id changes with the row, so a destination that declined the old bundle is offered the new one), `getHandoffBundle` (filters on `expires_at` itself rather than trusting the ten-minute reaper), `deleteHandoffBundle`, `purgeExpiredHandoffBundles`.
- **`protocol.ts`** — four separately versioned signed messages: `handoff-put:v1`, `handoff-meta:v1`, `handoff-get:v1`, `handoff-delete:v1`. A signature captured off a download cannot be replayed as the delete that throws the bundle away. The delete message binds the bundle id, so acknowledging a restore cannot drop a newer bundle prepared in the meantime.
- **`http.ts`** — `PUT`/`GET`/`DELETE /api/machines/:id/handoff` and `GET …/handoff/bundle`. Ed25519-signed rather than cookie-authed, because the desktop app holds no cookie; the slot belongs to the machine's *user*, which is how a second machine under one identity reaches what the first left. `PUT` verifies the signature over a header-declared SHA-256 **before buffering any body**, then hashes the arriving bytes and refuses on mismatch — an unsigned caller cannot make the relay buffer 64 MiB on its way to a refusal. The delete id is constrained to a UUID shape before it reaches the signed lines, so a value carrying a newline cannot shift the later fields.
- Rate limits: 5 uploads per IP and 5 per machine, 30 reads per IP. The per-machine budget is charged *after* the signature verifies, so nobody can exhaust another machine's allowance by naming its id.
- **`config.ts`** — `handoffTtlSeconds` (7 days) and `handoffMaxBytes` (64 MiB), both env-overridable via `HANDOFF_TTL_SECONDS` / `HANDOFF_MAX_BYTES`. The existing reaper now also calls `purgeExpiredHandoffBundles`.

## Desktop

- **`transfer/recovery-phrase.ts`** — the phrase is generated, never chosen: 16 entropy words carrying a byte each plus 2 checksum words, over a 256-word list where no word is a prefix of another and no two are one edit apart. HKDF-SHA256 to the 32-byte key — the input is 128 CSPRNG bits, so there is no dictionary for a slow KDF to defend against and a deliberately slow derivation would only tax the person restoring.
- **`transfer/handoff-crypto.ts`** — seals through the existing `seal`/`open` in `api/relay/e2e.ts`, so there is one AEAD implementation rather than two. Nonce fixed at zero is safe here because the key comes from a phrase generated for this one bundle and never reused.
- **`transfer/handoff.ts`** — transport-agnostic mechanics. The phrase is decoded **before** the download, so a typo is caught by its checksum without the stored bundle ever being touched. `acknowledge` fires only after the import returns, so a failed restore leaves the bundle there to try again. A decline is remembered against the bundle id, not the slot, so it is durable across relaunch but does not silence a later bundle from the same machine.
- **`api/relay/handoff-transport.ts`** — the four calls, signed. `peek` returns null for the bundle this machine prepared itself. A `404` on acknowledge is treated as the state the caller wanted.
- **`machine-handoff.ts`**, **`preload.ts`**, **`src/main/index.ts`** — the Electron half and the four IPC channels.
- **`renderer/components/MachineHandoff.tsx`** — the offer, the phrase display, and the phrase entry with a live word count.
- **`group-import-export.ts`** — the only change to existing desktop logic: `askRootMappings` and `registerRestoredAccountHooks` are now exported, so the restore path uses the same prompts as the file picker instead of a second copy. `bundle-import.ts` is untouched.

## Verification

The harness copy pinned in `initiatives/201/team.yaml` is not present in the environment this was finished in, so `scripts/verify.sh` has not been re-run since the draft's report. What follows is direct `bun test` and `tsc`, not a harness verdict.

Everything below is **bun 1.4.0**, the version `test.yml` and `sonarqube.yml` both pin — not the 1.3.x a local checkout is likely to have. That distinction turned out to matter twice; see the comment.

```
$ bun test --isolate <the 14 spec files this PR adds or touches>

 144 pass
 0 fail
 375 expect() calls
Ran 144 tests across 14 files. [17.76s]

$ bun test --isolate                       # the whole repo, relay included

 1415 pass
 1 fail
Ran 1416 tests across 100 files.
```

Both typechecks clean, root and `relay/`.

12 new spec files, plus additions to `preload.test.ts` and the migration assertion in `relay.test.ts` moved to `user_version` 5.

Sonar reports **97.90% coverage on new code**, 0 new issues, 0 duplicated lines.

### The one failing test, and why it is not this branch

`login detection > a token file appearing still completes the flow`, in `src/main/__tests__/account-auth.test.ts`, times out at 15s under full-suite load. It passes 3/3 in isolation, nothing in this branch touches it or its subject, and it is the load-sensitive flake #228 was filed for — still alive after #232, which is worth knowing since #228 is closed. CI's own `test` job is green.

### What is not covered

`src/main/index.ts` — 4 `safeHandle` registrations. The file boots the Electron app at import, so no spec can load it. The four channel names are pinned a step away by `preload.test.ts`, which asserts the exact strings `handoff:prepare`, `handoff:peek`, `handoff:restore` and `handoff:decline` that `index.ts` registers. That is coverage of the contract, not of the wiring line — read it as such.

`relay/src/index.ts` is no longer in this list: `main()` is exported and guarded by `import.meta.main`, and `index.test.ts` starts the real server to prove it serves at the shipped body ceiling.

`lint` has no linter this repo can drive.

## One deviation from the brief

The brief called for sodium's CSPRNG for the phrase entropy. `crypto.randomBytes` is used instead, because **`sodium-native` is absent from both `package.json` files, is not installed in either `node_modules`, and is referenced nowhere in `src/` or `relay/src/`.** I checked that independently before writing it here. `crypto.randomBytes` is the correct call given that, not a shortcut around a missing dependency.

`CLAUDE.md` line 65 still lists `sodium-native` among the project's native modules. It is stale on that point, and this PR does not change it — flagging it rather than editing an instruction file from a feature branch.

https://claude.ai/code/session_01D3LTFJRHsTU7qGiz1C5Nge

Order: Bodhilander

### DIFF
--- a/relay/.env.example
+++ b/relay/.env.example
@@ -27,6 +27,15 @@ ALLOWED_ORIGINS=
 # Lifetime of a machine-linking code, in seconds.
 LINK_CODE_TTL_SECONDS=600
 
+# Machine handoff: how long a prepared bundle stays available, the ceiling on
+# one bundle, and the ceiling on the whole store. Bundles are written to
+# handoffs/ beside DB_PATH, so the disk behind that path needs the store size.
+HANDOFF_TTL_SECONDS=604800
+HANDOFF_MAX_BYTES=268435456
+HANDOFF_STORE_MAX_BYTES=8589934592
+# Override only if bundles should not live beside the database.
+HANDOFF_DIR=
+
 # GitHub OAuth app credentials — used by Milestone 2, optional for now.
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
--- a/relay/docker-compose.yml
+++ b/relay/docker-compose.yml
@@ -23,6 +23,8 @@ services:
       - path: .env
         required: false
     volumes:
+      # Holds relay.db and handoffs/, where sealed transfer bundles are written.
+      # Sized for HANDOFF_STORE_MAX_BYTES (8 GiB by default); see .env.example.
       - relay-data:/data
     restart: unless-stopped
 
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -9,10 +9,10 @@ import { ConfigError, loadConfig } from './config';
 const BASE = { NODE_ENV: 'test', PUBLIC_URL: 'http://relay.test' };
 
 describe('handoff storage limits', () => {
-  test('default to a week and 64 MiB', () => {
+  test('default to a week and 16 MiB', () => {
     const { config } = loadConfig(BASE);
     expect(config.handoffTtlSeconds).toBe(7 * 24 * 60 * 60);
-    expect(config.handoffMaxBytes).toBe(64 * 1024 * 1024);
+    expect(config.handoffMaxBytes).toBe(16 * 1024 * 1024);
   });
 
   test('are overridable per deployment', () => {
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -9,10 +9,17 @@ import { ConfigError, loadConfig } from './config';
 const BASE = { NODE_ENV: 'test', PUBLIC_URL: 'http://relay.test' };
 
 describe('handoff storage limits', () => {
-  test('default to a week and 16 MiB', () => {
+  test('default to a week, 256 MiB per bundle and 8 GiB across the store', () => {
     const { config } = loadConfig(BASE);
     expect(config.handoffTtlSeconds).toBe(7 * 24 * 60 * 60);
-    expect(config.handoffMaxBytes).toBe(16 * 1024 * 1024);
+    expect(config.handoffMaxBytes).toBe(256 * 1024 * 1024);
+    expect(config.handoffStoreMaxBytes).toBe(8 * 1024 * 1024 * 1024);
+  });
+
+  test('put bundles beside the database, which is what the volume holds', () => {
+    const { config } = loadConfig({ ...BASE, DB_PATH: '/data/relay.db' });
+    expect(config.handoffDir).toBe('/data/handoffs');
+    expect(loadConfig({ ...BASE, HANDOFF_DIR: '/elsewhere' }).config.handoffDir).toBe('/elsewhere');
   });
 
   test('are overridable per deployment', () => {
@@ -24,6 +31,7 @@ describe('handoff storage limits', () => {
   test('fail loudly rather than falling back to a default nobody chose', () => {
     expect(() => loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '0' })).toThrow(ConfigError);
     expect(() => loadConfig({ ...BASE, HANDOFF_MAX_BYTES: 'lots' })).toThrow(ConfigError);
+    expect(() => loadConfig({ ...BASE, HANDOFF_STORE_MAX_BYTES: '0' })).toThrow(ConfigError);
     expect(() => loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '-1' })).toThrow(ConfigError);
   });
 });
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -1,0 +1,29 @@
+/**
+ * The handoff knobs an operator can turn. Both have a shipped default the
+ * desktop assumes, so a deployment that overrides one and not the other is
+ * worth being able to see.
+ */
+import { describe, expect, test } from 'bun:test';
+import { ConfigError, loadConfig } from './config';
+
+const BASE = { NODE_ENV: 'test', PUBLIC_URL: 'http://relay.test' };
+
+describe('handoff storage limits', () => {
+  test('default to a week and 64 MiB', () => {
+    const { config } = loadConfig(BASE);
+    expect(config.handoffTtlSeconds).toBe(7 * 24 * 60 * 60);
+    expect(config.handoffMaxBytes).toBe(64 * 1024 * 1024);
+  });
+
+  test('are overridable per deployment', () => {
+    const { config } = loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '3600', HANDOFF_MAX_BYTES: '1048576' });
+    expect(config.handoffTtlSeconds).toBe(3600);
+    expect(config.handoffMaxBytes).toBe(1048576);
+  });
+
+  test('fail loudly rather than falling back to a default nobody chose', () => {
+    expect(() => loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '0' })).toThrow(ConfigError);
+    expect(() => loadConfig({ ...BASE, HANDOFF_MAX_BYTES: 'lots' })).toThrow(ConfigError);
+    expect(() => loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '-1' })).toThrow(ConfigError);
+  });
+});
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -22,6 +22,14 @@ describe('handoff storage limits', () => {
     expect(loadConfig({ ...BASE, HANDOFF_DIR: '/elsewhere' }).config.handoffDir).toBe('/elsewhere');
   });
 
+  test('refuse an in-memory database with nowhere named to put the bundles', () => {
+    // There is no directory to sit beside, and the obvious stand-in — a fixed
+    // name under the system temp directory — is a world-writable path holding
+    // other people's sealed bundles. Say so instead of picking it.
+    expect(() => loadConfig({ ...BASE, DB_PATH: ':memory:' })).toThrow(ConfigError);
+    expect(loadConfig({ ...BASE, DB_PATH: ':memory:', HANDOFF_DIR: '/elsewhere' }).config.handoffDir).toBe('/elsewhere');
+  });
+
   test('are overridable per deployment', () => {
     const { config } = loadConfig({ ...BASE, HANDOFF_TTL_SECONDS: '3600', HANDOFF_MAX_BYTES: '1048576' });
     expect(config.handoffTtlSeconds).toBe(3600);
--- a/relay/src/config.ts
+++ b/relay/src/config.ts
@@ -18,6 +18,10 @@ export interface RelayConfig {
   /** Extra allowed web origins; empty = same-origin only. */
   allowedOrigins: string[];
   linkCodeTtlSeconds: number;
+  /** How long a prepared machine handoff stays available to be restored. */
+  handoffTtlSeconds: number;
+  /** Ceiling on one handoff's sealed bytes. Larger states move by file instead. */
+  handoffMaxBytes: number;
   /** GitHub OAuth app credentials — consumed in M2, optional for now. */
   githubClientId: string | null;
   githubClientSecret: string | null;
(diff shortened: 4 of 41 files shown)


William-Long-II:
## Since the draft opened: CI now runs, and it is green except for one gate condition

The draft said it would leave draft when its checks were green. Here is where they actually stand, and what moved.

**`development` merged in** (`da18851`) — the branch was 6 behind, including #232's flake fixes and #235's preference allowlist. Clean merge.

### `test` — was failing, now passes

Two failures, and neither was visible on a Mac:

**1. `relay/src/server.test.ts` pinned a Bun behaviour that changed.** `does not apply the ceiling to a body read as a stream` was a characterisation of Bun, and Bun 1.4 — the version `test.yml` and `sonarqube.yml` both pin — no longer behaves that way. Measured on this branch:

| bun | streamed body over `maxRequestBodySize` |
|---|---|
| 1.3.14 | handler saw all 524288 bytes, `200` |
| 1.4.0 | reader threw, `413` with an empty body, generator stopped at 131072 |

No relay behaviour depends on which. `writeHandoff` throws `HandoffTooLarge` at the cap and Bun's ceiling sits 64 KiB above it, so an oversized upload is still refused by the handler with a reason before the socket can refuse it without one — and the partial is unlinked either way. The test is gone; what it was protecting (that `readJson` must count bytes itself, because the ceiling here *is* the handoff cap and not a JSON bound) is now asserted inside `bounds a chunked body`, where it can be seen happening.

**2. `src/main/api/relay/__tests__/relay-client.test.ts` was running zero tests on CI.** It reaches `pty-manager`, which loads node-pty's native binding at import — and this checkout ships prebuilds for darwin and win32 only. bun reports that as an unhandled error *between* tests rather than a failure, so the file passed locally and contributed nothing on every CI run since it was added. `handoffTransport()` had no coverage there at all.

Counted per new spec file in the original run: every other one reported its tests, this one reported none. Proved by removing the darwin binding locally, which reproduces the runner exactly — without the stub `0 pass / 1 fail / 1 error` with the same require stack, with it `2 pass`. Same `mock.module('node-pty', …)` that `sessions.test.ts` already uses.

### Sonar — ran for the first time, 5 issues, all cleared

It had never run: the scan job runs the suite first, so it died on the same two failures.

- **The restore offer is now a native `<dialog>` + `showModal()`**, which is the pattern `OwnerConfirmModal`, `ShareSessionModal` and `ProviderInstallModal` already use here — the browser owns modality, the focus trap and the backdrop instead of markup claiming them. Escape is refused rather than allowed to close it: declining is durable, so a stray keypress must not spend the offer, and the dialog must not vanish leaving it neither restored nor declined. That is exactly what the overlay did before, so no behaviour moves. Two tests, both checked against a control — remove `onCancel` and the Escape test fails, use `show()` instead of `showModal()` and the modality test fails. The Escape one asserts `defaultPrevented` and not `open`, because happy-dom does not close a dialog on `cancel` by itself and asserting it stayed open would have passed with no handler at all.
- `role="group"` on the phrase block → a named `<section>`; its spec asks for `region` now.
- `sweepOrphans`' suffix ternary chain → `idFromFilename`.
- The nested template literal in `refusal()` → a named `because`.
- `PHRASE_WORDS[b]!` asserted nothing — this project does not set `noUncheckedIndexedAccess`.

**Also fixed, and the more interesting one:** `loadConfig` fell back to a *fixed name* under the system temp directory when `DB_PATH` was `:memory:` — a world-writable path, shared by name with every other user on the host, holding sealed transfer bundles. Nothing reached it; every caller here already passes `HANDOFF_DIR` beside `:memory:`. It now raises `ConfigError`, which is the convention `config.test.ts` already calls *"fail loudly rather than falling back to a default nobody chose"*.

### What is still red, and why I cannot close it

```
Quality Gate failed
- 0.00% Security Hotspots Reviewed on New Code (is less than 100.00%)
```

That is the only remaining condition — new issues are at zero. **Security hotspots are resolved by a person reviewing them in the SonarQube UI, not by a code change**, and the list is not in the check run, the annotations, or the bot's PR comment. `api/hotspots/search` on `sonarqube.bodhilabs.dev` returns `401` without a token, so I could not enumerate them to say whether any deserve a fix rather than a review.

Test files are excluded by `sonar-project.properties`, so it is production sources only. The likeliest source is `os.tmpdir()` (S5443) — `src/main/machine-handoff.ts:120` stages a bundle in one via `mkdtempSync`, which is the same thing `group-import-export.ts:249` has always done and is the recommended mitigation for that rule. I removed the one `os.tmpdir()` use that was genuinely indefensible, above; that did not take the count to zero, so there is at least one more I cannot see.

### Verification

`1415 pass / 1 fail` under bun 1.4.0 across the whole repo. The one failure is `login detection > a token file appearing still completes the flow` timing out at 15s under full-suite load — it passes 3/3 in isolation, lives in `account-auth.test.ts`, and is the load-sensitive flake #228 was filed for. Nothing in these commits touches it.

Both typechecks clean, root and relay.

The harness copy pinned in `initiatives/201/team.yaml` is not present in this environment, so `scripts/verify.sh` has not been re-run since the draft's report — the numbers above are direct `bun test` and `tsc` runs, not a harness verdict.

brannon-bowden (approved):
Approve. This is a large, security-sensitive feature (relay-carried machine handoff) but it's implemented with real care: the relay is genuinely a blind courier (bundles are sealed client-side with a key derived from a phrase that never leaves the source machine, verified by the sweep test in `relay/src/handoff-http.test.ts` that tries every value the relay actually stores against the ciphertext), machine identity is proven via Ed25519 signatures scoped per-verb with anti-replay binding (the delete message binds the specific bundle id so a captured download signature can't be replayed as a delete), uploads are streamed to disk with a hard cap enforced mid-stream rather than trusting `content-length`, and cross-account isolation is enforced server-side (`getHandoffBundle(userId)`), not just client-side. Test coverage across the relay HTTP surface, the store, the crypto, the phrase encoding, and the Electron/renderer glue is unusually thorough. CI is green across all four required checks.

**Blocking issues**

None found.

**Non-blocking issues**

- `relay/src/repositories.ts:657` (`totalHandoffBytes`) / `relay/src/http.ts` (`handleHandoffPut`'s `tooFull` checks): the store-size cap is enforced by reading the current total, then later writing/committing — there's no transaction spanning the read and the eventual insert, so two concurrent uploads from different users can each pass the pre-check and jointly push the store past `HANDOFF_STORE_MAX_BYTES`. Given per-IP/per-machine rate limits (5/window) and that this is a soft disk-safety cap rather than a security boundary, the impact is limited and self-heals via TTL expiry — but worth knowing about if `HANDOFF_STORE_MAX_BYTES` is ever tightened to something operationally load-bearing.
- `src/shared/types.ts:536` (`HANDOFF_MAX_BYTES`) duplicates the relay's default `handoffMaxBytes` as a compile-time constant in the desktop build. The cross-tree test only pins that the two *defaults* agree; if a deployment overrides `HANDOFF_MAX_BYTES` on the relay without a matching desktop release, the client's pre-upload size check will drift from what the server actually enforces. Real enforcement stays server-side and the transport surfaces a clear 413 message either way, so this is cosmetic rather than a correctness/security gap.
- `relay/src/http.ts` (routing block): the DELETE `/handoff` acknowledgement path shares the `'handoff-read'` rate-limit bucket with GET requests. Not a bug, just a slightly surprising bucket name for a mutating call — a burst of restores-and-acknowledges could exhaust budget that reads also draw from.

**Tests**

No gaps worth calling out — the suite already covers digest mismatch, oversized/streamed uploads, store-full handling, TTL expiry (row + file), orphan sweep vs. in-flight uploads, replace-on-reupload semantics, cross-account isolation, signature/replay protection, chunked-body ceilings, phrase checksum/typo handling, and the renderer's modal/decline/restore flows.

PR #235, opened 2026-08-26
227 — Export preferences by allowlist, and make a new key impossible to forget

### LINKED ISSUE #227: Export preferences by allowlist rather than denylist
Carried forward from review on #205 (raised twice by @brannon-bowden, non-blocking both times).

## What

`isPortablePreferenceKey` in `src/main/transfer/bundle-format.ts` decides what goes into a transfer bundle by **denylist** — `EXCLUDED_PREFERENCE_PREFIXES` (`providerApiKey.`, `providerApiKeyUse.`, `relay.`) and `EXCLUDED_PREFERENCE_KEYS` (`teamsTokens`, `windowBounds`, and the rest).

It is correct today and well covered by tests. The problem is the direction it fails in: **a preference key that does not exist yet is exported by default.** Any future key that happens to hold a secret ships in the bundle unless whoever adds it remembers to come back to this file and exclude it. The bundle is a portable file explicitly meant to be moved between machines and plausibly handed to someone else, so that default is the wrong way round.

## Why it was not done in #205

Flipping to an allowlist there would have silently dropped legitimate settings from every bundle written by the current build, and the risk it addresses is hypothetical rather than present. It is worth doing deliberately against the whole preferences surface, with a pass over every key that exists today, rather than as a rider on a feature PR.

## Suggested shape

Either:

1. An explicit allowlist of known-portable keys, with a test that fails when a key exists in the preferences table that is on neither list — so adding a preference forces a decision instead of defaulting to "export it"; or
2. An enforced naming convention (`secret.*` / `local.*`) applied at the point preferences are written, with the export filter reading the convention.

Option 1 is the smaller change and gives the forcing function; option 2 is better if the preferences surface is going to keep growing.

## Acceptance

- A new preference key cannot reach a bundle without someone explicitly classifying it.
- Every key currently exported still is (no silent regression for existing users).
- The canary-style test from #205 still passes — a seeded secret is absent from the raw bytes AND from the decompressed, readable text.

William-Long-II:
Closes #227

Carried forward from review on #205, raised twice as non-blocking.

## The bug it predicted had already happened

The issue argues the denylist "fails in the wrong direction" — a preference added later ships in the bundle unless someone remembers this file. Enumerating the whole preferences surface to build the allowlist turned up exactly that, live on `development`:

**`quotaCooldownsCleared` was being exported.** It's a one-time migration marker. Its three siblings — `legacyCodeSearchCleanupDone`, `legacyMemoryCleanupDone`, `legacyMemoryMcpCleanupDone` — are all explicitly excluded, so the intent was never ambiguous; this one was simply missed when it was added (by me, earlier today). Carrying it tells a destination machine a cleanup already ran on a database where it hasn't.

Under an allowlist it never shipped in the first place. That is the whole argument for the change, and it didn't need a hypothetical.

## Shape — option 1 from the issue

`PORTABLE_PREFERENCE_KEYS` now decides; `LOCAL_PREFERENCE_KEYS` and `LOCAL_PREFERENCE_PREFIXES` are kept as the written record of what was considered and rejected, which is what lets the policy test tell "deliberately local" from "nobody has looked at this yet".

The safe default is silent on its own — a legitimate new setting would quietly stop travelling and nobody would find out until a user noticed. So `isClassifiedPreferenceKey` plus `bundle-preference-policy.test.ts` turn that silence into a failing build: the test scans source for preference literals and fails on any key on neither list.

It's a lint rather than a unit test, deliberately. It makes "I added a preference" a decision written down in the file where the consequence lives.

**It caught a bug in itself on the first run** — the constant regex matched `*_PREFIX` names (`TRANSCRIPT_PREFIX`, `INSTALL_PREFIX`) and dragged in three non-keys. Tightened to names ending `PREF`/`PREF_KEY`. There's now a guards-the-guard test asserting the scan finds a known-portable and a known-local key, so a regex that silently matched nothing can't make the whole file pass forever.

## The existing tests were asserting on keys that do not exist

`bundle-format.test.ts` checked that `theme`, `soundEnabled`, `betaChannel` and `apiPort` were portable. **None of those four appear anywhere in non-test source** — this app has never written them. Under a denylist they passed for the same reason any unknown string did, so "keeps the ordinary settings a user would want back" was verifying nothing at all. The `db-fixture` seeded `theme` too, which is why four tests failed the moment the default flipped.

Fixtures and assertions now use real keys (`fontSize`, `closeToTray`, `soundVolume`, `updateChannel`, `webglRenderer`). This is the part I'd most like a second opinion on — it's the change that alters what the existing suite covers, and it's worth confirming I classified the real surface correctly rather than plausibly.

## Acceptance, against the issue's own list

- **A new key cannot reach a bundle without classification** — allowlist decides; policy test fails the build on an unclassified key.
- **Every key currently exported still is** — full inventory below. The only behaviour change is `quotaCooldownsCleared`, which is the bug.
- **The canary test still passes** — a seeded secret absent from the raw bytes and from the decompressed text; untouched and green.

### Classification

**Portable (16):** `fontSize`, `webglRenderer`, `closeToTray`, `autoLaunchClaude`, `sidebar.showActiveOnly`, `updateChannel`, `enableNotifications`, `notificationSound`, `soundVolume`, `soundDebouncePreset`, `soundWaitingEnabled`, `soundErrorEnabled`, `soundStartEnabled`, `soundCompleteEnabled`, `accountFailoverEnabled`, `accountFailbackEnabled`

**Local:** `teamsTokens`; the machine-shaped ones (`windowBounds`, `customShellPath`, `preferredEditor`, the four `sound*CustomPath`); the four migration markers; and the `providerApiKey.` / `providerApiKeyUse.` / `relay.` namespaces.

Note the sound settings split deliberately: the on/off choice travels, the custom **file path** behind it does not, because it names this machine's disk.

## Testing

`bun test --isolate` — **1298 pass, 0 fail**. Typecheck clean.

### DIFF
--- a/src/main/transfer/__tests__/bundle-export.test.ts
+++ b/src/main/transfer/__tests__/bundle-export.test.ts
@@ -160,7 +160,7 @@ describe('what must never reach the archive', () => {
     const tables = JSON.parse(bundle.read(TABLES_ENTRY)!.toString('utf-8'));
     const keys = tables.preferences.map((p: { key: string }) => p.key);
 
-    expect(keys).toContain('theme');
+    expect(keys).toContain('fontSize');
     expect(keys).toContain('closeToTray');
     expect(keys).toHaveLength(2);
   });
--- a/src/main/transfer/__tests__/bundle-format.test.ts
+++ b/src/main/transfer/__tests__/bundle-format.test.ts
@@ -78,11 +78,19 @@ describe('the container', () => {
   });
 });
 
-describe('the preference exclusion policy', () => {
+describe('the preference portability policy', () => {
+  // These used to be 'theme', 'soundEnabled', 'betaChannel' and 'apiPort' —
+  // none of which this app has ever written. The assertion passed under the
+  // old denylist for the same reason any unknown string did, so it proved
+  // nothing about real settings. They are real keys now (#227).
   test('keeps the ordinary settings a user would want back', () => {
-    for (const key of ['theme', 'closeToTray', 'soundEnabled', 'betaChannel', 'apiPort']) {
+    for (const key of ['fontSize', 'closeToTray', 'soundVolume', 'updateChannel', 'webglRenderer']) {
       expect(isPortablePreferenceKey(key)).toBe(true);
     }
+  });
+
+  test('a key nobody has classified stays home instead of travelling', () => {
+    expect(isPortablePreferenceKey('somePreferenceAddedLater')).toBe(false);
   });
 
   test('refuses every safeStorage-sealed key', () => {
--- a/src/main/transfer/__tests__/bundle-import.test.ts
+++ b/src/main/transfer/__tests__/bundle-import.test.ts
@@ -103,7 +103,7 @@ describe('what comes back', () => {
     await restore(exportBytes(), mappedToDest());
 
     const keys = (destination.prepare('SELECT key FROM preferences').all() as { key: string }[]).map((r) => r.key);
-    expect(keys).toContain('theme');
+    expect(keys).toContain('fontSize');
     expect(keys).not.toContain('teamsTokens');
     expect(keys).not.toContain('windowBounds');
   });
@@ -218,7 +218,7 @@ describe('transactional', () => {
       arenaRuns: [],
       // No matching run, so the foreign key fires after the rows above landed.
       arenaResponses: [{ id: 'r1', runId: 'missing-run', provider: 'claude', status: 'done', responseText: '', ttftMs: null, totalMs: null, inputTokens: null, outputTokens: null, costUsd: null, error: null, round: 0, prompt: null, sessionRef: null }],
-      preferences: [{ key: 'theme', value: 'dark' }],
+      preferences: [{ key: 'fontSize', value: '14' }],
       accounts: [],
     };
     const manifest: TransferManifest = {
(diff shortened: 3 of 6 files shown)


brannon-bowden (approved):
The skill's workflow file isn't accessible in this sandboxed review environment, so I completed an equivalent-depth manual review directly against the diff.




Approve — this is a well-reasoned, well-tested switch from a denylist to an allowlist for preference portability, with a self-enforcing lint that keeps the classification honest going forward. CI is fully green (SonarQube, quality-gate, test, GitGuardian).

The core logic in `bundle-format.ts` is straightforward and correct: `isPortablePreferenceKey` now checks membership in `PORTABLE_PREFERENCE_KEYS` only, and `isClassifiedPreferenceKey` correctly unions the allowlist, the local-keys set, and the local-prefixes check. The retained `LOCAL_PREFERENCE_KEYS` set (even though redundant for filtering) is a good call — it's what lets the new lint distinguish "deliberately local" from "nobody has looked at this yet." The `quotaCooldownsCleared` fix (the bug that motivated this change) is correctly added to the local set and covered by a dedicated regression test. The `'theme'` → `'fontSize'` swap across the existing test suite is a legitimate fix (replacing an assertion against a key the app never actually wrote) rather than a behavior change.

## Suggestions (non-blocking)

- `bundle-preference-policy.test.ts`'s source scan (`declaredPreferenceKeys`) only matches single-quoted string literals (`'([^']+)'`) for both patterns. If any call site in the codebase uses double quotes or template literals for a static preference key, that key would silently evade the lint rather than failing it — the one blind spot in an otherwise "impossible to forget" mechanism. Given CI is currently green, this isn't affecting any key today, but worth a comment or a broader pattern (e.g. matching both quote styles) so the safety net doesn't quietly degrade as the codebase evolves. (`src/main/transfer/__tests__/bundle-preference-policy.test.ts:33-46`)
- Minor: the constant-matching regex (`[A-Za-z0-9_]*PREF(?:_KEY)?`) doesn't enforce the naming convention it documents ("Name must END in PREF / PREF_KEY") — it would also match a lowercase `somePref` binding. Harmless in practice since it only widens the match, not a correctness issue.

PR #236, opened 2026-08-26
233 — Give the real-corpus regression a ceiling it cannot miss

### LINKED ISSUE #233: Bug: ChatParser real-corpus test can exceed bun's 5000ms per-test ceiling under coverage + load
Surfaced while verifying #228. Not caused by that work — observed on `development` at `24b37d3`, unmodified.

## Symptom

```
(fail) ChatParser real-corpus regression (BDHLNDR-72) > 5-min real Claude session
       captures: extracts real content, suppresses noise [5016.05ms]
```

`5016.05ms` against bun's default 5000ms per-test timeout — the test did not assert wrong, it ran out of clock.

## Conditions

Observed once, in 2 runs, under the harshest condition I have:

- `bun test --coverage --coverage-reporter=lcov --isolate` from the repo root
- all 16 cores saturated with busy loops throughout

It did **not** appear in 10 consecutive load-only runs (no coverage), nor in any idle run. So the trigger looks like coverage instrumentation plus contention, not contention alone — which matters because coverage is exactly what the SonarQube workflow runs.

## Why it is the same class as #228, but not covered by it

#228's four fixes were all *tests waiting on something* with a sleep or a deadline sized for an idle machine. This one is different in shape: it's a test doing genuine work (parsing a real captured corpus) that is simply slower than the ceiling when the machine is busy and every line is instrumented. There is nothing to "wait on the condition" for — the fix is a realistic per-test timeout, or a smaller corpus, or splitting it.

I have not characterised which. One observation is enough to file, not enough to prescribe.

## Notes for whoever picks it up

- Confirm it reproduces before changing anything: `bun test --coverage --coverage-reporter=lcov --isolate` with the machine loaded. It is intermittent.
- If the honest answer is "this test legitimately takes ~5s under instrumentation", raise its ceiling explicitly rather than trimming the corpus — the real-corpus regression is the point of it.
- Worth checking sibling tests in the same file for headroom rather than fixing only the one that tripped.

## Related

- #228 — the four wall-clock dependencies fixed in the load-sensitive suites
- #206 — the other coverage-sensitive failure, on Linux

William-Long-II:
Closes #233

The last known member of the flake family from #228. Filed during that work with one observation and no diagnosis; this is the diagnosis.

## Measured, and my first hypothesis was wrong

The obvious suspect was the test's chunking loop — 340KB at 4096 bytes is ~83 iterations, each doing `await new Promise((r) => setTimeout(r, 1))`. A per-chunk event-loop yield under contention looked like an easy answer.

It isn't the cause:

| | 83 × `setTimeout(1)` |
|---|---|
| idle | ~95ms |
| all 16 cores saturated | ~140ms |

Nowhere near the budget. The time is the **parsing itself**: ~2s wall in isolation under `--coverage` on a loaded machine. That leaves bun's 5000ms default only about 2× headroom, and an 88-file parallel run consumes it — which is exactly why this only ever appeared in the full-repo coverage run and never in isolation.

## Raised the ceiling rather than making the test cheaper

The issue asked for this call to be made deliberately, so: the test parses a 340KB real PTY capture, and **the size of that corpus is the regression**. It's the test that caught a parser producing 1572 events with a longest payload of one character. Trimming it to fit a timeout would quietly weaken the only thing standing between that bug and a rerun of it.

`REAL_CORPUS_TIMEOUT_MS = 30_000` — roughly 15× the measured isolated cost. It's a backstop against a hang, not a performance budget; if it's ever reached, something is genuinely wrong.

This is different in kind from the four fixes in #232. Those were tests *waiting* on something with a sleep sized for an idle machine, and the fix was to wait on the condition. There is nothing to wait on here — it's real work that is simply slower than the default when the machine is busy and every line is instrumented.

## Siblings

Checked, as the issue asked. `real-corpus-bdhlndr-72.raw` is the **only test fixture in the repo over 50KB**, so nothing else has this exposure. The other 41 tests in the file share ~1.5s between them.

## Verification — the condition that actually reproduced it

3 consecutive full-repo runs with `bun test --coverage --coverage-reporter=lcov --isolate` under full core saturation — the exact combination that produced the original failure:

**1298 pass / 0 fail**, 153–166s per run.

### DIFF
--- a/src/main/api/chat-parser/__tests__/parser.test.ts
+++ b/src/main/api/chat-parser/__tests__/parser.test.ts
@@ -424,6 +424,23 @@ describe('ChatParser wrap-space recovery (BDHLNDR-73 #40)', () => {
   });
 });
 
+/**
+ * Ceiling for the real-corpus test (#233).
+ *
+ * bun's default is 5000ms, and this test overran it in the full-repo run under
+ * coverage instrumentation — `5016.05ms`, i.e. it did not assert wrong, it ran
+ * out of clock. Measured: ~2s wall in isolation under `--coverage` with every
+ * core saturated, so the default left only about 2x headroom, which an 88-file
+ * parallel run consumes.
+ *
+ * Raised rather than making the test cheaper on purpose. It parses a 340KB real
+ * PTY capture, and the size of that corpus is the point of the regression —
+ * trimming it to fit a timeout would quietly weaken the only test that has ever
+ * caught this class of parser bug. This number is a backstop against a hang,
+ * not a performance budget: if it is ever reached, something is genuinely wrong.
+ */
+const REAL_CORPUS_TIMEOUT_MS = 30_000;
+
 describe('ChatParser real-corpus regression (BDHLNDR-72)', () => {
   test('5-min real Claude session captures: extracts real content, suppresses noise', async () => {
     // Real PTY capture from one of @Will Long's Bodhilander sessions on
@@ -502,5 +519,5 @@ describe('ChatParser real-corpus regression (BDHLNDR-72)', () => {
     expect(allText).toContain('memory usage');
     expect(allText).toContain('concerns (windows');
     expect(allText).toContain('Slack, and Figma');
-  });
+  }, REAL_CORPUS_TIMEOUT_MS);
 });


brannon-bowden (approved):
This is a small, well-justified change and I'm approving it.

The PR raises the timeout on the real-corpus regression test from bun's 5000ms default to 30s, with a clear explanation for why: the test measured ~2s in isolation but tipped over 5000ms (5016ms) when run under coverage instrumentation across the full 88-file suite with cores saturated. Raising the ceiling rather than trimming the corpus is the right call here, since the corpus size is the point of the regression test (BDHLNDR-72). The code change itself is a straightforward one-line addition of a `timeout` argument to the `test()` call, which matches bun/Jest's `test(name, fn, timeout)` signature.

All 4 CI checks (SonarQube, quality-gate, test, GitGuardian) are passing.

**Suggestions (non-blocking):**
- The comment is thorough and includes good context, but consider trimming it slightly in the future if it invites drift (e.g., the exact `5016.05ms` figure will age once the CI environment changes). Not a concern for merging as-is.

PR #240, opened 2026-08-28
202 — Arrival polish: propose the mapping, sign in inline, and say what is left

### LINKED ISSUE #202: Feature: Arrival polish — auto-remap, re-auth, and an honest restore report
### User story / problem

As someone who has just restored onto a new machine, I want the remaining manual steps to be short and obvious, so that "restored" and "actually working" are the same moment rather than an afternoon apart.

After Phase 1 (#200) and Phase 2 (#201), three things still need a human: mapping working-directory roots, re-authenticating each Claude account, and re-entering provider API keys. This phase shrinks the first and makes the other two impossible to miss.

### Proposed solution

**Auto-detect repo roots.** Most remapping is one prefix swap. Scan the obvious candidates on the destination — the user's home, common project roots — for directories matching the leaf names the manifest recorded, and propose the mapping instead of asking for it. The user confirms or overrides; nothing is applied silently.

**Inline re-auth.** A restored `claude_accounts` row has no credentials on this machine. Surface each one as an explicit "sign in" affordance that runs the existing login flow in `account-auth.ts`, rather than letting the user discover the gap when a session fails to launch.

**An arrival report.** One screen after restore, stating plainly: sessions restored, of those how many are resumable, which are marked "needs relink" and why, which accounts need signing in, and which provider keys were intentionally not carried. This is the honest counterpart to the epic's promise — the user should not have to infer what survived.

### Acceptance criteria

- [ ] Import proposes a working-directory mapping per root where a confident match exists on the destination; matches are shown for confirmation, never applied silently
- [ ] A rejected or absent proposal falls back to the Phase 1 manual mapping without a dead end
- [ ] Each restored account without credentials on this machine offers an inline sign-in that runs the existing `account-auth.ts` flow
- [ ] Each provider that had a stored API key on the source is listed as needing re-entry, and the "use API key" toggle is not left on with no key behind it
- [ ] The arrival report states counts for restored, resumable, and needs-relink sessions, and names the accounts and provider keys awaiting action
- [ ] The report is reachable again later, not a one-shot dialog that vanishes on dismiss
- [ ] Resolving every item in the report leaves no session in a "needs relink" state and no account unauthenticated

### Scope

Minor feature

### Priority

Medium

William-Long-II:
Closes #202. Phase 3 of #199 — Phase 1 (#200) made a machine's state portable, Phase 2 (#201) removed the USB stick, and this is the part that makes "restored" and "actually working" the same afternoon.

## What it does

**Auto-detect the roots.** Most remapping is one prefix swap, and the manifest already records the roots. `root-suggest.ts` looks for each of them here and proposes an answer, so the prompt is a confirmation rather than a form. A root already at the same path is not asked about at all.

**Sign in from the report.** A restored account arrives with no credentials — those live in the source machine's keychain. Each one gets an inline sign-in that runs the existing `account-auth.ts` flow against the account that already exists.

**Say what is left.** One kept report: what came across, how much of it can actually start, which sessions need their folder, which accounts need signing in, and which provider keys were deliberately not carried.

## The parts

- **`transfer/root-suggest.ts`** — scores candidates by shared *trailing* path components, so `Work/Repos` beats a bare `Repos` somewhere else, reusing `working-dirs`' case-folding rather than a second copy. Breadth-first under a visit ceiling, skipping the directories a checkout is never the root of, and treating a directory it cannot list as empty rather than fatal.

  The property that matters is not how many it finds — it is that **it declines when it does not know**. Two places matching equally well produce no proposal, because picking one silently relocates every session under the root and the failure is quiet: the sessions start, in the wrong tree.

- **`transfer/arrival-report.ts`** + **`arrival.ts`** — `resumable` is derived from the relink list rather than asserted. An account whose login evidence could not be *read* is never reported as needing a sign-in: `undefined` is "unreadable", not "logged out", and Claude rewrites that file about once a minute. `recordArrival` never throws — the restore has already happened, and failing to describe it must not turn a completed import into a reported failure.

  Kept under `arrival.`, now a local preference prefix: a report describes a restore that happened *here*, and carried onward it would name folders to relink that machine never had.

- **`providersWithApiKeys` on the manifest** — ids only, derived from the `providerApiKey.` preference rows. The values are excluded by that same prefix and a test re-checks it against the produced bytes.

- **`resumeLoginFlow`** — shares the pty and watcher with `startLoginFlow` via `beginLoginPty`, and deliberately does **not** share the rollback. That one deletes the row and its directory on a spawn failure, which is right for an account it minted a moment ago and catastrophic for one a restore just brought back.

- **`ArrivalReport.tsx`** — a native `<dialog>`, the pattern `OwnerConfirmModal`/`ShareSessionModal`/`MachineHandoff` already use. Escape closes and keeps, the same as Close; only the explicit button forgets it. Raised at launch only while something is outstanding, and reachable from Settings → Import/Export either way.

## Two things worth reading closely

**A spec that silently ran nothing.** Wiring the export to `listVaultStatuses()` put the OS keychain in the import graph of everything importing `group-import-export`, and `menu.test.ts` and `group-import-export.test.ts` both dropped to **zero tests** — reported by bun as an unhandled error *between* tests rather than as a failure. Same shape as #234's `relay-client.test.ts`.

Patching three electron stubs would have hidden it. The real answer is that answering "which providers had a key" needs no decryption: `hasKey` is a lookup for a `providerApiKey.` row, so the exporter derives the names from rows it already reads and the vault leaves the graph entirely.

**A test at the mercy of the machine running it.** Left to itself the suggester walked the real home directory and proposed *this repo's own checkout* for the fixture's root. That is the feature working end to end, and a test that cannot be trusted. Both entry points now take the suggester as a parameter, defaulting to the real one.

## Verification

Everything below is **bun 1.4.0**, the version `test.yml` and `sonarqube.yml` pin — not the 1.3.x a local checkout is likely to have.

```
$ bun test --isolate            # whole repo, relay included

 1471 pass
 0 fail
 5196 expect() calls
Ran 1471 tests across 103 files. [126.06s]
```

No unhandled errors, so no spec is silently contributing nothing. Both typechecks clean, root and `relay/`.

63 new tests across 3 new spec files plus additions to `bundle-export`, `account-auth`, `preload` and `machine-handoff`. Two were checked against controls rather than assumed: wiring the rollback into `resumeLoginFlow` fails the test that says a spawn failure must not delete the account, and the report's Escape test asserts `defaultPrevented` rather than `open`, because happy-dom does not close a dialog on `cancel` by itself and asserting it stayed open would pass with no handler at all.

### Not covered

`src/main/index.ts` — three `safeHandle` registrations. The file boots the Electron app at import, so no spec can load it; the channel names are pinned a step away by `preload.test.ts`. Coverage of the contract, not of the wiring line.

The harness copy pinned in `initiatives/202/team.yaml` is not present in this environment, so `scripts/verify.sh` has not been run — the numbers above are direct `bun test` and `tsc`.


brannon-bowden (changes requested):
This is a well-built feature (arrival report, resumed logins for restored accounts, root-mapping suggestions) with strong test coverage across the new modules. The main blocker is CI, plus a couple of things worth fixing before merge.

## Issues

- **CI is red**: SonarQube Code Analysis and quality-gate are both failing (test and GitGuardian pass). Please resolve whatever the quality gate is flagging before merge.
- **`src/renderer/components/ArrivalReport.tsx:181-183`** — `ArrivalReportModal`'s `onSignIn` calls `window.electronAPI.resumeAccountLogin(accountId)` and discards the result. `resumeLoginFlow` returns `{ account, ptyId }`, but nothing here uses the `ptyId` to open a terminal/OAuth view — the modal doesn't navigate anywhere or render a pty. As written, clicking "Sign In" fires off the login pty in the background with no visible way for the user to complete the OAuth flow from this dialog. If there isn't a global listener elsewhere that surfaces an active login pty automatically (not visible in this diff), the button appears to do nothing after the brief "Opening…" state. Worth confirming this wires into the same terminal view `startAccountLogin`/`ClaudeAccountsPanel` uses.
- **`src/renderer/components/SettingsModal.tsx:604`** — the "Show Restore Report" handler sets `arrivalMissing` to `true` when no report exists, but never resets it back to `false` when a report is later found (`if (report) setArrivalReport(report);` doesn't clear `arrivalMissing`). If a user clicks the button once with nothing restored yet, then later restores something and clicks again, the stale "Nothing has been restored onto this machine yet." hint keeps showing next to the now-successfully-opened report.
- **`src/main/arrival.ts:134-140`** — `accountItems()` reports on *every* account on the machine (filtered later to `loggedIn === false`), not just accounts that were part of the current restore. If there's a pre-existing, not-yet-signed-in account unrelated to this transfer, it will show up in the arrival report's "needs sign-in" list every time any future restore happens. May be intentional (a standing "who's logged in" snapshot), but worth confirming against intent, since the module doc frames this as "what a restore left outstanding."

## Suggestions (non-blocking)

- `src/main/transfer/bundle-format.ts:178-196` — the new `PROVIDER_KEY_PREFIX` JSDoc block is inserted between the existing `LOCAL_PREFERENCE_PREFIXES` doc comment and the `const LOCAL_PREFERENCE_PREFIXES` declaration, orphaning the first block (it no longer documents anything). Consider merging the two comments or moving `PROVIDER_KEY_PREFIX`'s doc above its own declaration only.
- `src/main/machine-handoff.ts` — `lastOfferedBy` is module-level mutable state set by `readHandoffOffer` and read later by `restoreMachineHandoff`. The code comment explains the invariant (offer is always read before restore), but this is fragile if `readHandoffOffer` is ever polled/re-invoked while a restore's confirmation dialogs are still awaiting user input (e.g., a second offer arriving mid-flow would silently mislabel the in-progress restore's `sourceLabel`). Consider threading the label through explicitly instead of relying on shared state, if that's feasible without a larger refactor.

William-Long-II:
Thanks — three of the four were real, and the sign-in one was the important catch. All four are addressed in `7d6d6fe`.

**`ArrivalReport.tsx` — the sign-in went nowhere.** You were right, and I confirmed the chain before changing anything: `startAccountLogin`'s result drives `ClaudeAccountLoginModal`, which embeds the `<Terminal>` where the OAuth flow actually happens. Nothing else surfaces an active login pty. So discarding the `ptyId` left a pty running with no terminal attached and a button that flashed "Opening…" and did nothing.

The report now raises that same modal (exported for it) over the top of itself. Cancelling routes to `cancelAccountLogin(ptyId, false)` — an interrupted sign-in is not an aborted one, and unlike a newly minted account this one was just brought back by a restore, so deleting it would be far worse than leaving it signed out.

Two tests, both checked against a control that discards the id again: with the fix reverted, `raises the login terminal, so the flow has somewhere to happen` and `abandoning the sign-in stops the pty and never deletes the account` both fail.

**`arrival.ts` — the account scope.** Not intentional; you read the module doc the way I meant it. Fixed at the source rather than at the filter: `ImportOutcome` now carries `accountIds`, the ids the bundle actually named — inserted or already present here — and the report is built from those. An account this machine happens to hold and the bundle never mentioned is not something this restore left outstanding, and listing it would have put a standing "you have a logged-out account" notice inside every future arrival report. Control-verified: removing the filter fails `reports only the accounts the bundle carried, not every account here`.

**`SettingsModal.tsx` — the stale hint.** Real. Both pieces of state are set on every click now. Control-verified.

**CI.** Stale — the review is against `e429823`, which is the commit *before* the Sonar fixes. All four checks pass on the branch as it stands, and the quality gate reports 0 new issues and 0 hotspots.

Both suggestions taken as well:

- **`lastOfferedBy`** is gone. The source machine's name now comes from one signed `peek` inside `restoreMachineHandoff`, so a poll landing while a confirmation dialog is open can no longer relabel an in-flight restore. A failed peek just means the report does not name the source, which it already handles.
- **The orphaned JSDoc** — you were right that I inserted `PROVIDER_KEY_PREFIX` between the `LOCAL_PREFERENCE_PREFIXES` comment and its declaration. Reordered so each block documents the thing under it.

---

One thing worth recording, because it cost the most time and the first explanation was wrong. The two new settings tests appeared to hang the suite outright — no output, and bun's `--timeout` never fired, which pointed at import time. It was neither. The settings window is a large DOM, and two things over it are slow enough to read as a hang: `*ByRole`, which computes the accessibility tree, and testing-library pretty-printing that DOM into a failure message.

I had first pinned it on `<output>` — the element from the earlier Sonar fix — on the strength of one bisect that happened to change two things at once. That was wrong, and `<output>` stays. The assertions are now booleans and a direct selector, and the control fails in 294ms instead of never finishing.

**Verification:** 1475 pass / 1 fail under bun 1.4.0. The failure is `account-auth`'s `a token file appearing still completes the flow` timing out at 15s under full-suite load — the #228 flake, untouched by this branch and green in isolation. CI's own `test` job is green. Both typechecks clean.

brannon-bowden (approved):
Approve — this delta resolves every blocker from the last pass, with tests backing each fix, and CI is green across all four checks.

All three previously-raised issues are addressed:
- **Sign-in flow**: `ArrivalReportModal` now captures `{ account, ptyId }` from `resumeAccountLogin` and raises `ClaudeAccountLoginModal` over the report (`src/renderer/components/ArrivalReport.tsx:168-217`), reusing the same login UI `ClaudeAccountsModal` uses, with a cancel path that stops the pty but never deletes the account. Covered by new tests in `ArrivalReport.test.tsx`.
- **Stale "nothing restored" hint**: `SettingsModal.tsx:604-611` now sets both `arrivalReport` and `arrivalMissing` unconditionally on every click, so a later successful restore clears the stale hint. Covered by the new "reopening the restore report" test suite.
- **Arrival report scope**: `accountItems()` in `arrival.ts` is now scoped to `input.outcome.accountIds` (the ids the bundle actually carried), rather than every account on the machine, with a regression test asserting an unrelated local account never leaks into the report.

## Suggestions (non-blocking)

- `src/main/machine-handoff.ts:140-146` — `sourceMachineName` does a second signed `transport.peek()` at restore time, in addition to the one `readHandoffOffer` already did. The comment explains why (avoids the stale-module-state race from before), which is a real improvement, but it's worth a quick check that `peek()` on the relay is safe to call more than once per offer (not a one-time-use token) — if it's idempotent this is fine as-is.
- `src/main/transfer/root-suggest.ts:105-127` — in `childrenOf`, once `budget.left` hits 0 mid-walk, remaining parents in the current frontier still get `subdirsOf(options, parent.path)` invoked (a readdir) before the loop body's `budget.left <= 0` check short-circuits it. The original code broke the outer parent loop as soon as the budget was exhausted, avoiding those calls. Purely a minor efficiency regression at one depth level, not a correctness issue.

PR #242, opened 2026-08-28
241 — Relink a session from the report that named it

### LINKED ISSUE #241: Relink a session from the arrival report, rather than sending the user to find it
Follow-up from #202 (PR #240). Flagged during the work rather than found afterwards.

### The gap

#202's acceptance criteria included:

> - [ ] Resolving every item in the report leaves no session in a "needs relink" state and no account unauthenticated

The report delivers the *naming* half of that. It lists every session whose working directory is not on this machine, with the path it looked for:

```
2 sessions need their folder
  api   /Users/will/Work/api
  web   /Users/will/Work/web
```

And it resolves the account half inline — each account without credentials here offers a "Sign In" that raises the real login terminal.

Sessions do not get the same treatment. To fix one, the user reads the name off the report, finds that session in the sidebar, and sets its working directory through the existing per-session UI. The report tells them what is wrong and then steps out of the way.

That is a real gap against the criterion as written, and it is the asymmetry that makes it obvious: the harder of the two problems (re-authenticating an account) is one click, and the easier one is a manual hunt.

### Why it was left

Scope. The relink control is a folder picker plus an `updateDbSession` call, but doing it well means the report has to reflect its own progress — a row that is fixed should stop being listed, the "N sessions need their folder" heading has to follow, and `hasOutstandingWork` decides whether the report is raised at all on the next launch. That is a live-state surface rather than the static snapshot #202 built, and bolting it on at the end of that PR would have been the worse version of it.

Everything it needs already exists: `arrival.lastReport` is a kept preference that can be rewritten, and `remapWorkingDir` is not involved — this is one session, one directory.

### Suggested shape

A "Set Folder…" button per row, next to the path. Picking a directory updates that session and drops the row from the report; when the last one goes, the section goes with it. The kept report is rewritten rather than recomputed — recomputing would need the whole import outcome again, which is gone by then.

Worth deciding at the same time: whether a row should also offer "this session is gone, delete it", since a restore onto a machine that genuinely does not have a checkout any more is a real case and the alternative is a session that is permanently marked and never fixed.

### Acceptance criteria

- [ ] Each session listed as needing a folder can have one set from the report itself
- [ ] Setting a folder updates that session and removes its row, without reopening the report
- [ ] The count in the heading follows, and the section disappears when the last row is resolved
- [ ] The kept report reflects the resolutions, so reopening it later does not list work already done
- [ ] A report with nothing left outstanding stops being raised at launch

William-Long-II:
Closes #241. The follow-up I flagged while building #202 rather than found afterwards.

## The gap

#202's report delivered the naming half: it lists every session whose working directory is not on this machine, with the path it looked for. Then it stepped out of the way. The user read a name off the report, found that session in the sidebar, and set its folder there.

That left the asymmetry that made it obvious — the *harder* of the report's two jobs, re-authenticating an account, was one click; the easier one was a manual hunt.

## What it does

A **Set Folder…** button per row, opening the picker at the path the session is looking for.

`resolveRelink` does both halves in one act, because they are one: fixing the session without rewriting the report leaves it listing work already done, and rewriting the report without fixing the session is a lie. It points the session at the folder, marks it `stopped` — which is what the sidebar's parked marker derives from — strikes the row off the kept report, and returns what it stored, so the window redraws from the record rather than from a guess at it.

The heading follows (`2 sessions need their folder` → `1 session needs its folder`), and the section disappears with the last row.

## Three decisions worth disagreeing with

**`resumable` is recomputed, not incremented.** Two windows with the report open — App raises it at launch, Settings can reopen it — could otherwise both resolve the same session and walk the count past the number of sessions that exist. Checked against a control: incrementing fails the double-resolve test.

**The report is not closed when the last row goes.** The counts are still worth reading, and a dialog that vanishes as you finish with it reads as a crash. It stops being *raised* on the next launch instead, which the launch check already decides via `hasOutstandingWork`.

**A relink with no kept report still relinks.** The session is parked whether or not anything is keeping score — someone who dismissed the report should not be unable to fix the sessions it named.

## Keeping the sidebar honest

The IPC handler broadcasts `sessions:refresh` after the write. `workingDirMissing` is derived in main on every read, so without it the sidebar keeps its parked marker until something else happens to reload. Every renderer store already listens to that channel, which is why this needed no callback threaded through both of the report's hosts.

## Verification

Under **bun 1.4.0**, the version `test.yml` and `sonarqube.yml` pin:

```
$ bun test --isolate

 1483 pass
 1 fail
Ran 1484 tests across 103 files.
```

The failure is `account-auth`'s `a token file appearing still completes the flow` timing out at 15s under full-suite load — the #228 flake, untouched by this branch and green in isolation. No unhandled errors, so no spec is silently contributing nothing. Typecheck clean.

8 new tests: 4 on the view (the chosen folder reaches the right session, the row and count follow, the section goes with the last one, cancelling changes nothing) and 4 in main (both halves applied, idempotence, no-report, and the channel pinned in `preload.test.ts`).

### Worth knowing

`tsconfig.json` excludes `src/**/__tests__/**`, so a spec that fails to pass a newly-required prop is **not** caught by `tsc` — only by running it. Adding `onRelink` to `ArrivalReportViewProps` typechecked clean while the spec's own render helper was still missing it. Worth remembering the next time a component prop becomes required.

### DIFF
--- a/src/main/__tests__/machine-handoff.test.ts
+++ b/src/main/__tests__/machine-handoff.test.ts
@@ -54,6 +54,7 @@ const { declineMachineHandoff, prepareMachineHandoff, readHandoffOffer, restoreM
 const noSuggestions = () => [];
 
 const { dismissArrival, readArrival, resolveRelink } = await import('../arrival');
+const sessionsRepo = await import('../repositories/sessions');
 
 const SOURCE_ROOT = '/src-machine/Work/Repos';
 
@@ -356,13 +357,14 @@ describe('restoring one', () => {
     expect(report!.needsRelink).toEqual([]);
     expect(report!.resumable).toBe(1);
     expect(readArrival()!.needsRelink).toEqual([]);
-    const row = db.query('SELECT working_dir, state FROM sessions WHERE id = ?').get('s1') as {
+    // The directory is the whole fix: `workingDirMissing` is derived from the
+    // filesystem on every read, so a session pointed at a real folder stops
+    // being parked without anything touching its state.
+    const row = db.query('SELECT working_dir FROM sessions WHERE id = ?').get('s1') as {
       working_dir: string;
-      state: string;
     };
     expect(row.working_dir).toBe(here);
-    // `stopped`, which is what the sidebar's parked marker is derived from.
-    expect(row.state).toBe('stopped');
+    expect(sessionsRepo.getSession('s1')!.workingDirMissing).toBe(false);
   });
 
   test('resolving the same session twice cannot walk the resumable count past the truth', async () => {
--- a/src/main/__tests__/machine-handoff.test.ts
+++ b/src/main/__tests__/machine-handoff.test.ts
@@ -53,7 +53,7 @@ const { declineMachineHandoff, prepareMachineHandoff, readHandoffOffer, restoreM
  */
 const noSuggestions = () => [];
 
-const { readArrival } = await import('../arrival');
+const { dismissArrival, readArrival, resolveRelink } = await import('../arrival');
 
 const SOURCE_ROOT = '/src-machine/Work/Repos';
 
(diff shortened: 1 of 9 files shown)


brannon-bowden (approved):
Approve — this is a well-scoped feature (relink a session from the arrival report) with solid test coverage across the main-process logic, IPC bridge, and renderer UI, and CI is green (SonarQube, quality-gate, test, GitGuardian all passing).

## Issues

- `src/main/arrival.ts:222` — `resumable: Math.max(0, report.sessions - needsRelink.length)` recomputes `resumable` as simply "total sessions minus still-unlinked sessions." I can't see how `resumable` was originally derived in `recordArrival` (not in this diff), so I can't confirm this holds if a session can be blocked from resuming for another reason (e.g. its account still needs sign-in) — if so this recompute could overcount `resumable` once relinked but before the account is signed in. Worth the author double-checking against the original computation; not blocking since I can't verify it from the diff.
- `src/main/index.ts:1059` / `src/main/arrival.ts:200` — `resolveRelink` unconditionally sets `state: 'stopped'` on whatever `sessionId` it's given, with no check that the session was actually in `needsRelink`. The UI only ever calls this with rows from the report, but the IPC handler itself doesn't enforce that, so a call with an arbitrary/currently-running session id would force its state to `stopped`. Likely consistent with this app's existing IPC trust model, so just flagging for awareness rather than as a blocker.

## Suggestions (non-blocking)

- `src/renderer/components/ArrivalReport.tsx:228` (`onRelink`) — if `arrivalResolveRelink` or `selectDirectory` rejects, the error isn't caught; it'll surface as an unhandled rejection with no user-facing feedback (the button just re-enables silently). A caught error with a small inline message would make failures visible instead of silently vanishing. No test currently exercises this failure path either.

PR #244, opened 2026-09-04
CI: shorten dist-* artifact retention to 1 day

brannon-bowden:
The `dist-${{ matrix.platform }}` artifacts uploaded in the `build` job exist only to hand build output to the `release` job in the same run, which immediately re-uploads everything to a GitHub Release (`softprops/action-gh-release`). Nothing downstream reads them after that.

With no `retention-days` set they defaulted to 90 days, and Bodhilander is currently the single largest Actions-storage line item in the org (33 GB of live artifacts, ~1.5 GB-hr-equivalent/day gross). Actual distribution to users is unaffected — that goes through Releases, not artifacts.

One-line change: adds `retention-days: 1`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -169,6 +169,9 @@ jobs:
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: dist-${{ matrix.platform }}
+          # The release job in this same run re-uploads everything to a GitHub
+          # Release immediately after; nothing downstream needs this copy.
+          retention-days: 1
           path: |
             release/*.AppImage
             release/*.deb


William-Long-II (approved):
I'm about to rewrite this review into a more natural, engineer-voiced version per the instructions, keeping every finding intact.

Retention drop to 1 day means if the release/publish step fails after this upload, you've only got a day to grab the raw build artifacts for manual recovery. Comment makes clear that's intentional since the release job re-uploads immediately, so not a defect, just know that's your recovery window if a release ever needs to be manually re-run.

PR title says "dist-* artifact retention" (plural) but the diff only touches one `upload-artifact` step. Are there other dist-* uploads elsewhere in this workflow or others that should get the same treatment? Not seeing them here — worth a check, otherwise nothing to do.

Also the `test` check is failing but looks unrelated to this change (it's just a comment + config addition to a different job) — confirm that's pre-existing before merging.

Otherwise this is fine, indentation's correct, `1` is a valid retention-days value, safe to merge.

Issue #248, opened 2026-09-08
Relay deployments predate the handoff endpoints, so "Send to Another Machine" cannot work


William-Long-II:
Both live relays are healthy but running a build from before the handoff work.

| host | `/health` | `/api/machines/:id/handoff` | uptime | container started |
|---|---|---|---|---|
| `bodhilander.bodhilabs.dev` | 200, v0.1.0 | 404 | 1,699,988 s | 2026-08-19 |
| `bodhilander.sytanek.tech` | 200, v0.1.0 | 404 | 3,533,847 s | 2026-07-29 |

The 404 is the route being absent, not auth: the body is byte-identical to a garbage path (`/api/zzz/nope` → `{"error":"not_found"}`). A deployed handoff route would reach `handleHandoffRead`, which calls `machineFromSignature` first and rejects an unsigned request with 401.

`matchMachineHandoff` accepts any machine id, so the probe path does match the pattern on a current build.

The endpoints merged 2026-08-26 (`b9532cf`, `fe857fd`). Both containers predate that, so neither has been restarted since.

Nothing in the repo needs to change — `HANDOFF_DIR` falls back to `dirname(DB_PATH)/handoffs` for a file-backed database, which is what `docker-compose.yml` already mounts. It only needs a pull and a restart.

Until then the file path (Settings → General → Data → Export/Import) is the only working handoff, and #199's relay leg cannot be exercised.

William-Long-II:
`bodhilander.bodhilabs.dev` is deployed and serving the handoff routes.

```
/health                       {"ok":true,...}
/api/machines/probe/handoff   {"error":"invalid_request"}  [400]
/api/zzz/nope                 {"error":"not_found"}        [404]
```

The 400 is `machineFromSignature` rejecting an unsigned request — the route exists. Before this it returned a 404 byte-identical to the bogus path.

Deployed from `9227ceb` via rsync + `docker build`, recreated from the container's own `docker inspect` rather than the README, since the two disagree: the README publishes `127.0.0.1:47393` and the live container publishes `0.0.0.0:47393`. Following the docs would have quietly changed the host binding.

- relay suite green before deploy — 308 pass, 0 fail
- previous image kept as `bodhi-relay:rollback-20260908` (rollback is retag + re-run)
- database backed up to `~/relay-backup-20260908/` — `relay.db` plus a 4 MB `-wal` and the `-shm`; a `relay.db`-only copy would have been short
[...]

PR #249, opened 2026-09-08
Point the default relay URL at a relay that exists

### LINKED ISSUE #247: Default relay URL points at a hostname that does not exist
`DEFAULT_RELAY_URL` in `src/main/api/relay/relay-client.ts:103` is `https://cl-relay.sytanek.tech`. That name is NXDOMAIN — confirmed against 1.1.1.1, while the parent `sytanek.tech` resolves fine (parked on Hostinger, `ns1.dns-parking.com`). It is not a transient resolver failure; the record does not exist.

This is the value every fresh install uses, because `relay.url` is in `LOCAL_PREFERENCE_PREFIXES` and deliberately does not travel in a transfer bundle. So the machine that most needs a working relay — a clean install being handed off to — is exactly the one that falls back to a dead hostname.

The failure a user sees is a DNS error, not "no relay configured", which sends them looking for a network problem instead of a setting.

Either point the constant at a relay that exists, or drop the default and have an unset `relay.url` say so plainly.

William-Long-II:
Closes #247

`DEFAULT_RELAY_URL` becomes `https://bodhilander.bodhilabs.dev`. One line; it was the only reference in `src/`.

The old value, `https://cl-relay.sytanek.tech`, was NXDOMAIN — verified against 1.1.1.1, with the parent `sytanek.tech` resolving fine, so the record genuinely did not exist rather than a resolver having a bad day. The new one resolves to `20.127.80.85` and answers `/health` with 200.

This matters most on a clean install. `relay.url` sits under `LOCAL_PREFERENCE_PREFIXES`, so it deliberately does not travel in a transfer bundle — meaning the machine being handed *to*, which is the one that most needs a working relay, is exactly the one that fell back to the dead name and got a DNS error instead of a usable default.

**Not fixed here:** that relay is still running a pre-handoff build and 404s on `/api/machines/:id/handoff`. Pointing at it is correct regardless; #248 covers the redeploy.

No test pins the constant. One would assert a hostname rather than a behaviour, and would have to be edited by hand every time the relay moves — the check that would actually have caught this is DNS, not the suite.

Verified: `tsc --noEmit` clean; `bun test src/main/api/relay` 161 pass / 0 fail.

### DIFF
--- a/src/main/api/relay/relay-client.ts
+++ b/src/main/api/relay/relay-client.ts
@@ -100,7 +100,7 @@ export interface AssertedOwner {
 
 export type { RelayStatus } from '../../../shared/types';
 
-const DEFAULT_RELAY_URL = 'https://cl-relay.sytanek.tech';
+const DEFAULT_RELAY_URL = 'https://bodhilander.bodhilabs.dev';
 
 const PREF = {
   enabled: 'relay.enabled',


brannon-bowden (approved):
This is a minimal, low-risk change that swaps a hardcoded default relay URL constant, and CI is green across all checks — approving.

The change is a single-line constant update (`DEFAULT_RELAY_URL`) with no logic changes, so no new tests are warranted here. One non-blocking note: since this is only the *default* relay URL, any existing user preferences/config already persisted with the old `cl-relay.sytanek.tech` value presumably aren't affected by this change (the default only applies when no preference is set) — worth a quick sanity check by the author if there's any migration or fallback logic elsewhere that reads this constant, but nothing in the diff suggests an issue.

William-Long-II:
On the note about persisted values — checked, and no migration is needed.

`PREF.url` has exactly two touch points: the getter at `relay-client.ts:159` and `setRelayUrl` at `:239`. The default is never written to the preference; the getter falls back to it in memory when the stored value is empty. So anyone who never set a relay by hand has no row at all and picks up the new default on next launch, which is the case this fixes.

The only people holding `cl-relay.sytanek.tech` are those who typed it into Remote Hosting, and I'd argue we should leave them alone rather than rewrite it. `setRelayUrl` clears every grant and the confirmed owner id when the origin changes, because certificates carry their relay origin in the signed bytes. A silent migration would fire that teardown on a user who never asked for it.

`DEFAULT_RELAY_URL` has no other readers — `:103` and `:161` are the whole story.

PR #253, opened 2026-09-08
Show Settings off macOS instead of hiding it behind a chord

### LINKED ISSUE #252: Settings is invisible in the Windows and Linux menu, reachable only by knowing Ctrl+,
On macOS, Settings lives in the app menu as `Preferences...`. That menu does not exist off macOS, so the non-mac branch mirrors its entries into the Session menu — except Settings, which is created with `visible: false`.

`src/main/menu.ts:181-190`:

```ts
{
  // Hidden menu item to register Ctrl+, accelerator on Windows/Linux
  label: 'Settings',
  accelerator: 'CmdOrCtrl+,',
  visible: false,
  ...
}
```

The comment says what happened: the item was added to register an accelerator, not to be a menu entry. The effect is that on Windows and Linux there is **no visible path to Settings at all** — not in Session, not in Edit, not in View. A user who does not already know `Ctrl+,` cannot open it, and Settings is where Remote Hosting, Updates, and the Export/Import handoff controls live.

Reported against `beta.10` on Windows; the code is unchanged on `development`, so every build is affected.

Scope is just this item. Everything else in the macOS app menu is accounted for off macOS: `About Bodhilander` is in Help on every platform, `Quit` is mirrored into Session, `Claude Accounts...` is mirrored and visible, and `services`/`hide`/`unhide` are macOS-only roles with nothing to mirror.

Making it visible is the fix. Windows convention would also accept it under Edit, but Session already holds the mirrored `Claude Accounts...` and `Quit`, so keeping the pair together is more consistent than splitting them.

No test caught this — `menu.test.ts` asserts the items exist, which a `visible: false` item does.

William-Long-II:
Closes #252

Drops `visible: false` from the non-mac Settings item and names it `Settings...`, matching its sibling `Claude Accounts...`.

The macOS app menu does not exist on Windows or Linux, so the non-mac branch mirrors its entries into the Session menu. Settings was mirrored as an invisible item — added to register `CmdOrCtrl+,`, per its own comment, not to be reachable. The result was that Windows and Linux had **no visible path to Settings anywhere**: not Session, not Edit, not View. Reported from `beta.10` on Windows, and the code is unchanged on `development`, so it has been true of every build.

That matters more than a missing menu entry usually would, because Settings is where Remote Hosting, Updates and the Export/Import handoff controls live. A user who did not already know `Ctrl+,` could not reach any of them.

Scope is this one item. Everything else in the app menu is already accounted for off macOS — `About Bodhilander` sits in Help on every platform, `Quit` and `Claude Accounts...` are mirrored and visible, and `services`/`hide`/`unhide` are macOS-only roles with nothing to mirror.

**On the tests.** `isMac` is read per call, so forcing `process.platform` around `createApplicationMenu` is enough to build the other branch — the platform is restored in a `finally` so a failure cannot leak it into later tests.

The assertion is `visible !== false`, not "the item exists". Existence is precisely what the old tests checked and precisely what an invisible item satisfies, which is how this survived. A macOS case is included so the fix cannot drift into duplicating Settings into Session on the platform that already has it in the app menu.

Control-verified: with `visible: false` restored, the win32 and linux cases fail and the other six still pass.

Verified: `bun test --isolate` 1490 pass / 0 fail; `tsc --noEmit` clean.

### DIFF
--- a/src/main/menu.test.ts
+++ b/src/main/menu.test.ts
@@ -110,16 +110,12 @@ describe('reaching Settings without the macOS app menu', () => {
       // The bug was an item that existed only to register its accelerator, so
       // asserting it exists proves nothing on its own.
       expect((item as { visible?: boolean }).visible).not.toBe(false);
+
+      sent.length = 0;
+      item!.click!();
+      expect(sent).toContain('open-settings');
     });
   }
-
-  test('clicking it opens Settings', () => {
-    sent.length = 0;
-    const item = sessionMenuOn('win32').find((i) => i.label?.startsWith('Settings'));
-    item!.click!();
-
-    expect(sent).toContain('open-settings');
-  });
 
   test('macOS keeps it in the app menu and does not duplicate it into Session', () => {
     const labels = sessionMenuOn('darwin').map((i) => i.label);
--- a/src/main/menu.test.ts
+++ b/src/main/menu.test.ts
@@ -35,9 +35,30 @@ mock.module('electron', () => ({
 
 const { createApplicationMenu } = await import('./menu');
 
+const sent: string[] = [];
+
 function buildMenu(): void {
-  const fakeWindow = { webContents: { send: () => {} } } as never;
+  const fakeWindow = {
+    isDestroyed: () => false,
+    webContents: { send: (channel: string) => { sent.push(channel); } },
+  } as never;
   createApplicationMenu(fakeWindow);
+}
+
+/**
+ * The app menu is macOS-only, so everything it holds has to be mirrored
+ * elsewhere off macOS. Building under a forced platform is the only way to see
+ * that other branch from here.
+ */
+function sessionMenuOn(platform: string): MenuItem[] {
+  const real = process.platform;
+  Object.defineProperty(process, 'platform', { value: platform, configurable: true });
+  try {
+    buildMenu();
+    return template.find((m) => m.label === 'Session')?.submenu ?? [];
+  } finally {
+    Object.defineProperty(process, 'platform', { value: real, configurable: true });
+  }
 }
 
 function sessionMenu(): MenuItem[] {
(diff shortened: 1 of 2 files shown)


brannon-bowden (approved):
Approve — this is a small, well-contained fix that un-hides the Windows/Linux Settings menu item and backs it with solid tests.

The change to `src/main/menu.ts` simply drops `visible: false` and renames the label to `Settings...` to match the macOS `Preferences...` convention; the click handler and accelerator are untouched. The test additions exercise exactly the behavior that regressed (item exists *and* is visible, click fires `open-settings`, macOS still routes through the app menu without duplicating into the Session menu), which directly targets the described bug ("an item that existed only to register its accelerator") rather than just checking existence.

All listed CI checks (SonarQube, quality-gate, test, GitGuardian) are passing.

## Suggestions (non-blocking)
- `menu.test.ts`: the "clicking it opens Settings" test only exercises `win32`; since the click handler is platform-agnostic this is very likely fine, but a quick `linux` case would remove any doubt for future refactors.
- The `sessionMenuOn` helper mutates `process.platform` via `Object.defineProperty` and restores it in a `finally` block — good practice, but worth double-checking that no other test file in the suite runs concurrently and reads `process.platform` while this is temporarily overridden.

William-Long-II:
Both suggestions addressed.

**Linux click case** — folded the click assertion into the platform loop rather than adding a second test, so `win32` and `linux` now each check visibility *and* that clicking reaches `open-settings`. Seven tests, one fewer than before, one more assertion.

**`process.platform` concurrency** — checked rather than assumed, and it holds: `bun test --isolate` is the runner in both `package.json` and `.github/workflows/test.yml` (twice — root and `relay/`). `--isolate` gives each file its own process and module registry, so no other file can observe the override. Within a file bun runs tests serially, and the `finally` covers the failure path.

Control re-run after the change: restoring `visible: false` still fails exactly the win32 and linux cases, five pass.

PR #259, opened 2026-09-08
Say what deleting a Claude account destroys

### LINKED ISSUE #257: Removing a Claude account destroys the conversation transcripts of every session that used it
Removing a Claude account deletes its config directory with `fs.rmSync(root, { recursive: true, force: true })` (`src/main/account-auth.ts`, `deleteAccountAndDir`). That directory holds `projects/**/*.jsonl` — the conversation transcripts. There is no warning, no count of what is about to go, and no check for sessions that reference the account.

Sessions keep their `claude_session_id`, so starting one afterwards asks the CLI to resume a conversation whose transcript no longer exists. The error surfaces as **"session id not found"** at spawn time, which names neither the account nor the deletion, so the cause is not discoverable from the message.

**Why this is worse than a normal destructive action.** It sits directly next to the arrival flow. After a machine handoff the arrival report lists the carried accounts as needing sign-in. Removing them instead of signing in is a reasonable misreading of "these are stale" — and it destroys precisely what the handoff just carried across. The transcripts are the single most valuable thing the transfer moves; everything else can be rebuilt.

Observed on a real Mac -> Windows handoff. Source machine had 432 transcripts across its accounts. After removing the carried accounts on the destination, 10 remain, all under the account that was signed in fresh.

**Suggested fix.** Warn before deleting, naming the cost concretely — how many sessions reference this account and how many conversations are about to be destroyed — rather than a generic confirm. Better still, leave the `projects/` tree in place and remove only the credentials, so the account can be re-added and its history resume; nothing about signing out requires the history to go.

At minimum the resume failure should say the conversation's account was removed, instead of "session id not found".

William-Long-II:
Closes #257

The delete confirmation said `Sessions themselves are kept.` and listed what it unsets. Meanwhile `deleteAccountAndDir` runs `fs.rmSync(root, { recursive: true, force: true })` over the config directory, which holds `projects/**/*.jsonl` — every transcript. The one sentence that came close to the truth ("their account directory goes away with it") lived in a branch that fires only when sessions are *running*, so the quiet case said nothing at all.

Sessions keep their `claude_session_id`, so starting one afterwards asks the CLI to resume a conversation with no transcript and surfaces `session id not found` — a message naming neither the account nor the deletion.

Found on a real Mac -> Windows handoff. The source had 432 transcripts; after removing the carried accounts on the destination, 10 remained. The arrival report lists carried accounts as needing sign-in, and removing them instead is a reasonable misreading of "these are stale" — which destroys exactly what the handoff just carried.

**What changes.** A new `accounts:removalCost` IPC returns the sessions bound to the account and the transcripts under its config dir; the dialog names both:

> This permanently deletes 12 saved conversations. The 3 sessions bound to this account stay in the sidebar, but their history is gone and they cannot be resumed.

and when there is nothing to lose it says so, rather than warning about a cost that does not exist.

The count is computed in main because the renderer cannot stat the disk. `countSessionsUsingAccount` uses the same predicate as the `UPDATE` in `deleteAccount`, so the number cannot drift from what actually gets unset.

**Deliberately not changed.** The delete still removes the directory. Keeping `projects/` while dropping credentials would preserve the history, but leaves an orphaned tree with no owner and no UI to manage it — a design change that deserves its own issue rather than riding in on a confirmation-copy fix. This makes the destruction visible; it does not make it optional.

Also removed a comment that named a ticket and described the copy it no longer matches.

**Tests.** Control-verified: with the old dialog restored, the four assertions on the new copy fail and the rest pass. Both plural and singular are pinned, since the singular branch is the one nobody exercises by hand.

**Verification.** `tsc --noEmit` clean. Full suite on this Windows machine: **1484 pass / 6 fail**, and the same **6 fail on a clean tree at `2de2d8b`** — pre-existing, Windows-only, none in files this touches. CI runs ubuntu, so they are invisible there; filed separately.

### DIFF
--- a/src/main/account-auth.ts
+++ b/src/main/account-auth.ts
@@ -21,7 +21,7 @@ import * as accountsRepo from './repositories/accounts';
 import { AccountIdentity, LOGIN_ARTIFACTS, resolveAccountIdentity } from './account-identity';
 import { registerHooks } from './mcp-config';
 import { seedLegacyConversations } from './legacy-claude-seed';
-import { ClaudeAccount } from '../shared/types';
+import { AccountRemovalCost, ClaudeAccount } from '../shared/types';
 
 interface LoginFlow {
   accountId: string;
@@ -297,6 +297,41 @@ export function cancelLoginFlow(
   }
 
   activeFlows.delete(ptyId);
+}
+
+/**
+ * What removing an account costs, for a confirmation that can name it. The
+ * conversations are the part worth counting: they live under the config dir
+ * and go with it, and nothing else on disk records them.
+ */
+export function accountRemovalCost(accountId: string): AccountRemovalCost {
+  return {
+    sessions: accountsRepo.countSessionsUsingAccount(accountId),
+    conversations: countTranscripts(configDirFor(accountId)),
+  };
+}
+
+/** Transcripts live at <configDir>/projects/<cwd-slug>/<uuid>.jsonl. */
+function countTranscripts(configDir: string): number {
+  const projects = path.join(configDir, 'projects');
+  let slugs: fs.Dirent[];
+  try {
+    slugs = fs.readdirSync(projects, { withFileTypes: true });
+  } catch {
+    // No projects/ is the ordinary "never used" case, not a failure.
+    return 0;
+  }
+
+  let total = 0;
+  for (const slug of slugs) {
+    if (!slug.isDirectory()) continue;
+    try {
+      total += fs.readdirSync(path.join(projects, slug.name)).filter((f) => f.endsWith('.jsonl')).length;
+    } catch {
+      // A slug we cannot read contributes nothing and stops nothing.
+    }
+  }
+  return total;
 }
 
 /**
--- a/src/main/repositories/accounts.ts
+++ b/src/main/repositories/accounts.ts
@@ -125,6 +125,14 @@ export function touchAccount(id: string): void {
   );
 }
 
+/** Counts what deleteAccount() is about to unset, using the same predicate. */
+export function countSessionsUsingAccount(id: string): number {
+  const row = getDatabase()
+    .prepare('SELECT COUNT(*) AS n FROM sessions WHERE claude_account_id = ?')
+    .get(id) as { n: number };
+  return row.n;
+}
+
 /**
  * Delete an account and NULL out any sessions/groups that referred to it.
  * SQLite ALTER TABLE can't add a real FK, so we enforce SET NULL semantics here.
@@ -137,14 +145,6 @@ export function touchAccount(id: string): void {
  * neither a live account nor an assigned one — so the user is told their
  * account was removed and offered nothing to move to.
  */
-/** Counts what deleteAccount() is about to unset, using the same predicate. */
-export function countSessionsUsingAccount(id: string): number {
-  const row = getDatabase()
-    .prepare('SELECT COUNT(*) AS n FROM sessions WHERE claude_account_id = ?')
-    .get(id) as { n: number };
-  return row.n;
-}
-
 export function deleteAccount(id: string): void {
   const db = getDatabase();
   const tx = db.transaction(() => {
--- a/src/main/repositories/accounts.ts
+++ b/src/main/repositories/accounts.ts
@@ -137,6 +137,14 @@ export function touchAccount(id: string): void {
  * neither a live account nor an assigned one — so the user is told their
  * account was removed and offered nothing to move to.
  */
+/** Counts what deleteAccount() is about to unset, using the same predicate. */
+export function countSessionsUsingAccount(id: string): number {
+  const row = getDatabase()
+    .prepare('SELECT COUNT(*) AS n FROM sessions WHERE claude_account_id = ?')
+    .get(id) as { n: number };
+  return row.n;
+}
+
 export function deleteAccount(id: string): void {
   const db = getDatabase();
   const tx = db.transaction(() => {
--- a/src/renderer/components/ClaudeAccountsModal.tsx
+++ b/src/renderer/components/ClaudeAccountsModal.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { ClaudeAccount, LiveAccountBindings } from '../../shared/types';
+import { AccountRemovalCost, ClaudeAccount, LiveAccountBindings } from '../../shared/types';
 import Terminal from './Terminal';
 import { AccountChip } from './AccountChip';
 import './ClaudeAccountsModal.css';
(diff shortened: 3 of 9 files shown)


src/renderer/components/ClaudeAccountsModal.tsx
```diff
@@ -96,21 +96,31 @@ export const ClaudeAccountsPanel: React.FC = () => {
     await refresh();
   }, [refresh]);
 
-  // The panel computes "N sessions are running on this account" two inches
-  // above this button, and deleting removes the on-disk config dir out from
-  // under those live ptys (#165). Withholding the number here while the dialog
-  // reassures the user that "sessions themselves are kept" is the one place
-  // that count actually had a job to do.
+  // Deleting removes the on-disk config dir, which is where the transcripts
+  // live, so the confirmation has to name what that costs rather than only
```
brannon-bowden: No try/catch around this await — if accountRemovalCost() rejects, the delete flow silently fails before the confirm dialog even appears, with no user feedback.

src/main/repositories/accounts.ts
```diff
@@ -137,6 +137,14 @@ export function touchAccount(id: string): void {
  * neither a live account nor an assigned one — so the user is told their
  * account was removed and offered nothing to move to.
  */
+/** Counts what deleteAccount() is about to unset, using the same predicate. */
+export function countSessionsUsingAccount(id: string): number {
```
brannon-bowden: This function is inserted between the pre-existing deleteAccount docblock and deleteAccount itself, orphaning the docblock onto this function instead. Move the new function above that docblock or add a separate one for deleteAccount.

src/main/account-auth.ts
```diff
@@ -299,6 +299,41 @@ export function cancelLoginFlow(
...
+ */
+export function accountRemovalCost(accountId: string): AccountRemovalCost {
+  return {
+    sessions: accountsRepo.countSessionsUsingAccount(accountId),
+    conversations: countTranscripts(configDirFor(accountId)),
+  };
+}
+
+/** Transcripts live at <configDir>/projects/<cwd-slug>/<uuid>.jsonl. */
+function countTranscripts(configDir: string): number {
```
brannon-bowden, line 315: No direct unit test for countTranscripts' error-handling branches (missing projects/ dir, unreadable slug dir) — only covered indirectly via mocked renderer tests.

brannon-bowden (changes requested):
This PR improves the account-deletion confirmation by naming what's actually destroyed (saved conversations), backed by new main-process helpers and updated tests — a good change in intent, but it can't merge with CI red and has an error-handling gap worth fixing first.

## Issues

- **Failing CI checks**: SonarQube Code Analysis and quality-gate are both failing on this PR. These need to be resolved (or the findings addressed) before merge.
- **Unhandled rejection can silently kill the delete flow** (`src/renderer/components/ClaudeAccountsModal.tsx:100`): `handleDelete` now does `const cost = await window.electronAPI.accountRemovalCost(id);` with no try/catch around it. Previously, deleting an account depended only on `deleteAccount` succeeding; now it also depends on this new IPC round-trip succeeding *before the user even sees the confirm dialog*. If `countSessionsUsingAccount`'s SQL call throws (e.g. a transient DB error), the promise rejects unhandled, the confirm dialog never appears, and there's no user-facing feedback — the "Delete" button just silently does nothing. Worth wrapping in a try/catch with a fallback (e.g., proceed with a cost-less confirmation, or surface an error) so a failure to *compute* the cost doesn't block being able to delete at all.
- **Documentation misattached to the wrong function** (`src/main/repositories/accounts.ts:141`): the pre-existing docblock ending "...offered nothing to move to." was written to document `deleteAccount`. The new `countSessionsUsingAccount` (with its own one-line doc comment) is inserted between that docblock and `deleteAccount`, so the original docblock now visually documents `countSessionsUsingAccount` instead, and `deleteAccount` is left with no docblock directly above it. Move `countSessionsUsingAccount` above the pre-existing docblock (or relocate the docblock) so each function keeps its own documentation.
- **No direct unit tests for the new main-process logic** (`src/main/account-auth.ts:315`, `src/main/repositories/accounts.ts:141`): `countTranscripts` has non-trivial branching (missing `projects/` dir, unreadable per-slug directory, filtering by `.jsonl` extension) and `countSessionsUsingAccount` is a raw SQL query — neither has a dedicated test. Coverage only comes indirectly through the renderer test's mocked `accountRemovalCost`, which doesn't exercise the actual filesystem/DB code paths (e.g., the two `catch` branches in `countTranscripts` are untested).

## Suggestions

- (non-blocking) The count shown in the confirmation dialog can go stale between the `accountRemovalCost` fetch and the user confirming/`deleteAccount` running (e.g., a new conversation is written in that window). Fine for a confirm dialog, but worth being aware of if this pattern is reused somewhere with higher stakes.

William-Long-II:
All three blocking findings addressed, plus the Sonar failures.

**Sonar (7 findings, all nested ternaries / nested template literals).** The copy is now built by two pure functions, `removalWarning(cost)` and `runningWarning(n)`, with early returns instead of a ternary pile. That also gave the wording somewhere to be tested directly, which is where the next point lands.

**Unhandled rejection could kill the delete.** Correct, and worse than it looked: the rejection happened *before* the confirm, so the button did nothing at all. Now wrapped, with the failure falling through to a confirmation that claims no number rather than a wrong one — `removalWarning(null)` returns empty. A cost we could not measure must not be reported as a cost of zero, which is the one wrong answer that reads as safe.

The test for this was **vacuous on the first attempt**: my control edit silently no-op'd on a line-ending mismatch, so the "control" ran against the unmodified file and passed. Re-run properly, the test fails with the guard removed.

**Docblock misattached.** Right — `countSessionsUsingAccount` went in between `deleteAccount`'s docblock and `deleteAccount`. Moved above the docblock so each function keeps its own.

**No direct tests for the main-process logic.** Added to `account-auth.test.ts`, which already stands up a temp `userData` dir and the real repository against in-memory sqlite — so these exercise the actual filesystem walk and the actual SQL, not a mock:

- counts transcripts across several slugs, and only the sessions on this account (`other` and `null` bindings excluded)
- ignores files that are not `.jsonl`
- an account that never ran, and one with no config dir at all, both report zero rather than throwing
- a plain file among the slug directories is skipped

**One correction to the review's framing, and to my own first attempt.** The two `catch` branches are not both reachable from a test. The outer one is covered by the missing-`projects/` cases. The inner one guards a directory that stops being readable *after* it was listed — a race — and the plain-file case I first wrote to cover it actually exercises the `isDirectory()` guard instead, never reaching the catch. I renamed that test to say what it really does and left a comment stating the inner catch is uncovered, rather than leaving a test whose name claims coverage it does not have.

Also dropped three panel tests that now duplicate the pure-function tests, keeping the one that proves the panel wires the cost through, plus the new error-path test.

**Verification.** `tsc --noEmit` clean. Full suite on Windows: **1496 pass / 6 fail** — the same 6 that fail on a clean `2de2d8b`, tracked in #260. Controls re-run: dropping the `.jsonl` filter fails the extension test; removing the try/catch fails the error-path test.

The stale-count note is fair and left as is — the window is between the fetch and the confirm, and a re-count after confirming would only move it.

brannon-bowden (approved):
This update resolves everything blocking from the previous pass — CI is green, the cost lookup no longer risks killing the delete flow, the docblock is back on the right function, and the new logic now has real unit coverage — so this is good to merge.

Previously raised issues, checked against this diff:

- **CI red**: now passing across the board (SonarQube, test, quality-gate, GitGuardian). Resolved.
- **Unhandled rejection could silently kill delete** (`ClaudeAccountsModal.tsx`): `accountRemovalCost` is now wrapped in try/catch, falling back to `cost = null`, and `removalWarning(null)` returns `''` so the confirm dialog still appears with a generic (non-numeric) warning instead of hanging. Backed by a new test (`still offers the delete when the cost cannot be measured`) that asserts the delete still proceeds when the IPC call rejects. Resolved.
- **Docblock misattached to the wrong function** (`accounts.ts`): `countSessionsUsingAccount` (with its own one-line doc comment) now sits above the `deleteAccount` docblock instead of between it and the function, so `deleteAccount`'s docblock is directly above `deleteAccount` again. Resolved.
- **No direct tests for the new main-process logic**: the new `accountRemovalCost` describe block in `account-auth.test.ts` exercises real filesystem/DB paths — multi-slug transcript counting, non-`.jsonl` filtering, missing `projects/` dir, missing config dir entirely, and a stray non-directory file among slugs — using real fixtures rather than mocks. Largely resolved; one branch (a directory that becomes unreadable *after* being listed) is explicitly left uncovered in a comment since it requires a race to hit, which is a reasonable, clearly-flagged trade-off rather than an oversight.

## Suggestions (non-blocking)

- The inner "directory becomes unreadable mid-listing" catch branch in `countTranscripts` is still untested, as the added test comment acknowledges. Fine to leave as-is given how hard it is to hit deterministically.
- As noted last round, the cost shown can go stale between the `accountRemovalCost` fetch and the user confirming (e.g., a new conversation written in that window). Not worth solving for a confirm dialog, just worth remembering if this pattern gets reused somewhere higher-stakes.

Issue #260, opened 2026-09-08
Six tests fail on Windows and CI cannot see them


William-Long-II:
Running `bun test --isolate` on Windows 11 against a clean `development` (`2de2d8b`) gives **1481 pass / 2 skip / 6 fail**. The same tree is green in CI, which runs `ubuntu-latest`, so nothing surfaces these.

```
(fail) handoff storage limits > put bundles beside the database, which is what the volume holds
(fail) ensureTranscriptInConfigDir > reports present and copies nothing when the transcript is already in the target
(fail) ensureTranscriptInConfigDir > skips the target when it also appears among the candidates
(fail) ArenaEngine > vault env is merged into the contestant spawn and wins over ambient env
(fail) POST /sessions/:id/start > a session whose folder is present still starts
(fail) POST /sessions/:id/start > a spawn failure reports what actually went wrong
```

The grouping suggests path handling rather than six unrelated defects — bundle paths, transcript locations and session working directories are all places a POSIX assumption (`/` separators, a `/tmp`-shaped root, case sensitivity) would survive review and fail only here. Some of these were also seen alongside `ENOTDIR: not a directory, mkdir '...\legacy-seed-f0rays\account\.claude\projects'` in the log noise.

**Why it matters beyond tidiness.** Windows is a shipped platform, and this run is the only signal that would catch a Windows-only regression before a user does. Right now a contributor on Windows cannot tell their own breakage from the standing six, which is the failure mode that makes people stop running the suite locally.

Two things worth deciding together: fix the six, and add a Windows job to `test.yml` so they cannot come back silently. Fixing without the job just resets the clock.

Found while verifying #257 on the destination machine of a Mac -> Windows handoff; that PR's own changes are unrelated and its file set is disjoint from these.

PR #261, opened 2026-09-08
Report group roots that landed without their folder

### LINKED ISSUE #258: Groups with missing working directories are invisible to the arrival report
After a Mac -> Windows handoff, 14 restored groups still carry macOS working directories on the Windows machine:

```
MISS Work                  -> /Users/will/Work/Repos
MISS Bodhilander           -> /Users/will/Work/Repos/Bodhilander
MISS Care Service          -> /Users/will/Work/Repos/bodhi-service-care
MISS Front End             -> /Users/will/Work/Repos/bodhi-web-apps
...
```

Groups are remapped on import — `bundle-import.ts:192` runs `remapWorkingDir(group.workingDir ?? '', options.mappings)` — so the mechanism exists and the mappings simply came back empty for these roots. That part may be operator choice at the prompt.

**The gap is that nothing surfaces the result.** `needsRelink` is computed only for sessions (`bundle-import.ts:229-230`), and `relinkItems` in `arrival.ts` takes `sessionIds`. A group whose root does not exist produces no report row and no **Set Folder…** button, so there is no way to discover or fix it from the arrival flow. The sessions in this restore all resolved, so the report was clean while 14 groups were pointing at a filesystem that does not exist on this machine.

**Impact.** A group's working directory is the default cwd for new sessions in it. Creating a session in one of these throws `Working directory does not exist` at `pty-manager.ts:297`. That is the exact failure #199 called out as load-bearing — "a session whose directory is missing must land in a visible needs-relink state rather than erroring at launch time" — reached through groups rather than sessions.

**Suggested fix.** Extend the arrival report to cover group roots as well as session directories: same missing-directory test, same row, same **Set Folder…**. A group root is arguably the more useful thing to relink, since fixing one root fixes every session later created under it.

Follow-up from the #199 handoff, found on the destination machine.

William-Long-II:
Closes #258

Groups were already remapped on import (`bundle-import.ts`), but nothing recorded which ones landed on a directory this machine does not have. `needsRelink` was sessions only, so the report could read clean while every group pointed at a filesystem that is not here — which is what happened on the Mac -> Windows handoff that found this: 3 sessions all resolved, 14 groups on `/Users/will/...`, report silent.

A group root fails nothing until someone creates a session under it, and then it throws `Working directory does not exist` at `pty-manager.ts:297`. That is the failure #199 named as load-bearing — "a session whose directory is missing must land in a visible needs-relink state rather than erroring at launch time" — reached through groups instead of sessions.

**What changes.** `restoreGroups` collects group ids whose remapped directory is absent, carried on `ImportOutcome.groupsNeedingRelink`. `ArrivalRelinkItem` gains `kind: 'session' | 'group'`, and the report lists both. Rows are marked `(group)`, and the heading names the kind when the list is all one thing rather than being uniformly vague. `resolveRelink` takes a kind and writes the group row instead of the session row.

A group with **no** directory at all is not reported — that is a plain folder in the sidebar, not a broken root.

**The bit that needed care: `resumable`.** It was `sessions - needsRelink.length`, and the list now carries non-sessions, so counting the length would report conversations as unrecoverable that came back fine. It now counts session items only, in `buildArrivalReport` and in `resolveRelink`, which have to agree or the number changes when a row is struck off.

Items written before `kind` existed carry none and were all sessions, so an absent kind counts as a session rather than being dropped. There is a test for that specifically.

**A vacuous test I wrote and removed.** My first `resumable` test went through the full restore, where the fixture has one session and one group. Correct code gives `max(0, 1-1) = 0`; broken gives `max(0, 1-2) = 0`. The clamp hid the bug and the control passed. Replaced with a direct `buildArrivalReport` test using 4 sessions and 2 group items, where the two answers are 3 and 1 — and confirmed it fails against the old derivation.

**Controls, each verified to fail.** Never flagging a group root: 2 tests fail. Counting groups in `resumable`: the new derivation test fails. `resolveRelink` writing a session for a group: the group-relink test fails. Hardcoding the click's kind to `'session'`: the renderer test fails.

**Verification.** `tsc --noEmit` clean. Full suite on Windows: **1502 pass / 6 fail** — the same 6 that fail on a clean tree, tracked in #260.

### DIFF
--- a/src/main/__tests__/machine-handoff.test.ts
+++ b/src/main/__tests__/machine-handoff.test.ts
@@ -320,8 +320,13 @@ describe('restoring one', () => {
     // not one that can start — and the report says so rather than reporting
     // a restore of one session and leaving the rest to be discovered.
     expect(report!.resumable).toBe(0);
-    expect(report!.needsRelink.map((r) => r.sessionId)).toEqual(['s1']);
+    // The group root is missing too, and is listed beside the session: nothing
+    // fails until someone creates a session under it, which is precisely why
+    // waiting for a failure is no way to find out.
+    expect(report!.needsRelink.map((r) => `${r.kind}:${r.sessionId}`))
+      .toEqual(['session:s1', 'group:g1']);
   });
+
 
   test('reports only the accounts the bundle carried, not every account here', async () => {
     const { bytes, phrase } = preparedElsewhere();
@@ -347,16 +352,16 @@ describe('restoring one', () => {
     messageBoxResponses = [1];
     await restoreMachineHandoff(relay.transport, phrase, legacyDir, noSuggestions);
 
-    expect(readArrival()!.needsRelink.map((r) => r.sessionId)).toEqual(['s1']);
     const here = path.join(tmp, 'dst', 'api');
     fs.mkdirSync(here, { recursive: true });
 
     const report = resolveRelink('s1', here);
 
-    // Both halves, or the report is describing work that is already done.
-    expect(report!.needsRelink).toEqual([]);
+    // Both halves, or the report is describing work that is already done. The
+    // group row survives: resolving a session says nothing about a group root.
+    expect(report!.needsRelink.map((r) => r.kind)).toEqual(['group']);
     expect(report!.resumable).toBe(1);
-    expect(readArrival()!.needsRelink).toEqual([]);
+    expect(readArrival()!.needsRelink.map((r) => r.kind)).toEqual(['group']);
     // The directory is the whole fix: `workingDirMissing` is derived from the
     // filesystem on every read, so a session pointed at a real folder stops
     // being parked without anything touching its state.
@@ -365,6 +370,31 @@ describe('restoring one', () => {
     };
     expect(row.working_dir).toBe(here);
     expect(sessionsRepo.getSession('s1')!.workingDirMissing).toBe(false);
+  });
+
+  /**
+   * The same act on a group: it writes the group row, not a session row of the
+   * same id, and strikes only the group off. Fixing a root is worth more than
+   * fixing one session — every session made under it later inherits it.
+   */
+  test('relinking a group root writes the group and leaves the session listed', async () => {
+    const { bytes, phrase } = preparedElsewhere();
+    const relay = standIn(bytes);
+    messageBoxResponses = [1];
+    await restoreMachineHandoff(relay.transport, phrase, legacyDir, noSuggestions);
+
+    const here = path.join(tmp, 'dst', 'work');
+    fs.mkdirSync(here, { recursive: true });
+
+    const report = resolveRelink('g1', here, undefined, 'group');
+
+    expect(report!.needsRelink.map((r) => r.kind)).toEqual(['session']);
+    // The session was never touched, so nothing about it became resumable.
+    expect(report!.resumable).toBe(0);
+    const group = db.query('SELECT working_dir FROM groups WHERE id = ?').get('g1') as {
+      working_dir: string;
+    };
+    expect(group.working_dir).toBe(here);
   });
 
   test('resolving the same session twice cannot walk the resumable count past the truth', async () => {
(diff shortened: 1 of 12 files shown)


brannon-bowden (changes requested):
This PR extends the arrival-report machinery to also track group roots whose working directory didn't survive a restore, keeping them separate from sessions for the `resumable` count. The core logic (in `arrival.ts`, `arrival-report.ts`, `bundle-import.ts`) is internally consistent — `kind` defaults to `'session'` everywhere it's read, the resumable count is recomputed rather than decremented, and the id-collision case between a group and session sharing an id is handled by filtering on `(sessionId, kind)` pairs together. Test coverage across the main process, IPC surface, and renderer is thorough and exercises the new behavior end to end.

## Issues

- CI is currently red: **SonarQube Code Analysis** and **quality-gate** are both failing. These need to be resolved (or the underlying findings addressed) before this can merge; `test` and `GitGuardian Security Checks` are passing.
- `src/shared/types.ts:591` — `ArrivalRelinkItem.kind` is typed as required (`kind: 'session' | 'group'`), but the runtime code treats it as optional everywhere it's consumed (`item.kind ?? 'session'` in `arrival.ts` and `ArrivalReport.tsx`, `item.kind !== 'group'` in `arrival-report.ts`) to stay backward-compatible with arrival reports persisted before this field existed. The new test in `arrival-report.test.ts` has to bypass the type system with `as BuildArrivalReportInput['needsRelink']` to construct a kind-less item, which is a sign the type doesn't match reality. Consider making it `kind?: 'session' | 'group'` so the type documents the actual contract instead of requiring a cast at every legacy-data call site.

## Suggestions (non-blocking)

- The field name `sessionId` is now overloaded to also hold a group id when `kind === 'group'` (e.g. `arrival.ts`'s `groupRelinkItems` pushes `{ kind: 'group', sessionId: groupId, ... }`). It works correctly, but a more neutral name like `id` would read less confusingly to future maintainers scanning for where group ids and session ids are handled.

William-Long-II:
Both blocking findings fixed.

**Sonar** — the nested ternary in `relinkLabel` is now a `let` with two guards. That was the only annotation.

**`kind` should be optional.** Agreed, and the cast in my own test was the tell — I wrote `as BuildArrivalReportInput['needsRelink']` to build a kind-less item and did not stop to ask why the type was fighting me. It is `kind?: 'session' | 'group'` now, with the reason in the doc comment: reports persisted before this field existed carry no kind, every one of those was a session, and readers default accordingly. The cast is gone from the test, which is the real confirmation the type now matches the contract.

**On renaming `sessionId` to `id`** — not doing it, and the reason is that this shape is persisted. `ArrivalRelinkItem` is serialised into the `arrival.lastReport` preference and read back on a later launch, so the field name is on-disk data, not just an internal name. Renaming it either breaks reading every report written by an earlier build or needs a migration that reads both — which is a lot of machinery to buy a better name for a field whose doc comment already says "the session or group id, per `kind`". If the shape gains a version, this is worth doing in the same pass.

**Verification.** `tsc --noEmit` clean. Full suite on Windows: **1502 pass / 6 fail** — the same 6 that fail on a clean tree, tracked in #260.

brannon-bowden (approved):
This resolves both previously-raised concerns and introduces no new issues, so it's ready to merge.

Both items from the prior review are addressed by this diff:

- **CI**: previously SonarQube Code Analysis and quality-gate were failing; the current CI status shows all four checks (SonarQube Code Analysis, quality-gate, test, GitGuardian Security Checks) passing.
- **Type mismatch**: `ArrivalRelinkItem.kind` (`src/shared/types.ts:591`) is now `kind?: 'session' | 'group'`, matching the optional-in-practice handling already present in `arrival.ts` and `ArrivalReport.tsx`. The test in `arrival-report.test.ts` no longer needs the `as BuildArrivalReportInput['needsRelink']` cast to construct a kind-less item, confirming the type now documents the real contract.

The `relinkLabel` rewrite in `ArrivalReport.tsx` from a nested ternary to an if/else is logically equivalent (`groups === 0` → `'session'`, `groups === items.length` → `'group'`, else `'item'`) — no behavior change.

The overloaded-`sessionId`-for-group-id naming note from the prior review remains a non-blocking style observation and doesn't need to hold up this PR.

PR #264, opened 2026-09-08
Make the suite pass on Windows, and run it there

### LINKED ISSUE #260: Six tests fail on Windows and CI cannot see them

William-Long-II:
Closes #260

The suite now passes on Windows — **1508 pass / 0 fail** for the app and **308 / 0** for the relay — and `test.yml` runs on `windows-latest` alongside `ubuntu-latest` so it stays that way.

**The headline is that no product code changed.** All six failures were POSIX assumptions in *test* code. Windows users were never broken by any of them; the gate was. That is a better outcome than the issue guessed, and worth saying plainly rather than dressing six test fixes up as six bug fixes.

| test | assumption |
|---|---|
| `ensureTranscriptInConfigDir` (x2) | the `tree()` helper compared `path.relative` output against `/`-joined literals; Windows hands back `\` |
| `ArenaEngine > vault env is merged` | the fake contestant ran `echo "key=$FAKE_PROVIDER_KEY"` — `$VAR` is POSIX syntax that cmd.exe leaves as literal text, so the value read back empty |
| `POST /sessions/:id/start` (x2) | fixtures used `/tmp` as "a folder that is present". It is not present on Windows, so `/start`'s parked-session guard correctly returned 409 where the test wanted 200 and 500 |
| `handoff storage limits` | asserted `handoffDir === '/data/handoffs'`, but `path.resolve('/data/relay.db')` is `C:\data\relay.db` on Windows |

**On the relay one specifically.** The relay only ever runs in a Linux container, so its production behaviour was never in question. Rather than rebuild the implementation's own `path.join(dirname, 'handoffs')` in the expectation — which would assert the code against itself — the contract is now stated as two independent facts: the handoff dir's parent is the database's directory, and its name is `handoffs`. Either half breaking still fails the test.

The arena fix reads the variable through `node -e` instead of shell interpolation, which keeps that test about **env merging** rather than about which shell ran it — the thing it was actually written to prove.

**The workflow.** A matrix over `[ubuntu-latest, windows-latest]` with `fail-fast: false`, so a Windows failure never hides a Linux one or vice versa. `timeout-minutes` goes 15 → 25 because Windows runners are markedly slower at install.

This renames the check from `test` to `test (ubuntu-latest)` / `test (windows-latest)`. Verified safe first: the `development` ruleset carries only a `pull_request` rule, and classic branch protection reports no required status check contexts, so nothing requires the old name. If required checks are added later, both names need listing.

Fixing these without the job would only have reset the clock — which is the whole point of the issue.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,11 +18,8 @@ concurrency:
 
 jobs:
   test:
-    # Windows is a shipped platform, and until this ran there nothing stood
-    # between a Windows-only regression and a user: six tests failed there for
-    # months while this gate was green. Path handling is where it goes wrong —
-    # POSIX literals, `/tmp`, `$VAR` in a shell command — and none of it is
-    # visible from Linux.
+    # Windows is a shipped platform, and its failures — path separators, 8.3
+    # short paths, POSIX-only shell syntax — are invisible from Linux.
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,8 +18,18 @@ concurrency:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # Windows is a shipped platform, and until this ran there nothing stood
+    # between a Windows-only regression and a user: six tests failed there for
+    # months while this gate was green. Path handling is where it goes wrong —
+    # POSIX literals, `/tmp`, `$VAR` in a shell command — and none of it is
+    # visible from Linux.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    # Windows runners are markedly slower at install and native-free spawns.
+    timeout-minutes: 25
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -19,10 +19,9 @@ describe('handoff storage limits', () => {
   });
 
   test('put bundles beside the database, which is what the volume holds', () => {
-    // "Beside the database" as two independent facts. The old assertion was a
-    // POSIX literal, and `path.resolve('/data/relay.db')` is `C:\data\relay.db`
-    // on Windows — so it failed there for a reason the relay, which only ever
-    // runs in a Linux container, does not actually have.
+    // Stated as two independent facts rather than one path literal: the
+    // config resolves the db path, so a literal only holds on the platform
+    // that shaped it.
     const dbPath = path.join(os.tmpdir(), 'relay-config-test', 'relay.db');
     const { config } = loadConfig({ ...BASE, DB_PATH: dbPath });
     expect(path.dirname(config.handoffDir)).toBe(path.dirname(dbPath));
--- a/relay/src/config.test.ts
+++ b/relay/src/config.test.ts
@@ -4,6 +4,8 @@
  * worth being able to see.
  */
 import { describe, expect, test } from 'bun:test';
+import * as os from 'node:os';
+import * as path from 'node:path';
 import { ConfigError, loadConfig } from './config';
 
 const BASE = { NODE_ENV: 'test', PUBLIC_URL: 'http://relay.test' };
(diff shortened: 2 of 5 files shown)


brannon-bowden (changes requested):
Request changes — the PR's entire purpose is to make the suite pass on Windows and gate CI on it, but the CI status shows `test (windows-latest)` is currently failing.

## Blocking issues

- **CI: `test (windows-latest)` is failing.** This is the check this PR exists to add, and it's red. Given the diff's own reasoning ("six tests failed there for months while this gate was green"), a red Windows job here means either a fix in this diff is incomplete/incorrect, or there's a remaining Windows-only failure not addressed by these hunks. Since branch protection will presumably require this new matrix leg, merging with it red defeats the PR's stated goal and would leave main red or force someone to bypass the gate. Please paste/link the failure output and fix (or, if it's a known-flaky/unrelated infra issue, note that explicitly) before merge.

## Non-blocking issues

- `.github/workflows/test.yml:29`: `timeout-minutes: 25` — reasonable bump for a slower Windows runner, no concerns.
- `src/main/arena/__tests__/engine.test.ts`: swapping `echo "key=$FAKE_PROVIDER_KEY"` for a `node -e ...` invocation is a good portability fix, but it does subtly change what's under test — env vars are now read directly by Node rather than exercising shell variable interpolation through the spawned shell. Worth a one-line comment or confirming that's intentional (it looks like it is, given the accompanying comment), but flagging since the test's semantics shifted slightly, not just its syntax.
- The inline commentary throughout (config.test.ts, sessions.test.ts, engine.test.ts) is quite verbose for test-file comments. Not a blocker, just a style note — future diffs touching these files will carry a lot of prose to read through.

## Tests

No new test coverage gaps introduced — this PR is entirely about making existing tests portable/correct on Windows, not adding new logic. The `insertSession` default change (`/tmp` → `os.tmpdir()`) in `sessions.test.ts` is a sound fix since the old default assumed a Linux-only path existed.

William-Long-II:
Right on the blocking finding, and the red job is the best thing that happened to this PR.

**The failure, and why only CI could find it.** One test: `ArenaEngine > a scoped run spawns contestants inside the working directory`.

```
Expected to contain: "C:\Users\RUNNER~1\AppData\Local\Temp\arena-cwd-RzIlO2"
Received:            "C:\Users\runneradmin\AppData\Local\Temp\arena-cwd-RzIlO2"
```

`os.tmpdir()` returned an **8.3 short path** because the runner account is `runneradmin` — over eight characters — while the shell printed the long form. The test already called `fs.realpathSync`, which resolves symlinks but does not expand short names; `fs.realpathSync.native` goes through the OS call and does. It resolves symlinks on POSIX too, so it is strictly the better call on every platform.

My own Windows username is four characters, so no shortening happens locally and this passed here every time. It was only ever reachable on a machine with a long username — which is to say, on CI. The job justified itself on its first run, before it had even merged.

**On the arena semantics shift** — intentional, and you are right that it is a shift and not just syntax. The test is titled "vault env is merged into the contestant spawn and wins over ambient env"; what it exists to prove is that the vault's env reaches the child and outranks the ambient value. Shell interpolation was the vehicle, never the subject. Comment reworded to say that rather than to narrate the change.

**On the verbose comments** — fair, and it collides with a house rule I should have applied: comments explain the code, not the change. Several of mine were narrating what the old assertion used to be. Trimmed all four to the constraint that still matters, roughly half the prose.

**Verification.** Full suite on Windows locally: **1508 pass / 0 fail**. The real check is this PR's own `test (windows-latest)` leg, now re-running.

brannon-bowden (approved):
Approve — CI now shows `test (windows-latest)` passing along with the rest of the matrix, which resolves the blocking issue from the previous review, and the changes since then are sound.

## Blocking issues

None. The prior blocker was `test (windows-latest)` failing; CI status now lists it as passing (along with `test (ubuntu-latest)`, `SonarQube Code Analysis`, `quality-gate`, `GitGuardian Security Checks`), so the gate this PR exists to add is green.

## Non-blocking issues

- `src/main/arena/__tests__/engine.test.ts`: switching from `fs.realpathSync(dir)` to `fs.realpathSync.native(dir)` is a well-reasoned fix — the accompanying comment correctly explains that `os.tmpdir()` can hand back an 8.3 short path on Windows runners and only the native OS call expands it to match what the shell prints. Since the Windows job is passing, this appears to work as intended.
- The comment rewrites across `test.yml`, `config.test.ts`, `sessions.test.ts`, and `engine.test.ts` are noticeably more concise than the previous versions, which addresses the verbosity note from the last review.

## Tests

No new test coverage gaps — this delta is comment tightening plus a targeted portability fix to an existing assertion, not new logic.

PR #266, opened 2026-09-11
CO-722: the run state machine and its durable state

William-Long-II:
CO-722 phase 3, first slice. Part of Software-Development-LLC/bodhi-code#722.

**No processes, no spawning, no UI.** This is the foundation the engine sits on, and none of it is reachable from the app yet — nothing here changes behaviour for anyone.

## Why a state machine at all

`claude-team-workflow` documents a gate model and nothing enforces it. Its own docstrings record what that costs: owners dispatched serially where the command says concurrent (*"a paragraph read before the run did not prevent it"*), two owner dispatches that ended with "open a PR" so gates 3 and 4 never ran, 97 turns spent locating a tool across three reachable copies of it.

None of those is a reasoning failure. They are sequencing, ordering and isolation — which is what a state machine is for, and what a prompt cannot reliably be.

## What is here

**`transitions.ts`** holds the arrows and nothing else. It contains **no domain judgment**: it does not know what a reviewer looks for, what a seam is, or how to verify anything. That stays in the model and in the plugin's Python. Pure `(state, event) -> decision`, so every path — including every inconclusive one — is asserted in a unit test rather than discovered during a run that spends real tokens.

The invariants it encodes, each of which is a defect somewhere if dropped:

| invariant | what it prevents |
|---|---|
| `inconclusive` is not a shade of failure | read as pass → the green-having-checked-nothing defect filed seven times against the plugin; read as failure → an owner sent back to fix code that was never broken |
| provisioning precedes gate 2 | an unprovisioned worktree makes `verify.sh` report a missing runner as FAIL — measured as `'jest' is not recognized`, exit 1, over code that was fine |
| review is requested only once checks are green, and requesting is an **action** | `arbiter` does not fire until an approver is asked, so an un-requested PR is invisible to the bot *and* the human queue |
| a bot nit is recorded, a bot major returns to gate 2 | CO-447 measured `arbiter/review` at "major 0, minor 0, nit 1"; re-entering gate 2 for that burns a cycle on a note |
| approval releases the run; merge does not | nothing comes back after the last approval, and waiting for someone to press merge is unbounded — one initiative sat twelve hours there |
| a permission prompt parks with **no timeout** | a silent auto-deny is indistinguishable from a gate finding, so the run would carry a verdict nobody gave |

**No `(state, event)` pairing throws.** A run is durable and will be handed stale and duplicate events — a reconcile racing a webhook, a gate reporting twice after a restart. Throwing there turns a duplicate into an outage; advancing on one would be worse.

**`runs.ts`** persists it. Every transition writes its state and its reason in **one transaction**, because a state without its event is a run whose history has a hole exactly where someone will later ask what happened. `run_events` is append-only: it is the run view, the audit trail and resume at once.

`team.yaml` stays the source of truth on disk. `run_owners` mirrors it so a query need not parse YAML — and a re-run cannot erase a PR number the previous pass recorded, which matters because `spawn.sh` is idempotent, knows nothing about PRs, and *will* be re-run on resume.

## Tests — 59

`initializeRunTables` is **exported so tests execute the real schema**. A test that builds its own `CREATE TABLE` proves the repository's SQL matches itself, not the schema that ships.

- **33 transitions** — every state × every event, the four inconclusive paths, the bot-nit-vs-major pair (each a control for the other), and that no pairing throws.
- **17 repository** — the state-and-reason transaction, append-only ordering, a malformed payload not taking the run view down, and the PR-number preservation above.
- **9 schema** — the cascades and the composite key SQLite cannot add after the fact, plus a group deletion leaving the run (a run is not its sidebar entry).

Full suite: **1567 pass, 2 skip, 0 fail** across 106 files. `tsc -p tsconfig.main.json` clean.

## Not here, deliberately

- **Gate spawning.** Next slice: argv construction is pure and testable on its own, and it depends on flags verified in phase 2 (`--plugin-dir`, `--tools`, `--json-schema`, `--permission-prompts host`).
- **The reconciler and `expected_checks`.** Both need the exit-code contract that Software-Development-LLC/claude-team-workflow#285 is settling — writing the wrong number is worse than leaving it.
- **Any UI.** A fourth `ContentView` comes once there is something to show.
- **Multi-repo, seams, the barrier.** Phase 5.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/__tests__/run-schema.test.ts
+++ b/src/main/__tests__/run-schema.test.ts
@@ -1,0 +1,146 @@
+/**
+ * The run engine's real schema (CO-722).
+ *
+ * `runs.test.ts` builds its own CREATE TABLE so it can run without electron,
+ * which means it proves the repository's SQL agrees with ITSELF. This file
+ * executes `initializeRunTables` from database.ts, so a column added in one
+ * place and not the other fails here rather than on a user's machine.
+ *
+ * It also pins the two structural choices the engine depends on and SQLite
+ * cannot be made to add later: the cascades, and the composite key that makes
+ * the owners mirror re-runnable.
+ *
+ * Run with: bun test src/main/__tests__/run-schema.test.ts
+ */
+import { describe, expect, mock, test } from 'bun:test';
+import { Database } from 'bun:sqlite';
+import type BetterSqlite3 from 'better-sqlite3';
+
+type Db = BetterSqlite3.Database;
+const asDb = (d: Database) => d as unknown as Db;
+
+// Superset of the electron surface database.ts touches at import time —
+// the same shape database-cleanup.test.ts uses, for the same reason.
+mock.module('electron', () => ({
+  app: { getPath: () => '/nonexistent-bodhilander-test-userdata' },
+  safeStorage: { isEncryptionAvailable: () => false },
+}));
+mock.module('electron-log', () => ({
+  default: { info() {}, warn() {}, error() {} },
+}));
+
+const { initializeRunTables } = await import('../database');
+
+function freshDb(): Database {
+  const d = new Database(':memory:');
+  d.exec('PRAGMA foreign_keys = ON');
+  // runs.group_id references groups(id); the real database always has it.
+  d.exec('CREATE TABLE groups (id TEXT PRIMARY KEY, name TEXT NOT NULL)');
+  initializeRunTables(asDb(d));
+  return d;
+}
+
+function columns(d: Database, table: string): string[] {
+  return (d.prepare(`PRAGMA table_info(${table})`).all() as { name: string }[])
+    .map((c) => c.name)
+    .sort();
+}
+
+describe('the schema applies', () => {
+  test('creates every table the engine reads', () => {
+    const d = freshDb();
+    const names = (
+      d.prepare("SELECT name FROM sqlite_master WHERE type = 'table'").all() as {
+        name: string;
+      }[]
+    ).map((r) => r.name);
+    for (const t of ['runs', 'run_owners', 'run_gates', 'run_events']) {
+      expect(names).toContain(t);
+    }
+  });
+
+  test('is re-runnable, because it executes on every launch', () => {
+    const d = freshDb();
+    expect(() => initializeRunTables(asDb(d))).not.toThrow();
+  });
+
+  test('carries the fields that cannot be re-derived later', () => {
+    // harness_path and python_path are the two a resumed run cannot work out
+    // for itself: three plugin copies were reachable in one 18-hour window,
+    // and `python3` on PATH can be a Store alias that is not Python.
+    const cols = columns(freshDb(), 'runs');
+    for (const c of ['harness_path', 'python_path', 'bodhi_root', 'initiative_key',
+      'permission_posture', 'blocked_reason']) {
+      expect(cols).toContain(c);
+    }
+  });
+
+  test('records the permission posture per GATE, not only per run', () => {
+    // If you cannot tell afterwards whether an owner ran unsandboxed, you
+    // cannot trust what it produced.
+    expect(columns(freshDb(), 'run_gates')).toContain('posture');
+  });
+});
+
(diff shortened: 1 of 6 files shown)


brannon-bowden (approved):
Solid piece of work — the state machine is genuinely pure and exhaustively tested (every `(state, event)` pair, terminal absorption, the taxonomy partition), and the repository layer correctly makes state + event durable in one transaction. CI is green across the board. Approving, with a few points worth the author's attention.

## Issues
None blocking.

## Suggestions

- **`src/main/run-engine/transitions.ts`** — the `running` state doesn't track which gate is currently in flight; `onGateFinished` reacts purely to the `gate` number in the incoming event. If a stale/duplicate `gateFinished` for gate 2 arrives after the run has already advanced to gate 3 or 4 (e.g. a late report from a retried/duplicated process, which the module's own docstring calls out as an expected failure mode), the machine will regress the run back to a spawn of gate 3, overwriting further-along progress. The tests only cover replaying the *same* event twice, not an out-of-order report from an earlier gate arriving after a later one has already started. Worth confirming whether the caller filters events against the currently active gate (e.g. via `run_gates.status`) before invoking `transition` — if not, that's a real regression path.
- **`src/main/repositories/runs.ts:174-179`** (`recordTransition`) — `blockedReason` is optional with no type-level link to `state`. Nothing stops a caller from transitioning to `'inconclusive'` without supplying `blockedReason`, silently producing the "inbox says blocked but not why" case the code comments explicitly warn about. A stronger type (e.g. requiring the reason when `state === 'inconclusive'`) would make that omission a compile error instead of a runtime gap.
- **`src/main/run-engine/transitions.ts` (`onReviewNotRequested`)** — this handler only reacts to `reviewRequested`; a `checksFailed` (or `checksGreen`→already green, new commit invalidates checks) arriving while sitting in `reviewNotRequested` is silently ignored rather than looping back to gate 2. If new commits can land checks-red after entering this state but before review is actually requested, the run would stay parked in `reviewNotRequested` waiting on a review request that may never come cleanly. Worth confirming this is an intentionally unreachable transition given the surrounding orchestration.
- Nit: `src/main/database.ts` has a stray extra blank line after the new `initializeRunTables` closing brace.

William-Long-II:
All four taken, in 665391d. The first was a real regression path and the suite did not cover it — thank you.

**The late gate report.** You are right, and the gap in the tests is the part that stings: they covered replaying the *same* event twice, which passes either way, and never an earlier gate reporting after a later one started. With the module's own contract saying it **will** be handed out-of-order events, that was a hole in exactly the place it claims to be careful.

`transition()` now takes a required `RunContext` carrying the gate in flight; a report from any other gate is ignored with a note. **Required rather than optional deliberately** — you suggested confirming whether the caller filters first, and leaving it there would be a rule kept by convention, which is the thing this machine exists to replace.

Four cases, and the last two are the ones that keep it honest:

- a stale gate 2 **pass** while gate 4 is in flight
- a stale gate 2 **fail and inconclusive** — the guard has to hold for the verdicts that *stop* a run too, or a stale report parks a healthy gate 4
- a report with **nothing** in flight, which belongs to nothing this run started
- a **control** asserting the gate actually in flight is still acted on. Without it, a guard that ignored everything would pass the other three and stall every run.

**`blockedReason`.** Now required for the states that block, via overloads, so the omission is a compile error rather than an inbox entry saying a run needs a person and not what for. Verified both directions rather than assumed:

```
recordTransition(r, inconclusive, k)
  -> error TS2345: Argument of type inconclusive is not assignable ...
recordTransition(r, inconclusive, k, { blockedReason: ... })   // compiles
recordTransition(r, running, k)                                  // compiles
```

**`reviewNotRequested` and `checksFailed`.** Not intentionally unreachable — you were right to ask. It is *meant* to be transient, entered on green and left as soon as an approver is asked, but a commit landing in that window would park the run waiting for a review request that should no longer be made. It now returns to gate 2.

**Blank line** removed.

38 transition tests; full suite 1572 pass. Your approval survived the push — re-request if you would rather look again, since the context parameter changes the public signature.

PR #267, opened 2026-09-12
CO-722: build a gate command line from the pinned harness

William-Long-II:
CO-722 phase 3. Part of Software-Development-LLC/bodhi-code#722. Still no spawning and no UI — this is the pure function that turns a gate into an exact `claude` command line.

## Three measurements contradicted the design

Every flag here was verified by running it, not read from documentation. Three came back the opposite of what the design assumed:

**`--allowedTools` does not restrict anything.** With `--allowedTools "Read,Grep,Glob" --permission-mode manual`, Bash ran to completion with `permission_denials: []`. It is a pre-approval list. This module never emits it, and a test asserts that.

**`--agent` and `--json-schema` do not compose.** A run with `--agent bodhi:reviewer` and a schema returns `subtype: success`, `is_error: false`, exit 0 and **no `structured_output`** — even when the agent completes the task in two turns.

That one is worth dwelling on: **a verdict can simply be absent while every success signal is set.** It is why "no verdict is inconclusive, never a pass" is load-bearing rather than defensive, and it is the same shape as the seven `exits 0 having checked nothing` defects filed against the plugin.

**`--append-system-prompt` cannot carry a real agent.** `reviewer.md`'s body is **37,429 characters**; inline it exceeds Windows' 32,767-character command line and fails with `ENAMETOOLONG` in `uv_spawn` before the process even starts. No unit test would have found that.

## So the modes differ, each using what works

| | reading gates (3, 4) | owner gate (2) |
|---|---|---|
| invocation | `--print` | `--bg` |
| agent | `--append-system-prompt-file` + `--tools` | `--agent bodhi:…` |
| verdict | `--json-schema` → `structured_output` | receipt file |
| grant | the agent file's `tools:`, forwarded | resolved by the plugin |

The owner gate can afford `--agent` because its verdict is a receipt file anyway, and keeping it preserves the plugin's own composition and hooks. An unresolvable name there **fails loudly and lists the alternatives** rather than falling back to a full grant — verified.

**Neither path restates a grant.** `agent-definition.ts` reads it from the agent file and **refuses** when it cannot: no front matter, no `tools:`, an empty list, or a body-less file all throw. Defaulting would undo the plugin's own CI check — *"a missing `tools:` line silently grants an agent everything, including Skill and Agent"* — for every gate this engine spawns.

## Verified end to end

Against the real reviewer, through the real CLI:

```
structured_output present: true
  verdict  : inconclusive
  tools    : Bash, Glob, Grep, Read, StructuredOutput
  role_seen: # Purpose
```

`role_seen` is the control that makes it real: it quotes the first heading of `reviewer.md`'s body, proving the 37,429 characters actually reached the model. **An earlier run of this probe passed on an empty file** — a `sed` had extracted nothing — which is precisely the vacuous pass this initiative keeps finding, so the control is not decoration.

No `Write`, no `Edit`.

## Also

`--setting-sources ""` so two machines cannot reach different verdicts on one branch and both report honestly. `--max-budget-usd` so the plugin's documented ceiling ("a ceiling, not a target") is enforced rather than described. `PYTHONUTF8`/`PYTHONIOENCODING` in the child env, without which every em-dash the plugin prints lands in receipts as a replacement character. `BODHI_PYTHON`, so the wrappers get the interpreter this run resolved rather than trusting the name `python3`.

## Tests — 80 in the module

The ones that matter assert what must **not** be there: no `--allowedTools`, no `Write`/`Edit` in a reading gate's grant, no body inline in argv, and no posture other than `bypass` skipping permission checks. Each has a control, because an assertion that something is absent passes just as well when everything is absent.

Full suite: **1614 pass, 0 fail** across 108 files.

## Not here

Spawning, the reconciler, `expected_checks`, and any UI. The middle two need the exit-code contract Software-Development-LLC/claude-team-workflow#285 is settling.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/run-engine/gate-command.ts
+++ b/src/main/run-engine/gate-command.ts
@@ -1,0 +1,206 @@
+/**
+ * Turning a gate into the exact `claude` command line (CO-722).
+ *
+ * Pure: it builds an argv, an env and a cwd, and runs nothing. That is what
+ * makes "does a reviewer get Write?" a snapshot assertion instead of something
+ * you find out by reading a diff an agent should never have been able to
+ * produce.
+ *
+ * Every flag here was verified against Claude Code 2.1.263 rather than taken
+ * from documentation, and three of those measurements contradicted what the
+ * design assumed:
+ *
+ * - `--allowedTools` does NOT restrict anything. With
+ *   `--allowedTools "Read,Grep,Glob" --permission-mode manual`, Bash ran to
+ *   completion with `permission_denials: []`. It is a pre-approval list. This
+ *   module never emits it, and a test asserts that.
+ * - `--bg` and `--print` conflict outright, so a gate cannot be both
+ *   attachable and verdict-shaped. Owners get `--bg`; the reading gates get
+ *   `-p`.
+ * - a variadic option swallows a trailing positional, so a prompt placed
+ *   after one is consumed as a value. Print mode therefore passes the prompt
+ *   on **stdin**, which also keeps it out of argv entirely.
+ *
+ * The two modes also differ in how the agent reaches the CLI, and that is a
+ * measurement rather than a preference. `--agent` and `--json-schema` do NOT
+ * compose: a run with `--agent bodhi:reviewer` and a schema returns
+ * `subtype: success`, `is_error: false`, exit 0 and **no**
+ * `structured_output` — even when the agent completes the task. So:
+ *
+ * - the reading gates pass the agent's own body via
+ *   `--append-system-prompt-FILE` and its own `tools:` via `--tools`, which
+ *   returns the validated object and the same restricted set (`Bash, Glob,
+ *   Grep, Read` — no Write, no Edit). The file matters: reviewer.md's body is
+ *   37,429 characters and passing it inline exceeds Windows' 32,767-character
+ *   command line, failing with `ENAMETOOLONG` before the process starts;
+ * - the owner gate keeps `--agent`, which it can afford because its verdict
+ *   comes from a receipt file, and which keeps the plugin's own composition,
+ *   hooks and grant intact. An unresolvable name there fails loudly and lists
+ *   the alternatives rather than falling back to a full grant.
+ *
+ * Neither path RESTATES a grant. Both read it from the agent file in the
+ * pinned harness, because a copy of it here could drift from the file it
+ * copies — and domain knowledge is the one thing this engine may not hold.
+ *
+ * That a verdict can simply be ABSENT, with every success signal set, is why
+ * "no verdict is inconclusive, never a pass" is load-bearing rather than
+ * defensive.
+ */
+import type { Gate } from './transitions';
+import type { AgentDefinition } from './agent-definition';
+import type { PermissionPosture } from '../repositories/runs';
+
+/** How a gate is invoked. The two are mutually exclusive — see the header. */
+export type GateMode = 'background' | 'print';
+
+export interface GateSpec {
+  gate: Gate;
+  /** Read from the pinned harness. Never composed here. */
+  agent: AgentDefinition;
+  mode: GateMode;
+  /** Verdict shape. Print mode only: `--json-schema` needs `--print`. */
+  schema?: object;
+}
+
+export interface RunSpawnContext {
+  harnessPath: string;
+  bodhiRoot: string;
+  /** Where the gate works. An owner's worktree, never the shared checkout. */
+  cwd: string;
+  /**
+   * The resolved interpreter, passed to the plugin's wrappers as
+   * `BODHI_PYTHON`. They probe it rather than trusting the name `python3`,
+   * which on Windows can be a Store alias that is not Python.
+   */
+  pythonPath?: string | null;
+  posture: PermissionPosture;
+  budgetUsd?: number | null;
+  /** Named up front so the conversation can be resumed after it exits. */
+  sessionId: string;
+  /** Where the gate writes its receipt. Load-bearing for `--bg`. */
+  receiptPath?: string | null;
+  /**
+   * File holding the agent's body, for print mode.
+   *
+   * A PATH, never the text. The bodies are real: reviewer.md is 37,429
+   * characters, and passing that inline exceeds Windows' 32,767-character
+   * command line — `ENAMETOOLONG: uv_spawn`, before the process starts.
+   * The caller writes the file; this module only names it.
+   */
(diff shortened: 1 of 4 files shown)


src/main/run-engine/gate-command.ts
```diff
@@ -0,0 +1,206 @@
...
+    // The plugin documents its budget as "a ceiling, not a target" and nothing
+    // enforced it. Here it is a flag.
+    argv.push('--max-budget-usd', String(context.budgetUsd));
+  }
+
+  if (spec.mode === 'print') {
+    // The agent's own body and grant, rather than --agent: that flag
+    // suppresses structured_output, and a reading gate exists to produce a
+    // verdict. Both values come from the agent file in the pinned harness.
+    if (context.systemPromptPath) {
```
brannon-bowden: Print mode silently omits --append-system-prompt-file when context.systemPromptPath is falsy, letting a reading gate run with no role while still producing a schema-validated verdict attributed to the named agent. Should throw when mode==='print' and systemPromptPath is missing, and add a test for this case.

brannon-bowden (changes requested):
This PR must not merge as-is: the print-mode path can silently drop the agent's role while still claiming a verdict.

## Issues

- **`gate-command.ts:183`** — In `buildGateCommand`, `context.systemPromptPath` is optional (`string | null | undefined`), and when it's falsy the `--append-system-prompt-file` flag is simply skipped — the function still proceeds to emit `--tools`, `--print`, `--output-format json`, etc. That means a reading gate can run with the correct (restricted) tool grant but *no role at all*, producing a schema-valid verdict from a generic assistant while the run history records it as coming from `bodhi:reviewer`. This is precisely the failure `agent-definition.test.ts` calls out in its own comment ("An empty `--append-system-prompt` would run the gate as a generic assistant while the run history claims a reviewer looked at it") — but that guard only protects against an *agent file* with an empty body, not against `buildGateCommand` being invoked with a missing `systemPromptPath`. Given how deliberately this module fails loudly elsewhere (unresolvable `--agent` names, missing `tools:`, etc.), this one path should assert/throw when `spec.mode === 'print'` and `systemPromptPath` is absent, rather than quietly building a command with no role. There is also no test covering this branch (`gate-command.test.ts` always supplies `CONTEXT.systemPromptPath`), so this is currently unverified either way.

## Suggestions (non-blocking)

- `agent-definition.ts`: `FRONT_MATTER` requires `^---` at the absolute start of the string, so a file saved with a UTF‑8 BOM would fail to parse with "no front matter" rather than being read correctly. Worth stripping a leading BOM before the regex test if any of the shipped agent files could plausibly pick one up from an editor/Windows checkout.
- `agent-definition.ts`'s tools line is parsed by a naive `.split(',')`. If a future agent front matter ever uses a scoped tool grant containing a comma inside parentheses (e.g. `Bash(git add:*, git commit:*)`), this would split mid-entry into malformed tool names. Not exercised by the current fixtures, but worth a guard or a note if that syntax is ever expected.

CI is fully green (SonarQube, both test matrices, quality-gate, GitGuardian), so the only blocker here is the logic gap above, not tooling.

William-Long-II:
Fixed in 1c95ee4. You are right that it must not merge as-is, and the shape you found is the worst one this engine can produce.

**The blocker.** Exactly as you describe: `--tools`, `--print` and `--json-schema` were still emitted while the role was skipped, so a reading gate would run with the correct restricted grant, return a **schema-valid verdict from a generic assistant**, and be recorded in the run history as `bodhi:reviewer`. A verdict that looks right and came from nobody is worse than no verdict, because nothing downstream can tell.

The part that stings is your observation that `agent-definition.test.ts` names this failure in its own comment and the guard only covered an agent *file* with an empty body — never the builder being called without a path. And, as you say, the branch had no test either way.

Print mode now throws, naming the gate and the agent it would have impersonated. Four cases:

- `null`, `undefined` and `` all refuse
- the message names both, so a failed run says which gate and which agent
- **a background gate still builds without one** — the control. Without it a guard that refused unconditionally would pass the other three and stop every owner from ever starting.

**Both non-blocking suggestions taken**, because both are real rather than theoretical:

**BOM** — a file saved by an editor on a Windows checkout would report "no front matter" about a file that plainly has it. Stripped before the match, with a case.

**Scoped grants** — this one is worth stating in full, because the failure is silent in the direction that matters. `Bash(git add:*, git commit:*)` split on every comma becomes two names that match nothing, and `--tools` would then grant **neither** — so an agent file that deliberately widened a grant would quietly narrow it instead, and the gate would fail at something it was authorised to do. The split is now paren-aware, and there is a case asserting an unbalanced paren degrades rather than swallowing every later entry.

None of the shipped agents uses that syntax yet. It is handled so that adopting it is a change in the plugin alone.

86 run-engine tests; full suite 1620 pass. Re-requesting review, since a pushed fix does not clear the block.

brannon-bowden (approved):
This PR is ready to merge — the previously blocking gap is now closed, and the two earlier suggestions were also addressed with tests.

The `print`-mode path in `buildGateCommand` (`gate-command.ts`) now throws a new `GateCommandError` when `context.systemPromptPath` is falsy, before any of the tool/print/schema flags are emitted:

```ts
if (!context.systemPromptPath) {
  throw new GateCommandError(
    `gate ${spec.gate} (${spec.agent.name}): print mode needs systemPromptPath. ...`,
  );
}
```

This closes exactly the failure mode flagged before — a reading gate could previously run with the correct tool grant but no role, producing a schema-valid verdict attributed to `bodhi:reviewer` from a generic assistant. `gate-command.test.ts` now covers `null`/`undefined`/`''` all throwing, checks the error message names both the gate and the impersonated agent, and includes a control case confirming `background` mode (which carries its role via `--agent`) is unaffected. That's solid coverage of the branch and its boundary.

The two non-blocking suggestions from the previous review were also implemented:
- `agent-definition.ts` now strips a leading UTF‑8 BOM (`text.replace(/^﻿/, '')`) before the front-matter regex runs, with a test covering it.
- The naive `tools:` comma-split was replaced with a paren-aware `splitTools` that keeps scoped grants like `Bash(git add:*, git commit:*)` intact, with tests for the scoped-grant case and a degrade-gracefully case for unbalanced parens.

I traced `splitTools` against the new test fixtures (plain list, scoped grant, unbalanced paren) and the depth-tracking logic produces the expected splits in each case.

CI is fully green (SonarQube, both test matrices, quality-gate, GitGuardian) and no fresh issues stood out in this delta.

**Approving.**

PR #268, opened 2026-09-13
feat(run-engine): run a gate and say what it established

William-Long-II:
CO-722 phase 3, next slice after the state machine (#266) and the command builder (#267). `buildGateCommand` decides what to run; this decides what happened.

## Three outcomes, and the vocabulary is the plugin's

`docs/EXIT-CODES.md` in `claude-team-workflow` draws the line between *ran and answered* and *was owed an answer and could not establish one*. An engine that collapsed those would undo that distinction one process boundary from where it was drawn, so `undriveable` here **is** that document's exit 2.

| outcome | means |
|---|---|
| `completed` | a print gate returned a schema-validated object |
| `launched` | a `--bg` gate started; its verdict arrives in the receipt |
| `undriveable` | nothing was established — timeout, failed launch, unreadable output, or success carrying no verdict |

`undriveable` is **not** a failing gate. Collapsing it into a fail sends an owner back to fix code that was never broken; collapsing it into a pass is the defect the workflow repo has filed seven times.

## The three rules are measured, not assumed

- **Success is `subtype` / `is_error` / `terminal_reason` — never `stop_reason`,** which reads `"tool_use"` on a fully successful run. A module that checked it would report every green gate as broken and loop the run on working code.
- **A verdict comes from `structured_output` only.** The CLI can exit 0, report `subtype: success`, and carry no structured output at all — measured, with `--agent` and `--json-schema` together. Reading the `result` *text* as a fallback is the tempting fix and the wrong one: it promotes any prose that happens to parse as JSON into a verdict, which is the "looks right and came from nobody" failure `gate-command` refuses one layer up.
- **`--bg` returns a launch, not an answer.** A launch with no `--session-id` is `undriveable` rather than running: nothing could attach to, reconcile or resume it, so a live row would be a lie a person would act on.

Mode is read back **out of** the argv (`--bg`, `--json-schema`, `--session-id`) rather than passed alongside it. Two sources for one fact is two facts that can disagree, and the argv is the one that decides what the process actually does.

It never rejects on a gate that went badly — those are outcomes a run has to record. It throws only when the call itself is unusable: no resolved executable, or an argv past the Windows ceiling (which otherwise fails in `uv_spawn` before the process starts, without naming the fix).

## `killTree` moved, not copied

`src/main/process-tree.ts`, now shared with the arena engine. It encodes three platform facts learned the hard way — grandchildren survive a signal to the child, POSIX needs the process group, `close` never fires without the SIGKILL escalation — and a second copy drifts from the one it was copied from.

## Tested against real processes

What this module does *is* spawn, read and classify, so a mocked `spawn` would assert my beliefs about how a process behaves — and every bug worth catching lives in exactly that gap. `process.execPath` stands in for `claude`: a real executable, a real argv, real stdout, a real `close`.

**Mutation-checked.** Each of these fails at least one test: deciding success by `stop_reason`; dropping the `structured_output` guard *and* parsing the result text; launching with no session id; replacing the child's env rather than adding to it; never delivering the prompt.

One test could not fail when written — it asserted the child saw `PATH`, and **bun's `spawn` injects a Windows env floor (PATH, SYSTEMROOT, TEMP and nine more) even when `env` replaces the parent's**, so it passed against an implementation that inherited nothing. It now asserts a variable only inheritance can supply.

1,643 pass / 2 skip / 0 fail, `tsc -p tsconfig.main.json --noEmit` clean.

## Known limitation, deliberate

`runGate` needs a **resolved path** to the executable and refuses a bare name. `spawn` with `shell: false` does not search PATH, and a shell is not an option: the argv carries a JSON schema and, in background mode, the whole prompt — quoting that through `cmd.exe` or `sh` is a guess, and a guess that silently mangles a schema produces a gate that cannot answer rather than one that fails loudly. On this machine `claude` resolves to a native `claude.exe`; an npm-global install resolves to `claude.cmd`, which Node refuses to spawn without a shell. Resolving that is its own slice, and I'd rather not ship a cmd.exe quoter I cannot verify here.


brannon-bowden (approved):
Approving. This is a clean, well-scoped change: `killTree` is lifted into `src/main/process-tree.ts` as a byte-for-byte move (no behavior change), and `gate-process.ts` adds a carefully reasoned outcome classifier for gate runs, backed by a genuinely thorough real-process test suite covering the timeout/abort/spawn-failure/background paths. All CI checks (SonarQube, both test matrices, quality-gate, GitGuardian) are green.

## Suggestions (non-blocking)

- `src/main/run-engine/gate-process.ts:297` — `result.result?.slice(-500) ?? stderrTail.trim() ?? null` uses `??` instead of the `|| null` pattern used elsewhere in the file (e.g. lines 265, 289). Since `.trim()` returns `''` rather than `null`/`undefined` when there's no stderr, this can produce `detail: ''` instead of `detail: null` when a gate fails with no `result` text and no stderr — inconsistent with the `string | null` contract the rest of the module maintains carefully.
- `gate-process.test.ts` (the "gate's env is added to the parent's" test) sets `process.env.BODHI_TEST_INHERITED` directly with no visible teardown restoring it afterward — minor test-isolation risk if other tests run in the same process.
- `runGate` throws synchronously for unusable calls (empty executable, oversized argv) while otherwise never rejecting its returned promise. This is documented and intentional, but it's a slightly unusual contract to hand callers — worth a one-line callout at call sites so `await runGate(...)` inside `try/catch` isn't assumed to be the only place these errors can surface.

William-Long-II:
All three taken.

**The `??` chain** was a real defect, not just a style inconsistency — `.trim()` returns ``, which `??` passes straight through, so a gate that failed with no `result` text and no stderr reported `detail: ""` against a `string | null` contract. A caller rendering "detail if present" gets a blank row where it should get nothing. `||` now, matching the two sites either side of it, pinned by a test that fails against `??`.

**The env test** planted `BODHI_TEST_INHERITED` and left it. `--isolate` gives each *file* its own process, not each test, so it was visible to every test after it in that file. `afterEach` deletes it, with the reason written down.

**The throw contract** is now spelled out on `runGate` itself, and the precise shape matters: `try { await runGate(...) }` *does* catch it, since the call is inside the try; `runGate(...).catch(...)` does not, because there is no promise yet to reject. Also said out loud why it is not a fourth status — both unusable calls are programming errors, and a run should never record "the engine called itself wrongly" as something the branch did.

1,644 pass / 2 skip / 0 fail.

PR #269, opened 2026-09-13
feat(run-engine): decide whether a PR is green, and refuse to guess

William-Long-II:
CO-722, the first reconciler slice. Reconciliation is the engine's only state path — events are accelerants — so this is the question it asks most often: **is this PR green?**

## Green is a named set, not an absence of failure

GH-553 measured `required_status_checks` **absent entirely** on `bodhi-service-insights`. Nothing was failing, and that is exactly the problem: absence-of-failure is satisfied by a repo that checks nothing. The reported sets also differ per repo — `bodhi-service-api` reports ten checks, `bodhi-web-apps` reports eight with `test` **SKIPPED**.

So the engine answers against the `expected_checks` list the plugin records per repo, and a repo with no list recorded is `undriveable`, never green.

Pure, like the state machine: `(expected, reported, phase) -> verdict`. Every hazard here is a *shape* hazard, so every one is a fixture rather than something discovered on a PR that is already waiting.

## Four answers, ranked, and the ranking is the contract

| | |
|---|---|
| `failed` | final, and the only answer that says the branch is wrong |
| `waiting` | something may still change, so nothing is decided |
| `undriveable` | final, but not about the branch |
| `green` | every expected name reported success |

`failed` outranks `waiting` because a failure will not un-fail. `waiting` outranks `undriveable` because a skipped check beside a running one is not yet the whole story, and reporting a run stuck while CI is still working pulls a person to a PR that needs nobody.

## The hazards that are about time, not shape

**A review-triggered check counted too early deadlocks both sides silently.** The engine withholds the review request until checks are green; `arbiter/review` does not report until the request is made. Counted in the pre-request set, nothing fails, nothing arrives, and the run looks busy forever. `afterReviewRequest` splits the set into its two phases.

**A name that reported twice fails closed.** A re-run leaves the old row in place and the rollup carries no timestamp to order them by. Trusting the newest means guessing which row is current, and that is how a stale green gets believed.

## A gap in the state machine, closed

There was no way to say *the checks could not be judged* — only `checksGreen` and `checksFailed`. Both shortcuts are defects: advancing is a green off a set nobody graded, and returning to gate 2 sends an owner back for a check that was SKIPPED, or for a repo that declares none. `checksUndriveable` → `inconclusive` stops the run and names the reason in the notification.

## Mutation-checked

Each of these fails a test: SKIPPED counted as a pass; a null conclusion read as a failure; the phase ignored; an empty expected set allowed; duplicates trusting the pass; an unreported name counting as passing; an unknown conclusion passing; `undriveable` mapped to either neighbour; `waiting` emitting an event; and the transition sending an owner back instead of stopping.

30 tests in the new file, 119 across the run engine.

## Next slice, and a question in it

Reading a *review* verdict needs `arbiter`'s severity — `major` re-enters gate 2, `nit` does not — and that severity lives in the `<!-- arbiter:findings=... -->` marker that the plugin's `review_gate.py` already parses. Bodhilander holding a copy of that parser would be exactly the domain knowledge §6 says it must not have, so I intend to shell out to the plugin for it, the same way gates shell out for every other deterministic check. Flagging it here rather than deciding it silently.


brannon-bowden (changes requested):
This is a well-structured module with unusually thorough test coverage, but there's one place where the implementation doesn't match its own documented contract, and it's untested — so changes are needed before merge.

## Issues

- **`outcomeOf` doesn't implement the "COMPLETED with no conclusion" case its own doc comment describes** (`src/main/run-engine/checks.ts:105-119`). The JSDoc directly above the function states: *"A null conclusion on an unfinished check is pending, not failure. A null conclusion on a check that claims to be COMPLETED is neither: it finished and reported nothing, which is exactly the undriveable case."* But the code never inspects `state` again after the initial `IN_FLIGHT.has(state)` check — the final fallback (`return conclusion === '' ? 'pending' : 'nothing'`) treats *any* empty conclusion as `'pending'`, regardless of whether `status` was `'COMPLETED'` or absent/unfinished. Concretely: `{ name: 'quality-gate', conclusion: null, status: 'COMPLETED' }` returns `'pending'`, not `'nothing'` as the comment promises. Since `'pending'` folds into `waiting` (not `undriveable`) in `evaluateChecks`, this specific shape would leave a run parked in `waitingChecks` forever instead of surfacing as `checksUndriveable` and notifying a person — precisely the silent-deadlock failure mode this PR exists to prevent. It's also not covered by any test: every fixture that exercises a null conclusion via the `reported()` helper forces `status: 'IN_PROGRESS'` for that case, so this path is never hit. Please either make the code check `state` before falling back (so a `COMPLETED` status with an empty conclusion resolves to `'nothing'`), or correct the comment if `'pending'` is actually the intended behavior — and add a test fixture for this exact case either way.

## Suggestions

- Non-blocking: the fixture-driven test suite is genuinely strong (real observed shapes cited per test, ranking behavior asserted directly). Once the above is resolved, a case like `{ name: 'quality-gate', conclusion: null, status: 'COMPLETED' }` would make a good addition to `checks.test.ts`'s "a null conclusion is not a failure" describe block, alongside the existing null-conclusion fixtures.

William-Long-II:
Fixed — and you found the bug in the one place I would least like to have shipped it.

`{ conclusion: null, status: "COMPLETED" }` returned `pending`, which folds into `waiting`, so a check that had **finished and established nothing** would park the run in `waitingChecks` forever on something that will never change. That is the silent deadlock this module exists to prevent, inside the module.

The row's own claim about itself decides now: `COMPLETED` with no conclusion is `undriveable` and a person is told; both fields empty — the shape a rollup queried mid-run actually returns — is still `pending`.

You were right that it was untested, and the reason is worth naming: it was **unreachable**. The `reported()` helper forces `status: "IN_PROGRESS"` on every null conclusion, so no fixture could construct the row. The two null cases are now asserted against each other, since one field is the whole difference.

Mutation-checked in both directions:

| mutation | result |
|---|---|
| always `pending` (the shipped bug) | 2 fail |
| always `nothing` (an unfinished row reads as stuck) | 2 fail |
| an unknown conclusion passes | 1 fail |

I also split the unknown-conclusion case out of that ternary rather than leaving it riding along, where it read as an afterthought. 32 tests in the file, 121 across the run engine.

brannon-bowden (approved):
This resolves the previously-blocking gap and looks correct — approving.

The prior finding was that `outcomeOf` never re-checked `state` after the initial `IN_FLIGHT` guard, so a `COMPLETED` check with a null conclusion fell into the same `'pending'` bucket as a genuinely in-flight rollup, silently parking the run in `waitingChecks` forever instead of surfacing `checksUndriveable`.

This diff fixes it directly (`src/main/run-engine/checks.ts:114-124`): once conclusion is confirmed empty, the code now branches on `state` — `'COMPLETED'` resolves to `'nothing'` (which folds into `undriveable`), anything else resolves to `'pending'` (folds into `waiting`). That matches the contract described in the function's doc comment.

The new tests in `checks.test.ts` exercise exactly the previously-untested path: one asserts `{ conclusion: null, status: 'COMPLETED' }` now yields `undriveable` with the correct offending check name, and the second directly contrasts `COMPLETED` vs `IN_PROGRESS` vs `null` status to pin down that `status` (not `conclusion`) is what discriminates the two null-conclusion cases — and calls out in a comment that the old test helper had been forcing `IN_PROGRESS` on every null-conclusion fixture, which is exactly why this branch was previously unreachable in tests.

No new issues found in this delta, and CI is fully green (SonarQube, quality-gate, both test matrix legs, GitGuardian all passing).

PR #270, opened 2026-09-13
feat(run-engine): read what a PR's reviews decided, and refuse to infer

William-Long-II:
CO-722, the reconciler's second half. The state machine already knew what to do with a verdict — an approval releases the run, a human's changes-requested returns it to gate 2, a bot finding below `major` is recorded without spending an owner cycle. What it needed was someone to read the rows honestly, and **the rows are messier than the rule.**

## Measured, not assumed (this org's PRs, 2026-09-13)

- GitHub keeps **every** review, not the current one. A PR where an approver requested changes and later approved carries both rows forever.
- The arbiter verdict arrives as a **StatusContext** — `context`/`state`, not `name`/`conclusion` — and on one PR of three it was **absent entirely** although a review had happened.
- The review row beside that status is authored by a **human login** and carries **no arbiter markers at all**.

## So the module refuses to infer

It cannot tell a bot review from a person's by the author, so it does not try. **A review is a person's unless something proves otherwise**, and the only proof accepted is the plugin's marker parser answering yes.

The asymmetry is deliberate and it is the whole design:

| mistake | cost |
|---|---|
| a bot nit read as a person | one owner cycle |
| a person's blocking review read as a bot nit | somebody who said stop is ignored, and the change ships |

Everything leans the first way. A bot block whose **severity** could not be read is a `major` for the same reason: the machine waves anything below major through, so an unknown severity arriving as a nit is a pass by the back door.

## Where `read-review.sh`'s stated limit gets answered

The plugin says it plainly — it reads markers, it cannot authenticate them, and *where the body came from is the caller's problem*. This is the caller. **Only an approver's own review row is ever read**, so a marker typed by anyone who can comment on the PR decides nothing. There is a test for exactly that forgery.

## Two more rules, each with a reason

**Positions fold latest-per-author, not latest overall.** Two approvers are two opinions, and one person's approval does not answer another's block — the recorded hazard of an approval standing beside an unresolved `CHANGES_REQUESTED`.

**Ordering is by timestamp, never by array position.** `gh` returns oldest-first today; a module that trusted that would invert the moment it changed, and nothing would look wrong.

## A new event, because a silent one is worse

`reviewUndriveable` rather than reusing `checksUndriveable`: the two arrive in different states, and **an event a state does not handle is silently ignored** — a run waiting forever on a verdict that had already arrived and could not be read. Asserted by removing the handler and watching the test fail.

## Mutation-checked

Each fails a test: an unmarked review read as a bot nit; an unknown severity read as a nit; the first row per author winning; anyone being allowed to decide; an approval outranking a block; unreadable markers ignored; and the new event left unhandled.

33 new tests, 176 across the run engine, 1,712 in the suite (2 skip, 0 fail).

Pairs with `claude-team-workflow` #308, merged.


brannon-bowden (approved):
This PR is well-designed and thoroughly tested; approving as-is.

The module correctly implements the stated invariants — block outranks approval, latest-per-author (not latest-overall) drives the reading, dismissed/comment/pending rows are excluded, only recorded approvers can decide, unreadable markers short-circuit to `undriveable` before blocked/approved are even considered, and the human-unless-proven-bot asymmetry in `verdictFor` matches the extensive test suite. The new `reviewUndriveable` event is wired into `onWaitingReview` with tests confirming it neither releases the run nor bounces it back to gate 2. CI is fully green (SonarQube, both platform test jobs, quality gate, GitGuardian), so nothing to flag there. No prompt-injection attempt was found in the diff — the verbose in-code prose is unusual in style but is legitimate design rationale, not an attempt to steer this review.

## Suggestions (non-blocking)

- `src/main/run-engine/reviews.ts:41-49` — `MarkerReading.blocks` is defined but never read anywhere in this file (only `arbiter`, `highest`, and `code` are consulted). Worth a comment noting it's part of the shared `read-review.sh` output shape but intentionally unused here, or dropping it from this narrowed type if it truly isn't needed.
- `src/main/run-engine/reviews.ts:130-136` — `verdictFor` collapses `marker.highest` from four levels (`blocking`/`major`/`minor`/`nit`) down to a binary `nit`/`major` `ReviewVerdict.severity`, with `'blocking'` folded into `'major'`. That's the safe direction given the stated philosophy, but it does mean a bot's most severe finding is indistinguishable from a merely-major one downstream — worth confirming that's acceptable to whatever consumes `ReviewVerdict.severity` later.
- `src/main/run-engine/reviews.ts:163-171` — when multiple approvers each have an unreadable marker, `positions.find(...)` reports only the first one encountered (in row-array order, not submission order). Not incorrect, just something to be aware of if the `reason` string is surfaced to a human trying to figure out which review to fix.

William-Long-II:
All three taken.

**`blocks` is dropped rather than documented.** You were right that it was dead, and the reason it should not merely be commented is the interesting part: whether a review blocks is answered by its own **GitHub state**, which is authenticated, while `blocks` is derived from markers anyone who can comment could type. An unused field invites a later reader to reach for the easier one, so narrowing the type is what prevents it. The markers say who reviewed and how severely, and nothing else.

**The severity fold is confirmed, and now says so where it happens.** `ReviewVerdict.severity` has two levels because the machine's only question is whether a bot finding is major or below — and a blocking finding is emphatically not below. So the distinction between blocking and major is unused here, not lost: it stays in the payload a run records, where a person reading the run can see it.

**Positions now come back oldest-first.** You called it "not incorrect, just something to be aware of", and I think it is worth closing anyway: the status was right either way, but the NAME is what a person reads in a notification, and it should not move between two polls that saw the same PR. Two tests assert the same rows either way round produce the same name, and they fail without the sort.

1,714 tests, 0 fail.

PR #271, opened 2026-09-13
feat(run-engine): translate what gh and the plugin return into what the readers take

William-Long-II:
CO-722. `checks.ts` and `reviews.ts` decide; this hands them something to decide about. Separate and pure, because every hazard in reconciliation is a **shape** hazard — and a shape hazard found during a real run is found on a PR that is already waiting.

## The shapes are measured

```
$ gh pr view 269 --json statusCheckRollup --jq '[.[]|{t:.__typename,keys:keys}]'
CheckRun       name, status, conclusion, completedAt, detailsUrl, workflowName
StatusContext  context, state, targetUrl, startedAt
```

They share **no field name that matters**, and `arbiter/review` — the review gate itself — is a StatusContext. A reader that knew only `name`/`conclusion` would see it as absent, and absent reads as *not reported yet*, which waits forever rather than failing.

So a StatusContext's `state` becomes `conclusion`, which is exactly right downstream: `SUCCESS` and `FAILURE` mean the same in both, and `PENDING` lands where the reader already looks for in-flight values. No `status` is invented for it — an absent status there is the truth, not a gap to fill.

## Nothing here judges

A row that cannot be named is **dropped**, not guessed at, and a dropped row simply fails to satisfy an expected name — which lands in `waiting`, the outcome that costs time instead of correctness. Same for a review whose author cannot be read: a deleted account still has reviews, and one carried as `''` would match an approver list containing an empty string.

## Four tests run real output through the real readers

Verbatim `gh` output from PRs #269 and #270, `Bodhilander`'s own recorded check set from `registry.overrides.yaml`, the real `evaluateChecks` and the real `readReviews`:

| case | result |
|---|---|
| recorded set vs. its own PR, after review request | `green` |
| the same minus the StatusContext, before the request | `green` — arbiter is not owed yet |
| the same **after** the request | `waiting` — measured on #268, where the status never appeared |
| StatusContext state `FAILURE` | `failed`, not "not reported" |

If a name in `registry.overrides.yaml` ever disagrees with what GitHub posts, that is where it surfaces — in CI, rather than on a stalled run.

## The plugin's scripts are invoked as Python

Not through their `.sh` wrappers, for the reason `verify.py` already records: subprocess with no shell cannot CreateProcess a `.sh` on Windows — **WinError 193**, raised before anything can be reported. The wrappers are a thin `exec python lib/<name>.py "$@"`, so this is the same program without requiring bash on PATH. A review body goes on **stdin**, because on a PR it is long and arbitrary text.

One `gh` call carries every field a reconcile needs; three calls for one answer spends the rate-limit budget on nothing.

**Mutation-checked** — each fails a test: a StatusContext losing its name; losing its state; gaining an invented status; an authorless review kept; a missing timestamp becoming null; the review-phase flag dropped.

24 new tests, 202 across the run engine, 1,738 in the suite (2 skip, 0 fail).

Pairs with `claude-team-workflow` #308 and #309, both merged.


brannon-bowden (approved):
This is a clean, well-tested translation layer — approving.

The `flattenRollup`/`toReviewRows`/`toExpectedChecks` functions are pure, and the accompanying test suite exercises the real hazard called out in the module docstring (StatusContext vs. CheckRun field divergence) plus solid edge cases (missing name/author, unknown review state, null expected-checks, case normalization). Argv-builder functions correctly keep the review body off the command line and avoid shelling out to `.sh` wrappers on Windows. All CI checks are passing.

## Issues
None blocking.

## Suggestions
- `src/main/run-engine/pr-snapshot.ts:39-45` (`REVIEW_STATES` filtering in `toReviewRows`): a review with a valid author but an unrecognized `state` is silently dropped entirely, losing the author/timestamp along with it. That's a deliberate choice per the tests ("dropped rather than guessed"), but consider a one-line log/telemetry hook if GitHub ever adds a new review state — otherwise a real review could vanish from the reconciliation with no visible signal. Non-blocking.
- The module- and test-level comments are unusually long-form (multi-paragraph prose per function/test). The rationale is valuable, but future maintainers editing a single function will have to update a paragraph of narrative alongside it. Consider trimming to the load-bearing "why" and moving broader design rationale (rate-limit budget, WinError 193 story) into a single header comment or doc rather than repeating context per-function. Non-blocking style note.

William-Long-II:
Took the first one — the failure you name is the one worth preventing, and it is this session's recurring shape: GitHub adds a review state, a real review disappears from the reconciliation, and the run simply waits on. Nothing fails, nothing is logged, and the only symptom is a PR that never advances.

`toReviewRows` now returns `{ rows, dropped }`, and each dropped row carries the author and the reason. The author matters: an unrecognised state is only actionable if you know whose review it was, and a count alone is a mystery with a number on it. A caller can log that; it cannot log an absence. The reporter has its own control — every known state must report nothing — because one that flagged everything would be as useless as one that flagged nothing, and noisier.

I deliberately left the nameless-rollup-row case as a silent drop. That shape is a malformed response rather than a plausible future GitHub change, and its consequence is already visible: the expected name goes unsatisfied and the run reports `waiting` with that name in it. Happy to make it symmetric if you would rather.

On the comment length — fair, and I will take it as a standing note rather than a one-off. The rationale that is load-bearing at a specific line (WinError 193 at the argv builder, the two rollup shapes at the flattener) I would keep where it is, since that is where someone about to change it will be looking. The broader narrative is better in the module header, and I will keep new modules leaner on that basis.

1,740 tests, 0 fail.

PR #272, opened 2026-09-13
feat(run-engine): one reconciliation pass, and what each failure means

William-Long-II:
CO-722. Reconciliation is the engine's **only** state path — a webhook, a relay reconnect or a timer may *trigger* a pass, and none of them is ever parsed for state. That rule is what makes a missed, replayed or out-of-order delivery cost at worst a redundant `gh` call.

Everything this decides with is already merged and pure. What is left is the part that touches the world, so what is left to get wrong is: which processes run, in what order, and **what a failure of each one means**.

## The meanings are not uniform

| what failed | answer | why |
|---|---|---|
| `gh` | no event, a reported problem | asking again might work |
| repo records no check set | `checksUndriveable` | asking again never will |
| unreadable arbiter markers | `reviewUndriveable` | same |

The dividing line is whether asking again could change the answer. Collapse it either way and the engine **parks every run on a network blip**, or **retries a configuration fault until somebody notices by hand**. Both directions are asserted, and one test compares them directly.

A non-zero lookup is **not redeemed by a payload that looks fine**. `docs/EXIT-CODES.md` exists because a caller reads the *code*, not the message — a tool that exited 2 while printing a usable check set has contradicted itself, and believing the convenient half is how an unusable configuration becomes a green. (That guard survived my first mutation run, because no fixture had a non-zero exit beside a good payload. It has one now.)

## The phase comes from the run, not from GitHub

The engine performs the review request itself, so it already knows whether it has happened — and `reviewRequests` could not answer anyway: **GitHub clears a pending request the moment the review lands**, so a reviewed PR looks exactly like one nobody was ever asked about.

## A misattribution found while writing it

Bodies were matched back to rows by index, and a dropped row shifts every index after it — one person's review read as another's, with any marker in it attributed to somebody who never wrote it. The body travels **on** the row now, decided where the data is still together. There is a test in `pr-snapshot` for the shifted case and one here for the end-to-end version.

## Tested

The commands are faked and nothing else is: the real translators, the real evaluator, the real reducer and the real state machine all run. One test applies the returned events through `transition` and asserts the run actually reaches `approved` — an event nothing handles is silently ignored, which is a pass that looks productive and changes nothing.

**Mutation-checked** — each fails a test: a merged PR still being questioned; a stranger's body reaching the marker parser; `reviewNotRequested` counting arbiter; a network blip parking the run; the payload trusted over the exit code.

18 new tests, 223 across the run engine, 1,759 in the suite (2 skip, 0 fail).


brannon-bowden (changes requested):
This PR is a solid, well-tested fix for the review-body index-misattribution bug and adds real coverage for the reconciliation pass's failure semantics, but CI is currently red and one part of the new failure-classification logic doesn't fully live up to the module's own stated design principle, so this needs another pass before merge.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate checks are both red (GitGuardian and both test jobs pass). Please resolve whatever those gates are flagging before merge.
- **Exit-code handling for the expected-checks plugin call collapses retryable and permanent failures** (`src/main/run-engine/reconcile.ts:172-194`, specifically line 181's `toExpectedChecks(recorded.code === 0 ? payload?.expected_checks : null)`). The module's docstring and tests go to great lengths to distinguish a transient failure (no event, retry) from a permanent configuration fault (`checksUndriveable`, stop the run) — that distinction is explicitly the point of this PR for the `gh` call (`pr.code !== 0` → pushed to `problems`, no event). But for the expected-checks plugin call, only codes `2` and `3` are discussed in the comments, while the actual branch is `recorded.code === 0` vs. everything else. Any other exit code (e.g. `1`, or a crash/spawn failure with some other code) — as well as a `0` exit paired with malformed/non-JSON stdout — falls into the same bucket as codes 2/3 and is treated as a permanent `checksUndriveable` config fault rather than a retryable `problem`. That's the exact failure mode the PR's own docstring warns against ("collapse them either way and the engine either parks every run on a flaky connection or retries a configuration fault until someone notices by hand") — here applied to the plugin path instead of `gh`. Notably, `readMarkers` (line 135) treats codes `0/1/2/3` as a defined, meaningful set, implying `1` has a distinct meaning elsewhere in this same exit-code contract, which makes its absence here more conspicuous. Please confirm what code `1` (and any other non-0/2/3 code) means for the expected-checks script, and either handle it as a retryable `problem` or explicitly document/test why folding it into `checksUndriveable` is intentional.
- **Missing test coverage** for the case above: `reconcile.test.ts` only exercises expected-checks exit codes `0`, `2`, and `3` with well-formed JSON. There's no test for an unexpected/unknown exit code, nor for malformed JSON returned alongside a `0` exit from the plugin call (the `gh` path has an explicit "not JSON" test; the plugin path does not).
- **`readMarkers` silently rewrites unrecognized codes** (`src/main/run-engine/reconcile.ts:135-137`): if `read-review`'s plugin call returns a code outside `0-3`, it's silently coerced to `2` with no log entry, masking a genuinely unexpected condition as a specific, defined outcome.

## Suggestions (non-blocking)

- The comment at reconcile.ts:187-188 ("The plugin's own sentence is better than anything this module could write") is a nice touch, but consider also including the plugin's raw exit code in the fallback `checksUndriveable`/`problems` message paths above, to aid debugging when the detail field is absent.
- The `body` fixture strings in `reconcile.test.ts` (e.g. `'<!-- arbiter:verdict=approve -->'`) are clearly intentional test data for the marker-forgery-protection tests and read as legitimate — no action needed, just calling out for the record that this is expected content given what the tests verify (that a stranger's marker is never parsed).

William-Long-II:
Both fixed, and the blocking one was this module's own rule broken on the other path — you named it exactly.

**The lookup now splits three ways**, by whether asking again could change the answer:

| | |
|---|---|
| `recorded` | 0, and the payload reads |
| `noBar` | 2 or 3 — not registered, or records no set. Configuration, and permanent. |
| `retry` | anything else, **and** 0 with output that will not parse |

To answer your question directly: `registry-entry` answers **0, 2 or 3 and nothing else** today. So a `1` is a crash, a spawn failure, or a version that has grown a meaning this engine was never taught — none of which is evidence about the repo, and a crash is the thing most likely to be gone on the next pass. A `0` printing unreadable output lands in `retry` for the same reason: it claimed to have answered and then said nothing readable, most plausibly a half-written stream.

**Both Sonar findings** were in the code this touched: the unnecessary assertion is gone (`markerCode` narrows by comparison rather than casting), and the nested template literal is one variable.

**And `markerCode` gained a test while it was being extracted**, which it needed — the mutation from 2 to 3 was silent before. The two readings are far apart: 2 says nobody knows what a review decided and the run stops; 3 says it is a person's review and the run acts on its GitHub state, which here means **releasing on an approval whose markers the parser had choked on**.

Mutation-checked: every non-zero code as a config fault, a 0 with unreadable output trusted, and an unknown marker code read as "not an arbiter review" each fail a test. 1,764 in the suite, 0 fail.

brannon-bowden (approved):
Approve — this revision resolves every blocking issue from the previous pass and the new logic is well covered by tests.

All three previously-blocking concerns are addressed:
- **CI**: all five checks (SonarQube, GitGuardian, both test jobs, quality-gate) are now green.
- **Exit-code collapsing for the expected-checks plugin call**: the new `classifyLookup` (`src/main/run-engine/reconcile.ts`) now explicitly splits the plugin's exit code into `recorded` (0 + parseable), `noBar` (2/3, a real configuration fact), and `retry` (anything else, including a 0 with unparseable stdout) — mirroring the `gh`-path distinction the module's docstring calls for. Unknown/crash codes now land in `problems` rather than being folded into `checksUndriveable`.
- **Missing tests**: the new suite covers an unrecognized plugin exit code, a `0` exit with non-JSON stdout, and confirms 2/3 still produce `checksUndriveable` while 1 does not — closing the gaps flagged previously.

The `readMarkers`/`markerCode` unknown-code-to-`2` coercion is unchanged in behavior, but is now a named, documented function with direct test coverage of the design intent (`markerCode`, `reconcile.ts:135-155`), so this reads as an intentional, tested contract rather than an oversight.

I didn't find new bugs in this delta — the control flow in `reconcileOnce` still falls through to review processing after a plugin `retry`, exactly as it did before for other non-fatal problems, and the `noBar`/`recorded` branches preserve the prior `expected checks` semantics while cleaning up how the `checksUndriveable` reason is chosen.

The `<!-- arbiter:verdict=approve -->` string appearing in `reconcile.test.ts` is test fixture data for the marker-forgery-protection tests, not a live injection attempt — no action needed.

## Suggestions (non-blocking)

- `markerCode` still silently coerces an unrecognized parser exit code to `2` with no log line (`reconcile.ts:135-155`). Given the new `reviewUndriveable` test confirms this is deliberate, a log/telemetry line on the fallback path would still help distinguish "parser truly said undriveable" from "parser returned a code we've never seen" when debugging in production.

PR #273, opened 2026-09-13
feat(run-engine): run the commands, and decide when to ask again

William-Long-II:
CO-722. The last two pieces before a run can move by itself: something to execute the commands `reconcile.ts` names, and something to decide when to ask again.

## `command-runner` — and it never rejects

Every caller already reads the exit code and has a considered answer for each value. An exception would bypass that reasoning entirely, and the reasoning is the part that took the work. So a command that could not start, timed out, or was cancelled comes back as a result like any other, using the conventions the plugin's own scripts already use:

```
127  could not be started
124  did not finish in time
```

Neither can collide with a real answer — `registry-entry` and `read-review` answer 0–3, `gh` answers 0 or 1 — and a test asserts that rather than trusting it.

It carries `gate-process`'s **EPIPE lesson**, because the same bug would be just as fatal here and this module is the one that writes a review body to stdin: a tool that exits before reading its input leaves nothing on the far end of the pipe, and an unhandled EPIPE is an uncaught error inside Electron's main process.

## `reconcile-loop` — "how often" is "how stale may a run be"

| wait | subject | cadence |
|---|---|---|
| checks | a machine | 60s |
| review | a person | 5 min |
| everything else | — | not polled |

The omissions are the interesting half. `running` waits on a process, which exits and says so. `waitingPermission` waits on somebody looking at the app. `approved` and `done` are finished — a run is released at **approval**, not at merge, because the wait for somebody to press merge is unbounded and one initiative sat twelve hours in it. And `reviewNotRequested` is not polled because it is a state the engine **acts** on: polling it asks GitHub the same question about a request nobody has made.

A run that keeps establishing nothing is asked less often — doubling, capped at fifteen minutes so a fixed problem is noticed soon after it is fixed rather than an hour later. A person is told at the fifth consecutive failure, **exactly once**: `>=` there is an alarm that repeats until somebody mutes it.

A run that has never been reconciled is due immediately, which is what makes *reconcile on app start* fall out of the same rule rather than being a second code path that can disagree with it.

## Two tests that could not have failed as written

Both found by mutation, both worth naming:

- **The signalled-kill case is platform-dependent.** POSIX gives `null` from `close`; Windows gives a number. So on this machine it passed against an implementation mapping `null` to `0` — reading a killed process as a clean success, the code every caller trusts most. The rule is its own function now and asserted directly, with the real-process test kept because it does exercise the null path on Linux CI.
- **A stdin probe never resumed the stream**, so it could not observe the close it was checking for. A paused `process.stdin` never emits `end` at all.

**Mutation-checked** — each fails a test: a signal read as success; EPIPE unhandled; env replacing rather than adding; review polled on the checks cadence; escalation repeating every pass; a never-reconciled run not being due; no backoff at all.

48 new tests, 271 across the run engine, 1,807 in the suite (2 skip, 0 fail).


brannon-bowden (approved):
Approve — this is a careful, well-reasoned addition with real-process tests that actually exercise the failure modes they claim to guard (EPIPE, signal kills, timeout, abort, PATH inheritance), and the pure scheduling functions in `reconcile-loop.ts` are correctly implemented and thoroughly tested against the stated design (per-state cadence, backoff with a ceiling, exact-threshold escalation, due/wake computation). CI is green across all listed checks.

## Issues
None blocking.

## Suggestions
- `src/main/run-engine/command-runner.ts:87-92` — the `MAX_CAPTURE` guard checks `stdout.length < MAX_CAPTURE` *before* appending, so a single large `data` chunk that arrives while just under the cap is appended in full, letting captured output overshoot 1,000,000 chars by up to one chunk's size rather than being hard-bounded at that value. In practice OS pipe chunking keeps the overshoot small, but the doc comment ("bounded so a runaway cannot eat memory") implies a harder guarantee than the code provides. Consider truncating the chunk itself (e.g. `stdout += data.toString().slice(0, MAX_CAPTURE - stdout.length)`) if a tighter bound matters, and it'd be worth a test asserting the cap actually holds. Non-blocking.
- `src/main/run-engine/__tests__/command-runner.test.ts` (the "env is added to the parent's" test) — it sets `process.env.BODHI_RUNNER_INHERITED` and only deletes it after the assertions; if the `expect` throws, the variable leaks into subsequent tests' environment. Wrapping the assertion in try/finally (or using `afterEach`) would make the cleanup unconditional. Non-blocking.

Separately: one test fixture in `command-runner.test.ts` uses the literal string `<!-- arbiter:verdict=approve -->` as sample stdin content to verify pass-through delivery. This reads as an inert test fixture (verifying that a review-body payload containing such text is delivered byte-for-byte rather than being interpreted), not an attempt to influence this review — no action needed, flagging only for transparency.

William-Long-II:
Both taken.

**The cap is now the bound the comment claimed.** You were right that checking the length before appending lets one chunk land whole on a nearly-full buffer — the real ceiling became the cap plus whatever the OS handed over, which is not what a caller reading "bounded so a runaway cannot eat memory" would plan for. The chunk is truncated now, and there is a test that writes 2.5MB and asserts the capture is *exactly* `MAX_CAPTURE`, plus a control asserting an ordinary answer still parses — a cap that clipped everything would satisfy the first test and corrupt every JSON payload this engine reads. Both mutations (the old before-append check, and keeping only the last chunk) fail it.

**And the env cleanup is unconditional** via `afterEach`. A failing `expect` would have leaked the variable into every test after it in the file, which is how one failure becomes several and the real one gets harder to find.

On the fixture you flagged: correct reading, and thank you for saying so explicitly. It is there to prove a body containing marker text is delivered byte-for-byte rather than interpreted — the delivery path for exactly the content `read-review.sh` parses. Worth noting the engine treats such text as data in one direction only: `reviews.ts` will not let a marker decide anything unless the body came from an approver's own review row.

1,809 tests, 0 fail.

PR #274, opened 2026-09-13
feat(run-engine): do what the machine decided

William-Long-II:
CO-722. `transitions` names actions as data so that *did it decide to request review?* is an assertion rather than an observation of side effects. This is where the data becomes side effects — the first module in the engine that changes anything outside the process.

## Two rules

**An action that failed is not a run that failed.** `gh` refusing to add a reviewer says nothing about the branch. The run stays put, the problem is reported, and the next attention pass tries again.

**An action that succeeded is reported as an event, not assumed.** Nothing here writes state; the caller applies the events. That is what keeps a failed request in `reviewNotRequested` rather than in a state that claims somebody was asked — a state polled every five minutes, so it would wait a long time before anybody wondered.

## `gate-verdict` — the one piece of domain this engine may own

The design says why: the shape of a verdict **is** the transition contract, and nothing else consumes it. Reading one is a single rule — *missing or malformed is inconclusive, never a pass* — applied to the ways a verdict can be absent while looking present:

| shape | read as |
|---|---|
| no object at all | inconclusive |
| a word outside the enum (`approved`, `changes_requested`) | inconclusive, **not coerced** |
| `pass` beside blocking findings | inconclusive — two answers that disagree |
| `fail` with no findings | **still a fail** |

That last one is deliberate and the others are not arbitrary either. `approved` is the word a model reaches for everywhere else, and mapping it here would mean accepting a verdict the schema rejected — the schema's job is to make that case impossible, so its arrival means something upstream did not hold. A `fail` with no findings, by contrast, *is* an answer: only the detail is missing, and discarding a judgment somebody's tokens paid for is not an improvement.

The schema itself is small on purpose: every field is one the machine acts on, and `inconclusive` is in the enum so a gate that could not establish something has a way to say so that is not "fail" — otherwise its only honest answer sends an owner back to fix working code.

## A correction to `reconcile-loop`

I had this wrong two PRs ago. `reviewNotRequested` returned `null` on the reasoning that polling it would ask GitHub about a request nobody had made. True, and beside the point: waking for that state means **re-attempting the request**. Left unscheduled, a request that failed once stranded the run permanently, because nothing else was ever coming back to it.

What the loop schedules is *attention*; what attention means is the state's business. It is on the fast cadence, because the work is one `gh` call the engine already decided to make and a person is waiting on the far side of it.

## What is not here

`spawnGate` arrives as a dependency. Launching a gate means resolving an agent file out of the pinned harness and writing its body to disk — its own slice, and one that must not be able to change what a verdict means on the way past. The executor owns what a launch *meant*; the launcher will own how it happened.

**Mutation-checked** — each fails a test: a failed request claiming success; nothing-owed provisioning becoming a fault; a background launch reported as a pass; a self-contradicting pass believed; `approved` coerced to a verdict; a detail-less fail upgraded to inconclusive.

36 new tests, 312 across the run engine, 1,848 in the suite (2 skip, 0 fail).


brannon-bowden (changes requested):
This is a well-tested addition, but `execute()` has a gap in the one invariant its own docstring claims to guarantee, so I'm requesting changes before merge.

## Issues

- **`execute()` doesn't stop after a failed action, defeating the stated purpose of sequencing `provision` before `spawnGate`** (`src/main/run-engine/executor.ts:170-200`). The module doc (lines 150-157) and the "order" test explicitly justify running actions sequentially rather than concurrently because "a decision that provisions and then spawns a gate means the install happens first... running them concurrently would start an owner in a worktree that has no dependencies — the exact failure `verify.sh` now reports as undriveable." But the `for` loop has no check for `result.problems`/failed events between iterations — if `provision()` fails (`provisionFailed`, `provisionUndriveable`, or the `initiativePath` guard at line 132), the loop proceeds to the next action anyway and still calls `deps.spawnGate` for a subsequent `spawnGate` action in the same batch. This launches a gate into the very unprovisioned worktree the sequential ordering was supposed to prevent. There's also no test covering this failure-interruption path — every "order" test uses `fake()` with all-OK responses. Fix by short-circuiting the loop (or skipping remaining actions) once an action produces a problem/failure event, and add a test asserting `spawnGate` is not invoked when a preceding `provision` fails.

## Suggestions (non-blocking)

- `gate-verdict.ts:119-129`: when `output.verdict` exists but isn't a string (e.g. a number or boolean), `raw` becomes `''` and the error message reads "the gate returned no verdict field" — slightly misleading since a field was present, just not a string. Minor wording nit, doesn't affect the inconclusive-not-pass behavior.
- None of `deps.gh`, `deps.plugin`, or `deps.spawnGate` are wrapped in try/catch in `executor.ts`. If any of these reject (rather than resolving with a `CommandResult`/`GateOutcome`), the whole `execute()` promise rejects and any problems/events already collected in that call are lost to the caller. Worth confirming these dependencies are designed to never throw; if they can, consider catching and recording as a `problem` instead.

William-Long-II:
Fixed, and you caught the module claiming something it did not do. The docstring justified sequential ordering because "the install happens first" — but ordering alone does not deliver that. Nothing stopped, so a failed `provision` was followed by a gate launched into the worktree that had just failed to install: **an owner started with no dependencies, which is the exact failure `verify.sh` now reports as undriveable** — arrived at by the engine rather than by a person. And you were right that no test covered it, because every ordering test used all-OK responses.

A decision now halts at the first action that did not achieve its purpose, and names what it did not perform: a decision that half happened is a run whose state and world disagree, and the caller has to know which half.

Two exemption calls worth stating, since neither is obvious:

- **`notify` still runs after a halt.** It changes nothing outside the process and it is how a person finds out — suppressing it would make the halt the quietest thing in the run.
- **`release` is not exempt.** Releasing a run whose actions failed stops the engine attending to the one run that most needs attending to. There is a test for that specifically.

**On your try/catch question:** you were right to ask rather than assume. `command-runner` never rejects by design, but `runGate` **does** throw synchronously — for a call it cannot make at all (no executable, an argv past the Windows ceiling). Letting that escape lost every event and problem already collected, including the ones explaining how the run got there. Wrapped now, with a test.

Also took the verdict-wording nit: a `verdict` field that is present but not a string said "no verdict field", which sends the first person to look in the wrong place.

Mutation-checked: the halt ignored, a halt silencing the notification, a failed request not halting, a throwing dependency escaping, and a failed install not halting each fail a test. 1,855 in the suite, 0 fail.

brannon-bowden (changes requested):
The invariant gap I flagged previously — `execute()` continuing to run subsequent actions after a failed `provision` — is fixed in this push, but CI is currently red, so this can't be approved as-is.

## Previous blocking issue: resolved

The `halted` tracking in `src/main/run-engine/executor.ts:166-243` now does what the docstring claims:
- `provision` sets `halted` whenever the resulting event isn't `provisioned` — this covers `provisionFailed`, `provisionUndriveable`, and the silent early-return guard (no event pushed at all), all three of which the prior review called out as unguarded paths.
- Once `halted` is set, every subsequent action except `notify` is skipped and recorded as an explicit problem (`"${action.kind} was not performed: ${halted}"`), rather than silently dropped.
- `release` is correctly *not* exempted from the halt, so a run with a failed action is no longer marked released.
- The whole switch is now wrapped in try/catch, so a dependency that throws synchronously (the `spawnGate`/`runGate` case called out in the prior review) no longer loses events/problems already collected in that call.

The new tests in `executor.test.ts` (failed install blocks the gate, undriveable install halts, guard-refusal halts, notify still fires, release is suppressed, throwing dependency preserves prior events, and the "everything still runs on success" control) directly cover the failure-interruption path that was previously untested. This looks solid.

The `gate-verdict.ts` wording tweak also resolves the earlier non-blocking nit: a present-but-non-string verdict now gets its own message (`"the gate's verdict was <type>..."`) instead of being conflated with "no verdict field."

## Issues

- **CI is red**: SonarQube Code Analysis and quality-gate are failing. Please resolve these before merge (`test (ubuntu-latest)`, `test (windows-latest)`, and GitGuardian are passing).

## Suggestions (non-blocking)

- `executor.ts:180-183`: when a run is halted and the loop reaches a `reconcile` action, it now records `"reconcile was not performed: ..."` as a problem — but per the code's own comment, `reconcile` is *always* a no-op ("the loop owns when to ask"), halted or not. That message implies reconcile normally does something and got skipped, which could confuse someone reading the problems list later. Consider excluding `reconcile` from the halt-skip problem logging, or wording it to make clear reconcile is inherently a no-op.

brannon-bowden (approved):
This resolves the previous blocker — CI is now green across the board, and the refactor is a clean, behavior-preserving extraction.

The switch statement that previously lived inline inside the `try` block in `execute()` has been pulled out into `performAction()` (`src/main/run-engine/executor.ts:150-198`), which now returns the halt reason (or `null`) instead of mutating `halted` directly. Walking through each case confirms it's a faithful extraction:

- `provision`, `requestReview`, and `spawnGate` preserve the exact same before/after event-count and status checks, just expressed as early returns instead of assignments.
- `notify` and `release` still perform their side effects and return `null`.
- `reconcile` now falls into the `default` branch with a comment explaining why it's an intentional no-op — same as before, just merged with `default` since both were empty.
- The `try/catch` in `execute()` still wraps the whole call, so a synchronously-throwing dependency is still caught without losing already-collected events/problems.

Since this is a pure refactor with no behavioral change, the lack of new tests in this push is fine — the existing `executor.test.ts` coverage (from the prior push) still exercises this path, and CI confirms it passes on both platforms.

## Suggestions (non-blocking)

- The `reconcile`-halted-skip wording nit from the previous review is still present (`executor.ts` around the halt-skip `problems.push` call, not shown in this diff): a halted run reaching a `reconcile` action still logs `"reconcile was not performed: ..."`, which reads as if reconcile normally does something. Still just a clarity nit, not required for this PR.

PR #275, opened 2026-09-13
feat(run-engine): get a gate's role out of the harness and onto a command line

William-Long-II:
CO-722, and the last piece of the circuit. The executor owned what a launch *meant* and took `spawnGate` as a dependency; this is that dependency, kept separate so resolving a role cannot change what a verdict means on its way past.

## Everything about the agent is read, never composed

The name, the grant and the role all come out of the agent's own file in the pinned harness. A copy of any of them in this repo could drift from the file it copies — and a drifted tool grant is a reviewer that can rewrite the branch it is reviewing.

**Including which agent serves which gate, which I expected to have to hardcode and did not.** The agent files declare it:

```
agents/arch.md       gate: 1
agents/reviewer.md   gate: 3
agents/verifier.md   gate: 4
agents/scribe.md     gate: 0,4
agents/staff/*.md    no gate — the owners, chosen per repo
```

So the mapping is discovered, and a plugin that moves a role to another gate moves it here too. What stays this side is **sequence** — which of gate 4's two agents runs first is the engine's business and nobody else's, which is why `agentsForGate` sorts for determinism rather than for order, and says so.

An agent that is not in **this** harness is refused rather than searched for. Three copies of the plugin were reachable in one 18-hour window, and "find it somewhere" is how a gate runs a role nobody pinned.

## The prompt file

Print mode writes the body to disk before the command is built, because the builder refuses print mode without it — rightly: without the file the gate answers as a generic assistant while the run records the agent that never saw it, and a schema-valid verdict from nobody is worse than no verdict because nothing downstream can tell.

It is named for the **session**, not the role: two gates in one run can share a role, and a shared file would be rewritten under a gate still reading it. Background mode writes none — `--bg` resolves the agent itself through `--plugin-dir`, and a file nothing reads is a role left on disk for no reason.

## Tested against a real harness on disk

Because what this module does is *read one*. A fixture of hand-written strings in memory would be exactly the copy this module exists to avoid, one layer up. The fixture agents are shaped like the plugin's own, `gate:` lines included.

**Mutation-checked** — each fails a test: a substring counting as a gate (`gate: 14` answering for 4); an agent with no gate serving every gate; a background launch writing a prompt file; the file named for the role instead of the session; the staff folder not searched.

**One mutation is equivalent and stays untested**, rather than gaining a test that cannot fail: passing a schema in background mode changes nothing observable, because `buildGateCommand` reads `spec.schema` only inside the print branch, which returns first. I checked that by reading the consumer rather than assuming it.

18 new tests, 330 across the run engine, 1,873 in the suite (2 skip, 0 fail).


src/main/run-engine/gate-launcher.ts
```diff
@@ -0,0 +1,167 @@
...
+ * plugin were reachable in one 18-hour window, and "find it somewhere" is how
+ * a gate ends up running a role from a checkout nobody pinned.
+ */
+export async function loadAgent(
+  harnessPath: string,
+  name: string,
+): Promise<AgentDefinition & { path: string }> {
+  const tried: string[] = [];
+  for (const dir of AGENT_DIRS) {
+    const file = path.join(harnessPath, dir, `${name}.md`);
```
brannon-bowden: `name` is interpolated into a filesystem path with no traversal check (no rejection of `..` or path separators), undermining this function's own stated guarantee that only agents inside the pinned harness can be loaded.

src/main/run-engine/gate-launcher.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (launch.mode === 'print') {
+    await fs.mkdir(launch.promptFileDir, { recursive: true });
+    // Named for the session rather than the agent: two gates in one run can
+    // share a role, and a shared file would be rewritten under a gate that
+    // is still reading it.
+    systemPromptPath = path.join(
+      launch.promptFileDir,
+      `${launch.context.sessionId}-gate${launch.gate}.md`,
+    );
+    await fs.writeFile(systemPromptPath, agent.body, 'utf8');
```
brannon-bowden, line 194: The written system-prompt file is never removed after the gate runs (success or failure). Confirm cleanup happens at the call site, or these files will accumulate under promptFileDir.

src/main/run-engine/gate-launcher.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import type { Gate } from './transitions';
+import { parseAgentFile, type AgentDefinition } from './agent-definition';
+import { buildGateCommand, type GateMode, type RunSpawnContext } from './gate-command';
+import { runGate, type GateOutcome, type GateSpawnOptions } from './gate-process';
+import { GATE_VERDICT_SCHEMA } from './gate-verdict';
+
+export class GateLaunchError extends Error {}
```
brannon-bowden: GateLaunchError doesn't set this.name, so it will report as "Error" in logs/stack traces despite the custom class.

brannon-bowden (changes requested):
This PR is well-tested and the design intent is clearly documented, but it can't be approved yet: CI is red, and there's a defensive-coding gap in the new harness-file resolver that's worth closing before merge.

## Issues

- **CI is failing.** SonarQube Code Analysis and quality-gate are both red on this PR. Both need to pass before merge (GitGuardian and the Windows/Ubuntu test jobs are green).
- **`loadAgent` doesn't validate `name` before using it in a file path** (`src/main/run-engine/gate-launcher.ts:61`). `path.join(harnessPath, dir, \`${name}.md\`)` interpolates `name` (and `launchGate`'s `agentName`) directly into a filesystem path with no check for `..` or path separators. The whole point of this module, per its own doc comment, is that "an agent that is not in the pinned harness is not an agent this run may use" — but a `name` like `../../../etc/passwd%00` (or any traversal string) would resolve outside `AGENT_DIRS` and defeat that guarantee if it's ever attacker-influenced upstream. I can't see the call site that supplies `agentName` in this diff, so I can't confirm it's exploitable today, but given the stated security intent of this exact function, rejecting names containing `/`, `\`, or `..` before building the path would make the guarantee actually hold rather than relying on callers to always pass safe values.

## Suggestions (non-blocking)

- `launchGate` writes the system-prompt file (`gate-launcher.ts:149`) but nothing in this diff removes it afterward. If nothing downstream cleans these up, `promptFileDir` will accumulate a file per gate run indefinitely. If cleanup happens at a call site not shown here, disregard.
- `GateLaunchError` (`gate-launcher.ts:42`) doesn't set `this.name = 'GateLaunchError'`, so `error.name` will read `"Error"` in logs/stack traces even though `instanceof` checks work fine. Minor, but worth the one-line fix for anyone triaging logs later.
- Nice test coverage overall — the print/background/missing-agent/ordering cases are all exercised, and the "no cleanup, no spawn" ordering guarantee is explicitly tested.

William-Long-II:
All three fixed, and the security one turned into a lesson about my own test.

**The traversal guard.** You were right that the guarantee rested on callers — and names come from `team.yaml`, which a person writes, so "sensible" was a hope rather than a property. Two guards now: a name pattern, and a containment check that the resolved path is under the agents directory. The second is the one that holds. A pattern can be out-argued by an encoding nobody thought of; *"does this resolve under that directory"* is the question actually being asked.

**The test I wrote for it was too loose, and only the mutation run showed it.** It asserted `rejects.toBeInstanceOf(GateLaunchError)` — but a **missing file raises GateLaunchError too**, so it passed against a version with no guard at all. It asserts the reason now. With both guards removed four tests fail; with either one alone removed the other still holds, which is what defence in depth should look like and is why the single-mutation run looked clean at first.

**Non-blocking items, both taken.** The prompt file is removed once the gate has finished with it — the body is reproducible from the harness at any moment, so a copy per run is a growing pile of duplicates of a file that already exists, and one holding a role a later harness may have changed. A run interrupted before that leaves its file behind, which is why the directory is the caller's: it can be swept on start. And `GateLaunchError` sets its own `name`.

That cleanup also exposed two tests asserting the file existed *after* the run — which the cleanup made unobservable. `promptFilePath` is split out so naming is asserted on the pure function and cleanup on the directory being empty, rather than one test that quietly checked neither.

Sonar's `localeCompare` is in too. 1,877 tests, 0 fail.

brannon-bowden (approved):
Both blockers from the previous review are resolved, and the remaining suggestions were also addressed — this is good to merge.

- **CI is now green.** All five checks (SonarQube, both test jobs, quality-gate, GitGuardian) are passing.
- **The path-traversal gap is closed.** `agentFile()` (`gate-launcher.ts:64-80`) now validates `name` against `SAFE_NAME` (`^[A-Za-z0-9][A-Za-z0-9._-]*$`), explicitly rejects any `..` substring (catching cases like `a..` that the charset alone wouldn't stop), and then double-checks the resolved path is actually contained under the harness directory via `path.resolve`/`startsWith`. This is applied uniformly through `loadAgent`, so it covers `launchGate`'s `agentName` too — confirmed by the new `'a launch with a traversal name spawns nothing'` test, which exercises the full call path and asserts no prompt file is written. The test suite is thorough here: it covers `../` traversal, backslash separators, a traversal that resolves to a file that actually exists (the sharpest case), and an empty name, plus a positive control asserting real role names still load.
- **The two non-blocking notes were also fixed:** `GateLaunchError` now sets `name = 'GateLaunchError'` (`gate-launcher.ts:42-45`), and `launchGate` now removes the prompt file in a `finally` block after the gate runs (`gate-launcher.ts:211-221`), addressing the unbounded-accumulation concern. Both are backed by tests (`'print mode leaves no role behind once the gate is done with it'`, the promptFilePath uniqueness test).

## Suggestions (non-blocking)

- `agentsForGate`'s sort was changed from the default `.sort()` to `.sort((a, b) => a.localeCompare(b))` (`gate-launcher.ts:148`). This is unrelated to the traversal fix and makes ordering locale-dependent (ICU collation can vary by runtime/OS locale), whereas the previous default sort was a deterministic UTF-16 code-unit comparison. Worth reverting unless there's a specific reason for locale-aware ordering here.

PR #276, opened 2026-09-13
feat(run-engine): advance a run durably, so it survives the process

William-Long-II:
CO-722. Every other module in this engine is a half: `transitions` decides and touches nothing, `executor` acts and decides nothing, the repository writes and knows about neither. The driver is where they meet, and **the order they meet in is the design**.

## Persist before acting

The transition is written first, in one transaction with the event that caused it, and only then are the actions performed. The actions here start agents and ask people for things, so *"did that already happen?"* is otherwise a question somebody answers by looking at GitHub.

The cost is that an action can fail after its transition is recorded, leaving a run whose state says more than the world does. That is the lesser problem, and it is already handled: the executor reports rather than assumes, so nothing false is written, and the run sits in a state the loop re-attends. **A run a minute stale beats a run whose history has a hole in it.**

The same rule applies to gates: a `run_gates` row is opened *before* the spawn, so a gate that starts and dies leaves something to see. Ordering it the other way looks identical on every happy path — which is why the test asserts a crash rather than a success.

## The loop is bounded

The machine has cycles by design: gate 3 sends a run back to gate 2, which comes forward again. One that does not settle would spin for as long as the process lives. Eight rounds is a **tripwire, not a tuning knob** — hitting it is reported as a fault, and a control test asserts an ordinary advance does not come near it.

## Three things the tests found, all mine

- **"Record only a real state change" was wrong.** A gate-2 pass leaves a run in `running` and spawns gate 3 — no move, and the most important line in the log. It records when something *happened* now (a state change **or** an action); an event that does neither still writes nothing.
- **Closing a gate row when its report arrived** cleared `activeGate` before the guard read it, so the machine rejected the very report that was closing it: a gate finishes, its verdict is discarded, and the run sits in `running` forever. Closed on *acceptance* now.
- **A gate with no recorded role was left unrecorded but still spawned**, which the docstring claimed it was not. A gate that runs with no row is worse than one that does not run: it spends tokens, changes a worktree, and leaves nothing saying it happened. It is dropped now — and the test that found this had asserted the wrong behaviour first, which is how the gap surfaced.

## A fixture that could drift

`repositories/__tests__/runs.test.ts` declared the four tables by hand, so a column added in `database.ts` and missed there produced tests passing against a schema nobody ships. The SQL now lives in one module that both the fixture and `initializeRunTables` read.

**Mutation-checked** — each fails a test: gate rows opened after the spawn; rows never closed; progress without a state change left unlogged; the state not re-read each round; a cycle not reported; a gate with no role spawned anyway.

19 new tests, 360 across the run engine, 1,896 in the suite (2 skip, 0 fail).


brannon-bowden (changes requested):
This is a solid piece of work — the SQL-schema extraction removes a real drift risk between the production schema and the test fixture, and the driver's persist-before-act ordering is well tested for the crash/regression scenarios it's meant to guard against. There's one concrete bug in the round-limit/cycle-detection logic that should be fixed before merge.

## Issues

- **`src/main/run-engine/driver.ts:122-174` — the round-cap loop always reports `runaway` when it exits via exhausting `MAX_ROUNDS`, even if the run actually settled on the last permitted round.** The only place `pending.length === 0` is checked is at the top of each `for` iteration (line 123). If processing during the final iteration (`round === MAX_ROUNDS - 1`) produces an empty `next` array — i.e., the run genuinely finished, it just took exactly `MAX_ROUNDS` decide-act rounds to do so — the loop exits by falling off the `for` condition rather than via the early `return result` on line 123, and control falls straight into the unconditional `result.runaway = ...` / `result.problems.push(...)` at lines 171-173. A run that completed correctly would be misreported as stuck in a cycle. Fix: after the loop (or on final assignment), only set `runaway` when `pending.length > 0`, e.g. check `pending.length === 0` once more before declaring the fault. The existing `MAX_ROUNDS` test only exercises the ordinary 1-2 round case, not this exact-boundary case, so it wouldn't have caught this.

## Suggestions (non-blocking)

- `advance()` reads `runs.getRun(runId)` and `runs.activeGate(runId)` as two separate, non-transactional queries per event (driver.ts:127-133), and the loop `await`s `execute(...)` in between rounds. If `advance()` can ever be invoked concurrently for the same run (e.g., a reconcile pass and a gate callback landing close together), there's a window for one call to decide against a state the other is about to change. Worth confirming that callers serialize `advance()` per run; if they don't already, this is worth a lock/transaction around the read-decide-write portion.
- `startGate`'s attempt-count is a read-then-insert (`runs.ts` new `startGate`) rather than a single atomic statement; fine under the current single-threaded/synchronous SQLite access pattern, but worth keeping in mind if this ever moves to a connection pool or multi-process writer.

William-Long-II:
Fixed — and the boundary you identified is exactly right: the empty-queue check only ran at the *top* of a round, so a run that settled on the last permitted round fell off the `for` condition and got a fault for finishing correctly the long way.

**Both non-blocking notes taken as well.**

`advance()` is serialised per run. Worth stating precisely what that does and does not buy, because I checked rather than assumed: reading, deciding and writing are *synchronous* within a round, so a single advance was already atomic. What the queue adds is that a **multi-round** advance is atomic too — and a reconcile pass and a gate reporting back do arrive independently. Per run id rather than global, with a control test asserting a slow run does not hold up a different one.

`startGate` is one statement now. Read-then-insert let two spawns racing for the same gate both read 1 and both record attempt 2, which quietly undoes the thing the attempt number exists for.

**Three of my tests in this change could not fail as written, and the mutation run caught all three:**

| test | why it could not fail |
|---|---|
| the boundary test | settled a round early, so the loop returned from the check at the top and never reached the boundary |
| the serialisation test | asserted a timing overlap, which passes with no queue at all whenever the second call decides nothing |
| the cycle fake | juggled `run_gates` itself and broke the chain at round one |

The serialisation one is the instructive one: I wrote it twice before noticing the property I was asserting was not the property the code provides. It now asserts what is actually observable — the second call does not begin until the first returns.

1,900 tests, 0 fail.

brannon-bowden (approved):
Approve — the round-limit bug from the previous review is fixed correctly, and the two concurrency suggestions from that review have both been implemented well.

Checking the fix: in `driver.ts:198-207`, the `runaway` fault is now only recorded `if (pending.length > 0)` after the loop exits, so a run that empties its pending queue on exactly the last permitted round is no longer misreported as a cycle. The new test (`driver.test.ts`, "a run that settles on the last permitted round is not a cycle") exercises exactly the boundary the old code got wrong — spawning until `spawns >= MAX_ROUNDS` before returning a no-event result — which is the right way to pin this down.

The two non-blocking suggestions from last time are also addressed:
- `advance()` is now serialized per run id via an `inFlight` promise-chain queue (`driver.ts:106-138`), closing the read-decide-write race between a reconcile pass and a gate callback. The "one run advances at a time" tests confirm ordering is preserved within a run and that separate runs don't block each other, and that a thrown error doesn't wedge the queue.
- `startGate` (`runs.ts:351+`) is now a single `INSERT ... SELECT` statement instead of a separate `SELECT COUNT` followed by `INSERT`, closing the window where two racing gate starts could both read the same prior count and claim the same attempt number.

## Issues

- `driver.ts:132` — `inFlight` is a module-level `Map<string, Promise<void>>` keyed by `runId`, and entries are never deleted after a run's queue settles. Over a long-lived process handling many distinct runs, this grows without bound (a small leaked promise per run id, forever). Not urgent given the size of each entry, but worth a cleanup path (e.g., delete the map entry once the stored promise for that key settles and no newer entry has replaced it) if run volume is expected to be high over the process lifetime.
- No test accompanies the `startGate` change verifying the new atomic attempt-numbering under concurrent/interleaved calls for the same `run_id`/`gate` (e.g., two `startGate` calls for the same gate yielding attempts 1 and 2, not both 1). The single-statement approach is sound, but a regression test would guard against a future refactor reintroducing the read-then-write split.

## Suggestions (non-blocking)

- The `inFlight`-map leak above could be resolved cheaply by having the stored promise remove its own map entry on settle, guarded by an identity check so a newer queued entry for the same run isn't accidentally deleted.

PR #277, opened 2026-09-13
feat(run-engine): start a run, and refuse to

William-Long-II:
CO-722. Every other module in this engine assumes a run exists. This is the one that makes one — and almost all of it is about **not** making one.

## Half a working orchestrator is worse than none

The design says it and it is right. A run that starts without a Python it can drive does not fail at the start, where somebody would understand it. It fails four gates in, having cut worktrees and spent tokens, with a message about whatever happened to break first.

So everything is checked before anything is written, **every check runs even after one fails** — a person fixing a machine wants the whole list, not three round trips for one problem — and each refusal names both what is missing and what to do about it.

**Python is checked by running it.** That is the plugin's own rule and not a general nicety: on Windows `python3` resolves to a WindowsApps alias that exits non-zero with a Store advertisement. It is on PATH, it is not Python, and a check that asked `command -v` passed on it.

When a run is refused, **nothing is written**. A half-written run is worse than none: it appears in every list, and the next thing to look at it has no idea it never passed a check.

## Who owns gate 2 is discovered, never guessed

Staff agents declare `repo:` in their front matter exactly as gate agents declare `gate:`, so the engine holds no list of who owns what.

| candidates | answer |
|---|---|
| one | decided |
| several | **a question** — named for this run, or refused |
| none | refused |

Several is a question rather than a coin toss because a lead owns what cuts across a repo while domain owners own their own modules, and picking wrong is an agent editing a module that is not theirs. Naming one answers it — and a name that does *not* own the repo is still refused, because naming is a choice between candidates, not a way past the check.

## Armed, not started

A created run has its rows, its owners, and nothing else: `preparing`, no events, no gates. Starting cuts worktrees and launches an agent that writes — the first genuinely irreversible thing this system does — and the design says phase 3 ships dark until one real run has passed. There is a test asserting the absence.

## Also

`gate-launcher`'s front-matter reading collapses into one reader. The regex was about to exist twice, and the second copy was wrong before it was ever committed — the ordinary fate of a duplicated pattern, caught this time by the compiler rather than by a run.

**Mutation-checked** — each fails a test: python never probed; refusals not stopping the write; the first of several owners picked; a named owner unchecked; the initiative's exit code ignored.

13 new tests, 1,913 in the suite (2 skip, 0 fail).

Pairs with `claude-team-workflow` #310, merged.


src/main/run-engine/ignition.ts
```diff
@@ -0,0 +1,230 @@
...
+    request.initiativePath,
+  ]);
+  const payload = parse<InitiativePayload>(read.stdout);
+  if (read.code !== 0 || !payload) {
+    refusals.push({
+      what: `the initiative at ${request.initiativePath} could not be read`,
+      // The plugin's own sentence: it knows whether this is a directory
+      // nobody started or one whose owners spawn.sh has not filled in yet,
+      // and those have different answers.
+      fix: payload?.detail ?? read.stderr.trim() ?? 'Check the path.',
```
brannon-bowden: `read.stderr.trim() ?? 'Check the path.'` — trim() never returns null/undefined, so this fallback is dead code and `fix` can end up empty when there's no `detail` and empty stderr. Use `||` (or an explicit non-empty check) instead of `??`.

brannon-bowden (changes requested):
This is a well-tested, thoughtfully documented change, but there's a concrete logic bug in the refusal-message construction that should be fixed before merge.

## Issues

- **`src/main/run-engine/ignition.ts:189`** — `fix: payload?.detail ?? read.stderr.trim() ?? 'Check the path.'` uses `??` after `.trim()`. `String.prototype.trim()` always returns a string (never `null`/`undefined`), so when `payload` has no `detail` and `stderr` is empty, this evaluates to `''` — the `'Check the path.'` fallback is unreachable dead code, and the refusal ships with a blank `fix` field. This is exactly the scenario the module's own doc comment promises to avoid ("each one says what to do about it"). It's reachable whenever `initiative.py` exits non-zero (or prints unparseable JSON) without writing to stderr — e.g. an uncaught Python exception whose traceback goes to stdout, or a silent non-zero exit. Fix: use `||` or an explicit `read.stderr.trim() || 'Check the path.'` / length check instead of `??`. No test currently covers this empty-stderr path.

## Suggestions (non-blocking)

- The `armed` result's `owners` map (repo → chosen agent name, e.g. `{ 'demo-repo': 'demo-lead' }`) is returned to the caller but the `runs.upsertOwner(...)` calls in `armRun` only persist `worktree`/`branch`/`base`/`scratch` — the resolved agent name doesn't appear to be written to the owner row. Worth confirming that this assignment (especially for repos where a human had to disambiguate between multiple candidates) survives somewhere for when the run actually starts gate 2 — otherwise a repo with multiple candidate owners may need to be re-resolved (or re-asked) later.
- `armRun` still shells out to run `initiative.py` via `request.pythonPath` even when `pythonWorks()` has already determined that interpreter doesn't run. If the real `deps.run` implementation can throw (e.g. `ENOENT` for a truly missing executable) rather than resolving with a non-zero `CommandResult`, that exception would propagate out of `armRun` unhandled, breaking the "always refuse cleanly" contract this module advertises. Can't confirm from this diff since `run`'s implementation isn't shown — worth double-checking.
- `createRun` followed by a loop of `upsertOwner` calls are separate statements; if the process dies mid-loop you could end up with a run row and only some owner rows. Consider whether these should be wrapped in a single transaction (may already be handled inside the repository layer, not visible here).
- `resolveOwners` calls `agentsForRepo`, which itself calls `eachAgent` (a full harness directory + file re-read) once per repo. For harnesses with several repos this re-reads and re-parses every agent file N times; caching `eachAgent`'s result once per `armRun` call would avoid the redundant I/O.
- The field name `owners` is used for two different shapes in this module: `InitiativePayload.owners` (repo → worktree/branch/base/scratch) and `IgnitionResult.owners` / `resolveOwners`'s return (repo → agent name). Both are legitimately called "owners" by the domain, but the overload may trip up a future reader scanning for `owners` usages.

William-Long-II:
All three addressed, and your second suggestion was a real gap rather than a question.

**The blocking one.** You are right that `'Check the path.'` was unreachable: `.trim()` returns a string, `??` passes `''` straight through, and the refusal shipped with a blank `fix` — precisely what the module docstring promises not to do. Reachable whenever `initiative.py` exits non-zero without writing to stderr, which an uncaught traceback does.

This is the **second time** this exact mistake has been made in this engine and caught in review (the first was `gate-process`, also by you). So the fix is a shared function rather than a shared intention: `firstText()` states the rule once — *emptiness disqualifies a candidate, not nullness* — and `gate-process` now uses it instead of its private copy. Worth writing down why neither operator alone is right: `??` passes `''` through, and `||` reads as a defect to any linter seeing a nullable left operand (S6606), so it invites being "fixed" back into the first case by somebody who was not here.

**The owner-role gap was real.** It was returned and not stored. Where a person chose between a lead and a domain owner, that choice is the run's and has to survive the process — otherwise gate 2 asks again and the second answer need not match the first. `run_owners` gains an `agent` column with the ALTER-ladder entry `database.ts` already establishes, and the upsert `COALESCE`s it like the PR columns beside it, because `spawn.sh` is idempotent and a re-mirror that does not carry the role must not erase it.

**On the third:** `command-runner` never rejects — that was verified when it was written, and its tests assert every failure path (missing binary, timeout, abort, signal) returns a `CommandResult` rather than throwing. The contract holds for the real dependency.

Mutation-checked: the `??` chain restored, the role not persisted, and the `COALESCE` dropped each fail a test. 1,917 in the suite, 0 fail.

brannon-bowden (approved):
The previously blocking bug is fixed and the owner-persistence gap I flagged as a suggestion is now implemented, so this is ready to merge.

- **The `fix` fallback bug is fixed.** `ignition.ts` now uses the new shared `firstText(payload?.detail, read.stderr) ?? 'Check the path.'` helper (`src/main/run-engine/first-text.ts`), which trims each candidate and treats an empty/whitespace-only string as "no candidate," so the `'Check the path.'` fallback is reachable again. `gate-process.ts`'s near-identical `firstDetail` was deleted and replaced with the same shared helper, removing the duplicated logic that let this bug recur in the first place. Two new tests (`ignition.test.ts:180-201`) cover exactly the previously-broken path (empty stderr → fallback text) and the "prefers real output when present" control case.
- **The resolved `agent` is now persisted, not just returned.** `run_owners` gains an `agent` column (`run-tables-sql.ts`, plus a `PRAGMA table_info` migration in `database.ts` for pre-existing databases), `upsertOwner`/`listOwners` read and write it, and `armRun` writes `resolved.owners[repo]` into it. The `COALESCE(excluded.agent, run_owners.agent)` upsert clause correctly preserves a previously-chosen agent across a re-mirror that doesn't carry the value, matching the existing pattern used for `pr_number`/`pr_url`. Tests confirm both persistence (`ignition.test.ts:249-260`) and non-erasure on re-mirror (`ignition.test.ts:262-270`).
- The migration is safe for fresh installs: `RUN_TABLES_SQL` already declares the `agent` column, so `CREATE TABLE IF NOT EXISTS` produces a table that already has it, and the follow-up `PRAGMA table_info` check correctly finds it present and skips the `ALTER TABLE`. For pre-existing databases the column is absent and gets added.

I didn't find anything new to block on in this diff.

## Suggestions (non-blocking)

- The remaining suggestions from the previous review (transaction boundaries around `createRun`/`upsertOwner`, `eachAgent` re-read cost in `resolveOwners`, the `owners` naming overload between `InitiativePayload.owners` and `IgnitionResult.owners`, and whether `deps.run` can throw for a missing interpreter) apply to code this diff doesn't touch, so I'm not re-raising them as part of this pass — worth revisiting separately if still relevant.

PR #278, opened 2026-09-13
feat(run-engine): what needs a person, and why

William-Long-II:
The first half of the run inbox — the query, and a read-only channel the renderer can reach it through. The view follows in its own PR.

## What is in it

The states come from the state machine's own `NEEDS_A_PERSON`, not a list repeated in the query. A state added there and missed here is a run **nobody is ever told about** — the exact failure an inbox exists to prevent, arriving through the inbox. The test iterates that constant rather than naming states, so adding one there without handling it here fails.

**Oldest wait first.** The run that has waited longest is the one most likely to have been forgotten, and a newest-first inbox buries it exactly as it becomes urgent.

**A run merely waiting carries no reason, and that is not a gap.** A review in progress has not gone wrong. Inventing a sentence for it would make the column meaningless where it does matter — the blocked runs, where `blockedReason` is the thing the type system already forces to exist.

## What is deliberately absent

| | why |
|---|---|
| `preparing`, `provisioning`, `running`, `waitingChecks` | the engine is working; nobody is needed |
| `reviewNotRequested` | a state the engine **acts** on — a person seeing it would be told to do something already under way |
| `approved`, `done`, `failed` | finished |

An inbox that lists everything is a list, and a list is what the run table already is.

## Read-only, on purpose

`db:runs:inbox` is the only channel, and there is no companion that starts, stops or advances a run. That is what lets somebody watch the first real run without being able to set one off from a window while the engine ships dark.

`RunInboxRow` is declared in `shared/types` rather than shared by reference with the repository's `InboxRow`. The duplication is deliberate: the shape a window is allowed to see is a decision, and a type shared by reference makes every column added in main visible in a window by default.

**Mutation-checked** — each fails a test: newest-wait-first; every state included; the repos dropped; a reason invented for a run that merely waits.

9 new tests, 1,926 in the suite (2 skip, 0 fail).


brannon-bowden (approved):
This is a clean, well-scoped, read-only addition — the inbox query is parameterized (no injection risk), the new IPC channel does nothing but read, and the test suite is genuinely useful (it asserts inbox membership is driven by `NEEDS_A_PERSON` itself rather than a re-derived list, sorting by oldest-first, blocked-reason nullability, and repo attribution). Approving.

## Suggestions (non-blocking)

- `listInbox()` in `src/main/repositories/runs.ts:497-511` issues one `listOwners(row.id)` query per row after the main `SELECT`, so building the inbox is O(N) queries for N pending runs. Fine at the scale this is likely used at today, but worth a join or a single batched lookup if the inbox is ever expected to grow large.
- Minor formatting nit: the new `InboxRow` doc comment in `src/main/repositories/runs.ts:461` and the new `RunInboxRow` interface in `src/shared/types.ts:669` both land directly against the closing brace of the preceding declaration with no blank line, unlike the spacing used elsewhere in these files.

William-Long-II:
Both taken.

**The N+1 is gone** — one query for the runs, one for their owners, grouped by run id. You are right that it is fine at today's scale; what decided it was that `listOwners()` per row is fine at five and silly at fifty, and nobody notices the moment in between.

The grouping is the part that deserved a test, and it now has one: grouped wrongly, **every row carries the same repos** — which reads as a bug in the engine rather than in a query, because the names would be real ones belonging to a real run. A person would go looking in the wrong place. The mutation that ungroups them fails two tests.

Blank lines fixed too. 1,925 in the suite, 0 fail.

PR #279, opened 2026-09-14
feat(runs): the inbox, as a view you can open

William-Long-II:
The surface that answers the question a person actually has — *is anything waiting on me?* — and the answer it gives most of the time should be **no**.

## The failure that matters is a wrong "no"

An empty inbox is the good state and says so. But a load that fails and renders empty tells somebody everything is fine at exactly the moment it is not, so a failure shows the error, and the first paint does not claim an answer before it has one.

| state | what it says |
|---|---|
| loaded, empty | *Nothing is waiting on you* |
| loading | *Reading the inbox…* |
| failed | the error, with a retry |

Three tests, one per way of saying "no" wrongly.

## `waitingReview` leaves the inbox

It still belongs in `NEEDS_A_PERSON` — the machine is right that a run cannot advance until a reviewer answers. But that person is a reviewer working from a queue that already exists, not an operator at this window, and an inbox listing things you cannot act on from where you are standing stops being read.

The list is **derived** — `NEEDS_A_PERSON` minus named exclusions — rather than restated. A state added to the machine and missed here is a run nobody is ever told about, so a new state appears by default and leaving one out has to be a deliberate line with a reason beside it. There is a test asserting exactly that, and a control asserting the exclusion list has not quietly grown.

## What a row says

A blocked run shows **its own sentence**, because that was composed where the run stopped by whatever knew why; the state description is the fallback, never the other way round. Waits are rounded to units a person thinks in — "43 seconds" invites watching something that changes in minutes — and a timestamp that will not parse says `unknown` rather than rendering `NaNm`, which reads as a number.

Read-only throughout, matching the channel behind it: no control that starts, stops or advances a run, and a test that counts the buttons.

## The menu entry is wired, deliberately

The design said ship dark until one real run has passed. Will's call is that the beta channel is the vehicle for that test, so it is reachable — `Runs`, `Mod+4` — to be looked at rather than waiting to be remembered. Recording the override rather than quietly diverging.

Two `ViewSwitcher` tests named `arena` as the last tab, so a fourth destination broke them — which is not a wrapping regression. They derive the last tab now.

**Mutation-checked** — each fails a test: a failed load rendering empty; empty before loading; seconds shown as `0m`; an unparseable date rendering `NaN`; the run's own reason dropped; `waitingReview` back in the inbox.

19 new component tests, 1,948 in the suite (2 skip, 0 fail).


brannon-bowden (approved):
This PR is well-scoped and well-tested; approving with a few non-blocking notes.

The feature adds a read-only "Runs" inbox view (menu entry, keyboard shortcut, IPC plumbing, and a new `RunInbox` component), plus a `NOT_THE_OPERATOR`/`INBOX_STATES` derivation in `runs.ts` that excludes `waitingReview` from the inbox query. The exclusion is correctly derived from the existing `NEEDS_A_PERSON` list rather than duplicated, the SQL still uses parameterized placeholders (no injection risk), and the test suite (`runs.test.ts`, `RunInbox.test.tsx`) exercises the exclusion, ordering, empty/loading/error states, and the reason/wait-time formatting logic thoroughly, including a "control" test that pins `NOT_THE_OPERATOR` to its expected single entry so a silent expansion gets caught. CI is fully green (SonarQube, both test matrices, quality gate, GitGuardian).

## Issues
None blocking.

## Suggestions
- `src/renderer/components/RunInbox.tsx:85` — the render logic checks `failed !== null` before checking `rows`, so a transient poll failure (e.g. the DB briefly locked) after a successful initial load replaces the previously-shown list with a full-page error rather than keeping the stale list with a lighter inline indicator. This may be intentional given the "shown, not swallowed" philosophy in the comments, but it's worth confirming that's the desired UX for periodic background refresh failures, not just the first load.
- `src/renderer/components/RunInbox.tsx:76-83` — the 60-second polling interval (and its cleanup via `clearInterval`) isn't covered by a test; consider a fake-timers test asserting `load` is called again after the interval fires and is not called after unmount.
- `src/renderer/components/RunInbox.tsx:66` — `window.electronAPI.getRunInbox` is invoked detached from its object (`(load ?? window.electronAPI.getRunInbox)()`); harmless if the preload implementation doesn't rely on `this`, but worth a quick check since it's an easy thing to break silently if that function is ever refactored to a class method.

William-Long-II:
All three taken, and you were right to ask whether the first was intended — it was not.

**The refresh failure was a real regression I had not thought through.** The failure check ran before the rows check, so a transient poll failure (the database briefly locked) replaced three waiting runs with a full-page error. That loses the answer to keep the warning, which is the wrong way round.

Now: a failure with **nothing to fall back on** is still the whole page, because an empty list would be the wrong answer and there is nothing else to show. A failure **over a list we already have** is a note beside it — the list is a minute stale, and the note says so. Both directions have a test, and they are each other's control.

**The poll has tests now, both halves.** It polls, because nothing tells this window a run stopped; and it stops on unmount, because an interval left running is a query a minute for a window nobody is looking at, forever. The interval is a prop so the test does not wait a minute to find out.

**On the detached call:** checked — `getRunInbox` is an arrow function in the preload object literal and does not use `this`, so it was harmless today. Called through the object now anyway, because "harmless today" is exactly the kind of thing that breaks silently later.

Mutation-checked: the refresh failure blanking the list, the interval never cleared, and never polling each fail a test. 1,952 in the suite, 0 fail.

PR #280, opened 2026-09-14
feat(tools): drive a run by hand, one step at a time

William-Long-II:
Nothing in this engine has ever spawned a real gate. 1,952 tests and a complete circuit — and every assumption it makes about paths, flags and which copy of the harness is on disk is still unverified, because **a test that supplied those would be supplying the thing in doubt.**

So: a console, and deliberately not a button. It arms a run and stops. Each step afterwards is a separate invocation, and the event log can be read between any two of them — the first failures are likelier to be environmental than logical, and those are read from a terminal rather than a window.

```
bun run console -- arm    <initiative-dir> [--owner repo=agent]
bun run console -- status
bun run console -- events <run-id>
bun run console -- step   <run-id> <event>
```

## It runs inside Electron, and that is not a preference

`better-sqlite3` is built for Electron's ABI by `postinstall`, so neither bun nor plain node can load it. Running here means the console drives **the same database module, the same native driver and the same migrations the app does** — a console that swapped any of those would be testing something other than what ships.

Two other routes were tried first and are worth recording so nobody repeats them: bun cannot load the driver at all, and bun's runtime `onResolve` does not intercept relative specifiers, so `database.ts` could not be redirected to a `bun:sqlite` handle from a preload.

## It writes to a scratch store

`app.setPath('userData', …)` runs **before anything opens a database** — `getDatabase` resolves its path on first call and caches the handle. A first run that goes sideways cannot leave rows in the file the app reads on start.

## One production change

`database.ts` requires `electron` lazily. `app` is used on exactly one line, inside a function, and an import that one line needs should not decide who may load the module. Inside the app it resolves exactly as before; outside it, a tool can now load the module at all.

## Proven on this machine, which is the whole point

```
$ bun run console -- arm C:/work/repos/nothing-here
store     C:\work\repos\Bodhilander\.run-console
REFUSED. Nothing was written.
  - the initiative at C:/work/repos/nothing-here could not be read
    fix: no team.yaml at …\team.yaml. init-task.sh writes it; without it
         there is no record of where an owner's worktree is.
```

That sentence came from the plugin's own `initiative.py`, through real Python, on real disk — the first time this engine and that harness have spoken to each other outside a fixture. With `BODHI_PYTHON` pointed at nothing, **both** refusals arrive together rather than one per attempt, which is the behaviour `armRun` was written for and had only ever been asserted against a fake.

1,952 tests unchanged, 0 fail.


scripts/run-console.ts
```diff
@@ -0,0 +1,247 @@
...
+    return;
+  }
+
+  console.log(`ARMED     ${result.runId}`);
+  console.log(`initiative ${result.initiativeKey}`);
+  for (const [repo, agent] of Object.entries(result.owners)) {
+    console.log(`  ${repo} -> ${agent}`);
+  }
+  console.log('\nNothing has been cut, launched or pushed. To move it:');
+  console.log(`  bun scripts/run-console.ts step ${result.runId} prepared`);
```
brannon-bowden: This printed next-step command runs the raw TS file via bun instead of through `bun run console --`, which bypasses Electron. Per this file's own header, `app` from `require('electron')` outside Electron is just a path string, so `app.setPath` will throw. Update the hint to go through the `console` npm script.

scripts/run-console.ts
```diff
@@ -0,0 +1,247 @@
...
+  const kind = process.argv[4];
+  if (!runId || !kind) throw new Error('usage: step <run-id> <event>');
+
+  const run = runs.getRun(runId);
+  if (!run) throw new Error(`no run ${runId}`);
+  const owners = runs.listOwners(runId);
+  const agents: Record<number, string> = {};
+  // Gate 2's role is the owner's, recorded when the run was armed. Gates 3
+  // and 4 are read from the harness, so they are not named here.
+  for (const owner of owners) if (owner.agent) agents[2] = owner.agent;
```
brannon-bowden: This always writes to `agents[2]`, so with multiple owners only the last one with an `.agent` wins silently for gate 2. Confirm this is intended for the single-relevant-owner case, or scope the assignment per owner.

brannon-bowden (changes requested):
This PR adds a well-documented manual run-console tool plus a small, sensible change to lazily require `electron` in `database.ts`, but the console's own printed next-step command will crash if followed literally, so changes are needed before merge.

## Issues

- **`scripts/run-console.ts:125`** — After a successful `arm`, the console prints `bun scripts/run-console.ts step ${result.runId} prepared` as the next command to run. This bypasses the `console` npm script (`bun run build:console && electron dist/run-console.js`) and invokes the raw TS file with `bun` directly. The file's own header explains why this can't work: `import { app } from 'electron'` at the top of this file resolves, outside an actual Electron process, to the path string the `electron` npm package exports rather than the app API — so `app.setPath(...)` in `useScratchStore()` throws immediately (`app.setPath is not a function`). This contradicts the documented "it runs inside Electron, and that is the point" design and will break the exact interactive hand-off this tool exists to support. The hint should read something like `bun run console -- step ${result.runId} prepared`.
- **`scripts/run-console.ts:165`** — `for (const owner of owners) if (owner.agent) agents[2] = owner.agent;` assigns every owner with an `agent` to the same key (`2`), so with more than one `--owner repo=agent` pair, only the last one silently wins for gate 2. If a run can legitimately have multiple owners with independently meaningful agents, this needs to select/scope the right owner rather than last-write-wins; if it's intentional (single relevant owner per run today), it'd help to guard or comment why collapsing is safe.

## Suggestions (non-blocking)

- `scripts/run-console.ts:170` — the default approvers fallback (`'brannon-bowden,William-Long-II'`) hardcodes specific personal GitHub usernames into shared script code. Consider requiring `BODHI_APPROVERS` to be set explicitly (or documenting the default prominently) so other engineers running this don't silently get these two names.
- `scripts/run-console.ts:51-54` — the `flag()` helper is defined but doesn't appear to be called anywhere in this file (owner parsing uses its own inline loop in `chosenOwners()`); consider removing it if unused.
- CI is fully green (SonarQube, both `test` matrix legs, quality-gate, GitGuardian) — no gating concerns there.

William-Long-II:
Both fixed, and the first was the tool contradicting its own header — a hand-off that crashes if followed is worse than none. It prints `bun run console -- step ...` now.

One correction to the mechanism, since I checked rather than taking it on faith: the raw-file route fails with a **SyntaxError from the import**, not the `app.setPath is not a function` TypeError —

```
$ bun scripts/run-console.ts status
SyntaxError: Export named 'app' not found in module '…\node_modules\electron\index.js'.
```

bun resolves `electron` to the package entry, which exports a path string and no `app`, so it dies before any of the file runs. Your finding is unaffected; the fix is the same.

**On multiple owners — it refuses now rather than collapsing.** You asked whether it was intentional, and the honest answer is that the limitation is the *engine's*, not the console's: `ExecutorTarget.agents` is keyed by **gate**, so it holds exactly one role for gate 2. That is sound for slice one, where the design scopes a run to a single repo, and wrong the moment a run has two owners. Quietly keeping the last would run one repo's gate under another repo's role — so it names the repos and stops. Papering over it here would be the console hiding the thing a first run exists to find.

**And the approvers default is gone.** You are right that baking two real usernames into a tool makes it request review from people who did not ask for it. It reads `BODHI_APPROVERS` or records none, and the engine already refuses clearly in that case — a tested path.

1,952 tests unchanged.

brannon-bowden (approved):
Both previously blocking issues are resolved by this update, and CI is green, so this is good to merge.

- **Next-step hint (`run-console.ts:125` area)** — now prints `bun run console -- step ${result.runId} prepared`, going through the npm script instead of invoking the raw TS file with `bun` directly. This avoids the `app.setPath is not a function` crash from the `electron` import resolving to a path string outside an Electron process. Resolved.
- **Multi-owner gate-2 role collapse (`run-console.ts:163-181` area)** — the silent last-write-wins loop is replaced with an explicit check: if a run has more than one owner, `step()` now throws with a clear message instead of quietly picking the last owner's agent for gate 2. For the single-owner case, `owners[0]?.agent` is used directly. This turns a silent correctness bug into a loud, actionable failure, which is the right tradeoff until multi-owner runs are actually supported. Resolved.
- **Hardcoded approver default** — `BODHI_APPROVERS` no longer defaults to specific personal GitHub usernames; it now defaults to an empty list (`env('BODHI_APPROVERS', '').split(',').filter(Boolean)`). This addresses the earlier non-blocking note about baking real logins into shared script code.

No new issues stand out in this delta. CI is fully green (SonarQube, both `test` matrix legs, quality-gate, GitGuardian).

PR #281, opened 2026-09-14
fix(run-engine): a repo claim is a list and may hold globs

William-Long-II:
Found on **first contact with the real harness**, which is what first contact is for. This is the first defect in the engine that no test could have caught, because every fixture supplied the shape the code expected.

## What the harness actually declares

```
repo: bodhi-service-api                                  one name
repo: bodhi-service-api, bodhi-service-notify-v2, …      a list
repo: bodhi-provider-*, bodhi-service-connector*         globs
```

`agentsForRepo` compared the whole value as one exact string. So:

| repo | before | after |
|---|---|---|
| `bodhi-service-api` | bsa-care, bsa-lead | bsa-care, bsa-lead, **bsa-platform** |
| `bodhi-service-insights` | **(none)** | bsa-platform |
| `bodhi-provider-demo-service` | **(none)** | integrations |
| `Bodhilander` | (none) | (none) |

A repo reporting no owner refuses to start, which is loud. **The sharper half is the other direction:** `bsa-platform` was missing from `bodhi-service-api`'s candidates, and fewer candidates turns *"several — ask a person"* into *"one — decided"*. A run could pick an owner nobody chose, silently, with a real agent's name on it.

## What counts as a claim

Only `*` is a wildcard; everything else is escaped. A `repo:` value is a repository name, and letting a dot act as a pattern would hand one agent repos it never named. A plain claim is **exact** for the same reason — `bodhi-service-api` must not own `bodhi-service-api-v2`, or a new repo joins somebody's remit on the day it is created without anyone saying so.

## `Bodhilander` having no owner is a gap, not a bug

The harness staffs the product repos and not the tooling ones. Ignition refuses on it, which is the right answer — whether this repo gets an owner agent is a decision about who owns it, not something the engine should infer.

**Mutation-checked** — each fails a test: lists not split; globs not honoured; a plain claim matching by prefix.

8 new tests, 1,959 in the suite (2 skip, 0 fail).

Related: `claude-team-workflow` #311, filed from the same session — `tracking_id` silently mis-derives a 7-letter key prefix, and `BDHLNDR` is one this org uses.


src/main/run-engine/gate-launcher.ts
```diff
@@ -185,6 +185,40 @@ function declared(front: string, key: string): string {
...
+function declaresRepo(front: string, repo: string): boolean {
+  return declared(front, 'repo')
+    .split(',')
+    .map((claim) => claim.trim())
+    .filter(Boolean)
+    .some((claim) => {
+      if (!claim.includes('*')) return claim === repo;
+      const pattern = claim
+        .split('*')
+        .map((part) => part.replace(/[.*+?^${}()|[\\]\\\\]/g, '\\$&'))
```
brannon-bowden: The escaping regex here is malformed: `[.*+?^${}()|[\\]\\\\]` closes the character class one `]` too early, so the class only contains `.*+?^${}()|[\` plus a literal escaped backslash, and everything after (`\\\\]`) becomes a literal 4-char sequence outside the class. As written, `.replace()` is a no-op for ordinary repo/glob strings, so metacharacters like `.`, `+`, `(` in a glob claim are never escaped before being embedded in `new RegExp(...)`, contradicting the docstring's 'matched literally' guarantee. Use the standard escape: `/[.*+?^${}()|[\]\\]/g` (single `\]` inside the class, single `\\` before the closing bracket).

brannon-bowden (changes requested):
This PR fixes a real and well-documented bug (treating `repo:` as a single exact string instead of a comma-separated list of literal names/globs), and the test suite for the list/glob/exact-match semantics is thorough — but the glob-escaping regex introduced to protect the "no accidental wildcards" guarantee is itself broken, so I'm requesting changes.

## Issues

- **`src/main/run-engine/gate-launcher.ts:216`** — the character class in the escaping regex is malformed:
  `/[.*+?^${}()|[\\]\\\\]/g`. Walking the ECMAScript regex grammar: the class opens at `[` and the first `\\` (an escaped backslash) adds a literal `\` to the class, then the very next `]` closes the class early (it is not itself escaped). Everything after that — `\\\\]` — is parsed *outside* the class as a literal 4-character sequence (`\`, `\`, `]`). The net effect is a regex that only matches a class-member char immediately followed by `\\]`, which essentially never occurs in a repo name or glob. So `.replace()` becomes a no-op for real input, and none of `.`, `+`, `?`, `^`, `$`, `{`, `}`, `(`, `)`, `|`, `[` get escaped before being spliced into the final `RegExp`.
  This directly contradicts the docstring above it ("Only `*` is a wildcard. Anything else is matched literally... treating a dot or a dash as a pattern would let one agent claim repos it never named"). A glob claim containing a literal dot, e.g. `repo: bodhi-service-api.internal-*`, would compile to `^bodhi-service-api.internal-.*$`, where the unescaped `.` matches *any* character — so it would also silently match `bodhi-service-apiXinternal-anything`, reintroducing exactly the "candidate list quietly changes" failure mode this PR is trying to close, just via globs instead of the list-parsing bug.
  Fix: correct the character class to the standard escape (`/[.*+?^${}()|[\]\\]/g` — a single `\]` inside the class, a single `\\` before the closing bracket), and note it doesn't surface in the current tests because none of the sample globs (`bodhi-provider-*`, `bodhi-service-connector*`) contain a regex metacharacter other than `-`, which needs no escaping.

## Suggestions

- Non-blocking: once the regex is fixed, add a regression test with a glob claim containing a literal `.` (or another metacharacter), e.g. `repo: 'a.b-*'` matching `a.b-x` but not `aXb-x`, so this exact escaping regression can't reappear silently.

William-Long-II:
Correct on every point, and I verified it before accepting rather than taking the grammar walk on faith:

```
broken: /^a.b-.*$/    a.b-x=true  aXb-x=true
fixed:  /^a\.b-.*$/   a.b-x=true  aXb-x=false
```

The class closed early exactly as you described, the replace never fired on real input, and a claim of `a.b-*` matched `aXb-x` — the failure this PR exists to close, arriving through globs instead of through list parsing.

**The part worth dwelling on is your last sentence.** You noted the sample globs contain no metacharacter other than `-`. It is worse than that: the test named "a dot is not a wildcard" used a **plain** claim, and a plain claim never builds a regex at all — it returns from `claim === repo`. So the one test that appeared to cover escaping asserted a different branch entirely and could not have failed no matter how broken the class was.

Two tests now, both through a glob — one for `.`, one for `+` and `()` — and restoring the malformed class fails both. The docstring says what the class must escape and what happens when it does not, because it is easy to get wrong in the direction of silently doing nothing.

1,959 in the suite, 0 fail.

brannon-bowden (changes requested):
The escaping bug from the previous review is fixed and well covered by new tests, but CI is currently red, so this can't be approved as-is.

## Issues

- **CI is failing**: `test (ubuntu-latest)` is failing (windows-latest, SonarQube, quality-gate, and GitGuardian are all passing). This needs to be green before merge — worth checking whether it's a platform-specific flake/path issue or an actual regression from this change.

## Resolved from previous review

- **`src/main/run-engine/gate-launcher.ts:222`** — the character class is now `/[.*+?^${}()|[\]\\]/g`, which correctly escapes `]` (as `\]`) and `\` (as `\\`) inside the class instead of closing it early. Walking it through: `.`, `*`, `+`, `?`, `^`, `$`, `{`, `}`, `(`, `)`, `|`, `[`, then `\]` (literal `]`), then `\\` (literal `\`), then the closing `]` — every intended metacharacter is now a member of the class. This is the standard escape and matches what was needed.
- The new tests (`and it is not a wildcard inside a GLOB either`, `and neither is any other metacharacter`) exercise exactly the case the old bug missed — a glob claim containing `.`, `+`, `(`, `)` — and assert both that the literal claim still matches and that a naive-regex-interpretation match (`aXb-x`, `aab(c)-x`, `a+bc-x`) is correctly rejected. This directly closes the gap the prior review flagged (the old "dot is not a wildcard" test never went through the glob/regex path at all).

## Suggestions

- Non-blocking: the updated docstring is a good addition — explaining *why* the class is easy to get wrong (must escape both `]` and `\` or it closes early) will help the next person who touches this regex.

brannon-bowden (approved):
This is a solid, well-tested fix — approving.

The bug it addresses is real and clearly demonstrated: treating `repo:` as a single exact string silently dropped valid owners whenever the field held a comma-separated list or a glob, and (per the PR's own framing) the dangerous failure mode is under-inclusion, not over-inclusion. The new `declaresRepo` correctly splits on commas, trims entries, treats `*` as the only wildcard, and properly escapes the remaining regex metacharacters before building an anchored pattern — the character class `[.*+?^${}()|[\]\\]` is the standard, correctly-balanced escaping idiom (the PR's own tests catch the kind of malformed-class bug that would otherwise slip through, e.g. the `a.b-*` / `aXb-x` case). Exact-match behavior for non-glob claims, prefix-non-match (`bodhi-service-api` vs `bodhi-service-api-v2`), and the no-`repo:`-declared case are all covered.

## Suggestions (non-blocking)
- `declaresRepo` builds a regex out of `.*` joined between literal segments per glob claim. Since `repo:` values come from harness config files rather than external/network input, catastrophic backtracking risk is low, but if these files could ever be sourced from less-trusted content it'd be worth keeping an eye on (classic multi-`.*` ReDoS shape). Not a concern for the current trust model.
- The docstring on `declaresRepo` (gate-launcher.ts:188-208) is long and narrative; consistent with the file's existing style for `agentsForRepo`, so leaving as-is is reasonable, but a future pass could trim it to the essential contract now that the historical bug is fixed and covered by tests.

One thing outside what's shown in this diff: `agentsForRepo` can return multiple owners for a repo (the list test explicitly expects `['bsa-lead', 'bsa-platform']`). Worth the author confirming that downstream gate-2 owner-selection logic (not part of this diff) already handles the multi-candidate case correctly rather than assuming a single owner — but that's a question for confirmation, not something this PR needs to change.

William-Long-II:
On the one thing outside the diff — confirming: `agentsForRepo` returning several owners is handled, and not by picking.

`resolveOwners` (`src/main/run-engine/ignition.ts:121-146`) takes three paths. A repo a person named an owner for is checked against the candidate list and refused if the named role does not actually own it. Exactly one candidate is taken. **Anything else — zero candidates or more than one — is a refusal, not a default**, and the refusal carries the list so the fix is a copy-paste:

```
Bodhilander has 3 possible owners
  -> Name one for this run: bsa-care, bsa-lead, bsa-platform
```

`armRun` writes nothing when it refuses, so a run cannot start against an owner nobody chose. The decision about which of several owners should hold a gate is a person's, and this is one of the places the engine is deliberately not allowed to have an opinion.

Verified live on the real harness just now, via the run console: `Bodhilander -> bodhilander-lead` armed because there is exactly one; `bodhi-service-api` resolves to three and would refuse without `--owner`.

PR #283, opened 2026-09-14
test(relay): a port the runner's own connections cannot already hold

### LINKED ISSUE #282: relay entry-point test picks a random port inside Linux's ephemeral range, so it fails intermittently on ubuntu CI
## What happens

`relay/src/index.test.ts` fails intermittently on `test (ubuntu-latest)` and never on windows:

```
error: Failed to start server. Is port 47706 in use?
(fail) the entry point serves at the shipped body ceiling
```

Seen twice in one session, on unrelated PRs (#273 and #281), with different ports each time — 49826 and 47706.

## Why

```ts
const port = 40000 + Math.floor(Math.random() * 20000);
```

The test picks a random port in **40000–59999** and binds it with no retry.

Linux's default ephemeral range is **32768–60999**, which entirely contains that window. Any outbound connection on the runner — and CI makes many — can be holding the port the test then tries to bind, and `main()` fails to listen.

Windows' default dynamic range starts at 49152, so it overlaps only the top quarter of the window. That is consistent with what we see: the failure is ubuntu-only, and both observed ports fell inside the Linux range.

## Why it is worth fixing rather than re-running

A check that fails for reasons unrelated to the change trains people to re-run without reading the output. That is the habit through which a genuine red eventually gets waved through — the same failure mode the exit-code work (#305) addressed in the tooling, arriving instead through CI behaviour. It has already cost two re-runs.

## Suggested fixes, cheapest first

1. **Move the range below the ephemeral one** — e.g. `20000 + random(10000)`. One line, and avoids both platforms' dynamic ranges.
2. **Bind port 0 and read back what the OS assigned.** Strictly correct, but `main()` currently returns only `{ stop }`, so it would need to expose the bound port.
3. **Retry on `EADDRINUSE`.** Works, but hides a collision rather than avoiding one.

Not fixed in passing: the test belongs to the relay suite and the PRs that hit it were about something else.

William-Long-II:
## What this changes

One number in `relay/src/index.test.ts`: the random port the entry-point test binds moves from **40000-59999** to **20000-29999**.

## Why

Linux's default ephemeral range is **32768-60999**, which contains the old window entirely. CI makes many outbound connections, so one of them can be holding the port the test then tries to bind - and the test binds with no retry, so `main()` just fails to listen:

```
error: Failed to start server. Is port 47706 in use?
(fail) the entry point serves at the shipped body ceiling
```

Windows' dynamic range starts at 49152 (measured: `netsh int ipv4 show dynamicport tcp` -> start 49152, 16384 ports), overlapping only the top quarter of the old window. That matches what we saw: ubuntu-only, and both observed ports (49826, 47706) fell inside the Linux range.

The new range is below both.

## Why it was worth a PR rather than a re-run

It has now turned a green change red on two unrelated PRs in one session (#273, #281), and on #281 it also cost a review cycle - the reviewer correctly declined to approve over a red check, so the fix for the *actual* review finding sat behind a failure that had nothing to do with it. A check that fails for reasons unrelated to the change is the mechanism by which a real red eventually gets waved through.

## Why not `PORT=0`

`main()` already returns `server`, so reading back `server.port` would work and would be strictly correct. But `loadConfig` rejects `PORT=0` (`must be an integer between 1 and 65535`), and that rejection is doing a job: a deployment that typos the port should fail loudly, not silently listen somewhere nobody can find. Trading that for a test fixture is the wrong way round.

## Verification

No new test - the change is to a test's own fixture, and a flake has no failing case to pin. What is checkable is the claim about the ranges, and that is measured above rather than assumed. `bun test src/index.test.ts` passes.

Closes #282


brannon-bowden (approved):
This is a clean, well-justified test fix — approving.

The PR narrows the random port range used in the body-ceiling test from `40000-59999` down to `20000-29999`, specifically to sit below both Linux's (32768-60999) and Windows's (49152-65535) ephemeral port ranges. This directly addresses flaky CI failures caused by the test's `PORT` colliding with a port already claimed by an outbound connection from the runner itself, rather than a real regression in the code under test. The reasoning is sound and the accompanying comment does a good job explaining why `PORT=0` wasn't used instead (deliberate rejection in `loadConfig`).

No correctness issues with the change itself — the arithmetic checks out (`20000 + [0, 9999]` → `[20000, 29999]`), and it's a pure test-only diff with all CI checks green.

## Suggestions
- (non-blocking) The narrower range still leaves a small residual chance of two parallel test runs picking the same port if tests are ever parallelized within a single runner; not a concern introduced by this change, just worth keeping in mind if flakiness resurfaces.

brannon-bowden (approved):
This is a clean, well-justified test fix — approving.

The change narrows the random ephemeral port range used in the body-ceiling test from `40000-59999` (which overlaps Linux's dynamic port range `32768-60999`) to `20000-29999`, which falls below both Linux's and Windows' (`49152-65535`) ephemeral ranges. The accompanying comment clearly documents why the old range caused flaky CI failures (a runner's own outbound connection could occupy the port before the test bound to it) and why `PORT=0` isn't a viable alternative (rejected by `loadConfig` by design). The math checks out and the reasoning is sound.

All listed CI checks are passing, and the change is scoped to test-only flakiness mitigation with no production code touched.

**Suggestions (non-blocking):**
- Since this port-selection logic is now inline with a detailed rationale, it might be worth extracting it into a small shared test helper (e.g., `pickSafeTestPort()`) if other tests in the suite pick random ports the same way — that would centralize the "why" in one place rather than risk future copy-paste of the older, flaky range. Not necessary for this PR.

William-Long-II:
On the shared helper: I checked before deciding, and this is the only test in either suite that picks a port.

```
$ grep -rn "Math.floor(Math.random" --include=*.test.ts relay/src src | grep -i port
relay/src/index.test.ts:31
```

A `pickSafeTestPort()` with one caller would move the rationale one indirection away from the only line that needs it, which is the opposite of what the comment is for. If a second test ever needs a port, that is the moment the helper earns its place — and the comment as written is what will tell whoever writes it why the range is not arbitrary.

PR #284, opened 2026-09-14
fix(run-engine): provisioning is not a plugin read, and cannot share its clock

William-Long-II:
## What happened

This is the first finding from driving the run engine against the real machine rather than against fakes — the failure a test could not have found, because a test would have had to supply the thing in doubt.

The run armed cleanly and the first step reported:

```
state     inconclusive
applied   prepared, provisionUndriveable
notify    python did not finish within 60000ms
notify    nothing could be installed here - no result for the change
```

That message is true and useless. `processDeps` gave every plugin call the same 60-second deadline, and provisioning is the one call that does not belong on it: it runs the repo's own installer, which resolves a dependency tree and compiles native modules — minutes on a cold worktree.

Two things went wrong, and the second is worse than the first:

1. **The wrong diagnosis.** "python did not finish" sends whoever reads it to look at Python, which is fine. The problem was `bun install`.
2. **The evidence was destroyed.** Killing the process threw away the installer's own account of what failed. The engine knew only that something took too long, so it recorded `provisionUndriveable` — "no result for the change" — when there very much was a result.

## The change

`provision` becomes its own dependency on `ExecutorDeps` with a 15-minute default, built from the same factory as `plugin` so the two cannot drift on anything except the clock. Long enough that a slow install finishes; short enough that an install blocked on a prompt nobody will answer still ends.

One 60-second deadline for "the plugin" was a reasonable default while every plugin call was a read that answers in about a second. It stopped being reasonable the moment one of them started doing real work.

## The same run afterwards

```
state     failed
applied   prepared, provisionFailed
notify    FAIL  Bodhilander: bun install exited 255
          node-gyp failed to rebuild '...\node_modules\node-pty'
          an install FAILED: Bodhilander
            That is a real result -- read it before re-running.
```

`failed` rather than `inconclusive`, because now we actually know. The exit-code contract did its job once it was given a code to read.

## Verification

Both halves are pinned, and both were mutation-checked by making the regression and watching the tests go red:

- The executor's fakes record `provision` under its own call kind rather than folding it in with `plugin`. Putting `deps.plugin` back in `provision()` fails **5 tests**.
- `processDeps` is checked by running one slow script on each clock: `plugin` at 250ms times out, `provision` at 30s returns `installed`. Giving `provision` the plugin timeout fails **1 test**, which is the one that matters.

Full suite: 1963 pass, 2 skip, 0 fail. `tsc -p tsconfig.main.json --noEmit` clean.

## Not fixed here

The install failure this uncovered is real and separate — `postinstall` forces `electron-rebuild -o better-sqlite3,node-pty`, and `node-pty` 1.1.0 is N-API with win32 prebuilds, so on a Windows machine with no Visual Studio it cannot rebuild. Filed separately; the worktree it produced runs the full test suite regardless, which is its own finding.

Refs #722


brannon-bowden (approved):
This PR splits provisioning's timeout off from the plugin-read timeout so a cold `bun install` (or equivalent) isn't killed by the 60s deadline meant for quick plugin reads, and the tests are updated in lockstep to reflect the new `provision` dependency. All 5 CI checks (SonarQube, both OS test runs, quality-gate, GitGuardian) are passing.

The fix itself is sound: `command-runner.ts` factors the spawn logic into `pluginCall(timeoutMs)` and instantiates it twice with independent timeouts (`pluginTimeoutMs` default 60s, `provisionTimeoutMs` default 15min), `executor.ts`'s `provision()` step now calls `deps.provision(...)` instead of `deps.plugin(...)`, and the new `command-runner.test.ts` cases directly pin the regression (a script that takes 1.5s times out under a 250ms plugin deadline but succeeds under a 30s provision deadline) and confirm the interpreter is still taken from argv rather than config. The executor/driver test updates consistently rename the provisioning-path mocks from `plugin` to `provision`, matching the production code path.

No correctness, security, or missing-test issues stood out — the new behavior is covered by tests that would fail if the clocks were folded back together.

## Suggestions

- `command-runner.ts`: the shared spawn helper is named `pluginCall` and is now used for both `plugin` and `provision`. Given the PR's own premise that provisioning is not a plugin read, a more neutral name (e.g. `spawnPythonLike`) would avoid the naming tension for future readers — non-blocking, since the two really do share the identical spawn shape.

PR #286, opened 2026-09-14
fix(build): node-pty ships N-API prebuilds and must not be rebuilt from source

### LINKED ISSUE #285: bun install fails on a fresh worktree without Visual Studio: node-pty is forced through electron-rebuild but ships N-API prebuilds
## What happens

On a Windows machine with no Visual Studio / Build Tools, `bun install` in a fresh clone or worktree fails:

```
$ bun install
$ electron-rebuild -o better-sqlite3,node-pty && node node_modules/node-pty/scripts/post-install.js
- Searching dependency tree

Error: Could not find any Visual Studio installation to use
    at VisualStudioFinder.fail (node_modules\node-gyp\lib\find-visualstudio.js:118:11)

x Rebuild Failed
node-gyp failed to rebuild '...\node_modules\node-pty'
error: postinstall script from "bodhilander" exited with 255
```

`better-sqlite3` rebuilds fine — a prebuilt for Electron's ABI is fetched. It is `node-pty` that goes to source.

## Why the rebuild looks unnecessary

`node-pty` is pinned at **1.1.0**, which is N-API (`node-addon-api ^7.1.0`) and ships prebuilds:

```
node_modules/node-pty/prebuilds/
  darwin-arm64  darwin-x64  win32-arm64  win32-x64
```

N-API is ABI-stable across Node and Electron — that is the point of it — so a prebuilt binary does not need rebuilding for Electron. `node-pty`'s own install script already prefers one anyway (`node scripts/prebuild.js || node-gyp rebuild`). Naming it in `electron-rebuild -o` is what forces the source build past that fallback.

Evidence that the prebuild is what actually gets used here: **this repo's working `node_modules` has no `node-pty/build/Release` at all**, only `prebuilds/`, and the app runs.

```
$ find node_modules/node-pty -name "*.node"
node_modules/node-pty/prebuilds/win32-x64/pty.node
node_modules/node-pty/prebuilds/win32-x64/conpty.node
...
```

`better-sqlite3` is a different story and must stay in the list — it is not N-API, and `CLAUDE.md` records that it is rebuilt for Electron's ABI on purpose.

## Why it has not been noticed

- CI's `windows-latest` runners ship the MSVC toolchain, so the rebuild succeeds there.
- A developer machine that installed once while a toolchain was present keeps a populated `node_modules`; `bun install` afterwards is close enough to a no-op that the postinstall failure does not resurface.

It shows up on a **fresh worktree**, which is exactly what the run engine provisions for every initiative (#722). That is how this was found: `provision.py` ran `bun install` in a new worktree and got 255.

## The odd part, and why this is not urgent

**The failed install still produces a working tree.** The worktree it left behind runs the full suite:

```
$ cd _wt-bdh-239-Bodhilander && bun test --isolate src/main/run-engine/__tests__
381 pass  0 fail
```

So `bun install` exits 255 having done everything that mattered. Nothing is broken except the exit code — but the exit code is the only thing an automated provisioner can read, so to the run engine this is a hard stop on a tree that was actually fine.

## Suggested fix

Drop `node-pty` from the `electron-rebuild -o` list:

```diff
-"postinstall": "electron-rebuild -o better-sqlite3,node-pty && node node_modules/node-pty/scripts/post-install.js"
+"postinstall": "electron-rebuild -o better-sqlite3 && node node_modules/node-pty/scripts/post-install.js"
```

Before doing that, two things need checking rather than assuming, and neither can be checked from this machine:

1. **Linux.** There are no linux prebuilds in the list above, so the ubuntu CI leg must be compiling `node-pty` from source today. Whether that happens via `electron-rebuild -o` or via node-pty's own install script needs confirming on a runner — if it is the former, removing the flag breaks that leg.
2. **What `post-install.js` expects.** It cleans `build/Release` and copies the conpty files; it needs confirming that it behaves the same when nothing rebuilt into that directory first.

Both are answerable by pushing the one-line change and reading all three CI legs, which is the cheapest honest way to find out.

## Not fixed in passing

Found while driving the run engine; the fix belongs to whoever can watch all three platforms confirm it, not to a PR about something else.

William-Long-II:
Closes #285.

## The change

One character class, effectively:

```diff
-"postinstall": "electron-rebuild -o better-sqlite3,node-pty && node node_modules/node-pty/scripts/post-install.js"
+"postinstall": "electron-rebuild -o better-sqlite3 && node node_modules/node-pty/scripts/post-install.js"
```

## Why

`bun install` in a fresh clone or worktree fails on a Windows machine with no Visual Studio:

```
Error: Could not find any Visual Studio installation to use
node-gyp failed to rebuild '...\node_modules\node-pty'
error: postinstall script from "bodhilander" exited with 255
```

`node-pty` is pinned at **1.1.0**, which is N-API (`node-addon-api ^7.1.0`) and ships prebuilds for `darwin-arm64`, `darwin-x64`, `win32-arm64`, `win32-x64`. N-API is ABI-stable across Node and Electron — that is the entire point of it — so those binaries do not need rebuilding for Electron. node-pty's own install script already prefers one (`node scripts/prebuild.js || node-gyp rebuild`); naming it in `electron-rebuild -o` is what drives past that fallback into a source build.

`better-sqlite3` stays. It is not N-API, and `CLAUDE.md` records that it is rebuilt for Electron's ABI deliberately.

## Measured, not argued

On a Windows machine with no toolchain at all (`vswhere.exe` absent; the VS directory is an empty leftover):

**The postinstall now succeeds.**
```
$ ./node_modules/.bin/electron-rebuild -o better-sqlite3 && node node_modules/node-pty/scripts/post-install.js
✔ Rebuild Complete
> Cleaning release folder...
> Moving conpty.dll...
  Copying ...\third_party\conpty\1.23.251008001\win10-x64\conpty.dll -> ...\build\Release\conpty\conpty.dll
EXIT=0
```

That also answers the open question in the issue about `post-install.js`: it cleans `build/Release` and copies the conpty files correctly with nothing having rebuilt into that directory first.

**And the prebuild actually works** — not just loads, but spawns a pty and reads its output back, under plain node *and* under Electron:

```
$ node -e "...pty.spawn(cmd, ['/c','echo','pty-ok'])..."
loaded, exit 0 saw: true

$ ./node_modules/.bin/electron ptycheck.js
ELECTRON loaded, exit 0 saw: true
```

with `build/Release` containing only `conpty` — no compiled `pty.node` anywhere in the tree.

## Why nobody hit this

- `windows-latest` runners ship MSVC, so the rebuild succeeds in CI.
- A developer machine that installed once while a toolchain was present keeps a populated `node_modules`, and `bun install` afterwards barely touches it.

It surfaces on a **fresh worktree** — which is what the run engine (#722) cuts for every initiative. `provision.py` ran `bun install` in a new worktree, got 255, and stopped the run. That is how this was found.

## What CI has to answer

This machine cannot settle the linux leg. There are no linux prebuilds, so ubuntu must be compiling node-pty from source today, and the question is whether `electron-rebuild -o` was doing that or whether node-pty's own install script does it either way. **If the ubuntu leg goes red, the answer is the former and this change is wrong as written** — it would then need to stay for linux and drop only on win32/darwin.

Reading all three legs is the cheapest honest way to find out, so that is what this PR is for. Please do not approve it on the reasoning alone.

## Aside worth recording

The failed install still produced a working tree — the worktree `bun install` exited 255 on runs the full suite (`381 pass, 0 fail`). So the only broken thing was the exit code. That is harmless to a person, who sees a green test run and moves on, and fatal to a provisioner, for which the exit code is the only thing there is to read.


brannon-bowden (approved):
This is a minimal, well-scoped fix: it removes `node-pty` from the `electron-rebuild -o` list so it's no longer rebuilt from source, while leaving `better-sqlite3` (which does need a native rebuild) and the existing `node-pty` post-install script (which fetches/verifies its N-API prebuild) untouched. The rationale in the title lines up with the change, the diff is a single line, and all CI checks (SonarQube, both platform test runs, quality gate, GitGuardian) are green. No correctness or security concerns in what's shown.

Suggestions (non-blocking):
- Since this is purely a build/packaging behavior change, consider a quick manual sanity check (or note in the PR description) that a fresh `npm install` / packaged build still produces a working `node-pty` binary on both Windows and Linux/macOS runners, given the diff itself can't show that runtime behavior.

William-Long-II:
## Correction: the green CI on this PR did not test the change

I said earlier that the ubuntu leg would settle the linux question. That was wrong, and worth stating plainly before anyone reads the check marks as evidence.

`test.yml` installs with `--ignore-scripts`:

```yaml
# --ignore-scripts skips postinstall's electron-rebuild, which needs an
# Electron download and a native toolchain. The suites never load a
# native module: they use bun:sqlite and mock node-pty.
- run: bun install --frozen-lockfile --ignore-scripts
```

So does `sonarqube.yml`. **No pull-request check has ever run `postinstall`** — the only thing that does is a release build. Every leg on this PR was structurally incapable of failing for the reason this PR is about, which is also the reason the defect survived this long.

## What I added instead

A `native-install` job that installs for real on `ubuntu-latest` and `windows-2022` — the runner `release.yml` pins on purpose (#109), since `windows-latest` now means VS 2026 and node-gyp cannot detect it. Testing on a runner the release does not use would answer a question nobody asked.

It does not assert that node-pty *loads*. A binary for the wrong ABI can be present, resolve, and still fail on the first real call. It spawns a pty and reads its own output back:

```
ok: node-pty spawned a pty on win32 and read "native-module-ok" back
```

Two things I only found by running it:

- A pty holds the event loop open after its child exits. Without an explicit `process.exit(0)` the script printed its success and then hung to the step timeout — a pass that reports as an infrastructure fault.
- The failure path works. Pointed at a marker that never arrives, it exits 1 and prints what it actually saw, escape codes and all.

## What is still open

Whether linux compiles node-pty via `electron-rebuild -o` or via node-pty's own `node scripts/prebuild.js || node-gyp rebuild` fallback. The reasoning says the latter and says the result is fine either way — node-pty builds against `node-addon-api`, so even a locally compiled linux binary is N-API and ABI-stable across Node and Electron.

But that is reasoning, and this session has already produced three tests that were structurally incapable of failing. **The ubuntu leg of `native-install` is the thing that actually answers it. If it goes red, the one-line change in this PR is wrong as written** and needs to drop node-pty on win32/darwin only.

William-Long-II:
## The linux question is answered, empirically

Running the check under node was not enough, and I should not have stopped there. On linux there is no node-pty prebuild, so node-gyp compiles it against **Node's** headers — and this PR is precisely what stops `electron-rebuild` from rebuilding it afterwards. So the binary linux ships **did** change. Running that binary under node proves nothing about the pairing that actually matters.

Added an Electron leg (xvfb on linux, since Electron initialises a display even for a script that never opens a window). From the ubuntu job, with `electron-rebuild -o better-sqlite3` only:

```
Install for real, postinstall included              -> success
  $ electron-rebuild -o better-sqlite3 && node node_modules/node-pty/scripts/post-install.js
  ✔ Rebuild Complete
  > Moving conpty.dll...  SKIPPED (not Windows)

The native module loads and works under node        -> success
  ok: node-pty spawned a pty on linux and read "native-module-ok" back

The native module loads and works under Electron    -> success
  ok: node-pty spawned a pty on linux and read "native-module-ok" back
```

Not "it loaded" — it spawned a pty and read its own output back off it, under the runtime that ships. That is the thing I could not answer from a Windows machine, and it says the change is right for linux as written rather than needing a per-platform form.

So the N-API reasoning held. But it is now checked rather than believed, and it stays checked: `native-install` runs on every PR, so the next transitive bump or runner-image move that breaks the postinstall fails here instead of at release time.

Windows legs still running; this comment is about the leg that was in doubt.

William-Long-II:
Answering the suggestion, and flagging that the approval above is on a SHA that no longer exists at head.

## The suggestion is now a check, not a manual step

> consider a quick manual sanity check ... that a fresh install still produces a working `node-pty` binary on both Windows and Linux, given the diff itself cannot show that runtime behavior

Agreed, and a manual check would not have held. Two commits since that review turn it into CI:

- **`native-install`** installs for real — no `--ignore-scripts` — on `ubuntu-latest` and `windows-2022`. Until now *nothing on a PR ran `postinstall` at all*; only a release build did, which is why this defect survived.
- It does not assert the module loads. It spawns a pty and reads its own output back, **under Electron** as well as node. On linux there is no prebuild, so node-gyp compiles against Node's headers while Electron is what loads the result — that pairing was the actual open question, and running under node alone would have answered the wrong one.

```
The native module loads and works under node     -> success
The native module loads and works under Electron -> success
  ok: node-pty spawned a pty on linux and read "native-module-ok" back
```

## One correction to the coverage claim

**macOS is not covered.** `native-install` runs ubuntu and windows-2022 only. `release.yml` builds on `macos-14` and `macos-15-intel`, and those are still the first place a darwin postinstall regression would show. The reasoning is stronger there — darwin-arm64 and darwin-x64 prebuilds both exist, so there is no source build to fail — but reasoning is what this PR has already been caught by once. Adding mac runners to a per-PR job costs paid larger-runner minutes, so I have left it out deliberately rather than by oversight. Worth a follow-up issue if you want it closed.

## Re-requesting review

The approval was cast on `c350f9d`, the single-line change. Head is `83a94fb` and adds a workflow and a script. Different diff, and CI changes are exactly the kind that should not ride in on a stale approval — so this is re-requested rather than merged.

brannon-bowden (approved):
Approve — this update replaces the prior "no way to see it work" gap with an actual CI check, which is exactly what the earlier review suggested.

The new `native-install` job installs dependencies with scripts enabled (the only place in this pipeline that does) and then verifies the resulting `node-pty` binary is genuinely usable — first under plain Node, then under Electron, on both ubuntu-latest and windows-2022. Running the check under both runtimes is the right call given the stated concern (N-API is supposed to make a node-gyp build against Node headers loadable from Electron, but "supposed to" isn't a check). The script itself is sound: it spawns a real pty rather than just requiring the module, asserts on the round-tripped marker text (not just a clean exit), guards against a hung pty with an explicit timeout, and calls `process.exit(0)` explicitly since a live pty would otherwise keep the event loop open. The constant `MARKER` is hardcoded, so there's no command-injection concern in either the `cmd.exe` or `/bin/sh` invocation, and `$RUNNER_OS` in the workflow's bash step is a GitHub-supplied value, not attacker-controlled input.

All 7 CI checks, including both new `native-install` matrix legs, are passing.

**Blocking issues**

None.

**Non-blocking issues**

- `.github/workflows/test.yml`: the newly introduced action pins (`actions/checkout@...# v7.0.0`, `actions/setup-node@...# v7.0.0`, `actions/setup-python@...# v6.3.0`) reference major versions higher than what I'd expect to be current for these actions. I can't verify the SHA-to-tag mapping from the diff alone, but it's worth the author double-checking these resolve to the intended real releases rather than a typo'd version comment, since a mismatched comment (even if the SHA itself is fine) can mislead future readers doing pin bumps.
- `.github/scripts/check-native.js:20-22`: `child.onData` accumulates `output` and the listeners are never removed, but since the process exits immediately in `onExit`, this has no practical effect — just flagging as something to be aware of if this script is ever extended to run multiple commands.

**Tests**

The new `check-native.js` is itself the test coverage for the postinstall/native-module behavior that this PR is about, run automatically via the new `native-install` job — no gap here.

William-Long-II:
Both non-blocking points checked rather than waved past.

## The action pins resolve to exactly what the comments claim

Not verifiable from the diff, as you said, so resolved against the upstream tag refs:

```
actions/checkout@9c091bb2...  -> refs/tags/v7.0.0
actions/setup-node@82076278... -> refs/tags/v7, refs/tags/v7.0.0
actions/setup-python@ece7cb06... -> refs/tags/v6, refs/tags/v6.3.0
```

Each comment matches its SHA. The reason the majors look ahead of expectation is that they are not new here — every pin I introduced is **byte-identical to one already in `release.yml`**, copied from it deliberately so the two workflows cannot drift on action versions:

```
release.yml:121  actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
release.yml:124  actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
release.yml:135  actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
```

A `native-install` job on different action versions from the release build it is meant to predict would be checking something other than what ships, which is the failure this whole PR is about.

## The unremoved listener

Correct on both counts — it accumulates, and it cannot matter, because `onExit` exits the process. Leaving it: a `removeListener` here would be dead code guarding against a second command the script does not run, and the next person to add one will see the accumulation in four lines of straight-line code. Noted rather than fixed, on purpose.

Merging.

Issue #288, opened 2026-09-14
an unattended gate is launched with a permission posture that guarantees it blocks


William-Long-II:
Found on the first real gate launch (#722). See #287 for why nothing noticed.

## What happens

The engine launched gate 2 as:

```
claude.exe --session-id 4601935b-... --plugin-dir C:/work/repos/claude-team-workflow
  --setting-sources "" --strict-mcp-config
  --permission-mode manual --permission-prompts host
  --agent bodhilander-lead "Work gate 2 for BDH-239."
```

spawned with `--bg`. The agent got two tool calls in and stopped:

```
TOOL Bash  ls .../initiatives/ 2>/dev/null || ls .../initiatives/ 2>/dev/null
  -> (no result, ever)
TOOL Bash  git log --oneline -5 && git status --short
  -> e7fcb35 Merge pull request #286 ...
```

One Bash call was auto-approved; the other was not, and has no `tool_result` in the transcript 20+ minutes later. CPU flat at ~0.02/s. It is waiting for a permission answer.

`--permission-prompts host` routes prompts to the host session. A `--bg` gate has no host with a person attached, so the prompt goes somewhere nobody is looking and the gate waits for the rest of its life.

## Why the posture exists, and why that is not the bug

Posture is recorded per gate on purpose, and the reason is good — from `executor.ts`:

> Recorded per gate rather than per run, because if you cannot tell afterwards whether an owner ran unsandboxed, you cannot trust what it produced.

That reasoning stands. The defect is narrower: **`manual` is not a posture an unattended gate can be launched under at all**, and nothing refuses the combination. The engine will happily spawn a background process that is guaranteed to block.

## What needs deciding

This is not "default to bypass". The whole point of recording posture is that running an owner unsandboxed is a claim somebody has to be able to audit afterwards. Options, and none is obviously right:

1. **Refuse the combination at launch.** `background` + a prompting posture is undriveable, in the exit-code sense — ask what nobody knows. Cheapest and most honest: the run refuses before spending anything, and the message says exactly which two settings conflict.
2. **Route prompts somewhere a person can answer them.** The correct long-run answer, and the one that fits the inbox: a gate blocked on permission is precisely a `waitingPermission` run, which is already in `NEEDS_A_PERSON`. But nothing currently surfaces the prompt, and inventing a prompt channel is its own slice.
[...]

PR #290, opened 2026-09-14
feat(run-engine): a channel a blocked gate can wait on and a person can answer

William-Long-II:
Refs #722. Addresses the mechanism half of #288.

## The problem

The first real gate launch stopped two tool calls in and never moved again. `--permission-prompts host` routes prompts to a host session, and a `--bg` gate has no host with a person at it, so one `Bash` call sat unanswered for twenty minutes at ~0.02 CPU-seconds a second.

The chosen direction (owner's call) is a **prompt channel feeding `waitingPermission`**, with an unattended posture as a later option rather than the default — so that "this gate ran without a human able to refuse it" stays a recorded fact instead of a convenience.

## The state machine already agreed

Nothing in `transitions.ts` changes. The path was already there — `permissionRequested` → `waitingPermission` → `permissionAnswered` → `running` — and already carried the right instinct:

> No timeout. A silent auto-deny is indistinguishable from a gate finding, and the run would then carry a verdict nobody gave.

What was missing was only the thing that emits those events.

## Measured, not assumed

None of this is documented where you would look, so it was established against `claude 2.1.270` before anything was designed around it:

| claim | how it was established |
|---|---|
| `--permission-prompt-tool` is a real flag | it has **no entry of its own** in `--help` — it appears only inside the description of `--permission-prompts`. Unknown flags *do* error (`error: unknown option '--definitely-not-a-flag'`), and this one does not. |
| it is invoked as a normal tool | `tools/call` with arguments `{ tool_name, input, tool_use_id }`, captured from a spike server's log |
| `{"behavior":"allow"}` grants | the blocked `Bash` call ran |
| `{"behavior":"deny","message":...}` refuses | the file was never created |
| the denial message reaches the model | the gate quoted it back |
| waiting is supported | a reply held **120 seconds** was honoured in full; no short cap on the permission path |

That last one is what makes the whole design viable. If the CLI had capped the permission call at 30s, "wait for a person" would have been impossible and this would need a different shape entirely.

## Then proven end to end

Not with fakes — with this broker, a real gate, and an answer written by hand:

```
$ ls chan/
toolu_01YDVjtMvgGqJ2W7pSowiEuS.request.json

{ "toolUseId": "toolu_01YDV...", "toolName": "Bash",
  "input": { "command": "echo broker-ok > proof.txt", ... },
  "askedAt": "2026-09-14T12:05:02.381Z" }

$ echo '{"behavior":"allow"}' > chan/toolu_01YDV....reply.json
-> "It worked. proof.txt now contains broker-ok."
```

And the refusal, which is the half that actually has to hold:

```
$ echo '{"behavior":"deny","message":"this run may not write outside its worktree"}' > ...
-> "It did not work. The command was blocked before it ran - the sandbox rejected it
    with 'this run may not write outside its worktree' ... I won't retry it or route
    around the restriction."
```

The message arrives whole. That is the difference between a gate that knows it was refused and one that only knows it failed — and a refusal with nothing in it teaches an owner to retry.

The pure reader was also pointed at the real file the broker wrote, rather than only at its own fixtures.

## Design notes worth arguing with

**The channel is a directory.** Not a socket, not a table. The broker cannot be part of the app — `better-sqlite3` is built for Electron's ABI, so a child process could not open the run store, and it should not be able to. Files also leave a stuck gate readable with `cat`, which matters for "you can't see what's happening".

**Unreadable requests are reported, not skipped.** The gate is blocked on that id whatever the file says. Dropping a malformed one would hang a run on something no screen would ever show — the exact failure this channel exists to end. So `isWaiting` is true for them too.

**The broker never answers on its own.** No timeout, no default. A broker that denied after a while would hand the model a refusal nobody gave.

**Tool input is carried whole.** A person approving a `Bash` call is approving its command line; an approval given against a summary is not an approval of what runs. There is a test that fails if anything narrows it.

## What this deliberately does not do

**Nothing calls it yet.** No flags are added to `gate-command`, nothing polls, nothing reaches the inbox. That is the next slice.

Landing the contract first is what keeps all 22 cases assertions rather than fixtures on disk — and this engine has now produced three tests that were structurally incapable of failing, so the shape that makes failure easy to demonstrate is worth the extra PR.


brannon-bowden (approved):
I'll proceed with the review directly based on my analysis of the diff.




Approve — this is a well-documented, thoughtfully tested addition (the file-channel abstraction and its MCP broker), with a few non-blocking hardening items worth addressing.

## Issues

- **Unsanitized `tool_use_id` used to build file paths (path traversal)** — `scripts/permission-broker.js:39-42`. `toolUseId` comes straight from the incoming `tools/call` arguments and is interpolated into `path.join(channelDir, \`${toolUseId}.request.json\`)` / `...reply.json` with no validation that it's free of `/`/`..` segments. In normal operation this id is CLI-generated and opaque, so the practical exploitability is low, but a broker that accepts arbitrary MCP `tools/call` input and writes wherever that value points is worth defending — a regex allowlist (e.g. `^[A-Za-z0-9_-]+$`) before building `requestPath`/`replyPath` would close this off. (major, non-blocking given the low real-world attacker control over this field)
- **Polling loop swallows all `readFileSync` errors, not just "file missing"** — `scripts/permission-broker.js:~95-101`. The `catch { return; }` around `fs.readFileSync(replyPath, ...)` treats `EACCES`/`EISDIR`/etc. identically to `ENOENT`, so a persistent filesystem problem looks exactly like "not answered yet" and polls silently forever instead of surfacing something a person could act on. Consider checking `err.code === 'ENOENT'` and treating other errors distinctly (e.g. logging to stderr). (minor)
- **`tools/call` doesn't check the requested tool name** — `scripts/permission-broker.js:~161`. `handle()` routes every `tools/call` straight into `ask()` regardless of `message.params.name`, even though only `ask` is advertised in `tools/list`. Harmless today since nothing else calls this server, but a stray/misrouted call would silently be treated as a permission ask. (nit)

## Suggestions (non-blocking)

- `permission-channel.ts` has thorough unit coverage, but `scripts/permission-broker.js` itself (the JSON-RPC framing, polling, and file-write logic) has no tests. It's harder to unit test given the stdin/stdout/fs side effects, but even a thin integration test (spawn the script, pipe `initialize`/`tools/call`, drop a reply file, assert stdout) would catch regressions in the polling/JSON-RPC plumbing that the pure-module tests can't reach.
- The module docblocks are excellent as design rationale, but are unusually long for inline comments; if this repo has an `ADR`/design-doc convention, consider moving the "measured against claude 2.1.270" provenance notes there and keeping the source comments focused on the non-obvious invariants.

brannon-bowden (approved):
I have no working tree here (this is a diff-only review), so I'm basing this entirely on the unified diff provided.




Approve — this wiring cleanly connects the file-channel abstraction into the actual gate lifecycle (`step`/`perms`/`answer`, plus `gate-launcher`'s conditional MCP config), and the new test pins down the exact failure mode (tool-name/server-key mismatch) that would otherwise be silent. A few things worth a look, none of them blocking.

## Issues

- **Attempt-number consistency between writer and reader isn't visible here** — `scripts/run-console.ts` (`channelKeyFor` call inside `step()`, and `activeChannel()`). The channel a gate is launched into is keyed by `runs.listGates(runId).filter(g => g.gate === gate).length` (a count computed at launch time), while `perms`/`answer` later resolve the same channel via `runs.activeGate(runId).attempt`. These are two independent ways of deriving the same "attempt" number. If they disagree by even one (e.g., the DB's `attempt` column is 1-indexed while the pre-launch count is 0-indexed for a first attempt), `perms`/`answer` would silently look in a directory `launchGate` never wrote to and report "no gate in flight" forever. I can't see `runs.ts` to confirm either numbering scheme, so this is a question rather than a confirmed defect — worth a quick check or a regression test that exercises the real `step` → `perms` path end-to-end rather than just the pure `channelDirFor` helper.
- **Gate 2's default mode leaves `manual` posture unfixed for #288** — `scripts/run-console.ts` (the `mode:` line in `step()`). The comment correctly notes `--permission-prompts` only "applies with --print," yet `permissions` is still unconditionally attached to gate 2's `launchGate` call even though gate 2 defaults to `background`. So for the common case (gate 2, `manual` posture, default env), the channel/mcp-config machinery is created but inert, and the original #288 symptom (gate blocks with nobody able to answer) persists unless an operator remembers to set `BODHI_GATE_MODE=print`. This is documented in the comment, so it may be intentional, but it's worth confirming that's the desired default rather than something to fix inline (e.g., forcing `print` when posture is `manual`).

## Suggestions (non-blocking)

- The new conditional branch in `launchGate` (`gate-launcher.ts`, the `posture === 'manual'` check that creates the channel dir and writes the `.mcp.json`) and the new `perms`/`answer` CLI commands have no test coverage in this diff — only the pure helpers in `permission-channel.ts` are tested. Given this is the actual glue that makes the feature work end-to-end, a test exercising "manual posture + permissions supplied → mkdir + mcp config written" (and the inverse for non-manual postures) would catch regressions the unit tests for `channelDirFor`/`mcpConfigText` can't.
- `mode: (env('BODHI_GATE_MODE', gate === 2 ? 'background' : 'print') as 'background' | 'print')` casts an operator-supplied environment variable straight to the literal union with no validation. A typo'd value would flow through untyped-but-trusted into `launchGate`/`gate-command`. Low risk given this is an internal dev CLI, but a small `if (mode !== 'background' && mode !== 'print') throw` would fail fast instead of producing a confusing downstream error.

brannon-bowden (approved):
Approve — this delta adds `gateBrief` and `rolesFromHarness`, both pure functions with thorough, well-pinned tests, and wires them into `step()` in a way that only improves on the prior behavior (gates 3/4 previously never got a role at all; now they do when the harness declares exactly one agent, and ambiguous/absent cases are surfaced via console notes instead of silently doing nothing). CI is green across the board.

## Issues

- **Sequence/unclaimed gates still get no role, same as before this diff** — `scripts/run-console.ts` (the `fromHarness.unclaimed`/`fromHarness.sequences` loops). When gate 4 is a sequence or nothing declares a gate, `agents[gate]` is left unset and only a `console.log` note is emitted — there's no visible `continue`/skip of the actual `launchGate` call for that gate in this hunk. This mirrors pre-existing behavior (gates 3/4 never had a role before this change either), so it isn't a regression, but I can't see the code that consumes `target.agents[gate]` downstream to confirm an undefined role degrades gracefully rather than crashing later in the launch path. Worth a quick check, not a blocker.
- **Two notes from the prior review are unresolved by this diff and still apply**: the attempt-number derivation between `launchGate`'s pre-launch count and `perms`/`answer`'s `runs.activeGate(runId).attempt` remains unverified (unchanged code, not touched here), and gate 2's default `background` mode still leaves the `manual` posture channel inert unless `BODHI_GATE_MODE=print` is set explicitly. Neither is addressed or regressed by this delta.

## Suggestions (non-blocking)

- `agentsForGate`'s sort is alphabetical, and `GateSequence.agents` inherits that ordering while being explicitly documented as "NOT run order" — reasonable for a reporting-only structure, just flagging that any future caller that tries to use `sequences[].agents` order for execution would need a different source of truth.
- The env-to-union cast on `BODHI_GATE_MODE` (`mode: (env(...) as 'background' | 'print')`) and the still-missing test coverage for `launchGate`'s manual-posture branch (mkdir + `.mcp.json` write) remain open suggestions from the previous review; this diff doesn't touch either, so they're unchanged in status.

William-Long-II:
Answering the one thing flagged as worth checking, and pointing at where the rest is going.

## An unset role is refused, not launched

`driver.ts` `openGates` (the code that consumes `target.agents[gate]`, not visible in this hunk):

```ts
const agent = target.agents[action.gate];
if (!agent) {
  result.problems.push(
    `gate ${action.gate} has no role recorded for this run, so it was not started`,
  );
  continue;
}
runs.startGate({ id: randomUUID(), runId, gate: action.gate, agent, posture: target.posture });
```

There is the `continue`. The gate is dropped from `allowed` before anything spawns, and the row is opened only for gates that survive — so a gate with no role cannot reach `launchGate`, and cannot leave a `run_gates` row claiming an agent ran.

Not just read but observed: this exact message is what the first real run printed when gate 3 had no role, before this change. That is what prompted `rolesFromHarness`.

## The alphabetical-order flag was the right instinct

> any future caller that tries to use `sequences[].agents` order for execution would need a different source of truth

Correct, and that source of truth is now in review upstream: **claude-team-workflow#323** adds `gate_order: 1` to `verifier` and `gate_order: 2` to `scribe`.

Deliberately the harness's to state rather than this repo's. **Which** agent must precede which is domain knowledge about what gate 4 means; running them in declared order is sequencing. A runner hardcoding "verifier then scribe" would hold a fact about that workflow which it would then have to keep in step with a plugin free to change it — and the two would disagree silently, because nothing would compare them.

So `GateSequence.agents` stays documented as not-run-order until the runner reads `gate_order`, at which point the sequence stops being reporting-only. That is the next slice.

## The two carried-over notes

Both stand, both tracked rather than forgotten:

- **gate 2's `background` default leaves the channel inert** — #291. The channel is not consulted at all for a `--bg` gate: `--permission-prompts` applies "with `--print`", and `--bg` conflicts with `--print`. Measured, not inferred: the broker process started and was never asked anything.
- **`launchGate`'s manual-posture branch has no test** — fair. It is the mkdir-and-write half, and it is currently only covered by having been driven end to end. Worth a real test; noting it rather than claiming otherwise.

PR #293, opened 2026-09-14
test(terminal): wait for the startup resize, not for 60ms

William-Long-II:
## What failed

`test (ubuntu-latest)` went red on #290 — a PR that touches the run engine and nothing else:

```
(fail) a guest asking to be fitted to their screen > nothing happens until the owner answers
Expected  - 1
Received  + 3        Terminal.resizeRequest.test.tsx:155
```

Line 155 is `expect(ptyResizes).toEqual([])`. Three entries in an array asserted empty.

## Why

```ts
async function renderTerminal() {
  const rendered = render(<Terminal ... />);
  await act(async () => { await new Promise((resolve) => setTimeout(resolve, 60)); });
  ptyResizes = [];      // <- the clear
  termResizes = [];
```

A **fixed 60ms** for the startup path to settle, then clear. On a loaded runner the pty's own startup resizes landed *after* that window closed, and therefore after the clear — so they were still arriving when the test made its assertion.

The failure message then named the prompt, which was the one part working correctly. That is the expensive kind of flake: it points at the wrong component.

## The change

Wait for the startup resizes to **arrive** rather than for a duration:

```ts
await waitFor(() => { expect(ptyResizes.length).toBeGreaterThan(0); });
await act(async () => { await Promise.resolve(); });
ptyResizes = [];
```

A condition, so a slow runner makes the test slower instead of red.

## Confirmed, not assumed

- Shrinking the old fixed wait to **0ms** reproduces a failure in the same file — so the fixed-duration wait is the mechanism, not the assertion.
- The same CI job **passed on re-run with no change at all** — so it is timing, not the diff.
- The replacement passed **three consecutive** local runs; renderer suite green at 458.

## Also here

`.run-console-permissions/` added to `.gitignore`. It is already ignored on the run-engine branch, but not on `development` — and an unrelated `git add -A` on this branch duly committed 56 permission receipt files before I caught it. Better to have the rule on `development` than to rely on noticing.

## Why a PR rather than a re-run

This is the **second** flake this week to turn a green change red on a PR about something else — #282 was the first, and it cost a review cycle because the reviewer correctly declined to approve over a red check. A check that fails for reasons unrelated to the change is the mechanism by which a genuine red eventually gets waved through.


brannon-bowden (approved):
This is a solid test-reliability fix: replacing an arbitrary 60ms `setTimeout` with an explicit `waitFor` on the actual startup resize event addresses a real, well-diagnosed CI flake, and the `.gitignore` addition is trivial and unrelated-but-harmless.

## Suggestions (non-blocking)

- `Terminal.resizeRequest.test.tsx:135-142`: `waitFor` stops as soon as `ptyResizes.length > 0`, i.e. after the *first* startup resize lands. The comment above it notes the original flake involved multiple startup resizes (3 entries) arriving together on a loaded runner. If those resizes can also arrive as a trickle across separate ticks rather than as one batch, this fix could still race: `waitFor` resolves after entry #1, the arrays get cleared, and entries #2/#3 land afterward and pollute the subsequent assertions — the same failure mode this PR is fixing. Worth confirming (from the `Terminal` component's startup path, not shown in this diff) that all startup resizes are dispatched within the same synchronous batch/tick before the `length > 0` check can observe any of them; if not, waiting for a stable count (e.g. no change over two consecutive polls) would be more robust than "at least one."

William-Long-II:
Rebased onto `development` after #290 merged. The `.gitignore` hunk is gone from this PR — #290 already carries the `.run-console-permissions/` rule, so keeping it here would have been a conflict over an identical line. The commit message no longer claims it either.

What is left is the one test-helper change this PR is actually about. Approval was on `4caca51`, which no longer exists; head is `a983d41` and the diff is strictly smaller. Re-requesting rather than merging on the old SHA.

brannon-bowden (approved):
This is a solid, well-justified test-reliability fix — replacing an arbitrary 60ms sleep with an explicit wait for the startup resize to actually arrive, and the docstring clearly documents the flake it's addressing. Approving.

- `src/renderer/components/__tests__/Terminal.resizeRequest.test.tsx:139-145` — the updated docstring says pty startup "sends resizes of its own" (plural) and describes the original flake as `ptyResizes` ending up with **3** entries after the clear. The new `waitFor` only waits for `ptyResizes.length` to become `> 0` (i.e., the *first* resize), then does a single extra microtask flush (`Promise.resolve()`) before clearing both arrays. If the remaining startup resizes are dispatched via separate timers/macrotasks rather than being chained off the first one synchronously, they could still land after the clear on a sufficiently loaded runner — reproducing the same flake this PR is fixing, just less often. Worth confirming this is stable under load (e.g. by waiting for the resize count to stop changing for a tick, or asserting the expected startup count) rather than just "at least one arrived."

PR #295, opened 2026-09-14
fix(auto-updater): report a macOS install that quit but never landed

### LINKED ISSUE #294: macOS self-update downloads and "installs" repeatedly but never actually swaps the app bundle
## Summary

On macOS, Bodhilander's auto-updater has been silently failing to install for roughly three weeks (since ~2026-08-23) on at least one user's machine. Every check-download-restart cycle reports success in the app's own logs, but the app relaunches on the same old version every time, with no error surfaced anywhere.

Observed on a real machine: installed app was stuck on `3.5.1-beta.6` (2026-08-19) while `3.5.1-beta.14` (2026-09-08) was the latest release — 8 versions and 3 weeks behind, with the updater believing every one of those had installed.

## Evidence

- `~/Library/Logs/bodhilander/main.log` shows the expected happy-path sequence every time: `Update available` → `Starting update download...` → `Update downloaded: X` → the native Squirrel.Mac proxy serves the zip (`... requested by Squirrel.Mac, pipe .../pending/....zip`) → `Proxy server for native Squirrel.Mac is closed` → `[shutdown] cleanup complete; exiting` (or `cleanup exceeded budget; forcing exit so the updater can install`). The process exits cleanly and a new process starts a few seconds later — but `Info.plist`'s `CFBundleShortVersionString` shows the OLD version every time.
- `~/Library/Caches/com.bodhilander.app.ShipIt/ShipIt_stderr.log` — the macOS Squirrel.Mac install helper's own log — has **no entries after 2026-08-23**, even though the app-side logs show many "successful" update-and-restart cycles since then (beta.7 through beta.14, 2026-08-25 through 2026-09-08). ShipIt is simply never being invoked/observed again after that date; it neither logs success nor failure nor a crash report (`~/Library/Logs/DiagnosticReports` has nothing for ShipIt or Bodhilander).
- `~/Library/Caches/com.bodhilander.app.ShipIt/ShipItState.plist` and the `update.<random>/Bodhilander.app` staging directory it points at *do* contain a fully-extracted, correctly-signed copy of the latest version (verified with `codesign --verify --deep --strict` and `spctl -a -vv`, both pass) — so the download/extract/verify side of the pipeline is fine. The failure is specifically in the handoff that's supposed to swap it into `/Applications/Bodhilander.app`.

## What I ruled out

Diffed `v3.5.1-beta.6..v3.5.1-beta.14` on `development` looking for a regression around when this started (2026-08-23/25):
- No changes to `src/main/shutdown.ts`, `src/main/auto-updater.ts`, or `src/main/quit-state.ts` in that range (unchanged since well before 2026-08-15).
- No changes to `electron-builder.yml`'s `mac:` section (signing/entitlements/hardenedRuntime all identical).
- No changes to `.github/workflows/release.yml` other than an unrelated `retention-days` addition to an artifact upload step.
- `electron`, `electron-updater`, and `electron-builder` versions unchanged in `bun.lock` across the same range.
- Not a macOS update on the affected machine — `softwareupdate --history` shows the last OS update was 2026-02-18, long before the failures started.

So whatever broke the ShipIt handoff on 2026-08-23 isn't visible as a diff in this repo, and doesn't appear to be a local OS update either. I was not able to fully pin the native-side root cause with the tooling available (no crash report, no ShipIt log entry to explain the silence).

## Impact

This isn't just "one slow update" — because the failure is invisible to the user (the app never says anything went wrong, it just keeps re-showing "Update downloaded, restart to install"), a user has no way to notice they're stuck on a 3-week-old build short of manually diffing `Info.plist` against GitHub Releases, which is how this was found.

## Ask

1. Root-cause why ShipIt is not being invoked (or is failing before logging anything) on the native side — likely needs to be reproduced with Console.app/`log stream` open live during an update-and-restart, since after-the-fact log inspection found nothing to go on.
2. At minimum, make a failed install **loud and self-recovering** instead of silent: the app should be able to tell, on the next launch, that it restarted to install version X and is still running the old version, and tell the user (and clean up whatever's stuck) instead of quietly repeating the same failed cycle forever.

## Repro / workaround used

The already-downloaded, already-verified update sitting in the ShipIt staging directory was manually moved into `/Applications/Bodhilander.app` (after backing up the old app bundle) to get the affected machine current. That's a workaround, not a fix — the next auto-update attempt would hit the same silent failure.

brannon-bowden:
## What

macOS auto-update was silently failing on a real machine for ~3 weeks (`3.5.1-beta.6` → 8 releases behind `3.5.1-beta.14`). Every check-download-restart cycle logged success; `ShipIt_stderr.log` (Squirrel.Mac's own native install helper log) had **zero entries** across all of it, and nothing in the app noticed or said anything. Full forensic writeup is in #294 — ruled out a code/config/dependency regression in this repo and a local macOS update in that window; the native-side root cause of why ShipIt isn't invoked couldn't be pinned down after the fact from logs alone.

Since the native failure can't be reliably root-caused from static analysis, this makes the failure **observable and non-repeating** instead of trying to guess a fix for a mechanism this app can't see into:

- `update-downloaded` now persists the version it's about to restart into (`pendingUpdateVersion` preference), before either restart path (main-window dialog or the About-dialog "Restart Now" button) can fire.
- On the next `initAutoUpdater()`, `checkPendingInstall()` (pure, unit-tested, electron-free — same pattern as `shutdown.ts`/`quit-state.ts`) compares that pending version against `app.getVersion()`. A mismatch means the restart-to-install didn't take effect: it's logged as an error and surfaced via a native `Notification` pointing at the manual-download fallback, instead of silently looping.
- `pendingUpdateVersion` is machine-local bookkeeping (like the existing one-time migration markers), so it's added to `LOCAL_PREFERENCE_KEYS` in `bundle-format.ts` — the export-allowlist policy test (`bundle-preference-policy.test.ts`, #227) fails on any new preference key that isn't classified either way, and it caught this one as expected.

## Testing

- `bun test src/main/update-verify.test.ts` — new pure-logic tests for the mismatch detection.
- `bun test src/main/transfer/__tests__/bundle-preference-policy.test.ts` — confirms the new preference key is classified.
- `bun test src/main/` — full main suite; `npx tsc -p tsconfig.main.json --noEmit` — clean. The handful of `src/main/` test failures unrelated to this change (`Export named 'safeStorage'/'ipcRenderer' not found in module 'electron/index.js'`, `getWSLDistros`) were confirmed pre-existing on `development` via `git stash` before/after comparison — not introduced here.

Closes #294

https://claude.ai/code/session_01HyYMxHde9mztcQFYXh1Jx6


src/main/auto-updater.ts
```diff
@@ -80,9 +85,43 @@ function broadcastToAllWindows(channel: string, ...args: unknown[]): void {
...
+  if (nextPendingVersion === null) {
+    deletePreference(PENDING_UPDATE_VERSION_PREF_KEY);
+  } else {
+    setPreference(PENDING_UPDATE_VERSION_PREF_KEY, nextPendingVersion);
+  }
+
+  if (!failedInstall) return;
+
+  log.error(
+    `[auto-updater] Restarted to install ${failedInstall.expected} but is still running ${failedInstall.actual} — the macOS install did not take effect`
```
William-Long-II, line 118: This message hardcodes "macOS" but verifyPendingInstall() has no process.platform guard, so it also runs (and can fire this exact text) on Windows and Linux, where it would be factually wrong. Either gate the check to darwin or generalize the wording.

src/main/auto-updater.ts
```diff
@@ -219,6 +258,11 @@ autoUpdater.on('download-progress', (progress) => {
 // Update downloaded
 autoUpdater.on('update-downloaded', (info: UpdateInfo) => {
   log.info('Update downloaded:', info.version);
+  // Record what we're about to restart into (#294) *before* any restart path
+  // (this dialog's "Restart Now", or the About dialog's restart button) can
+  // fire, so a silently-failed macOS install is caught on the next launch
+  // regardless of which path the user took.
+  setPreference(PENDING_UPDATE_VERSION_PREF_KEY, info.version);
```
William-Long-II: This sets the pending marker at download-completion rather than "right before quitAndInstall()" as the top-of-file comment states. If the app never actually installs (dismissed dialog + abnormal termination before an eventual auto-install-on-quit), the next launch will report a false failed install. Worth confirming this is covered by electron-updater's autoInstallOnAppQuit default, or clearing the marker on an explicit "decline this update" path if one exists.

src/main/auto-updater.ts
```diff
@@ -80,9 +85,43 @@ function broadcastToAllWindows(channel: string, ...args: unknown[]): void {
...
+function verifyPendingInstall(): void {
+  const { nextPendingVersion, failedInstall } = checkPendingInstall(
+    getPreference(PENDING_UPDATE_VERSION_PREF_KEY),
+    app.getVersion()
+  );
+
+  if (nextPendingVersion === null) {
+    deletePreference(PENDING_UPDATE_VERSION_PREF_KEY);
+  } else {
+    setPreference(PENDING_UPDATE_VERSION_PREF_KEY, nextPendingVersion);
```
William-Long-II: nextPendingVersion is always null in every branch of checkPendingInstall, so this setPreference branch is unreachable. Consider simplifying checkPendingInstall's return shape to drop the unused field, or document why it's kept for future use.

William-Long-II (changes requested):
The `pendingUpdateVersion` marker for catching a macOS install that silently failed to land (#294) is a good idea, and I like the pure-function split in `update-verify.ts` with unit tests, plus excluding the new pref key from portable bundle transfer in `bundle-format.ts`. Two things need fixing before merge.

`verifyPendingInstall()` in src/main/auto-updater.ts has no `process.platform` guard, so it runs on Windows and Linux too, but both the log line and the user-facing notification hardcode "macOS":

```ts
log.error(
  `[auto-updater] Restarted to install ${failedInstall.expected} but is still running ${failedInstall.actual} — the macOS install did not take effect`
);
```

CI already exercises `windows-latest`/`windows-2022` and `ubuntu-latest`, so if this mismatch is ever hit there, users get told their macOS install failed when it wasn't even a macOS box. The PR is scoped to "a macOS install that quit but never landed," so pick one: gate the check/marker-setting to `process.platform === 'darwin'`, or drop the platform-specific wording if this is meant to be platform-agnostic. Small fix, but as-is it's a guaranteed-wrong message on two of three build targets.

Second one: the file comment says the marker is set "right before every `quitAndInstall()`," but it's actually written in the `update-downloaded` handler, before the user has agreed to install anything:

```ts
autoUpdater.on('update-downloaded', (info: UpdateInfo) => {
  log.info('Update downloaded:', info.version);
  // Record what we're about to restart into (#294) *before* any restart path
  // (this dialog's "Restart Now", or the About dialog's restart button) can
  // fire, so a silently-failed macOS install is caught on the next launch
  // regardless of which path the user took.
  setPreference(PENDING_UPDATE_VERSION_PREF_KEY, info.version);
```

The comment itself admits a non-restart path exists (dismiss/defer the dialog). Nothing in this diff clears the marker on that path, so if the app quits or crashes abnormally after download but before an install is ever attempted, the next launch sees `pendingVersion !== currentVersion` and reports a failed install that never happened. This is probably mostly covered by electron-updater's default `autoInstallOnAppQuit` (installs on quit even without "Restart Now"), which narrows it to abnormal termination, and it's self-clearing after one false report. Still, can you confirm that assumption holds, or add a way to clear the marker when the update is abandoned rather than relying on eventual quit?

nit: in `update-verify.ts`, `checkPendingInstall` always returns `nextPendingVersion: null` — there's no branch that produces anything else:

```ts
export function checkPendingInstall(pendingVersion, currentVersion): PendingInstallCheck {
  if (!pendingVersion || pendingVersion === currentVersion) {
    return { nextPendingVersion: null, failedInstall: null };
  }
  return { nextPendingVersion: null, failedInstall: { expected: pendingVersion, actual: currentVersion } };
}
```

which makes the `else` in auto-updater.ts dead:

```ts
if (nextPendingVersion === null) {
  deletePreference(PENDING_UPDATE_VERSION_PREF_KEY);
} else {
  setPreference(PENDING_UPDATE_VERSION_PREF_KEY, nextPendingVersion);  // never reached
}
```

Not a bug, the always-delete behavior matches the tests, but it's indirection for what's really just `deletePreference(...)` unconditionally. Either collapse the return shape to `{ failedInstall }`, or if there's a retry-count plan that would eventually populate `nextPendingVersion`, drop a comment saying so.

brannon-bowden:
Pushed 7153c91 addressing all three points from the review:

1. **Mac-only guard** — `verifyPendingInstall()` and the `update-downloaded` marker-set now both check `process.platform === 'darwin'` before touching `pendingUpdateVersion`. The log line and notification are worded for Squirrel.Mac specifically, so this stops it from ever firing (with the wrong wording) on Windows/Linux CI or users.
2. **Dead branch** — `checkPendingInstall` no longer returns `nextPendingVersion` (it was always `null`); collapsed to `{ failedInstall }` and the caller now just calls `deletePreference` unconditionally.
3. **Abandoned-download path** — confirmed `autoInstallOnAppQuit = true` (set at module load, line 14) means a normal quit without "Restart Now" still installs, so the marker set on download matches an actual install attempt in that case. The only real gap is abnormal termination (crash/force-quit) between download and quit, which is now called out explicitly in the comment on `verifyPendingInstall`, and self-clears as a single false-positive report on the next launch — didn't add extra clearing logic for that narrow a window given it's self-healing already.

William-Long-II (approved):
Walked both hunks, isDownloading — nit: the self-clearing/single-false-report claim in the new comment depends on `autoInstallOnAppQuit` still being left at its default (`true`) wherever `initAutoUpdater`/`autoUpdater` config is set up elsewhere in this file, that configuration isn't in the shown hunks. If it's ever turned off, the "abnormal termination only" framing no longer holds, worth a quick self-check by the author.

Platform guard fix is clean: `IS_MAC = process.platform === 'darwin'` gates both places that touch the marker now, `verifyPendingInstall()` bails with `if (!IS_MAC) return;` and the `update-downloaded` handler only calls `setPreference(...)` when `IS_MAC`. Marker never gets written on Windows/Linux so there's nothing to misreport there either, and the mac-only log line/notification text is accurate now.

On the abandonment gap I flagged last time, the new docblock comment on `verifyPendingInstall` names the gap directly, dismissing "Restart Now" then dying abnormally before quit, credits `autoInstallOnAppQuit` for covering the normal-quit path, and states the residual exposure is one self-clearing false report. That's fine, an explicit documented tradeoff beats a silent gap, calling this resolved rather than blocking.

`nextPendingVersion` is gone too. `PendingInstallCheck` no longer carries it, `checkPendingInstall` just returns `{ failedInstall }`, caller does an unconditional `deletePreference(...)`, tests updated to match. Cleaner of the two options I suggested, no complaints.

Checked the diff line by line for anything new: early return sits before the `getPreference` call so no dead reads, `deletePreference` still fires unconditionally after the platform check, and the `update-downloaded` guard doesn't touch any other statement in that handler. lgtm.

PR #296, opened 2026-09-14
feat(run-engine): read the order a gate's agents run in

William-Long-II:
Refs #722. Depends on claude-team-workflow#323, which is merged.

## What this reads

Gate 4's order used to live only in prose. #323 made it declarable — `gate_order: 1` on `verifier`, `2` on `scribe` — and this reads it. Against the **real merged harness**, not a fixture:

```
roles      {"3":"reviewer"}
sequences  [{"gate":4,"agents":["verifier","scribe"]}]
unordered  []
```

Run order — which is the **reverse** of the alphabetical list this returned before. That is the entire point: the verifier judges whether green proves anything, and only the scribe opens the PR that carries that evidence. Reversed, the PR claims a verdict nobody reached.

## Read, never decided

Hardcoding "verifier then scribe" in this repo was the tempting fix and would have been wrong. **Which** agent must precede which is domain knowledge about what a gate means; running them in the declared order is sequencing. An engine holding that fact would have to keep it in step with a plugin free to change it, and the two would disagree silently, because nothing would be comparing them.

The reviewer of the previous slice flagged the same thing independently:

> any future caller that tries to use `sequences[].agents` order for execution would need a different source of truth

This is that source of truth.

## `unordered` is a separate list, not a flag

A `GateSequence` in `sequences` is in **run order**. One in `unordered` is alphabetical. Keeping them apart means a caller holding an unordered entry cannot iterate it and believe it sequenced anything — a boolean on one shared list would make that mistake a one-character oversight.

Three things land there, and they mean the same thing to a caller (*nobody said what order to run these in*):

- **a missing position** — one missing is worse than none: the gate *looks* ordered and is not.
- **a tie** — not an order. It leaves the sequence to whatever the filesystem listed first, which is how one initiative runs differently on two machines and both report honestly.
- **a non-numeric position**.

## The reader is strict on purpose

Anything that is not a run of digits is `null` rather than coerced. `Number('')` is `0` and `parseInt('2nd')` is `2` — a lenient reader turns a declaration nobody checked into a position this engine acts on, which is the exact failure #323's invariants exist to prevent, arriving through the reader instead.

Also covered: positions sort as numbers (`"10"` must not land between `"1"` and `"2"`), and they need not start at 1 or be contiguous — the harness declares relative order, not an index, so inserting an agent should not mean rewriting every other declaration.

## Mutation-checked — and one check caught me first

| mutation | tests failed |
|---|---|
| lenient reader (`Number(raw)` instead of digits-only) | **4** |
| drop the tie check | 1 |
| sort positions as text | 1 |

Worth recording how the first one went: my initial `sed` silently failed to match, and reported "no tests failed" for a mutation **that was never applied**. I nearly wrote it up as a weak test. Re-applied properly via an anchored replace that asserts the match, it fails 4.

A mutation that did not happen is not evidence, and it is indistinguishable from a test that cannot fail. Given this engine has already produced three tests that were structurally incapable of failing, that is worth more than a footnote.

## What this does not do

`ExecutorTarget.agents` still holds **one role per gate**, so gate 4 is reported rather than run:

```
note  gate 4 runs verifier then scribe, which this model cannot hold yet
```

Actually executing a sequence — two gate rows, an ordered launch, a short-circuit when the first fails — is the next slice. This one lands the reading, so that slice is about execution alone.


brannon-bowden (approved):
This PR is well-executed and I'm approving it — the ordering logic, null/tie handling, and test coverage all hold up under close inspection, and CI is fully green (7/7 checks passing).

## Analysis

- `gateOrder()` is appropriately strict: `/^\d+$/` after `.trim()` correctly rejects `''`, `'2nd'`, `'-1'`, `'1.5'` while accepting plain digit strings — the test-each block (`gate-launcher.test.ts:471-484`) directly exercises the cases the JSDoc calls out (`Number('')` and `parseInt('2nd')` footguns), and the logic matches.
- `sequence()` (`gate-launcher.ts:255-260`) correctly treats "any null" and "duplicate orders" (via `Set` size check) as equally unresolvable, falling back to `null` → caller routes to `unordered` with alphabetical fallback. This generalizes correctly beyond the 2-agent cases shown in the tests (a partial tie among 3+ claims, e.g. orders `[1,1,2]`, would still correctly invalidate the whole ordering via the `Set` size mismatch).
- `rolesFromHarness()` consistently threads `unordered` alongside `sequences`/`roles`/`unclaimed`, and the "everything gets accounted for" test (`gate-launcher.test.ts:531-536`) was updated to include the new list, so nothing silently falls through the cracks.
- `run-console.ts` consumes `sequences` and `unordered` with distinct, appropriately-worded messages, and doesn't conflate an ordered sequence with an unordered one.
- Numeric sort correctness ("10 must not land between 1 and 2") and sparse/non-contiguous positions are both explicitly tested.

## Suggestions (non-blocking)

- `rolesFromHarness` now re-implements the filter/sort-by-name logic that `agentsForGate` already encapsulates (`gate-launcher.ts:226-229` vs `gate-launcher.ts:149-152`), rather than reusing it. Minor duplication — worth a follow-up if `agentsForGate`'s behavior ever needs to change, since the two code paths could drift.
- Consider a test for a partial tie among three or more claims on one gate (e.g., orders `1, 1, 2`) to explicitly pin the "one tie poisons the whole gate" behavior, since it's currently only verified indirectly through the `Set`-based implementation rather than a dedicated test case.

William-Long-II:
Both suggestions answered; one is now a test.

## The partial tie is pinned

Taken — it was covered only by the `Set` implementation, which is not the same as being stated. Three claims, two tied at position 1, one untied:

```
one tie poisons the whole gate, not just the tied pair
  -> sequences: []   unordered: [{ gate: 4, agents: [a, b, c] }]
```

The untied third cannot be sequenced around an ambiguity either: *"a then (b or c)"* is not an order, and a caller handed a partial one would run it as though it were. All-or-nothing is the only answer that cannot be misread.

Mutation-checked: narrowing the tie check to compare two adjacent positions fails this test and **nothing else**, which is exactly why it earns its place.

## The duplication is real, and deliberate for now

> `rolesFromHarness` now re-implements the filter/sort-by-name logic that `agentsForGate` already encapsulates

Correct, and worth being precise about why rather than just agreeing. The two no longer want the same thing: `agentsForGate` returns **names**, while `rolesFromHarness` needs **claims** — name plus declared position — because it has to distinguish "no order declared" from "order declared and unusable". Routing the second through the first would mean reading every agent file twice, once for names and once for orders.

The right fix is to make the roster the shared primitive and have `agentsForGate` become `roster(...).map(c => c.name)`. I have not done it here because `agentsForGate` is public and separately tested, and folding it into a new primitive in the same PR that introduces the primitive makes both changes harder to review. Left as the first thing the sequence-execution slice does, since that slice already touches this file.

Noting the drift risk you raised is the real point: if `agentsForGate`'s filtering ever changes, these two paths diverge silently. That is worth a follow-up rather than a shrug.

brannon-bowden (approved):
Approving — this update only adds a test and it does exactly what I previously suggested was missing.

## Analysis

- The new test (`gate-launcher.test.ts:494-508`) exercises the three-claim partial-tie case (`gateOrder` values `1, 1, 2`) that my previous review flagged as untested. It correctly asserts `sequences` is empty and `unordered` contains all three agents, alphabetically sorted — consistent with the `Set`-size-based invalidation logic in `sequence()` that this PR's earlier commits introduced.
- No production code changed in this diff; the behavior under test was already reviewed and holds.
- CI is fully green (7/7).

No blocking issues, no new concerns.

PR #298, opened 2026-09-14
fix(accounts): require sustained idleness before a fail-back respawn

### LINKED ISSUE #297: Account fail-back respawns a session mid-task, killing its running subagents
## Summary

Account failover (moving a session to a backup account when the primary hits its usage limit) works as intended. Fail-**back** (moving the session home once the primary account is healthy again) does not respect an in-flight task: it can trigger a pty respawn while a session is actively running — specifically, while it's mid-way through orchestrating one or more subagents (parallel Task-tool dispatches) — which kills the CLI process and every subagent under it.

Reported symptom: "the harness seems to handle stopping, then auto switching well, the issue is when it's available again, it switches back mid stream pissing off all subagents."

## Root cause

`canFailBackNow(state)` in `src/main/account-failover.ts` only guards on `SessionState` (`idle` or `stopped`) before `sweepFailbacks()` in `src/main/index.ts` respawns a session onto its home account. The intent (per the existing doc comments) is that fail-back only acts on a session "demonstrably between turns."

In practice, `state === 'idle'` is not a reliable signal of that. There are two producers of session state, and only one of them is actually wired up:

- **`pty-manager.ts`'s silence heuristic** (`scheduleIdleTimeout`): after just **2 seconds** without "substantial" terminal output (>10 printable chars), a `working` session is marked `idle` (`event: 'idle_timeout'`). This is the *only* thing that currently drives `idle`/`working` state — see below.
- **`StateMonitor`** (`src/main/state-monitor.ts`): a Unix-socket server that index.ts's comments describe as "the hook-driven monitor [that] sees Claude's own stop events" — implying Claude Code's `Stop` hook is a second, more reliable idle signal. It isn't, anymore: `registerHooks`/`desiredHookConfigs` in `mcp-config.ts` route both configured hooks (`PostToolUse`, `Stop`) through `bodhilander-hook.ts`, which POSTs to `/api/v1/hooks/post-tool-use` and `/api/v1/hooks/stop` — and both of those routes (`src/main/api/routes/hooks.ts`) only write analytics rows (`turn_complete`, `tool_use` session events). Neither touches `session.state`, emits a `stateChange`, or calls `sweepFailbacks()`. I could not find anything in the current source that connects to `StateMonitor`'s socket and writes it a `StateEvent` — it appears to be dead code from an earlier hook design that predates the HTTP-based analytics hooks, and the comments describing it as active were never updated.

So today, 100% of `idle`/`working` state comes from a 2-second terminal-silence guess, with no corroboration from Claude Code's own hook events despite what the code comments claim.

That silence guess is exactly what the pty looks like while a session is between a parallel subagent's status updates — a Task-tool dispatch can easily go quiet on the parent's terminal for well over 2 seconds while the subagent(s) do their own work. During that blip, `session.state` genuinely reads `idle` in the database, even though the top-level turn is very much still in progress.

Making this worse: `sweepFailbacks()` isn't only called from a 60s timer (`FAILBACK_SWEEP_MS`) — it's also called, unscoped, from **every** session's `stateChange` handler (both wiring points in `index.ts`, around lines 509 and 742). So a session sitting in that few-second false-idle window doesn't need to wait for the timer; any other unrelated session transitioning state at the same moment is enough to trigger a sweep that finds it and respawns it. With several sessions active in parallel (which is exactly when someone would be running enough traffic to hit a usage limit and care about failover in the first place), the odds of some other session's state changing during that narrow window are not low.

## Ask

Make fail-back require sustained, not momentary, idleness before it's willing to respawn a session — the mechanical trigger fix. Separately (bigger, not required for this fix): either wire `Stop`/hook-driven signals back into session state so `idle` means what the comments already claim it means, or update those comments to stop claiming a corroborating signal exists when it doesn't.

brannon-bowden:
## What

Fail-back (moving a session home once its primary account's usage limit lifts) could respawn a session mid-task, killing whatever subagents were running under it. Full root-cause writeup in #297; short version:

- `canFailBackNow(state)` only checked `SessionState`, trusting `state === 'idle'` as proof a session is "between turns."
- `idle`/`working` state comes exclusively from `pty-manager.ts`'s silence timeout — 2s without substantial terminal output flips `working` → `idle`. That's exactly what the pty looks like while a session is between a parallel subagent's status updates, not evidence the top-level turn ended.
- The doc comments claimed a second, hook-driven signal corroborates this (Claude Code's `Stop` hook). It doesn't, anymore — `Stop`/`PostToolUse` hooks route through `bodhilander-hook.ts` to `/api/v1/hooks/*`, which only write analytics rows. Nothing connects to `StateMonitor`'s socket and feeds it a `StateEvent`. Comments corrected to stop claiming corroboration that isn't there; reconnecting that signal is a separate, bigger change not required for this fix.
- `sweepFailbacks()` also isn't only a 60s timer — it runs unscoped on every session's state change, so a session sitting in that few-second false-idle blip has many chances (any other session's transition) to get caught, not just the timer tick.

## Fix

`canFailBackNow(state, idleSinceMs)` now also requires sustained idleness: a `stopped` session (no pty to interrupt) is exempt and goes through immediately; an `idle` session only counts once `idleSinceMs >= FAILBACK_MIN_IDLE_MS` (60s, matching the existing sweep cadence). `working`/`waiting` are unaffected — always false. The caller in `sweepFailbacks()` (`src/main/index.ts`) computes `idleSinceMs` from `session.lastActivityAt`, which is already updated on every state transition including the idle one.

## Testing

- `bun test src/main/__tests__/account-failover.test.ts` — updated the existing gate test for the new signature, added a dedicated test for the exact bug (idle just under vs at vs over the threshold).
- `bun test src/main/` — full main suite; `npx tsc -p tsconfig.main.json --noEmit` — clean. Six pre-existing failures (electron `safeStorage` mock gaps, one missing built `dist/hooks` artifact in this checkout) are unrelated — none touch `account-failover`, `index.ts`, or fail-back, confirmed by file name.

Closes #297

https://claude.ai/code/session_01HyYMxHde9mztcQFYXh1Jx6


William-Long-II (approved):
I'm going to rewrite the review as a real dev's comment, keeping every finding/claim/reference intact.

Fix closes #297: `canFailBackNow('idle')` used to be an unconditional `true`, but pty-manager's idle detection is just a 2s silence timeout, so a subagent's normal output gap could trip a mid-task respawn and kill its work. Requiring idle to hold for a full 60s sweep interval (`FAILBACK_MIN_IDLE_MS`) turns that blip into a non-event while a real between-turns pause still clears the gate. Logic checks out: `stopped` short-circuits to `true` regardless of `idleSinceMs` (fine, no pty to interrupt), `working`/`waiting` stay `false` always, and `idle` requires `idleSinceMs >= FAILBACK_MIN_IDLE_MS`. Boundary is inclusive at exactly 60000ms and the tests pin it precisely (`MIN_IDLE_MS - 1` → false, `MIN_IDLE_MS` → true, `MIN_IDLE_MS + 1` → true).

Call site in `index.ts` derives `idleSinceMs` as `Date.now() - session.lastActivityAt.getTime()` after confirming the session exists, keeping the original early-exit and just adding the duration check ahead of `canFailBackNow`. Since `sweepFailbacks()` fires both on a timer and on every state-change event, computing idle time off a wall-clock timestamp instead of trusting the state machine alone is the right call — a rapid re-entry into `idle` won't have accumulated 60s yet.

Also like the doc comment fix — it was claiming a "hook-driven monitor" corroborates the pty's silence-based idle detection, which isn't true; nothing wires Claude Code's Stop hook to that socket right now, so the silence timeout is the only signal in practice. Good to have that written down honestly.

Two things I'd want answered before merging, though not blockers:

The diff only tests the pure `canFailBackNow` function, nothing exercises the `sweepFailbacks` wiring itself — that `Date.now() - session.lastActivityAt.getTime()` line could get a sign flip or field swap someday and the pure-function tests wouldn't catch it.

Bigger question: this whole thing hinges on `lastActivityAt` meaning "last terminal/output activity" and being reset exactly at the idle transition. If something else bumps `lastActivityAt` while already idle, the 60s window keeps resetting and the session could just never qualify for fail-back. Can you confirm the update semantics there? Not visible in this diff.

lgtm otherwise, core fix is solid.

PR #299, opened 2026-09-15
BDH-239 — the handoff DELETE draws its own rate-limit bucket

### LINKED ISSUE #239: The handoff acknowledgement (DELETE) draws from the read rate-limit bucket
Raised by @brannon-bowden as a nit on #234.

In `relay/src/http.ts`, all three non-upload handoff verbs share one bucket:

```ts
if (method === 'GET') {
  return limited(req, peerIp, 'handoff-read', HANDOFF_READ_PER_IP) ?? …;
}
if (method === 'DELETE' && !handoff.bundle) {
  return limited(req, peerIp, 'handoff-read', HANDOFF_READ_PER_IP) ?? …;
}
```

`DELETE` is the acknowledgement a machine sends once a restore has landed — a mutating call, charged against a budget named for reads. Not a bug: the limit is per-IP and generous (30/window) next to what a restore actually does, and the delete is idempotent enough that `handoff-transport.ts` treats a `404` as the state it wanted.

It is worth tidying for two reasons:

- A burst of restore-and-acknowledge cycles spends budget that metadata polling also draws from. `peek()` runs on every launch and on every relay status change, so the two genuinely compete.
- The bucket name is what someone reads when a 429 shows up in a log. `handoff-read` pointing at a delete costs a few minutes of confusion at exactly the wrong time.

### Suggested shape

Give the acknowledgement its own bucket — `handoff-ack`, with its own constant. Whether the limit differs from the read limit matters less than the two not sharing a budget or a name.

### Acceptance criteria

- [ ] The acknowledgement is charged against a bucket named for what it does
- [ ] Exhausting the read budget does not block a machine from acknowledging a restore it has already completed, and vice versa
- [ ] A test pins the two budgets as independent

William-Long-II:
Closes Software-Development-LLC/Bodhilander#239

The handoff `DELETE` — the acknowledgement a machine sends once a restore has
landed — was charged to `handoff-read` at 30/window, the same bucket metadata
polling draws from. A restore that polled for its offer could spend the budget
and then be refused the call that clears the bundle it had already taken, so
the bundle sat on disk until it expired and the account slot stayed occupied.

It now draws on `handoff-delete` at 10 per window: one acknowledgement per
restore, with room for replies lost in transit.

## Acceptance criteria

- [x] The acknowledgement is charged against a bucket named for what it does —
      `handoff-delete`, its own constant `HANDOFF_DELETE_PER_IP`
- [x] Exhausting the read budget does not block a machine from acknowledging a
      restore it has already completed — `but a restore that read its fill can
      still acknowledge` drives the read bucket to 429 and asserts the DELETE
      still returns 204
- [x] A test pins the two budgets as independent — the above, plus
      `acknowledging is bounded on its own`, which asserts the 11th
      acknowledgement is 429 while the read bucket is untouched

The issue suggested the name `handoff-ack`; this uses `handoff-delete`, which
names the verb the route actually is. The issue called the name secondary to
the two not sharing a budget.

## Verified

Harness contract check, at this branch:

    scripts/verify.sh Bodhilander=feat/BDH-239-Bodhilander --initiative <initiative>

    base origin/development · commits base 8ad3515 ref 75f66ad · 2 files changed
    PASS  tests      (6 specs enumerated)
    SKIP  typecheck  SKIP lint  SKIP seams
    GREEN for what ran.

Relay package, scoped:

    bun test relay/src/     →  205 pass, 0 fail, 2546 expect() calls, 11 files

Full repository, in the isolation mode CI uses:

    bun test --isolate      →  1965 pass, 2 skip, 0 fail, 6591 expect() calls, 121 files

The bare, non-isolated `bun test` reports 26 failures and 5 errors. That is a
pre-existing property of the suite, not this change: run at `origin/development`
(8ad3515) it produces the same 26/5, and the 21 named failing tests are
character-identical between the two runs. None of them are in `relay/src`.

Merge state: merges into `development` cleanly (`git merge-tree --write-tree`
exits 0), and nothing on `development` has touched `relay/src/http.ts` or
`relay/src/handoff-http.test.ts` since this branch's merge base.

## Not verified

- **Type-check over the changed file.** Nothing type-checks `relay/` on any
  branch — `tsconfig.main.json` includes only `src/main/**` and `src/shared/**`,
  and bun executes TypeScript without checking it. Not introduced here and not
  fixable from this change.
- **Lint.** This repo has no eslint config and no lint script, so the harness
  SKIP is verified-absent rather than a gap.
- **`HANDOFF_DELETE_PER_IP` is pinned at 10 only from above.** A mutation
  narrowing it from 10 to 1 survives the full relay suite, because the first
  test spends one acknowledgement and the second asserts only that the 11th is
  429. Widening it to 100 is caught. The shipped value is correct; a later
  narrowing would not be.
- **Sonar / CI quality gate.** Had not run at the time this was written — no PR
  existed. The checks on this PR are the first execution.
- **Board status for #239.** Could not be read or set: the available token
  lacks the `read:project` scope. The issue is assigned; its column is not
  confirmed moved off Todo.

No Design Document or Testing Plan exists for BDH-239 — it was bootstrapped as
a single-repo task, and `bodhi-code/initiatives/` has no BDH-239 directory to
link.


brannon-bowden (approved):
This is a clean, narrowly-scoped change — approving. Splitting the handoff acknowledge (DELETE) rate limit into its own `handoff-delete` bucket instead of sharing `handoff-read` is a straightforward, correct edit (`relay/src/http.ts:105-106`, `relay/src/http.ts:271-272`), and both new tests confirm the arithmetic exactly matches the stated constants (30 for read, 10 for delete), following the same "loop N+1, expect 429 on last" idiom already used elsewhere in the file.

- `relay/src/handoff-http.test.ts:328-334` — `acknowledging is bounded on its own` calls `acknowledge` 11 times on the same `id` but only asserts the status of the final call. It would more precisely verify the boundary (and the implicit assumption that repeated acknowledges on an already-deleted handoff still return 204 rather than erroring) if it also asserted the first 10 calls return 204.
- Coverage is currently one-directional: the new tests show that exhausting the read bucket doesn't block acknowledge, but not the reverse — that exhausting the new delete bucket doesn't affect reads. A quick test for that direction would round out confidence that the two buckets are fully independent.
- Nit: the test name `but a restore that read its fill can still acknowledge` (`relay/src/handoff-http.test.ts:320`) reads like a sentence fragment continuing from the previous test rather than a self-contained description — worth a small rename for clarity when skimming test output.

Nothing here blocks merge; CI is fully green across all 7 checks.

PR #300, opened 2026-09-15
feat(run-engine): run a gate's roles in sequence, and tell the machine once

William-Long-II:
Refs #722. Builds on #296 (reading `gate_order`); the state machine is untouched.

## What happened when this ran

The first complete run of the engine, gate 2 through gate 4, ending in a PR the engine opened itself — **#299**, for BDH-239:

```
31b5e5fb  waitingChecks  BDH-239
    gate 2 done (bodhilander-lead, attempt 1)
    gate 3 done (reviewer,         attempt 1)
    gate 4 done (verifier,         attempt 1)
    gate 4 done (scribe,           attempt 1)      <- its own attempt 1, not the gate's 2nd
```

```
00:06:41  prepared
00:06:42  provisioned
00:11:32  gateFinished      gate 2
00:19:14  gateFinished      gate 3
00:55:43  gateFinished      gate 4   <- ONE event for two roles
```

The verifier's pass did not produce a `gateFinished`. That is the whole change.

## Where the sequence lives

In the driver, between the executor's report and the machine's ear. A step passing is not the gate passing: when the verifier reports, gate 4 is half done, and telling the machine `gateFinished` there would send the run to `waitingChecks` with **no PR open** — the scribe, the only role that opens one, has not run. So the driver closes the step's row, opens the next, spawns it, and tells the machine nothing until the *last* role reports.

A failure at any step goes straight through. Red is red whoever found it, and the machine already knows what red means: back to the owner, and the scribe never runs on a branch the verifier called soft.

## What changed shape

- **`ExecutorTarget.agents` is a list per gate, in run order.** A one-element list for the common case; the harness's `gate_order` for a sequence, read and never decided here.
- **`spawnGate(gate, agent)`** — the dependency is told *which* role. The executor must not reach into `agents` and pick, or the sequencing decision lives in two places and the second wins silently. The driver resolves the role when it opens the row and hands the executor a `ResolvedAction` that names it.
- **Attempts are counted per role.** A scribe following a green verifier is the same attempt's second half. Counting per gate would make every sequence read as a retry loop — the one thing that column exists to make visible.
- **A report from a role outside the sequence** goes to the machine as the gate's own (what every gate did before sequences existed) *and* is a problem, not a silence.

## Mutation-checked

| mutation | tests failed |
|---|---|
| no step continuation — every step reported to the machine | 2 |
| a fresh spawn opens the *last* role instead of the first | 1 |
| attempts counted per gate again | 1 |

One test was wrong before the code was: a fake that reported every role green let a red verifier walk owner → reviewer → verifier → scribe in a single call, so the scribe appearing was *correct* rather than the thing the test forbids. The owner now reports `launched` so the run pauses where the assertion looks.

## What the gates themselves did, since it is the first time anyone can say

- **gate 2** verified the existing fix rather than redoing it, as asked.
- **gate 3** wrote its own mutation script and recorded that `HANDOFF_DELETE_PER_IP` is pinned only from above.
- **gate 4 / verifier** mutated the change three ways with `cmp`-restored copies and found the same gap independently.
- **gate 4 / scribe** cut a temporary worktree at `origin/development`, ran the full suite there as a baseline, and diffed the failing-test sets to prove the change adds no new failures; wrote `decisions.md`; and **deliberately dropped the stock Claude Code footer** from the PR body because `verify-merge-order.sh` treats it as proof the PR was *not* opened by the scribe.

Two honest gaps it left, both recorded in its receipt: the PR body still says CI "had not run" (it had — 7/7 green by then), and #299 has **no reviewer**, because GitHub refused the request that included the commit author and appears to have dropped the whole call. The engine's own `requestReview` action covers the second once the run leaves `waitingChecks`.

## Driven, not automated

Every permission request in this run was answered by me under one stated rule — allow anything confined to the disposable worktree or a plain `gh` read; stop on anything that pushes, creates, edits or comments — and every stop was read in full before it was approved. The policy stopped six times. Each was the right place to stop.


brannon-bowden (changes requested):
This PR is well-designed and thoroughly tested — the sequencing model (drive multiple roles per gate, tell the state machine only once) is implemented consistently across `driver.ts`, `executor.ts`, `runs.ts`, and the CLI, and the new `driver.test.ts` cases cover the pass/fail/inconclusive/attempt-counting/unknown-role edge cases well. I'm requesting changes only because required CI checks are currently red.

## Issues

- **CI is red**: `SonarQube Code Analysis` and `quality-gate` are both failing. Per this repo's merge policy, these need to be green (or the failures explained/waived) before merge — please check the Sonar findings and address them or report back if they're false positives.

## Suggestions (non-blocking)

- `scripts/run-console.ts` — in the `spawnGate` dep, `channelKeyFor(runId, gate, agent, turn?.attempt ?? 1)` silently falls back to attempt `1` if `runs.activeGate(runId)` returns nothing. Per the surrounding comment, the row is always expected to exist by this point (since `startStep` inserts it before `spawnGate` is invoked), so if that invariant is ever violated the fallback would produce a wrong/duplicate channel key rather than surfacing the problem. Consider throwing or logging loudly instead of quietly defaulting, so a broken invariant fails fast rather than manifesting as a mysterious channel collision.
- The reasoning-heavy comment style (e.g. in `driver.ts` and `executor.ts`) is thorough but quite long for what are otherwise fairly small functions (`stepAfter`, `startStep`); worth confirming that's the convention the team wants going forward, though it's consistent with the pre-existing style in this file.

William-Long-II:
Both points taken; the blocking one is fixed and the non-blocking one turned out to be worth fixing too.

## Sonar

One new violation, and a fair one:

```
[CRITICAL] typescript:S3776  driver.ts:145
  Refactor this function to reduce its Cognitive Complexity from 25 to the 20 allowed.
```

The step-continuation block was one more decision nested inside a loop that already held several. It is `continueSequence` now — returns the next role's events, or `null` when the machine should hear the event after all — which is the same shape the executor took for the same rule. The step boundary is a separate idea from the round loop and reads as one on its own.

The mutation that disables continuation still fails the same two tests it did before the extraction, so the behaviour moved without changing.

## The `?? 1` fallback

> if that invariant is ever violated the fallback would produce a wrong/duplicate channel key rather than surfacing the problem

Correct, and worse than "wrong": attempt 1 is a key another turn has already used, so a broken invariant would hand a role a channel with someone else's unanswered request in it — a stale request read as the new one's, which is precisely what keying by attempt exists to prevent. It throws now, naming the gate and role it was asked to launch and what row it actually found.

## Comment length

Noted. The convention in these files is that the comment carries the reasoning a reader would otherwise have to reconstruct from git history, and it is the pre-existing style here — but the point that `stepAfter` and `startStep` are small functions under long comments is fair, and I have not added to them in this commit.

Head is `82e5188`; the approval-with-changes was on `49ffbb0`. Re-requesting once CI re-analyses.

brannon-bowden (changes requested):
This update resolves both open items from the last pass — the `run-console.ts` fallback now hard-fails on a broken invariant instead of silently defaulting, and the `advanceOnce` sequencing logic has been extracted into `continueSequence` to address the Sonar complexity complaint — but I'm holding at `request-changes` because required CI checks are still red.

## Issues

- **CI is still red**: `SonarQube Code Analysis` and `quality-gate` are both failing on this push. These need to go green (or the failures need to be explained/waived per this repo's merge policy) before merge.

- `scripts/run-console.ts:271-291` — the `turn = runs.activeGate(runId)` lookup is now validated (`!turn || turn.gate !== gate || turn.agent !== agent`) and throws a descriptive error instead of falling back to attempt `1`. This directly fixes the silent-default concern raised previously; the channel key at `channelKeyFor(runId, gate, agent, turn.attempt)` can no longer collide with a stale attempt.
- `src/main/run-engine/driver.ts` — the inline gate-sequencing branch in `advanceOnce` was extracted into `continueSequence`. I traced the extracted logic against the original inline version: the `gate ? await continueSequence(...) : null` guard preserves the original `gate?.gate === event.gate` short-circuit, the early-return conditions inside `continueSequence` are the logical negation of the original `if`, and the finish/execute/collect/return sequence is unchanged. Behavior looks preserved; this reads as a pure refactor for readability/complexity, consistent with the comment explaining it was done to satisfy Sonar's complexity limit.

## Suggestions (non-blocking)

- `scripts/run-console.ts` — the new invariant check is good defensive coding, but as a behavioral change (a new thrown-error path) in a file with no test coverage in this diff, it'd be worth confirming this script has some test/manual-verification path, since a typo in the condition would only surface at runtime.
- The dense, narrative comment style in `driver.ts` (e.g. the `continueSequence` doc comment) continues to be verbose relative to the function size — still just a style note, not a blocker.

William-Long-II:
Sonar is green on the current head.

```
5af1de2  qualityGateStatus: OK   new issues: 0
```

The first extraction took `advanceOnce` from 25 to 23; it needed a second. What remained was the transition-recording block — a compound condition with a nested `if` — and the `gate ? … : null` ternary. `recordDecision` now holds the first, carrying the two reasons that block existed for (write when something *happened*, not only when the state moved; close the gate row once the machine has *accepted* the report). `continueSequence` takes a nullable gate, which removes the ternary.

Pure moves, checked as such: the no-continuation mutation still fails the same two tests, and a new mutation — recording only on a state change — fails the test that expects a gate-2 pass to appear in the log while the run stays `running`, plus the red-verifier test whose `backToOwner` leaves the state at `running` while spawning the owner.

On the note that `run-console.ts` has no test coverage for the new throw: correct, and deliberate about where the line is. The console is excluded from Sonar (`sonar.sources` does not include `scripts/`) and has no unit tests because it is the manual-drive harness — its verification path is running it, and it has been run end to end today through gates 2, 3 and both halves of 4 (#299 is the result). The invariant it now enforces is the driver's, and *that* is tested: `a fresh spawn of a sequenced gate starts with its first role` asserts the row exists with the right gate and agent before the spawn fires. A typo in the console's condition would surface on the very next drive, loudly, which is the console's job.

Re-requesting at `5af1de2`.

brannon-bowden (approved):
This resolves the outstanding CI-red blocker from the previous pass, and the new refactor extraction is behavior-preserving — approving.

- CI is now green across the board (SonarQube, quality-gate, tests on both platforms, GitGuardian, postinstall checks) — the previous blocking issue is resolved.
- `src/main/run-engine/driver.ts` — the inline `decision.state !== run.state || decision.actions.length > 0` branch (recording the transition, closing the gate) has been extracted into `recordDecision`. Tracing it against the removed inline code: the early-return guard `decision.state === before && decision.actions.length === 0` is the exact De Morgan negation of the original condition, and the `writeTransition` / `result.applied.push` / `runs.finishGate` sequence inside is unchanged. This looks like a pure extraction for Sonar complexity, not a behavior change.
- `continueSequence` now takes `gate: runs.RunGateRow | null` and does its own `!gate` check internally instead of relying on the caller's `gate ? await continueSequence(...) : null` ternary. The call site was updated to call it unconditionally. This is logically equivalent — the null check just moved from the caller into the callee — and is arguably clearer.

No behavioral change is evident in this diff, so the lack of new tests here isn't a concern.

## Suggestions (non-blocking)

- The narrative comment above `recordDecision` largely duplicates the reasoning already given inline at the original call site (and now removed from there) — worth trimming one of the two copies for a future pass, but purely stylistic.

Issue #287, opened 2026-09-14
a launched gate is never checked again: the engine reports 'gate 2 running' for a process that no longer exists


William-Long-II:
Found on the first real gate launch (#722).

## What happens

The engine launched gate 2, the gate blocked, I killed the process — and the engine's own view never changed:

```
$ bun run console -- status
5c1c16b0-...  running   BDH-239
    gate 2 running (bodhilander-lead, attempt 1)

inbox: 1 waiting on a person      <- a DIFFERENT run. Not this one.
```

The process is gone. `run_gates` still holds an open row. The run is still `running`, and `running` is not in `NEEDS_A_PERSON`, so it is not in the inbox either. Nothing will ever look at it again.

## Why this is the important one

Three pains motivated this engine. The third was **"you can't see what's happening."** A gate that dies silently and leaves a run parked in `running` forever is that pain, reproduced by the thing built to fix it — and worse than the manual version, because a person driving by hand would at least notice their terminal had stopped.

`launched` was designed as a legitimate outcome distinct from `completed`, and that is right. What is missing is the other half: **something has to ask, later, whether a launched gate is still alive.**

## What is actually absent

`reconcile-loop` gives a cadence to checks, review and retryable actions. It gives none to a running gate, because there is no signal to reconcile against. `gate-process.ts` returns `backgroundId` and `sessionId` on launch; neither is read again.

A liveness pass would need to decide three cases, and they are genuinely different:

- **alive and working** — leave it alone.
- **gone without a verdict** — the process exited, crashed or was killed, and no gate report was accepted. This is `gateUndriveable` territory: a result was expected and none exists.
- **alive but not progressing** — blocked on a permission prompt (see the sibling issue), or wedged. This is the one that needs a deadline rather than a liveness probe, because the process is perfectly healthy and will never finish.

The third is why "is the pid alive" is not sufficient on its own. A gate stuck on a prompt is alive, has a session file whose mtime stopped advancing, and is indistinguishable from a working gate by pid alone.

## Notes toward a fix

- `sessionId` gives a transcript path. Its mtime is a progress signal that costs nothing to read and distinguishes "working" from "wedged" in a way pid liveness cannot.
- A gate needs a deadline of its own. `BODHI_GATE_TIMEOUT` (900000) exists but only applies to `print` mode, where `runCommand` enforces it. Background mode returns immediately and the number is never used.
[...]

PR #301, opened 2026-09-15
feat(run-engine): a background gate reaches the permission channel through a hook

### LINKED ISSUE #291: the permission channel is not consulted for a background gate: --permission-prompts applies to --print, and --bg conflicts with it
Found while wiring the permission channel (#290), which works — in a mode gate 2 does not use.

## What happens

A gate launched with `--bg` starts its permission broker and then never asks it anything.

Same broker, same flags, same run. Only the mode differs:

| mode | broker started | broker asked | result |
|---|---|---|---|
| `--print` | yes | **yes** | request appeared, run reached `waitingPermission`, answering released it |
| `--bg` | yes | **no** | gate reached a `Write`, got no result, sat there |

In the background case the broker process was confirmed alive and bound to the right directory:

```
node C:\work\repos\Bodhilander\scripts\permission-broker.js
     C:\work\repos\Bodhilander\.run-console-permissions/b5d1ae9b-...-g2-a1
```

so `--mcp-config` is honoured. The channel directory stayed empty, `NOTES.md` was never created, and the transcript shows **1 `Write` tool call and 0 tool results**.

## Why

`--help`, in a clause that is easy to read past:

> `--permission-prompts <target>`  Who answers permission prompts **with `--print`**: "host" (the SDK host or `--permission-prompt-tool`) or "none"

The spike that proved the mechanism used `--print`, so it proved it in the one mode gate 2 cannot use.

## Why this is awkward rather than simply fixable

`--bg` and `--print` conflict — that is a measured constraint this engine already works around elsewhere. Gate 2 runs background on purpose: an owner's run is long, and the console should not be held open for it. So the two requirements collide:

- a **background** gate is the right shape for gate 2, and
- a background gate has **no way to route a permission prompt** to the engine.

`--permission-prompts none` (the `denyOnPrompt` posture) does work in background, but it fails closed on everything, which is a different product, not a fix.

## What needs finding out

1. **What a `--bg` session does with a prompt today.** It goes somewhere — the daemon holds it. Bodhilander is a Claude Code session manager and already renders prompts for its own PTY sessions, so the mechanism may already exist in this codebase and simply not be pointed at gates.
2. **Whether `--bg` can be told to route to a tool.** Nothing in `--help` says so, but `--permission-prompt-tool` is itself undocumented as its own entry, so absence from the help text is not evidence.
3. **Whether gate 2 must be `--bg` at all.** Print mode works today, end to end, including the human decision. What it costs is holding the caller open for the length of an owner's run — which the reconciler, unlike the console, may not care about.

(3) is the cheapest and may be the right answer; it is recorded here so it is a decision rather than a drift.

## Meanwhile

`BODHI_GATE_MODE` overrides the console's mode, defaulting to the existing behaviour. The working path can be driven and demonstrated without anything changing silently.

William-Long-II:
Closes #291. Refs #722, #288.

## The problem this closes

`--permission-prompt-tool` is not consulted for a `--bg` gate. On the first background run the broker *started* and was never asked; the gate hit a `Write` and sat on an interactive prompt in a terminal nobody was attached to. Gate 2 is background on purpose, so the channel from #290 could not reach the one gate that spends the most.

## Measured before designed

Every row was run against `claude 2.1.270`, `--setting-sources ""`, `--strict-mcp-config`:

| question | result |
|---|---|
| does a `PreToolUse` hook decide permissions? | **yes** — a `deny` held in `--print` even under `bypassPermissions`; the hook outranks the posture |
| in `--bg`? | **yes** — an `allow` under `prompts none` (which alone denies) wrote the file |
| can it wait for a person? | 150s held under `timeout: 900`; 70s under `3600` |
| trust prompt for a worktree cwd? | none — the one seen earlier was for an untrusted scratch dir |
| what if the hook itself times out? | **neither allow nor deny** — the call fell through to the interactive prompt and the session sat there |

That last row decides the design.

## The change

**The broker gains a hook mode.** Same file, same channel directory, same request shape — `fileRequest` is the single place both modes write, so `permission-channel.ts`, `perms`, `answer` and the inbox cannot tell which route a request took.

**The launcher routes by mode.** A background gate gets a per-gate `--settings` file carrying the hook (`--settings` loads even with setting sources disabled); a print gate keeps the MCP config it had. Never both — a background gate given an MCP config would start a broker that is never asked, beside the one that is.

**The hook mode has a deadline, and that is not a change of principle.** A killed hook becomes a hang, so the broker answers strictly inside the configured hook timeout (`HOOK_ANSWER_BY_SECONDS < HOOK_TIMEOUT_SECONDS`), and when it answers because nobody replied it says exactly that. The model quoted it back verbatim in the spike:

> The command did not run — the permission request was blocked with the reason "nobody answered this permission request within 15s; the run is waiting on a person"

That is not the broker deciding the request; it is the broker refusing to let the request become a hang.

## Driven live, gate 2 in its default `--bg`

```
01:44:25  prepared
01:44:27  provisioned
01:44:36  permissionRequested     <- a Write, filed by the hook
01:45:05  permissionAnswered      <- `answer ... allow`, console unchanged
```

The probe file was written and then removed by the gate's next call. `perms` showed the request in 10 seconds with no code change on its side.

## Tests

- **The broker is spawned as the real program it is**, in both modes: allow via a late reply, deny with its message, the deadline refusal (`within 1s`, under 5s wall clock), a half-written reply waited on rather than read, a payload with no `tool_use_id` refused, MCP `tools/list` unchanged.
- **Launcher routes**: background writes `.settings.json` and no `.mcp.json`; print writes `.mcp.json` and no `.settings.json`; `denyOnPrompt` gets no channel at all. Swapping the mode condition fails exactly the two route tests.
- **`--settings`** is on the command line only when supplied.
- **Composer**: the hook command, the empty matcher (deciding which tools are dangerous is not this module's), the deadline strictly inside the timeout with at least 60s of slack, the timeout in the range a person needs.

## Left deliberately

Print mode keeps the MCP route. The hook fired in `--print` too, so unifying both on it is possible and would remove a mechanism; it is a follow-up so this slice is about the background path alone. The gate row in the live run still says `running` after the gate finished — a `--bg` gate reports via receipt and nothing reads it yet. That is #287, unchanged by this.


brannon-bowden (approved):
Approve. This is a well-scoped, thoroughly-tested extension of the permission broker to support a `PreToolUse` hook route for background gates, and the reasoning in the comments (deadline placement, why the hook needs its own timeout strictly inside the CLI's, why matching is unscoped) is easy to follow and backed by tests on both the broker script and the settings/command-building code. CI is fully green (all 7 checks passing).

## Suggestions (non-blocking)

- `scripts/permission-broker.js:69` — `Number(process.argv[4] || 3300)` will silently produce `NaN` if a non-numeric value is ever passed as the seconds argument. Since `Date.now() - started > NaN` is always `false`, the poll would never hit the deadline and the hook would hang indefinitely — exactly the failure mode this feature exists to prevent. In practice the only caller is `hookSettingsText`, which always supplies a valid number, so this isn't reachable today, but a guard (e.g. `Number.isFinite` fallback) would make the script robust against manual/future invocation.
- `scripts/permission-broker.js:249-272` — the new `if (!hookMode) { ... }` block wraps the existing MCP stdin handling but its contents aren't re-indented, so the nesting isn't visually apparent. Minor readability nit for future maintainers scanning the file.
- `src/main/run-engine/gate-launcher.ts:422` — `permission` changed from `{ ... } | null = null` to a `Pick<...> = {}` default. If any code later in `launchGate` (not shown in this diff) relies on `permission` being falsy/`null` to decide whether to spread permission fields, switching the "no permission" default to a truthy empty object could change that branch's behavior. Worth a quick self-check that downstream usage treats `{}` and `null` equivalently (e.g. via spread rather than truthiness), though the added tests (`a posture that never asks gets no channel at all`) suggest the visible behavior is already covered.

Issue #292, opened 2026-09-14
the gate deadline killed a working reviewer, and the run reported it as 'could not establish a verdict'


William-Long-II:
Found on the first run that reached gate 3 (#722).

## What happened

A run went 2 → 3 and ended `inconclusive`:

```
gate 2 done (bodhilander-lead, attempt 1)
gate 3 done (reviewer, attempt 1)
gate 3 could not establish a verdict
```

The reviewer had not failed to establish anything. Its last words were:

> Full relay suite is green. Now I'll mutate the change to see whether the tests are load-bearing.

and the transcript stops mid-work. The two `gateFinished` events are **exactly 900 seconds apart** — 15:04:37 and 15:19:37 — which is `BODHI_GATE_TIMEOUT`'s default of `900000`. The reviewer was killed by our own deadline while doing the most valuable part of its job.

## Why this is the provisioning bug again

#284 was the same shape: a deadline sized for something smaller than the real work, producing an outcome that is *honest* but points away from the cause. There the message was `python did not finish within 60000ms` for a failed `bun install`; here it is `could not establish a verdict` for a reviewer that was interrupted.

The engine behaved correctly given what it knew — `inconclusive` rather than a fail, which is the exit-code contract working exactly as intended. The defect is that it did not know, and so could not say, that **it** ended the gate.

## Two things to fix, and they are separate

**1. The outcome should distinguish being killed from answering unreadably.**

`runGate` knows it timed out — `command-runner` already has `TIMED_OUT = 124` for this. That has to survive into the gate's outcome so a run can report "gate 3 was stopped after 900s without a verdict" rather than implying the reviewer produced something unreadable. A person reading the current message goes to look at the reviewer, which is the wrong place.

**2. The number is wrong, and probably wrong in a way no single number fixes.**

Gate 2 did real work — wrote a failing test, confirmed red, fixed `http.ts`, ran the full relay suite, typechecked, committed — in about 8 minutes. The reviewer was still going at 15. A review that mutation-tests the change is not obviously bounded by any figure we would pick, and a gate that is thinking is indistinguishable from one that is wedged **by the clock alone**.

That points at #287 rather than at a bigger constant: what distinguishes a working gate from a stuck one is whether it is still *doing* anything, and the session transcript's mtime is a progress signal that costs nothing to read. A deadline on *silence* is defensible; a deadline on *total elapsed time* will always be either too short for a thorough gate or too long to catch a wedged one.

[...]

PR #302, opened 2026-09-15
feat(run-engine): a launched gate is looked at again

William-Long-II:
Refs #287, #722. Builds on #301 (the hook channel), now merged.

## The gap this closes

`launched` was designed as an outcome distinct from `completed`, and that was right: a `--bg` gate reports later, by receipt. What was missing was the other half — something that asks, later, what became of it. The first real run's gate finished, its process exited, and the engine said `gate 2 running` for as long as anyone cared to look. "You can't see what's happening" was one of the three pains this engine exists to remove; this was that pain reproduced by the thing built to remove it.

## Three pieces, each pure until the last

**`gate-receipt.ts`** reads what `gate-receipt.sh` writes at `<initiative>/gates/<gate>-<agent>.json`. The receipt's vocabulary is the plugin's — `pass | fail | undriveable | skip` — and the machine's is `pass | fail | inconclusive`. They agree on the two that move a run; `undriveable` and `skip` both become `inconclusive`, each with a reason naming the word the receipt used, so whoever reads the note knows whether the gate could not run or had nothing to do. Nothing is coerced: `approved` is inconclusive with the word quoted, exactly as `gate-verdict.ts` treats structured output. A pass carrying blocking findings is two answers that disagree. A fail without findings stays a fail. **No receipt is `null`, not a verdict** — the caller decides what silence means.

**`gate-attention.ts`** decides from two facts — the receipt, and whether the session is alive:

| receipt | alive | decision |
|---|---|---|
| yes | any | `gateFinished` with the receipt's verdict |
| no | yes | nothing — the gate is working |
| no | no | `gateFinished` inconclusive — gone without a receipt |

A receipt is taken even while the gate is still alive: the harness says the receipt is the end of the work, and today's scribe made one more read after writing its — holding the run for the process to exit would be the engine deciding the gate had not really meant it. The note says the gate was still alive, so the log can be reconstructed.

**The driver writes down which session a launched gate became.** The executor *reports* launches (it writes no state); the driver records them on the row it opened before the spawn. A gate with no session recorded is a gate nothing can look at again — the receipt can be found by path, but whether the process is alive cannot. Removing the recording fails the test that reads the row back.

**The console gains `watch <run-id>`**, which gathers the two facts and applies what `attend` decides through the same path `step` uses — so a gate finishing by receipt spawns the next gate exactly as one finishing by structured output does. Tried on this morning's background run, whose row predates session recording:

```
gate 2 (bodhilander-lead, attempt 1)
  receipt none
  session (not recorded)  alive: unknown
nothing to do; as far as can be seen, the gate is working
```

A row that cannot be checked is not pronounced dead. "Gone" is a verdict about the gate and must not be reached by failing to look.

## What this found about the harness

This morning's full run left receipts for `3-reviewer`, `4-verifier` and `4-scribe` — and **none for gate 2**. Only `arch`, `reviewer`, `scribe` and `verifier` are told to write one; owners are not. So for a background owner, "gone without a receipt" is today the *normal* end, and `attend` will call it inconclusive.

That is the honest answer — nobody said the owner finished well — but it is noisy, and the fix belongs upstream: the receipt is the transition contract, so the role that finishes gate 2 should sign it. The schema already admits it (`gate ≥ 0`, any lowercase agent). A plugin PR follows. Until it lands, an owner gate driven in background lands in the inbox with a note that names exactly why.

## Not yet

A gate that is alive and silent is indistinguishable from one that is thinking by these two facts alone. Telling them apart needs a progress signal and a deadline on **silence** rather than elapsed time — #292's point, and the next slice.

## Tests

- receipt reader: 15 cases including every non-coercion, both "two answers" rules, and null-vs-malformed
- attention: 6 cases, one per row of the table plus the still-alive note and the naming
- repository: both ids land on the row; a print gate records null
- driver: a launch lands on its row; a completed print gate leaves the columns alone — the first fails when recording is removed
- full suite green; type-check clean


scripts/run-console.ts
```diff
@@ -386,6 +402,66 @@ function readActiveChannel(runId: string) {
...
+  const gate = runs.activeGate(runId);
+  if (!gate) {
+    console.log('no gate in flight, so there is nothing to look at');
+    return;
+  }
+  if (gate.gate !== 2 && gate.gate !== 3 && gate.gate !== 4) {
+    throw new Error(`the open row is for gate ${gate.gate}, which this engine does not know`);
+  }
+  const receiptPath = receiptPathFor(run.initiativeDir ?? '', gate.gate, gate.agent);
+  const receipt = readReceipt(fs.existsSync(receiptPath) ? fs.readFileSync(receiptPath, 'utf8') : null);
```
brannon-bowden: TOCTOU: fs.existsSync then fs.readFileSync can throw ENOENT if the receipt file is deleted in between, crashing watch() instead of treating it as "no receipt yet." Wrap the read and fall back to null on ENOENT.

src/main/run-engine/driver.ts
```diff
@@ -370,6 +372,23 @@ function stepAfter(
...
+ * Write down which session each launched gate became.
+ *
+ * The row was opened before the spawn and the launcher named the session
+ * during it, so this is the first moment both are known. A gate with no
+ * session recorded is a gate nothing can look at again -- the receipt can
+ * still be found by path, but whether the process is alive cannot.
+ */
+function recordLaunches(runId: string, performed: ExecutorResult): void {
+  for (const launch of performed.launched) {
+    const row = runs.activeGate(runId);
```
brannon-bowden: recordLaunches matches a launched gate to a row purely by gate number via a fresh runs.activeGate(runId) lookup each iteration. If execute() ever launches more than one gate in a batch, this can mis-associate or silently drop a session. Consider passing the row id through explicitly, or add a test proving multiple simultaneous launches can't occur.

brannon-bowden (changes requested):
This PR adds a solid, well-tested piece of the run engine (receipt parsing, the `attend()` decision table, and session bookkeeping for background gates), but it can't be approved while CI is red.

## Issues

- **CI is failing and must be green before merge**: `postinstall produces a working native module (windows-2022)`, `test (windows-latest)`, and `quality-gate` are all reported as failing. None of the Windows-specific failures are explained by anything visible in this diff, so please confirm whether they're pre-existing/flaky or caused by this change, and get `quality-gate` passing. (`GitGuardian Security Checks` is still pending — worth a glance once it resolves, but it isn't gating this review.)
- **`watch()` has a TOCTOU between the existence check and the read** (`scripts/run-console.ts:428`, the `fs.existsSync(receiptPath) ? fs.readFileSync(...) : null` line): if the receipt file is removed between the two calls (plausible here, since the receipt is written by a separate background process), `readFileSync` throws `ENOENT` and crashes the command instead of being treated as "no receipt yet." Consider wrapping the read in a try/catch that falls back to `null` on `ENOENT`.
- **`recordLaunches` re-resolves the active gate by number rather than by the row it just opened** (`src/main/run-engine/driver.ts`, the `for (const launch of performed.launched) { const row = runs.activeGate(runId); ... }` loop): it matches purely on `row.gate === launch.gate`. This is fine as long as at most one gate is ever launched per `execute()` batch and `activeGate` tracks a single row per run — which the visible tests only exercise for the single-launch case. If `openGates` (not shown in this diff) can ever open more than one gate at once, or two agents for the same gate number end up "running" simultaneously, this loop would either write the same session onto one row twice or silently drop the other launch's session — precisely the "gate nothing can look at again" failure this PR exists to fix. Worth a test that drives two simultaneous launches through `advance()` to confirm this can't happen, or explicitly documenting why it can't.

## Suggestions (non-blocking)

- `gate-receipt.ts`'s "not a word" message for an empty-string verdict (`typeof receipt.verdict` → `"string"`) reads a little oddly for that specific case, since it *is* a string, just an empty one — minor wording nit only.
- The new `gate-attention.ts` and `gate-receipt.ts` modules are cleanly separated and thoroughly tested; nice use of pure functions for the decision logic, which makes the edge cases (pass-with-blocking-findings, fail-without-findings, alive-with-a-receipt) easy to verify in isolation.

William-Long-II:
All three taken; two were real defects and the third was a true-but-unhelpful message.

## `recordLaunches` resolved by number

Correct, and the honest version of the answer is: today it *could not* go wrong — a decision spawns at most one gate, and the row is opened moments before the spawn, so "latest running row with this gate number" and "the row I just opened" agree. But they agree by circumstance, not by construction, and a session written onto the wrong row is exactly the gate-nothing-can-look-at-again this PR exists to end.

So `openGates` and `startStep` now hand back the id of the row they opened, and the launch is recorded on that row by id. The test you asked for constructs the disagreement rather than arguing about it: inside the fake `spawnGate`, a second `running` row for the same gate appears while the launch is in flight. The session must land on the opened row and leave the intruder blank.

```
mutation: resolve by number again  ->  2 tests fail
  the session it became lands on its row
  not on whatever running row is newest with the same gate number
```

## `watch()` TOCTOU

Yes. One read now, `ENOENT` mapped to `null`, anything else thrown — a missing receipt is the normal state of a working gate, not a reason for the command to die. `readIfPresent` is the only place the console touches that file.

## The empty-string message

It *is* a string, so "was string, not a word" was true and pointed nowhere. Three cases now read differently: key absent → "carries no verdict"; empty string → "carries an empty verdict"; non-string → "was number, not a word".

## CI

The Windows legs and `quality-gate` are green at head; the only red is `arbiter/review`, which is this review. Full suite 0 fail, type-check clean.

## One thing this PR found in the harness, for the record

Driven live after this landed locally: a background owner ran, **tried to write its receipt, and the harness tool refused it** — `agents/<name>.md does not exist or declares no gate` — because the tool resolves a flat path and owners live in `agents/staff/`. `watch` then did exactly what the table says: `gate 2 is no longer running and wrote no receipt, so it established nothing` → `inconclusive`. Right answer to the evidence. The fix is in the plugin (claude-team-workflow#357), not here; this engine's job was to make the gap visible, and it did.

brannon-bowden (approved):
Both blocking issues from the previous review have been addressed, and CI is now fully green — approving.

- **CI**: all 7 checks (SonarQube, postinstall on Windows/ubuntu, quality-gate, test on Windows/ubuntu, GitGuardian) are passing.
- **`watch()` TOCTOU**: fixed. `readIfPresent` (`scripts/run-console.ts:449`) now does a single `readFileSync` wrapped in try/catch, treating `ENOENT` as "no receipt yet" instead of racing an existence check against a concurrent writer.
- **`recordLaunches` gate-number ambiguity**: fixed properly, not just patched around. `openGates` now returns the `rowId` it opened for each spawn (`src/main/run-engine/driver.ts:227-260`), and `recordLaunches` (`driver.ts:391-403`) attributes each launched session to that specific row instead of re-resolving "the active gate" by number. The new test in `driver.test.ts` constructs exactly the scenario I was worried about (a second running row for the same gate appearing mid-spawn) and confirms the session lands on the correct row. `continueSequence` was updated consistently to thread its own single-entry map through the same path.
- **`gate-receipt.ts` verdict messaging**: the empty-string-verdict wording nit from before is also resolved — it now reads "the receipt carries an empty verdict" instead of the odd "not a word" phrasing.

## Suggestions (non-blocking)

- `openGates`'s new `opened` map is still keyed by gate number (`Map<Gate, string>`), so if `decision.actions` ever contained two `spawnGate` actions for the same gate number within a single batch, the second `opened.set` would overwrite the first and one of the two rows would end up without its session — the same class of bug just narrowed to a single `openGates` call instead of the whole run. I can't see the code that builds `decision.actions` to know whether that's reachable, so this isn't a blocker, just worth a quick gut-check or a comment noting why a batch can't contain two spawns for the same gate.

William-Long-II:
## Driven live, with both halves in place

Local harness carrying the owner sign-off instruction and the `gate_receipt.py` fix; Bodhilander on the attention branch. A fresh run, gate 2 in its default `--bg`, task: verify the already-pushed branch, write your receipt, stop.

```
RUN=e87031c3
  launching gate 2 as bodhilander-lead
after 266s: RECEIPT WRITTEN
requests: 28  replies: 28
  receipt verdict: pass | agent: bodhilander-lead | not_verified: 6
```

Then `watch`, the first time a background gate's finish has been noticed by the engine rather than by a person:

```
gate 2 (bodhilander-lead, attempt 1)
  receipt pass   .../gates/2-bodhilander-lead.json
  session ea15b328  alive: false
  note    gate 2 (bodhilander-lead) wrote a receipt with verdict pass
stepping e87031c3: running + gateFinished
  launching gate 3 as reviewer
state     running
applied   gateFinished
```

Three things this proves at once: the owner writes a receipt when told (the fragment), the tool accepts it from `agents/staff/` (the fix), and the engine reads it and moves the run (#302). The session id landed on the row (`ea15b328`) and `alive: false` was read from `claude agents`, not assumed.

The same run an hour earlier, before the tool fix, ended `inconclusive` with `wrote no receipt, so it established nothing` — the owner had tried and been refused. Same engine, same decision table; the only thing that changed is that the receipt now exists.

Gate 3 was stopped immediately after launch; the point was the transition, not another review.

PR #303, opened 2026-09-15
feat(run-engine): a launched gate is judged by what the daemon says it is doing

William-Long-II:
Refs #292, #287, #722. Builds on #302.

## Where #292 ended up

It began as *the gate deadline killed a working reviewer*: a cap on elapsed time cannot tell a thinking gate from a stuck one, so it is always too short for a thorough gate or too long to catch a wedged one. The answer turned out not to be a better deadline. The daemon already says what a session is doing.

**Measured**, `claude agents --json`, two background sessions read at intervals:

| session | doing | `status` |
|---|---|---|
| wedged on a permission prompt (`--bg`, `prompts host`, no hook — the #288 shape) | blocked, nobody attached | `waiting` at 10s, 25s, 45s |
| a Read-only task | working, then done | `busy` at 5s → `idle` at 20s, 40s |
| stopped | — | not listed |

## The change

`attend()` takes the status where it took a boolean, and the decision table gains two rows a pid check could never have told apart:

| receipt | status | decision |
|---|---|---|
| yes | any | `gateFinished` with the receipt's verdict |
| no | `busy` | nothing — thinking |
| no | `waiting` | `permissionRequested` — a prompt nobody is attached to; the note names the id to `claude attach` |
| no | `idle` | `gateFinished` inconclusive — finished its turn and never signed off |
| no | `gone` | `gateFinished` inconclusive — crashed or killed without a receipt |
| no | unknown | nothing — "gone" is never reached by failing to look |

`idle` and `gone` get different notes because the fix is different: a crash is the machine's, a missing sign-off is the role's. `waiting` goes to `waitingPermission` — the inbox — rather than being called finished or left to sit. With the hook channel (#301) a background gate should never show `waiting` at all, so it also means something bypassed the channel, which is worth saying loudly.

The console's `sessionAlive` becomes `sessionStatus`: the daemon's word, `gone` when unlisted, and **null — not gone —** when the daemon could not be asked or reports a word it has not shown us before.

## What is still owed

A ceiling on `busy` with no word, forever. That is a cadence, not a fact about the gate, and it belongs beside the other intervals in `reconcile-loop.ts` rather than in a decision module. It is the last piece of #292 and the only place a clock still runs.

## Evidence

- Receipt path proven live in #302: a background owner wrote its receipt, `watch` read it and launched gate 3.
- Status vocabulary measured above, on real sessions.
- The table is unit-tested one row at a time. Making `idle` finish nothing fails one test; ignoring `waiting` fails one. Full suite green; type-check clean.


brannon-bowden (approved):
Approve. This is a well-documented, behavior-preserving extension of session liveness tracking into a richer daemon-reported status vocabulary, and the test suite is updated in lockstep with the new `busy`/`waiting`/`idle`/`gone` states, including the important "unknown status must not be treated as gone" invariant.

## Issues

- **Diagnostic log fires out of order** (`scripts/run-console.ts:459-475`): `sessionStatus()` calls `console.log` directly when it encounters a status word it doesn't recognize. Since `sessionStatus` is awaited *before* the `gate ${gate.gate} (...)` header is printed in `watch()` (`scripts/run-console.ts:433-436`), this diagnostic line will print above the gate header it actually refers to, breaking the established "note" convention the rest of the file uses (notes are collected and printed after the header via `attend()`'s returned `note`). Worth routing this through the same `note` channel, or deferring the log until after the header prints, so console output stays chronologically readable.

## Suggestions (non-blocking)

- `src/main/run-engine/__tests__/gate-attention.test.ts:53` (comment above the `waiting` test) refers to the target state as "`waitingPermission`", but the actual emitted event kind is `permissionRequested` (`src/main/run-engine/gate-attention.ts`). Worth fixing the comment so it doesn't send a future reader hunting for a `waitingPermission` state that doesn't exist.
- The `permissionRequested` event is emitted as a bare `{ kind: 'permissionRequested' }` with no gate/agent/backgroundId payload — only the accompanying console `note` carries that context. If any downstream consumer of `RunEvent` needs to know *which* gate is waiting (e.g. to route it into a per-gate inbox), it can't get that from the event alone. Might be worth confirming the consumer doesn't need that identifying data threaded through the event itself, since it isn't visible in this diff.

PR #304, opened 2026-09-15
feat(run-engine): a running gate is looked at, and busy-with-no-word goes to a person

### LINKED ISSUE #292: the gate deadline killed a working reviewer, and the run reported it as 'could not establish a verdict'

William-Long-II:
Closes #292. Refs #287, #722. Builds on #302 and #303. Two commits.

## 1. A reversal, recorded where it happened

`baseIntervalFor('running')` was `null`, and the test said why: *"the process exits and says so."* True of a print gate. Exactly wrong of a `--bg` one, which reports by receipt and cannot be heard any other way — which is how the first real run said `gate 2 running` for hours after the process was gone. `running` is now looked at every minute (one file, one `claude agents`), and the test that asserted the opposite now asserts this and says what changed.

### The one clock left, and what it does

#292 is named for a reviewer that a fifteen-minute deadline killed mid-mutation-test. No elapsed-time clock could have told it was working. Nearly every case that deadline existed for is now read from the daemon instead (#303: `idle`, `gone`, `waiting`). What remains is `busy` with no receipt, indefinitely — and only a person can say whether that is thought or a loop.

So the ceiling **does not kill**. Past it, the run goes to `inconclusive` — the inbox — with the session id to attach to, and the gate is left running for the person to judge:

```
gate 2 (bodhilander-lead) has been busy for 180 minutes with no receipt,
past the 120-minute ceiling; it is still running, and whether that is
thought or a loop is a person's call -- `claude attach ea15b328` to see it
```

Two hours, because today's scribe ran thirty-five minutes and a ceiling a thorough gate can hit is a ceiling that will lie. Both halves of the clock are optional inputs to `attend`, and **both are required** for it to apply: half a clock is not a clock. A receipt outranks the ceiling.

Also from #303's review: the console's unknown-status diagnostic is returned as a note and printed under the gate header it belongs to, and the test comment that said "moves to waitingPermission" now also names the event that gets it there.

## 2. A receipt older than the gate is a previous attempt's

Found by running `watch` on a leftover row with the first commit built. `gates/<gate>-<agent>.json` has no run and no attempt in its path, so a fresh run's gate 3 found **this morning's** `3-reviewer.json` — a different run's sign-off at the same path — read `pass`, and launched gate 4 on it. Same class as the channel-key collision guarded earlier with attempt numbers. The verifier it started was stopped by hand; the worktree was untouched.

The guard needs nothing from the harness. Every receipt carries `written_at`, every gate row carries `started_at`, and a receipt written before the row was opened is not this gate's. `attend` decides from the status as though there were no receipt, and the note says a receipt was there and why it did not count — a person looking at the directory will see one and wonder why the run did not move.

Compared as **instants, not text**: SQLite's `YYYY-MM-DD HH:MM:SS` and the receipt's ISO 8601 do not sort against each other as strings, and the console stamps the row's UTC time as UTC before handing it over. Both timestamps are required to call anything stale.

Confirmed live on the same leftover row after the guard was built — see the comment below.

## Tests

- loop: `running` is polled; `preparing`/`provisioning` still are not; the ceiling is ≥ an hour and > 10× the cadence
- attention: busy past the ceiling → inconclusive with minutes and attach id; under → nothing; half a clock → nothing; a receipt wins over the ceiling; stale under busy → nothing with the note; stale under gone → inconclusive with both notes; a receipt after the row is this attempt's; the two timestamp formats compare as instants; either timestamp missing → nothing is stale
- receipt reader carries `written_at`, null when absent
- flipping the ceiling comparison fails two tests; making nothing ever stale fails two; full suite green; type-check clean


William-Long-II:
## The stale-receipt guard, confirmed live

Same leftover row that exposed the bug, with the guard built: gate 4's verifier row opened this evening, its session stopped, and this morning's `4-verifier.json` sitting at the shared path.

```
gate 4 (verifier, attempt 1, running 3m)
  receipt pass   .../gates/4-verifier.json
  session 4fcbb9c8  status: gone
  note    gate 4 (verifier): a receipt at this path was written 2026-09-15T00:27:53Z,
          before this attempt started 2026-09-15T02:49:11Z; it is a previous attempt's
          and was not used. gate 4 (verifier) is no longer running and wrote no
          receipt, so it established nothing
state     inconclusive
```

Before the guard, the identical look at this run's gate 3 read a stale `pass` and launched a verifier. After it, the stale receipt is named, refused, and the status decides — `gone`, no receipt of its own, `inconclusive`. Nothing was launched. The row's `started_at` is shown as UTC because that is what SQLite wrote and what the receipt's `written_at` is; the two compared as instants.

brannon-bowden (approved):
Approve — this is a well-reasoned, well-tested change (busy-gate polling cadence, the busy-ceiling escalation to a person, and the note-ordering fix in the console are all covered by new unit tests, and CI is fully green). A couple of non-blocking items worth a look before or shortly after merge.

## Suggestions (non-blocking)

- `scripts/run-console.ts:435` — `busyForMs` is derived from `gate.startedAt` via `Date.parse`, but there's no test covering this parsing (e.g. malformed/undefined timestamp producing `NaN`, which would silently just never trip the ceiling rather than fail loudly). Given the rest of this feature is carefully unit-tested, it'd be good to add a small test for this conversion, or at least guard/log on a `NaN` result.
- `src/main/repositories/runs.ts:458` — `toGateRow` now requires `row.started_at` and maps it straight through unchecked. The diff only shows the interface/mapper change, not the SQL query(ies) that produce these row objects — worth confirming the `SELECT` used to build rows passed here was updated to include `started_at`, since a missing column would surface at runtime (as `undefined`, causing `gate.startedAt.replace(...)` in `run-console.ts` to throw) rather than at compile time if the row type is cast.
- Minor semantic note: the "busy for N minutes" message in `gate-attention.ts` measures time since the gate *started* (`busyForMs`), not time continuously in the `busy` status. If a gate can cycle through `idle`/`waiting`/`busy`, the reported duration could overstate genuinely-busy time. Not necessarily wrong given the stated design intent ("busy with no word, indefinitely"), but worth a sentence in the doc comment if that distinction matters to whoever reads the note.

William-Long-II:
On the `started_at` question, since the diff cannot show it: both queries that build gate rows are `SELECT *` (`activeGate` and `listGates` in `runs.ts`), so the column is already in every row the mapper sees — and the two live `watch` runs in this PR printed it: `running 21m` on the gate-3 row, `running 3m` on the gate-4 row. Not inferred from the schema; read off the rows.

The other two are right and small — a `NaN` from `Date.parse` should be said out loud and treated as half a clock rather than silently never tripping the ceiling, and the note should say "running for N minutes and busy now" rather than imply N minutes of continuous busy. Both land in a follow-up rather than on this head, so the approval stays attached to what was reviewed.

William-Long-II:
Sonar was red on the head you approved: `readReceipt` at cognitive complexity 21 against 20, once `written_at` was threaded through every return. Split into `verdictWord` (the three ways a verdict fails to be a word, told apart because they send a reader to different places) and `judge` (where the receipt vocabulary maps onto the machine's and nowhere else). The pass-with-findings mutation still fails its test.

Since the head moved anyway, the two small notes from your review landed with it rather than in a separate PR: the ceiling note now says "running for N minutes and is busy now" rather than implying N minutes of continuous busy (the daemon reports a status, not a history — the field's doc says which it measures), and a start time `Date.parse` cannot read is now said out loud and handed to `attend` as null, so it cannot silently never trip the ceiling.

Full suite green, type-check clean. Re-requesting at the new head once Sonar has re-analysed it.

scripts/run-console.ts
```diff
@@ -430,13 +431,34 @@ async function watch(): Promise<void> {
...
+    console.log(`  note    started_at is ${JSON.stringify(gate.startedAt)}, which does not parse; the busy ceiling cannot apply`);
+  }
   console.log(`  receipt ${receipt ? receipt.verdict : 'none'}  ${receiptPath}`);
-  console.log(`  session ${gate.bgSessionId ?? '(not recorded)'}  status: ${status ?? 'unknown'}`);
+  console.log(`  session ${gate.bgSessionId ?? '(not recorded)'}  status: ${seen.status ?? 'unknown'}`);
+  // Printed here, under the header it is about, rather than from inside the
+  // lookup -- a note above its own gate reads as somebody else's.
+  if (seen.note) console.log(`  note    ${seen.note}`);
 
   const { event, note } = attend({
```
brannon-bowden, line 455: Duplicates the `gate.startedAt.replace(' ', 'T') + 'Z'` normalization already computed above for busyForMs — consider hoisting to a shared local variable.

src/main/run-engine/__tests__/gate-attention.test.ts
```diff
@@ -50,14 +50,113 @@ describe('a launched gate, looked at again', () => {
...
+    // '2026-09-15 02:23:22' sorts AFTER '2026-09-15T00:19:00Z' as text only
+    // by accident of the space; the comparison must not depend on it.
+    const { event } = attend({
+      ...BASE,
+      receipt: { verdict: 'pass', blocking: [], writtenAt: '2026-09-15T02:28:00Z' },
+      startedAt: '2026-09-15 02:23:22',
+      status: 'gone',
+    });
+    expect(event).toEqual({ kind: 'gateFinished', gate: 2, verdict: 'pass' });
+  });
```
brannon-bowden, line 181: This test passes a non-ISO, zone-less `startedAt` straight into attend(), so its result depends on Date.parse's implementation-defined local-time interpretation rather than genuinely proving 'instants, not text'.

brannon-bowden (changes requested):
This is a well-reasoned change (the `running`-state polling fix, the busy-ceiling escalation-to-a-person instead of a kill, and the stale-receipt guard against reading a previous attempt's receipt are all sensible, and the test coverage for `attend()` is thorough), but CI is currently red, so this can't go in as-is.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate are both failing on this PR. These need to be resolved (or the failures understood/justified) before merge.
- **Timezone-dependent test assertion** (`src/main/run-engine/__tests__/gate-attention.test.ts`, the `"SQLite's timestamp and ISO 8601 are compared as instants, not text"` test): it passes `startedAt: '2026-09-15 02:23:22'` (no zone) directly into `attend()`, which parses it with `Date.parse`. For a string in this non-ISO, space-separated format, `Date.parse` behavior is implementation-defined and commonly parsed as *local* time rather than UTC. In production this ambiguity never occurs because `scripts/run-console.ts` always normalizes to `...T...Z` before calling `attend`, but this specific unit test bypasses that normalization, so its pass/fail result is dependent on the test runner's local timezone rather than testing the "instants, not text" claim it documents.

## Suggestions (non-blocking)

- `scripts/run-console.ts:435` and `scripts/run-console.ts:451` both compute `` `${gate.startedAt.replace(' ', 'T')}Z` `` independently. Worth hoisting into a single local variable to avoid the duplication and guarantee both usages stay in sync.
- The new `busyForMs` computation and the `note` field threaded through `sessionStatus` in `scripts/run-console.ts` don't appear to have dedicated tests in this diff (the added tests all target `attend()` in `gate-attention.ts`). If there isn't existing coverage for `sessionStatus`/`watch()` elsewhere, it'd be worth adding a couple of unit tests for the date-parsing and note-ordering behavior described in the comments.
- Worth double-checking that the `started_at` column already exists on the gates table/schema — this diff only adds the TypeScript-side field and row mapping in `src/main/repositories/runs.ts`, and no migration/schema change is visible here. If the column isn't already present, reading `row.started_at` would fail at runtime.

brannon-bowden (approved):
Approve — CI is now green across the board (SonarQube and quality-gate both passing), which was the blocking issue from the prior review, and the new changes are solid, well-documented improvements with no fresh defects.

## Issues

None blocking. For completeness on the prior review:
- **CI failing** — resolved; all 7 checks (including SonarQube Code Analysis and quality-gate) are now passing.
- **Timezone-dependent test assertion** (previously flagged in `gate-attention.test.ts`) — that specific test isn't touched by this diff, so I can't confirm from what's shown here whether it was addressed elsewhere. Not treating it as blocking since I have no current evidence either way; worth a quick confirmation from the author but not a merge gate.

## What's new and looks good

- `scripts/run-console.ts`: the `Number.isNaN(startedAtMs)` guard correctly handles an unparseable `startedAt` instead of silently propagating `NaN` through the busy-ceiling math — a real robustness fix, and the accompanying note output is clear.
- `gate-attention.ts` / test: the note wording change (`running for … and is busy now`) more accurately reflects that `busyForMs` is "time since row opened," not "time continuously busy" — the updated JSDoc explains this well, and the test assertions were updated to match.
- `gate-receipt.ts`: `readReceipt` was split into `verdictWord` and `judge` helpers. I traced the branch logic against the original (undriveable → skip → fail → default-with-blocking → pass) and it's behavior-preserving, just reordered and more readable.

## Suggestions (non-blocking)

- The duplicated `` `${gate.startedAt.replace(' ', 'T')}Z` `` computation flagged in the prior review (a second call site further down in `run-console.ts`) isn't visible in this diff, so I can't tell if it was hoisted — worth a look if it's still duplicated.
- Still no dedicated unit tests for `busyForMs`/`ran` computation or the new "unreadable start time" branch in `scripts/run-console.ts` itself (coverage here is all through `attend()` in `gate-attention.ts`). Given this is a script rather than a library function it may be lower priority, but a couple of tests around the NaN path would be cheap insurance.

William-Long-II:
Both issues addressed; the second was more interesting than it looked.

## CI

The review was cast while Sonar was still analysing the new head. It finished green a few minutes later — `readReceipt` split into `verdictWord` + `judge` took it from 21 to under the limit — and all 7 checks pass at `106cde2`. Nothing to change there.

## The timezone-dependent test

Correct, and I measured how wrong it was rather than just fixing it. `Date.parse('2026-09-15 02:23:22')` reads that bare form as **local** time in both Bun and node — 06:23:22Z on this machine at UTC-4. With that value the old test's receipt (02:28Z) would have been *stale* and the assertion (`pass`) should have **failed here**. It passed. Why: `bun test` pins the runner to UTC — offset 0, and it stays 0 even with `TZ=Asia/Kolkata` set in the environment. So the test passed only because the test runner happened to sit in the one zone where the bare form and the ISO form agree. Under node, or any runner that honours the machine's zone, it would have flipped.

Replaced by two tests that hold by construction rather than by luck — both timestamps carry explicit zones, so the runner's zone cannot enter:

- the same instant written as `2026-09-14T22:23:22-04:00` and as `2026-09-15T02:23:22Z` gives the same answer, and against a receipt at `00:19:00Z` that answer is *stale* — which a **text** comparison would get backwards, because `"2026-09-14T22…"` sorts before `"2026-09-15T00…"`. That is the actual "instants, not text" claim, and the old test never exercised it.
- an unreadable start time is not a comparison at all: nothing is called stale, and nothing is called fresh for the wrong reason.

Worth saying plainly: I also ran the suite under three forced zones and it passed in all three, but given the above that proves nothing — `bun test` ignored the variable. The guarantee is the explicit zones in the fixtures, not the runs.

## Non-blocking

- The normalised start time in `watch()` is now computed once and used twice.
- `started_at` is in `RUN_TABLES_SQL` (`started_at TEXT DEFAULT CURRENT_TIMESTAMP`, line 72), and both row queries are `SELECT *`; the live runs printed it as `running 21m` / `running 3m`.
- The console has no unit tests by design — it is the manual-drive harness, excluded from Sonar, and verified by being driven; every path in this PR was. The invariants it enforces are tested where they live (`attend`, the repository).

Re-requesting at the new head once CI has run on it.

brannon-bowden (approved):
Approve — this delta directly resolves both non-blocking notes from the prior review with no fresh defects, and CI remains green across all 7 checks.

## What's resolved

- **Duplicated timestamp computation** (`scripts/run-console.ts:435-438` vs the old second call site): now hoisted into a single `startedAtIso` constant used both for `Date.parse` and the `startedAt` field passed to `attend`. The comment explaining the UTC-without-zone-marker reasoning was consolidated to the one call site instead of being duplicated. Clean DRY fix, no behavior change.
- **Timezone-dependent test assertion** (`gate-attention.test.ts`): the old test fed a bare SQLite-style timestamp (`'2026-09-15 02:23:22'`) with no zone marker into `attend`, so its parse — and therefore the test's outcome — depended on the host machine's local timezone. The rewrite replaces it with two proper ISO-instant comparisons (`'2026-09-14T22:23:22-04:00'` vs `'2026-09-15T02:23:22Z'`, both the same instant) that demonstrate instant-based comparison independent of the runner's clock, plus a cross-check that both forms yield the same event. This is a real fix, not just a relabeling.

## New test added

- `'a start time that does not parse makes nothing stale, and says nothing false'` exercises `attend()` directly with an unparseable `startedAt` (`'not a time'`) and asserts it neither falsely marks the receipt stale nor silently treats it as fresh for the wrong reason. Since `gate-attention.ts` itself isn't touched in this diff, I can't see the implementation being exercised, but the test's intent lines up with the "half a clock is not a clock" guard already documented in `run-console.ts`, and CI (including this test) is passing.

## Suggestions (non-blocking, carried over)

- Still no dedicated unit tests for the `busyForMs`/ceiling computation in `scripts/run-console.ts` itself (coverage remains indirect via `attend()`). Given this is a script rather than library code, low priority — a couple of tests around the NaN-guard path there would still be cheap insurance if it's ever refactored.

PR #305, opened 2026-09-15
feat(run-engine): the loop that calls the engine without a person

William-Long-II:
Refs #722. First of two: this puts the loop under `src/main/run-engine/` with tests and proves it headless through the console; the second wires it into the Electron main process and adds the permission UI.

## Why this is the next thing

`src/main/index.ts` references the run engine zero times. Every run to date — sequencing, permissions, receipts, the busy ceiling — advanced because someone typed `step`, `watch` or `answer` into the console. The engine works. The app does not run it. Of the three pains this engine was built to end, *"you are still the dispatcher"* was untouched.

Every function this PR adds a caller for already existed. What did not exist was anything that called them without a person.

## Four modules, each thin over what was there

**`gate-spawner.ts`** — the console's launch wiring as a module: roles per gate (owner from the run, 3 and 4 from the harness), the brief, the channel key, and the one refusal — a spawn whose open `run_gates` row is not the one the driver just opened. Every `BODHI_*` variable the console read is now a field on `SpawnConfig`. That is what "the app can run a gate" concretely means: the decisions deferred behind environment variables all day now have a type.

**`attention-pass.ts`** — the gathering half of `attend`: the receipt at the harness's path, the daemon's word for the session, the time since the row opened. One module so `watch` and the app cannot disagree about the same gate.

**`pr-discovery.ts`** — the scribe opens the PR and nothing tells the engine which. The console learned it from an environment variable typed by a person — the dispatcher problem in miniature. The run knows the **branch** (`run_owners.branch`), and `gh pr list --head` run *in the worktree* finds the PR without anyone naming the repository; the `owner/name` slug is read from the PR's own URL, because the registry knows paths, not slugs. Recorded once on the owner row — the columns were already there.

**`run-loop.ts`** — the scheduler. Per due run (by `dueRuns`, so a run never looked at is due now and a failing run waits longer):

| state | what a tick does |
|---|---|
| `running` | look at the gate in flight; apply what `attend` decides, if anything |
| `waitingChecks` / `waitingReview` / `reviewNotRequested` | find the PR if the run does not know it yet; reconcile; apply every event |
| anything else | nothing — `intervalFor` is null, and asking cannot unstick a person or a finished run |

Failures back a run off and are announced exactly once at `ESCALATE_AFTER`. A throw in one run's pass is that run's problem, not the tick's. A slow tick is not two ticks. Every dependency is injected, because a scheduler is the one part of an engine that cannot be tested against the real thing without waiting.

## The console

`step` and `watch` now call the engine modules instead of holding the wiring. It gains `loop [seconds]`: arm a run, step it once to `prepared`, then `loop`, and watch it go gate to gate with a person needed only for permissions. That is the demonstration this console was built toward, and exactly the shape the app wires in next.

## Tests

45 across the four modules:

- **loop** — every branch of a tick with fakes: a decided look advances; a busy gate is left alone *and counts as a pass*; running with no row is a skip; discovery → record → reconcile by slug; a known PR is not asked again; no PR yet is a skip and a failure, not a state; every reconcile event applied in order; a reconcile problem counted; cadence respected to the millisecond; escalation exactly once; forgotten when inactive; a person-wait never due; one run's throw isolated
- **discovery** — open PR preferred, most recent otherwise, every malformed shape null, slug from URL including a non-github host
- **attention** — the real path, `gone` when unlisted, UTC read of the SQLite stamp, unparseable start said aloud, stale receipt refused from the same row, unknown gate refused before anything is read, daemon unreachable / unknown word / wrapped-or-bare list
- **spawner** — roles from a real temp harness in run order, target from the owner's URL, `null` not `''` before a PR exists, the refusal naming both rows, the defaults the console used to read from the environment

Mutations: counting every pass as a failure fails 4; never discovering the PR fails 2. Full suite green; type-check clean.

## What this does not do yet

Nothing in `index.ts` calls it. That is PR two, and it is deliberately separate: this one has to be right before anything starts a timer inside Electron.


src/main/run-engine/gate-spawner.ts
```diff
@@ -0,0 +1,189 @@
...
+        `gate ${gate} (${agent}) was asked to launch but the open run_gates row is ${found}; ` +
+          'the driver opens the row before it spawns',
+      );
+    }
+    log(`launching gate ${gate} as ${agent}`);
+    return launchGate({
+      gate,
+      agentName: agent,
+      mode: modeFor(gate),
+      prompt: gateBrief(
```
brannon-bowden, line 164: Dropped the `?? '(not recorded)'` fallback that the pre-refactor code applied to `run.initiativeDir` here. If this field can be null, restore the fallback.

src/main/run-engine/attention-pass.ts
```diff
@@ -0,0 +1,116 @@
...
+  const startedMs = Date.parse(startedAtIso(gate.startedAt));
+  // Half a clock is not a clock: an unreadable start time means the ceiling
+  // cannot apply, and that is said rather than left as a NaN that quietly
+  // never trips it.
+  const runningForMs = Number.isNaN(startedMs) ? null : deps.now() - startedMs;
+  const attention = attend({
+    gate: gate.gate,
+    agent: gate.agent,
+    receipt,
+    status: seen.status,
```
brannon-bowden, line 100: Dropped the `?? ''` fallback that `watch()` used to apply to `run.initiativeDir` before calling `receiptPathFor`. If this field can be null, restore the fallback.

src/main/run-engine/run-loop.ts
```diff
@@ -0,0 +1,230 @@
...
+  async function tick(): Promise<TickReport> {
+    const at = deps.now();
+    const report: TickReport = { at, due: [], looked: [], reconciled: [], skipped: [], problems: [], escalated: [] };
+    const active = deps.listActiveRuns();
+    const seen = new Set<string>();
+    for (const run of active) {
+      remember(run);
+      seen.add(run.id);
+    }
+    // A run that finished is not remembered forever.
```
brannon-bowden, line 239: `void tick().then(...).finally(...)` has no `.catch()`. Anything thrown in `tick()` outside the per-run try/catch (e.g. `listActiveRuns()`) becomes an unhandled rejection here.

src/main/run-engine/run-loop.ts
```diff
@@ -0,0 +1,230 @@
...
+  let ticking = false;
+
+  function remember(run: RunRow): ScheduledRun {
+    const known = memory.get(run.id);
+    // The state is the database's, always; only the bookkeeping is ours.
+    const next: ScheduledRun = known
+      ? { ...known, state: run.state }
+      : { id: run.id, state: run.state, lastPassAt: null, failures: 0 };
+    memory.set(run.id, next);
+    return next;
```
brannon-bowden, line 108: Uses `owners[0]` with no check for multiple owners, unlike `executorFor`'s explicit throw for `owners.length > 1`. A multi-owner run can silently reconcile only the first owner's PR.

brannon-bowden (changes requested):
This is a solid, well-documented extraction of the console's gate-spawning/attention/PR-discovery logic into reusable modules, with good unit-test coverage for the new pure pieces (`pr-discovery`, `gate-spawner`, `attention-pass`, `run-loop`). However, CI is currently red and the refactor introduces a couple of real regressions and a resilience gap in the new unattended loop, so this needs changes before merge.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate are both red (windows-latest test is still pending). Please resolve before merge.
- **Dropped null-guard on `run.initiativeDir`** — the pre-refactor code explicitly guarded this field twice (`run.initiativeDir ?? '(not recorded)'` in the inline `spawnGate` prompt, and `run.initiativeDir ?? ''` in `watch()`'s receipt lookup), while every sibling field (`owner?.repo ?? '(not recorded)'`, `owner?.worktree ?? process.cwd()`) kept its fallback. Both guards for `initiativeDir` specifically were dropped in the extraction with nothing replacing them:
  - `src/main/run-engine/gate-spawner.ts`: `initiativePath: run.initiativeDir,` (in `spawnGateFor`)
  - `src/main/run-engine/attention-pass.ts`: `receiptPathFor(run.initiativeDir, gate.gate, gate.agent)` (in `lookAtGate`)
  If `initiativeDir` can be null (which is exactly what the removed fallbacks assumed), this is a behavior regression — gate briefs and receipt paths would get `null` where a string/placeholder was previously guaranteed. Please restore the fallback in both places, or if the field's nullability genuinely changed, say so and drop the dead pattern deliberately elsewhere too.
- **`RunLoop`'s tick can crash/hang the unattended loop on a transient error** — in `run-loop.ts`, only the per-run body (`lookAt`/`reconcileRun`) is wrapped in try/catch inside `tick()`. Calls like `deps.listActiveRuns()`, `deps.listOwners()`, and `dueRuns()` sit outside that try/catch, at the top of `tick()`. If any of those throw (e.g. a transient SQLite error), `tick()` rejects, and:
  - in `RunLoop.start()`, the chain is `tick().then(...).finally(...)` with no `.catch()`, so the rejection becomes an unhandled promise rejection;
  - in `scripts/run-console.ts`'s `loop()`, `await engine.tick()` inside the `while` loop would throw uncaught, killing the entire "loop for N seconds without a person" run on a single hiccup.
  Given the whole point of this PR is unattended operation, a single transient failure taking down the scheduler undermines that goal. Please wrap the full body of `tick()` (or at least the calls before the per-run loop) in a try/catch, and add a `.catch()` in `start()`'s promise chain.

## Suggestions (non-blocking)

- `run-loop.ts`'s `reconcileRun` picks `deps.listOwners(run.id)[0]` with no check for multiple owners, unlike `scripts/run-console.ts`'s `executorFor`, which explicitly throws for `owners.length > 1` ("this slice drives one repo"). For a multi-owner run in `waitingChecks`/`waitingReview`, `reconcileRun` will silently reconcile only against the first owner's PR; if `reconcile()` happens to return zero events that tick, nothing ever surfaces the mismatch (the multi-owner guard only fires later, inside `advance()`, and only if an event is produced). Worth mirroring the same guard here, or at least logging when there's more than one owner.
- `RunLoop.start()`/`stop()` (the interval-based scheduling with the `ticking` overlap-guard and `unref()`) has no test coverage — `run-loop.test.ts` only exercises `tick()` directly. Since this is exactly the concurrency logic most likely to have subtle bugs, a test or two for "a slow tick suppresses the next timer fire" would be valuable.
- `scripts/run-console.ts`'s `loop()` re-lists the same four "still movable" states (`running`, `waitingChecks`, `waitingReview`, `reviewNotRequested`) that `run-loop.ts` already encodes in its exported-adjacent `RECONCILES` set. Consider exporting/reusing one source of truth so the two lists can't drift.
- Minor formatting change: the harness "gate N runs X then Y" line used to print as its own line (`  gate N    X then Y`); it's now folded into the generic `notes` array and printed as `  note      gate N runs X then Y`. Fine either way, just flagging it's a visible console-output change.

William-Long-II:
All three blocking items fixed, and running the loop live surfaced a fourth that was the real reason it did not work — folded in here.

## The bug under the loop: `--bg` ignores our `--session-id`

I drove a fresh run through the loop to demonstrate it, and it marked gate 2 `inconclusive` six seconds after launching it. The cause, measured against `claude 2.1.270`:

```
$ claude --bg --session-id aaaaaaaa-...
warning: --bg manages the session id; ignoring --session-id (use --resume <id> to continue)
  claude attach 9e338a7c    open in this terminal
```

`--bg` assigns its **own** id (`9e338a7c`) and prints it; the `--session-id` we pass is discarded. So `gate-process` recording `ourSessionId.slice(0,8)` stored an id the daemon never knew, and `sessionStatus` — looking that id up in `claude agents` — found nothing and returned `gone` for **every** live background gate. Every prior run was carried by a receipt being present (receipt outranks status); the one gate with no receipt yet, a just-launched owner, exposed it.

`backgroundIdFromOutput` now reads the id the daemon actually assigned, from the launch banner, and records that. This is a pre-existing `gate-process` bug from the original launch slice — the loop is just the first thing that exercised status-without-a-receipt.

## The review items

- **`run.initiativeDir` guards** — I took your second option. It is a `NOT NULL` column (`run-tables-sql.ts`) and a non-optional `string` on `RunRow`; the `?? '(not recorded)'` and `?? ''` were dead code for a null that cannot occur. Deliberately not restored. The `owner?.repo ?? ...` fallback stays, because `owners[0]` genuinely can be absent.
- **tick() resilience** — the whole body is now wrapped, `deps.now()` included, so a transient store error is a reported `(scheduler)` problem, never a rejection; `start()` gained a `.catch`. A new test makes `listActiveRuns` and `now()` throw and asserts the tick still resolves with the problem recorded.
- **multi-owner in `reconcileRun`** — mirrored as a skip-with-reason (not a throw: a data condition the loop reports and moves past).

## The suggestions

- `RECONCILES` is exported; the console's loop reuses it via `isMovable`, one source of truth.
- The nested ternary is `passOne`; the unnecessary `unref` cast is gone.
- Added the `start()` overlap test you asked for: a slow tick suppresses the next timer fire (the `ticking` guard), asserted by holding a tick open across several fires and checking only one got in.

CI should be green now; re-requesting once it re-runs.

brannon-bowden (changes requested):
This update meaningfully improves the unattended-loop resilience and ships a well-justified, well-tested fix for how background-gate session ids are derived, but two items from the prior review are still open, so this isn't mergeable yet.

## Issues

- **CI is still failing**: SonarQube Code Analysis and quality-gate are both red per the latest run. Please resolve before merge.
- **`run.initiativeDir` null-guard still missing** — this diff doesn't touch `src/main/run-engine/gate-spawner.ts` or `src/main/run-engine/attention-pass.ts`, so the dropped fallbacks flagged previously (`initiativePath: run.initiativeDir` in `spawnGateFor`, and `receiptPathFor(run.initiativeDir, ...)` in `lookAtGate`) are unchanged. If `initiativeDir` can be null, gate briefs and receipt paths still resolve to `null` where a string/placeholder was previously guaranteed. This needs to be addressed (or explicitly ruled out) before merge.

Resolved from the prior review:
- **Tick crash/hang on transient error** — `tick()` in `run-loop.ts` now wraps the scheduling calls (`deps.now()`, `listActiveRuns`, `dueRuns`) in the same try/catch as the per-run body via the new `runTick` split, and `start()`'s promise chain now has a `.catch()` before `.finally()`. Backed by new tests (`a throw from listActiveRuns...`, `a throw from dueRuns...`) that assert the tick resolves with a `(scheduler)` problem instead of rejecting. This looks solid.
- **Multi-owner reconcile guard** (previously a suggestion) — `reconcileRun` now explicitly skips and logs when `owners.length > 1`, mirroring `run-console.ts`'s `executorFor` guard.
- **Duplicated "movable" state list** (previously a suggestion) — `isMovable`/`RECONCILES` are now exported from `run-loop.ts` and reused in `scripts/run-console.ts`'s `loop()`, removing the drift risk.
- **Timer/scheduling test coverage** (previously a suggestion) — `run-loop.test.ts` now has a test asserting the `ticking` overlap-guard suppresses a second concurrent tick.

## Suggestions (non-blocking)

- The `gate-process.ts` rewrite (`backgroundIdFromOutput` replacing `sessionIdIn`) is a genuine bug fix — `--bg` ignores the passed `--session-id` and assigns its own id printed in the launch banner, and the old code derived an id the daemon never recognized. The regex, its negative-lookahead guard against longer hex runs, and the accompanying tests all look correct and well-reasoned. No concerns here, just flagging it as a substantial behavioral change worth a second pair of eyes given it touches the session/attach lifecycle for every background gate.

William-Long-II:
Both remaining blockers handled at head a46dd63.

**initiativeDir** — addressed in the diff this time, at both sites you named. It is `NOT NULL` in the schema and a non-optional `string` on `RunRow`, so there is no null to guard; I have said exactly that in a comment at `gate-spawner.targetFor` and `attention-pass.lookAtGate`, and deliberately not restored a fallback for a value that cannot be absent. If you would rather see a runtime assertion than a comment, say so and I will add one, but adding `?? "(not recorded)"` back would be dead code Sonar itself would flag.

**Sonar** — the one new violation was the `as number` cast on `owner.prNumber`. It was there because `owner` is mutable (the discovery branch writes `prNumber` onto it), so TypeScript would not carry the null-narrowing across that write. Replaced with a local `const prNumber` plus an explicit guard: narrows honestly, drops the cast, and skips-with-reason if the invariant is ever broken rather than asserting through it.

Full suite green, tsc clean. Re-requesting once Sonar re-runs.

brannon-bowden (approved):
Both previously-blocking items are resolved, and the new `run-loop.ts` change is a genuine correctness improvement, so this is approvable.

## Resolved from the prior review

- **CI red** — the latest run reports all 7 checks green, including SonarQube Code Analysis and quality-gate.
- **`run.initiativeDir` null-guard** — `attention-pass.ts:82-84` and `gate-spawner.ts:116-118` now carry comments asserting `initiativeDir` is `NOT NULL` in the schema and typed as a non-optional `string` on `RunRow`. I can't see the schema or the `RunRow` type definition in this diff to independently verify that invariant, but if the type is genuinely non-optional the TypeScript compiler enforces it at every call site, which is a real guarantee, not just documentation. I'm treating this as resolved; flagging only as a note in case `RunRow` is ever loosened (e.g., via an `as any`/partial cast elsewhere) without updating these two sites.
- **`owner.prNumber as number` cast** — `run-loop.ts:172-179` replaces the unsafe cast with an explicit null check that skips and logs (`'a PR number was expected by now but is not recorded'`) before calling `deps.reconcile`. This is a legitimate fix: previously a null `prNumber` would have been force-cast to `number` and passed straight into `reconcile`, which could have propagated a `null` where a real PR number was assumed downstream.

## Suggestions (non-blocking)

- The new skip branch in `run-loop.ts` (prNumber missing after discovery) is new control flow and isn't covered by the diff's test files. Given this PR's pattern of adding targeted tests for each new skip/guard condition (e.g., the multi-owner guard), a small test asserting this path pushes the expected `skipped` entry and returns `false` would round things out, though the logic itself is simple enough that this isn't merge-blocking.

PR #306, opened 2026-09-15
feat(run-engine): the app arms, drives, and answers runs

William-Long-II:
Refs #722, #288. This is the app-wiring feature: the Bodhilander app now **operates the run engine** end to end — it arms runs, drives them, and lets a person answer the permissions they block on. It is the piece that finally retires "you are still the dispatcher" inside the app rather than the console. Three commits, each a coherent slice.

Before this, `src/main/index.ts` referenced the run engine **zero** times: every run advanced because someone typed `step`/`watch`/`answer` into a console.

## Drive — the app runs runs (`626afd8`)

`run-loop-service.ts` wires `run-loop.ts` (which knows nothing of Electron, the store, or `gh`) to the real app: the run repository, `gh` and the plugin through `command-runner`, a gate spawner per run, the attention pass. Every `BODHI_*` the console read from the environment is a resolved field — `claude`, `gh`, prompt and channel directories under `userData`, the broker path (dev vs packaged).

Started at `whenReady`, stopped in the quit cleanup; a failure to start is **logged, not fatal**. The timer fires every 15s; each tick acts only on runs that are **due**, so the engine's cadences hold. It spawns nothing on its own and cannot answer its own prompts.

The permission **broker now ships**: `build:broker` bundles it to `dist/scripts/`, unpacked by electron-builder because `claude` spawns it as a real child process.

## Answer — a person decides, from the inbox (`3dad146`)

Closes a gap the drive slice exposed: with the hook broker a blocked background gate reads `busy`, not `waiting`, so nothing moved the run to `waitingPermission`. The loop now checks the channel **before** the attention decision — a pending request means a person is needed whatever the daemon says — and drops the run into the inbox. Answering writes the reply the hook is polling for and returns the run to `running`.

`permission-inbox.ts` reads a gate's channel and carries a decision (disk behind a `ChannelIo` dep, so both are tested against a fake): pending is read **whole**, a bare deny is given words for the model, a reply for a non-pending request is refused. Two IPC channels — read, and the one write the app exposes. The Runs view renders allow/deny inside a `waitingPermission` row.

## Arm — create a run from the app (`95d492c`)

`arm-run.ts` reads the harness the initiative pins itself to (its own `team.yaml`, not app settings — a second copy could disagree) and derives the repo root beside it, then hands the rest to `armRun`, which checks the machine and writes the rows or **refuses with a list**. It refuses rather than throwing when the directory is not a prepared initiative. `armRun` and the disk are dependencies, so the path is tested without a real python or gh. The Runs view gains `RunArm` above the inbox: pick a directory, arm, and see the run and its owners — or the refusal list in full. Nothing is launched here; the loop picks the run up next tick.

## The shape it holds to

The app never decides and never spawns directly. It **starts** runs (arm), **schedules** them (the loop calls the engine's `advance`), and **carries a person's decision** (answer) — every gate launch still goes through the driver, every verdict is the engine's, and a run waiting on a person is out of the loop's reach by construction.

## Verified

- Full suite green (`bun test --isolate`): permission-inbox 9, arm-run 11, RunInbox 27 (incl. allow/deny), RunArm 5, the loop's pending-precedence tests.
- `tsc -p tsconfig.main.json` clean; full `bun run build` green with the broker in the chain.
- **Launched the built app** three times across the slices (unpackaged → isolated dev `userData`, no real data touched): `[RunLoop] started; ticking every 15s`, clean idle ticks, every IPC registered, no startup error, clean shutdown.

## What is deliberately left

- **Machine settings** (python path, approvers) are read from env with sane defaults; a Settings UI for them is a follow-up, not a blocker — arming refuses clearly when a default is wrong.
- **Owner disambiguation** for a multi-owner repo surfaces `armRun`'s refusal naming the choices; a picker to carry the choice back is a follow-up.


src/main/run-engine/run-loop-service.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/**
+ * Build the loop's dependencies from the real app.
+ *
+ * Exported so a test can assert the wiring -- which repo call each dependency
+ * makes, which cwd `gh` runs in -- against fakes, without a timer or a real
+ * `gh` in the room.
+ */
+export function loopDeps(config: SpawnConfig, ghPath: string): LoopDeps {
```
brannon-bowden: This module (and brokerPath's dev/packaged/unpacked branching in particular) has no test coverage despite being explicitly structured for testability. Add tests for brokerPath()'s three branches and for loopDeps()'s wiring.

src/main/run-engine/run-loop-service.ts
```diff
@@ -0,0 +1,179 @@
...
+ * The channel root a person's answer is written into, exported so the
+ * permission-answer IPC and the loop agree on one location.
+ */
+export function permissionsRoot(userData: string): string {
+  return path.join(userData, 'run-engine', 'permissions');
+}
+
+/** Approvers a review request goes to. Empty until configured; the engine refuses clearly then. */
+function approvers(): readonly string[] {
+  return (process.env.BODHI_APPROVERS || '').split(',').filter(Boolean);
```
brannon-bowden: BODHI_APPROVERS entries aren't trimmed after split(','), so values after the first comma retain leading whitespace.

src/main/run-engine/run-loop-service.ts
```diff
@@ -0,0 +1,179 @@
...
+ * `scripts/`; packaged, it is built to `dist/scripts/` and unpacked. Mirrors
+ * `getHookScriptPath` in `mcp-config.ts`, for the same reason.
+ */
+export function brokerPath(): string {
+  if (!app.isPackaged) {
+    return path.join(app.getAppPath(), 'scripts', 'permission-broker.js');
+  }
+  const unpacked = path.join(process.resourcesPath, 'app.asar.unpacked', 'dist', 'scripts', 'permission-broker.js');
+  if (fs.existsSync(unpacked)) return unpacked;
+  return path.join(process.resourcesPath, 'app', 'dist', 'scripts', 'permission-broker.js');
```
brannon-bowden: This fallback assumes a non-asar resourcesPath/app layout, inconsistent with the app.asar.unpacked path checked just above. Confirm this branch is actually reachable for this app's packaging config.

brannon-bowden (changes requested):
This PR wires the run-engine's scheduler into the live Electron app (timer-driven `advance` instead of manual console commands) and ships the permission broker as an unpacked file. The wiring is careful — idempotent start/stop, dependency injection for testability, `startRunLoopService()` guarded by try/catch so a failed loop doesn't take the app down — but CI is currently red and there are a couple of open questions worth resolving before merge.

## Issues

- **CI is failing**: SonarQube Code Analysis and quality-gate checks are red. This needs to be resolved (or the failures explained/waived) before merge.
- **No tests for the new wiring module** (`src/main/run-engine/run-loop-service.ts`, entirely untested): the module explicitly exports `brokerPath`, `loopDeps`, and `spawnConfig` "so a test can assert the wiring... without a timer or a real `gh` in the room" (lines 114-120), but no such test is included in this PR. In particular, `brokerPath()` (lines 60-67) has non-trivial dev/packaged/unpacked branching that would be easy to get wrong and silently break broker resolution in a packaged build — worth covering directly.
- **Possible shutdown race** (src/main/index.ts, `before-quit` handler): `stopRunLoopService()` is now called immediately before `closeDatabase()`. If a tick is mid-flight (e.g. `advance()` awaiting a `gh` subprocess) when quit is triggered, and `RunLoop.stop()` only clears the interval timer without draining in-flight work, that tick could still attempt a DB write after `closeDatabase()` runs. This isn't visible from `run-loop.ts` in this diff, so it's a question rather than a confirmed bug — worth confirming `stop()` awaits any in-flight tick before returning.
- `approvers()` (run-loop-service.ts:90) splits `BODHI_APPROVERS` on `,` without trimming whitespace, so `"alice, bob"` yields `["alice", " bob"]` — the second entry keeps a leading space, which could silently break exact-match comparisons downstream.

## Suggestions (non-blocking)

- `brokerPath()`'s final fallback (run-loop-service.ts:66) assumes a non-asar `resourcesPath/app/dist/scripts/...` layout, which is inconsistent with the primary `app.asar.unpacked` path used just above it. If asar packaging is always enabled for this app, that branch may be dead code — worth a comment clarifying when it's expected to trigger, or dropping it if it can't.
- `executorFor()` rebuilds the full executor (owners, roles via `agentsForRun`, target, spawn config) on every `advance` call for every due run. If `agentsForRun` shells out to `claude agents`, this could add up under many concurrently active runs — worth keeping an eye on tick latency once this is running for real.

brannon-bowden (changes requested):
This increment adds real, well-tested value — the permission-inbox read/answer logic (`permission-inbox.ts`) and the loop's `pending()` precedence check are both covered by solid unit tests with a fake channel, and the `executorFor` refactor to share deps between the automatic loop and the manual answer path is a sensible design. But CI is still red and there's a code-quality slip in the wiring file, so this isn't ready to merge yet.

## Issues

- **CI is still failing**: SonarQube Code Analysis and quality-gate remain red on this PR. This must be resolved (or explicitly waived) before merge.
- **Duplicate/misplaced import in `run-loop-service.ts`**: the file now imports from `./permission-inbox` twice — `import { pendingRequests, type ChannelIo } from './permission-inbox';` near the top, and later `import { pendingRequests as readPending, writeDecision } from './permission-inbox';` inserted mid-file, after the closing brace of `loopDeps`, right before `listRunPermissions`. Splitting one module's imports across two statements (with the same binding re-imported under a different alias) and placing the second import in the middle of the file rather than at the top is a real maintainability regression and a plausible lint offender — worth consolidating into a single top-of-file import (and note `readPending` looks unused as an alias if `pendingRequests` is already in scope from the first import).
- **Deny never carries a message from the UI**: `permission-inbox.ts`/`writeDecision` explicitly supports a custom deny message (and it's unit-tested — "a supplied deny message is kept"), but `RunInbox.tsx`'s `decide()` always calls `send(runId, toolUseId, verdict, '')`, so every deny falls back to the generic `'declined by a person from the inbox'`. There's no input for a person to explain why they're declining, which undercuts the stated design intent ("a refusal it cannot read teaches an owner to retry"). Worth adding a message field (even a simple prompt/textarea) before this ships, or explicitly deferring it in the PR description if intentional for v1.
- `approvers()` (`run-loop-service.ts`, unchanged by this diff) still splits `BODHI_APPROVERS` on `,` without trimming, so `"alice, bob"` yields `["alice", " bob"]`. Not touched by this increment, so not re-litigated in depth, but still open.

## Questions (non-blocking)

- `answerRunPermission` resolves `gh` via `process.env.BODHI_GH || 'gh'` directly, while the loop's own `ghPath` is presumably threaded in from wherever `loopDeps()` is called (not shown in this diff). Is that the same resolution the loop uses, or could the manual answer path invoke a different `gh` binary than the automatic tick does?
- `loopDeps`'s new `pending` fn reads `config.permissionsRoot`; the `SpawnConfig` type and its construction aren't shown in this diff, so I can't confirm that field exists — worth a quick check that this compiles against the actual type.
- Previously raised: whether `RunLoop.stop()` drains an in-flight tick before `before-quit` calls `closeDatabase()`. Not touched by this diff and `run-loop.ts`'s `stop()` isn't shown here, so still unconfirmed either way.

## Suggestions (non-blocking)

- The new `pending(run, gate) > 0` short-circuit in `run-loop.ts` cleanly takes precedence over the attention pass and is well covered by the two new `run-loop.test.ts` cases — no changes needed there.
- `PermissionRequests` disables Allow/Deny on *all* pending requests for a run while any one decision is in flight (`disabled={busy !== null}` isn't scoped to `req.toolUseId`). Probably fine for the common case of one pending request at a time, but worth scoping to the specific request if multiple concurrent prompts turn out to be common.

src/renderer/App.tsx
```diff
@@ -1745,7 +1746,12 @@ const App: React.FC = () => {
             initialGroupId={arenaGroupId}
           />
         )}
-        {contentView === 'runs' && <RunInbox />}
+        {contentView === 'runs' && (
```
brannon-bowden: RunArm's onArmed prop (meant to refresh the inbox after a successful arm) isn't passed here, so the documented refresh behavior never fires unless RunInbox polls independently.

src/main/run-engine/run-loop-service.ts
```diff
@@ -0,0 +1,253 @@
...
+      return advance(run.id, event, target, deps);
+    },
+    approvers,
+    log: (line) => log.info(`[RunLoop] ${line}`),
+  };
+}
+
+import { pendingRequests as readPending, writeDecision } from './permission-inbox';
+import type { PermissionRequest } from './permission-channel';
+import { armInitiative } from './arm-run';
```
brannon-bowden: New imports (armInitiative, armRun, IgnitionResult) are added to this mid-file import block rather than consolidated with the file's top-of-file imports, compounding the previously-flagged import placement issue.

src/shared/types.ts
```diff
@@ -676,6 +676,25 @@ export interface AccountRemovalCost {
...
  * window by default.
  */
+/**
+ * One tool call a gate is waiting for permission on (CO-722, #288).
+ *
+ * Mirrors `PermissionRequest` in the run engine. The input is carried whole:
+ * a person approving a Bash call is approving its command line, and a
+ * summary is not a thing to consent to.
+ */
+/** What arming a run reported: the run, or the list of things to fix (CO-722). */
```
brannon-bowden: RunArmResult and its JSDoc are inserted between the pre-existing comment block and RunPermissionRequest, the interface that comment documents — consider moving it above or below that block.

brannon-bowden (changes requested):
This increment adds a solid, well-tested "arm a run" feature (`arm-run.ts`, its refuse-not-throw design, and the accompanying unit/component tests are good), but CI is still red and a couple of gaps from the previous round remain unaddressed, plus one new wiring gap in this increment — so this still isn't ready to merge.

## Issues

- **CI is still failing**: SonarQube Code Analysis and quality-gate remain red on this PR. This must be resolved (or explicitly waived) before merge.
- **Deny still never carries a message from the UI** (carried over, not touched by this increment): the earlier finding that `RunInbox.tsx`'s `decide()` always sends an empty message for denies — losing the custom-deny-message support `permission-inbox.ts` already implements — stands, since `RunInbox.tsx` isn't part of this diff.
- **`RunArm`'s `onArmed` refresh hook isn't wired up**: `RunArm.tsx` documents `onArmed` as "Told once a run is armed, so the inbox beside this can refresh," but `App.tsx` renders `<RunArm />` with no `onArmed` prop (src/renderer/App.tsx:1749-1754), so that callback is never invoked. Unless `RunInbox` independently polls (not shown in this diff), a newly armed run won't appear in the inbox without an unrelated refresh trigger. Worth wiring `onArmed={refreshInbox}` (or equivalent) here.
- **Mid-file import block in `run-loop-service.ts` persists and now grows**: the previously-flagged `import { pendingRequests as readPending, writeDecision } from './permission-inbox';` placed after `loopDeps` rather than at the top of the file is still there, and this increment adds three more imports (`armInitiative`, `armRun`, `IgnitionResult`) right alongside it (src/main/run-engine/run-loop-service.ts:180-183) instead of consolidating at the top. Worth cleaning up while touching this file.
- **New `RunArmResult` type wedged between an unrelated comment and its target**: in `src/shared/types.ts`, the new `RunArmResult` type and its own JSDoc are inserted between the existing doc comment ending "...a summary is not a thing to consent to." and `RunPermissionRequest`, the interface that comment was documenting — splitting doc from target. Minor, but worth moving `RunArmResult` above or below that comment block.
- `approvers()` in `run-loop-service.ts` (unchanged, carried over) still splits `BODHI_APPROVERS` on `,` without trimming. Not touched by this increment.

## Questions (non-blocking)

- `armInitiativeDir` resolves `python`/`gh` via `process.env.BODHI_PYTHON || 'python'` and `process.env.BODHI_GH || 'gh'` directly (src/main/run-engine/run-loop-service.ts:191-195), independent of whatever `ghPath` the loop itself is threaded with via `loopDeps(config, ghPath)`. Is this guaranteed to resolve to the same binary the automatic loop uses, or could manual arming diverge from it?
- `armInitiativeDir` calls `readIfPresent` and `runCommand` — not shown being imported in this diff's hunk. Presumably already present elsewhere in the file for other loop functionality; worth a quick compile check if not.

## Suggestions (non-blocking)

- `arm-run.test.ts`'s success-path test stubs `io.readFile` to ignore the path argument entirely, so it doesn't verify `armInitiative` actually reads `path.join(initiativeDir, 'team.yaml')` specifically. A minor strengthening opportunity, not required.
- `harnessFromTeamYaml`'s quote-stripping regex (`/^["']|["']$/g`) will strip mismatched quote pairs (e.g. a leading `"` with a trailing `'`). Unlikely in practice but easy to tighten if it comes up.

William-Long-II:
Addressed the review on `04acd09`:

- **Deny now carries a reason from the UI** — RunInbox gained a per-request reason box; a blank deny still lets the broker supply its own words. Allow carries none.
- **`onArmed` wired** — `App.tsx` refreshes the inbox on arm, so a newly armed run appears at once.
- **Imports consolidated** at the top of `run-loop-service.ts`; the mid-file block is gone.
- **`RunArmResult` moved** out from between the `RunPermissionRequest` doc and its interface.
- **`approvers()` trims** entries.
- **`harnessFromTeamYaml`** strips only a matched quote pair now; arm-run tests strengthened (asserts the team.yaml path it reads; a mismatched-quote case).
- Comment answers the gh-resolution question: manual arming uses the same `BODHI_GH || 'gh'` the loop is threaded with.

Suite green (876), tsc clean.

brannon-bowden (changes requested):
This increment cleans up every carried-over gap from the previous round — the deny-message wiring, the `onArmed` refresh hook, the mid-file import block, the misplaced `RunArmResult` doc comment, the untrimmed `BODHI_APPROVERS` split, and the quote-stripping regex are all fixed and covered by new tests — but CI is still red, so this isn't mergeable yet.

## Issues

- **CI is still failing**: SonarQube Code Analysis and quality-gate remain red on this PR. This must be resolved (or explicitly waived) before merge.

## Resolved from the previous round

- `RunInbox.tsx`'s `decide()` now sends `denyReasons[toolUseId]` on deny and `''` on allow, with a per-request input box and tests covering both paths and the allow/deny split (`src/renderer/components/RunInbox.tsx:108-155`, `RunInbox.test.tsx`).
- `App.tsx` now wires `onArmed={() => setRunsRefresh((n) => n + 1)}` on `RunArm` and remounts `RunInbox` via `key={runsRefresh}`, so a newly armed run refreshes the inbox (`src/renderer/App.tsx:1753-1755`).
- `run-loop-service.ts`'s mid-file imports (`writeDecision`, `PermissionRequest`, `armInitiative`, `armRun`, `IgnitionResult`) are consolidated into the top import block; the `readPending` alias is gone and call sites use `pendingRequests` directly.
- `RunArmResult` and its doc comment are now placed together, after `RunPermissionRequest` and before the `RunInboxRow` doc comment — no more split docs.
- `approvers()` now trims each entry after splitting on `,`.
- `harnessFromTeamYaml` now strips only a matched quote pair via `/^(["'])(.*)\1$/`, with a new test confirming a mismatched leading `"`/trailing `'` pair is left intact.

## Notes (non-blocking)

- `armInitiativeDir` now carries a comment asserting its `BODHI_GH || 'gh'` resolution matches what `startRunLoopService` threads into `loopDeps` — that's a reasonable answer to the earlier question, though the actual call site isn't in this diff to confirm against.

William-Long-II:
Cleared the two failing Sonar conditions on `f995cd3` (fixed at source, not waived):

- **new_violations → 0**: S6606 ×6 (env reads now go through an `envOr` helper — default on unset *or empty*, the intent the `||` had), S6819 (armed banner is now an `<output>`), S6479 (refusal list keyed on the reason, not the array index).
- **security hotspot**: S5852 — `harnessFromTeamYaml` regex rewritten to `/^harness:(.*)$/` with trimming in code; no lazy quantifier left to backtrack over.

Suite green (876), tsc clean.

brannon-bowden (approved):
The previously-blocking CI failures are resolved — SonarQube Code Analysis and quality-gate now both report passing — and the rest of this increment is a clean set of targeted fixes (ReDoS-safe regex, `<output>`/`role=status` semantics, an `envOr` helper that de-duplicates the `env || default` pattern, and a stable composite list key), so this is approvable.

## Issues

- **`harnessFromTeamYaml` early-return regression on blank `harness:` lines** (`src/main/run-engine/arm-run.ts:34-42`): the old pattern `/^harness:\s*(.+?)\s*$/` required at least one non-whitespace character after the colon, so a line like `harness:` or `harness:   ` (no value) simply failed to match and the loop moved on to the next line. The new pattern `/^harness:(.*)$/` matches such a line too (capturing `""`), so `if (match)` is now true and the function `return`s `null` immediately — it never gets to scan any later line. If a team.yaml ever has a blank/whitespace-only `harness:` line preceding a real one, this now silently short-circuits and returns `null` instead of finding the later value, where before it would have kept scanning. The comment frames this purely as a ReDoS-safety change ("nothing to backtrack over... whitespace handled by trim()"), but it also changes match-vs-no-match semantics for the empty-value case. Likely a rare input shape given `harness:` should only appear once in well-formed YAML, and there's no test covering it either way, so this is a minor, non-blocking correctness note rather than a merge blocker.

## Suggestions (non-blocking)

- A quick unit test for `harnessFromTeamYaml` with an empty `harness:` line ahead of a valid one (and/or a whitespace-only value) would pin down the intended behavior and guard against future regressions in this parser.
- The `<div role="status">` → `<output>` and index-key → `${r.what}::${r.fix}`-key changes in `RunArm.tsx` are solid accessibility/correctness cleanups; worth double-checking that no existing test queries the removed `role="status"` attribute directly by selector rather than by computed role (should be fine via testing-library's implicit-role resolution, but easy to verify).

PR #307, opened 2026-09-15
feat(run-engine): use the run engine from the app — machine config + prepare a run

William-Long-II:
Refs #722. Makes the run engine **usable from the app** by resolving the three "able to use it" prerequisites that #306 left behind: the machine config was env-only, and preparing an initiative required a terminal. Two coherent commits.

## Settings-backed machine config (`733aec9`)

The engine's binaries and approvers were read only from `BODHI_*` env at the loop's call sites. Now they resolve **preference → env → default**:

- `machine-config.ts` is the one place that order lives — `claudePath`, `ghPath`, `pythonPath`, `approvers`, and the three prepare paths. The preference read is wrapped so the loop can tick before the DB is open without crashing: a missing preference means "fall back", never a throw.
- A **"Run Engine" Settings tab** (`RunEngineSettings`) writes those preferences through the existing generic prefs channel; each field saves on blur. Keys are named once in `RUN_ENGINE_PREF_KEYS`, so the form and the resolver cannot drift.

A machine that already exports the env behaves exactly as before — Settings is just the friendlier way in. This resolves prerequisites **#1 (binaries) and #3 (approvers)**.

## Prepare & arm a run from the app (`2c88f47`)

Resolves prerequisite **#2**: the app now bootstraps an initiative instead of requiring the harness CLI.

- `run-loop-service` resolves every binary/approver through `machine-config`, so the Settings tab takes effect for the loop, arming, and answering alike.
- `prepare-initiative.ts` runs the harness's own two scripts the same way `ignition` runs `initiative.py` — the python interpreter against the library file, never the `.sh` wrappers Windows cannot spawn: `init_task.py <issue> <repo> --dir <root>`, then `spawn.py <dir>` with `BODHI_ROOT` set. It **refuses with a fixable list** (a missing machine setting, a bad issue name, or a script's own error) and never leaves a half-created initiative. `reposFromRegistry` reads the harness's repo names for the picker.
- The Runs view gains **`RunPrepare`** above arm: pick a repo, name the issue, optional budget → prepare then arm in one gesture, owners shown or the refusal in full. The loop drives it from the next tick.
- IPC: `db:runs:repos`, `db:runs:prepare`; preload + types to match.

## The shape it keeps

The app still never decides a verdict and never spawns a gate directly. Preparing shells out to the **harness's own** bootstrap — the app does not reimplement worktree-cutting, which must run where the repos live — and arming/driving are unchanged. Preparing writes only what `init_task`/`spawn` write; on any failure it surfaces the script's message and stops.

## Verified

- **894 tests green** (`bun test --isolate`): machine-config 4 (the preference→env→default order walked down all three rungs), prepare-initiative 9 (registry parse; the init→spawn sequence with argv + `BODHI_ROOT`; each refusal path), RunPrepare 5, plus the existing suites.
- `tsc` clean on `tsconfig.json` and `tsconfig.main.json`; full `bun run build` green.
- **Launched the built app** (isolated `userData`): `[RunLoop] started; ticking every 15s`, every IPC registered, no startup error, clean shutdown.

## What is deliberately left

- The **owner picker** for a multi-repo initiative still surfaces `armRun`'s refusal naming the choices; carrying the choice back is a follow-up.
- Preparing is **single-repo** (what `init_task.sh` bootstraps); a multi-repo initiative is still an `arch` amendment, unchanged by this PR.


src/main/run-engine/run-loop-service.ts
```diff
@@ -206,7 +194,40 @@ export function armInitiativeDir(initiativeDir: string): Promise<IgnitionResult>
...
+  issueId: string,
+  repo: string,
+  budgetUsd?: number,
+): Promise<RunPrepareResult> {
+  return prepareInitiative(
+    { issueId, repo, budgetUsd },
+    { run: (exe, argv, opts) => runCommand(exe, argv, { timeoutMs: 15 * 60_000, env: opts.env }) },
+    {
+      pythonPath: machine.pythonPath(),
+      harnessPath: machine.harnessPath(),
```
brannon-bowden: runCommand is invoked with env: opts.env, and for the spawn.py call opts.env is exactly { BODHI_ROOT: bodhiRoot } with nothing else. If runCommand passes this straight to child_process.spawn, the child loses PATH/HOME/etc. since Node does not merge a supplied env with process.env by default. Please confirm runCommand merges with process.env, or merge explicitly here.

brannon-bowden (changes requested):
This is a well-structured addition (clean separation between `machine-config`, `prepare-initiative`, and the wiring in `run-loop-service.ts`, with solid unit test coverage for the resolution order and the refusal paths), but CI is currently red and one design point around environment handling needs a look before merge.

## Issues

- **CI is failing**: `SonarQube Code Analysis` and `quality-gate` checks are red on this PR. These need to pass (or be explicitly addressed) before merge.
- **Possible env clobbering on the `spawn.py` call** (`src/main/run-engine/run-loop-service.ts`, in `prepareInitiativeFromApp`): the `io.run` implementation forwards `opts.env` straight into `runCommand(exe, argv, { timeoutMs, env: opts.env })`. `prepare-initiative.ts` passes `env: { BODHI_ROOT: bodhiRoot }` for the `spawn.py` call — just that one variable, nothing merged from `process.env`. Node's `child_process.spawn`/`execFile` does **not** merge a caller-supplied `env` with the parent's environment by default; it replaces it entirely. If `runCommand`'s implementation (not shown in this diff) passes `opts.env` through to `spawn` unmerged, the `spawn.py` invocation would lose `PATH`, `HOME`/`USERPROFILE`, `SystemRoot`, etc., and likely fail to even locate `python`/`git` on the child's search path. Worth confirming `runCommand` merges with `process.env` internally; if it doesn't, this call site should pass `{ ...process.env, BODHI_ROOT: bodhiRoot }` instead.

## Suggestions (non-blocking)

- `prepare-initiative.ts` builds script paths with a hardcoded `/` (`` `${harnessPath}/scripts/lib/init_task.py` ``). If a Windows user pastes a backslash-style harness path (the `RunEngineSettings` placeholder itself suggests `C:\work\repos\...`), this mixes separators. It'll likely still resolve on Windows, but `path.join`/`path.posix.join` would be more robust and self-documenting than string concatenation.
- In `RunPrepare.tsx`, a non-numeric `budget` value silently resolves to `undefined` (harness default) rather than surfacing a validation message — a typo in the budget field fails silently instead of being flagged to the user.
- `RunEngineSettings.tsx`'s `save()` swallows failures with only a code comment explaining the rationale; consider a lightweight inline error indicator (mirroring the `saved` state) so a failed preference write isn't only a comment in the source.

William-Long-II:
Addressed the review on `d329a63`:

- **new_violations → 0**: S6606 (machine-config `bodhiRoot` uses an explicit non-empty check, no `||` on a maybe-undefined); S3358 (RunPrepare's label ternary extracted to `startLabel`).
- **Budget validation**: a typed-but-invalid or negative budget is now surfaced ("not a valid budget") instead of silently falling back; the field is text+numeric inputmode so a typo reaches the check.
- **Separators**: `prepare-initiative` builds the harness script paths with `path.join`.
- **Save feedback**: RunEngineSettings shows an inline "not saved" on a failed write, not just a comment.

On the env question — not a bug: `runCommand` merges `{ ...process.env, ...opts.env }` (command-runner.ts:76), so `{ BODHI_ROOT }` for spawn.py is **additive** — PATH/HOME/git still reach the child. Added a note at the call site.

Suite green (895), tsc clean, full build green.

brannon-bowden (approved):
Approve — CI is now fully green and this update addresses the previously raised design points.

The prior blocking CI failures (SonarQube, quality-gate) are resolved; all 7 checks are now passing. The two non-blocking suggestions from the last review have also been addressed: `prepare-initiative.ts` now builds script paths with `path.join(harnessPath, 'scripts', 'lib', ...)` instead of hardcoded `/` concatenation (and the test was updated in lockstep so it stays platform-agnostic), and `RunPrepare.tsx` now validates the budget field, surfacing an error and refusing to call `prepare()` when the input isn't a valid non-negative number (covered by a new test). `RunEngineSettings.tsx` also gained a visible "not saved" indicator on a failed preference write, replacing the comment-only rationale from before.

## Suggestions (non-blocking)

- `prepare-initiative.ts`'s new comment asserts that "`runCommand` merges [`BODHI_ROOT`] over the parent process env, so PATH, HOME and the rest still reach `spawn.py`." That's a claim about code outside this diff (`runCommand`'s implementation isn't shown here), so it can't be verified from what's visible. If that merging behavior is accurate, it'd be worth a one-line note or link near `runCommand`'s definition itself so the guarantee doesn't live only as an assertion at the call site; if it turns out `runCommand` does *not* merge, this comment would actively mislead the next reader.
- `startLabel()` in `RunPrepare.tsx` is a reasonable extraction, but the JSDoc-style comment above it ("a plain lookup rather than a nested ternary") explains a stylistic choice rather than a non-obvious constraint — could be dropped per the project's usual comment bar.

PR #308, opened 2026-09-15
feat(run-engine): per-owner gate state in the schema (multi-owner, slice 1)

William-Long-II:
Refs #722. **Slice 1 of driving multiple repos per run** (CO-722 Phase 1). The run engine drives one repo today — `run-loop.ts` explicitly skips any run with more than one owner, and `gate-spawner.ts` is hardwired to `owners[0]`. This slice lays the per-owner foundation and **ships dormant**: it adds the data model and the rollup, but nothing calls the new paths yet, and a single-owner run is byte-for-byte unchanged.

The design insight (validated in planning): the pure `transition(state, event, context)` machine is single-track and side-effect-free, so **multi-owner = run that same machine per owner**. This slice makes the *storage* per-owner; later slices make the driver and loop fan out over owners.

## Schema
- `run_owners` gains `state`, `blocked_reason`, `merge_order`; `run_gates` gains `repo`.
- The migration is **more than the nullable-column precedent**: it **backfills** `run_gates.repo` (from a single-owner run's sole owner) and `run_owners.state` (from the run's current state, skipping the run-level prelude states `preparing`/`provisioning`/`failed`). Without the backfill, an in-flight run would strand the moment the per-owner paths read a NULL. Idempotent (touches only NULLs).

## Rollup
- `rollupState(ownerStates)` (pure, `transitions.ts`): `runs.state` becomes a **lossy summary** of its owners'. Precedence — any owner needing a person surfaces the whole run; `done` = all owners done; `approved` = all approved (≥1 not merged). **`rollup([s]) === s`**, so single-repo is identical. It is **only** a coarse filter key for the active-runs list and the inbox; scheduling and per-owner dispatch (later slices) read each owner's own state, so one stuck owner never starves the healthy ones.

## Repository API
- `recordOwnerTransition(runId, repo, …)` — moves one owner and **recomputes the rollup in the same transaction**, so two owners advancing can't lose each other's write.
- `activeGate(runId, repo)` — isolates each owner's in-flight gate (two repos at gate 4 share the `verifier` role but each sees its own row); no-arg callers unchanged. `startGate` and the event log carry `repo`; attempts count per owner.

## Verified
- New tests: `rollupState` (precedence, singleton identity, all-done/all-approved, starvation-safety ordering); per-owner `activeGate` isolation + two verifiers not colliding; the rollup recompute; the migration backfill (single-owner backfilled, `preparing` left NULL, multi-owner gate not guessed, idempotent).
- **1405 tests green** (`bun test --isolate`), `tsc` clean on `tsconfig.json` and `tsconfig.main.json`, full `bun run build` green.

## Next slices
2. Per-owner driver + gate spawning (off `owners[0]`). 3. Loop fan-out + scheduling decouple (remove the `>1` skip). 4. Permission channel per owner + `merge_order` display.


brannon-bowden (changes requested):
This is a well-structured migration with strong test coverage for the tricky backfill cases, but the backfill drops data for one in-flight scenario it explicitly claims to handle.

## Issues

- **`src/main/database.ts` (the `run_owners` backfill `UPDATE`)**: the migration copies `runs.state` onto every owner row (`SET state = (SELECT state FROM runs r WHERE r.id = run_owners.run_id) ...`) but never copies the paired `runs.blocked_reason` onto the new `run_owners.blocked_reason` column. If an in-flight run is currently in a blocked, non-terminal state (`inconclusive`, `waitingPermission`, `waitingHumanGate`) at migration time, each owner ends up with `state = 'inconclusive'` (etc.) but `blocked_reason = NULL`, even though `runs.blocked_reason` (pre-existing column, visible in the `oldDb()` fixture) holds the real reason. The PR's own comment frames this backfill as "correctness-critical... a NULL would strand an in-flight run," and `blockedReason` is documented as something that's "paired with a blocked `state`" — so this is exactly the kind of stranding the migration set out to prevent, just for `blocked_reason` instead of `state`. It's also untested: `seed()` in `run-schema.test.ts` only exercises `running` and `preparing` run states, never a blocked one, so this gap has no coverage. Fix: extend the `run_owners` backfill to also set `blocked_reason = (SELECT blocked_reason FROM runs r WHERE r.id = run_owners.run_id)` under the same condition, and add a test seeding a run in a blocked state (e.g. `inconclusive`) with a non-null `blocked_reason` to lock this in.

## Suggestions (non-blocking)

- `src/main/repositories/runs.ts`, `recordOwnerTransition`: the `if (states.length === 0) return;` guard after the rollup `SELECT` looks unreachable in normal use — the preceding `UPDATE` always sets a non-null `state` for the very owner row being transitioned (assuming `run_id`/`repo` match an existing row), so `states` can only be empty if the `UPDATE` matched zero rows (a caller bug: unknown repo). Either drop it, or update the comment to reflect that it's a defensive guard against a mismatched `repo`, not a legitimate "still run-level" case.
- The `Exclude<RunState, BlockedState>` overload of `recordOwnerTransition` still permits passing run-level-only states (`preparing`, `provisioning`, `failed`) as an owner's state, even though `rollupState`'s ladder — and the surrounding comments — treat those as states that should never appear per-owner. If that ever happens, `rollupState` silently falls back to an arbitrary owner's state instead of surfacing the misuse. Worth tightening the type (or asserting) if it's easy, though not urgent since nothing in this diff calls it that way.

William-Long-II:
Fixed on `5008e92`:

- **The blocking finding**: the backfill now copies `runs.blocked_reason` onto `run_owners.blocked_reason` alongside `state`, under the same condition — a run blocked at migration time keeps its reason. Added a test seeding an `inconclusive` run with a reason and asserting the owner inherits both.
- **recordOwnerTransition** overloads now forbid `preparing`/`provisioning`/`failed` as an owner state (compile error, not a silent rollup fallback).
- Clarified the empty-states guard is defensive against a mismatched repo.

Suite green, tsc clean.

brannon-bowden (approved):
This update resolves the previously blocking backfill gap and cleans up the type overloads; nothing here blocks merge.

## Issues

- **`src/main/repositories/runs.ts` (the `recordOwnerTransition` overloads)**: the blocked overload narrowed from `state: BlockedState` to `state: 'inconclusive'`, and the permissive overload's exclusion list is now `Exclude<RunState, 'inconclusive' | 'preparing' | 'provisioning' | 'failed'>`. If `BlockedState` includes other blocked values (e.g. a `waitingPermission`/`waitingHumanGate`-style state) beyond `inconclusive`, those would now fall through to the permissive overload, which takes `detail?: EventDetail` with no required `blockedReason` — the same "state without a paired reason" hole the migration fix in this PR was written to close, just moved to the call-site type contract instead of the backfill. This diff doesn't show the `RunState`/`BlockedState` definitions, so I can't confirm whether `inconclusive` is in fact the only blocked owner-level state (in which case this is a correct, purely clarifying change) or whether other blocked states exist and are now under-typed. Worth a quick confirmation from the author; not blocking since CI (including the type-check-gated `quality-gate`) is green, which suggests no live call site is currently affected.

## Verification of the prior blocking finding

- The `run_owners` backfill now sets `blocked_reason = (SELECT blocked_reason FROM runs r WHERE r.id = run_owners.run_id)` alongside `state`, under the same `WHERE state IS NULL AND state NOT IN (...)` guard (`src/main/database.ts`). This is exactly the fix requested, and it's atomic with the `state` write so a blocked owner row can never land with a reason-less state.
- The new test (`run-schema.test.ts`, run `r4` with `state = 'inconclusive'` and a non-null `blocked_reason`) exercises the previously-uncovered blocked-at-migration-time path and asserts both fields land correctly on the owner row.
- The `if (states.length === 0) return;` guard's comment was updated to describe it as defensive against a mismatched repo rather than a legitimate run-level case, addressing the earlier readability note.

Both previously raised points are resolved by this delta.

PR #309, opened 2026-09-15
feat(run-engine): drive a run per owner (multi-owner, slice 2)

William-Long-II:
Refs #722. **Slice 2 of driving multiple repos per run** (CO-722 Phase 1), stacked on #308. The driver and gate spawner now act on **one owner named by repo**, so the per-owner state machine that #308 added is actually exercised. The loop still drives a single owner (its first) — fanning out over every owner is the next slice — so a single-repo run is unchanged, now flowing through the per-owner path.

## driver.ts
`advance(runId, repo, event, target, deps)` reads **that owner's** state (`run_owners.state`, falling back to `runs.state` for the first transition after the run-level bootstrap) and **that owner's** active gate (`activeGate(runId, repo)`). Transitions persist on the right plane:
- the prelude/terminal states (`preparing`, `provisioning`, `failed`) stay **run-level** — provisioning runs once over every worktree, and a failed install is not one repo's fault;
- everything else writes the **owner's** track and rolls the run up in the same transaction.

The gate-4 verifier→scribe sequence, gate-row opening, and the stale-report guard are all repo-scoped.

## gate-spawner.ts
`agentsForOwner`/`targetFor`/`spawnGateFor` take one owner instead of `owners[0]`. The spawn refusal compares repo, and **`channelKeyFor` + the permission channel dir carry the repo** — so two owners sharing the gate-3/4 roles (`verifier`, `reviewer`) no longer collide on one channel. Without it, a person's allow/deny at gate 4 could reach the wrong owner (a real correctness hazard the design review flagged).

## run-loop-service.ts
Builds a **per-owner** executor; the loop's `advance` and the permission-answer path name the (still single) owner's repo.

## Verified
- New driver integration tests: two owners advance on their own tracks without cross-talk, and a **stale gate-2 verdict for a repo at gate 4 is ignored while another repo advances** (proving `activeGate` is read per repo). `channelKeyFor` distinguishes repos. Every existing driver / gate-spawner / permission-inbox test updated to the per-owner shape.
- **1408 tests green** (`bun test --isolate`), `tsc` clean on `tsconfig.json` and `tsconfig.main.json`, full `bun run build` green.

## Next slices
3. Loop fan-out over all owners + scheduling decouple (remove the `>1` skip). 4. Permission channel per owner in the inbox UI + `merge_order` display.


brannon-bowden (approved):
**Approve.** This is a clean, mechanical propagation of a `repo` parameter through the run-engine driver, gate-spawner, and permission-inbox so that a run's per-owner gate/state tracking doesn't collide across owners. The state-plane split in `writeTransition` (run-level bootstrap/terminal states vs. owner-level track) is well-reasoned and matches the bootstrap fallback documented on `ownerState`. The new multi-owner tests exercise the actual cross-contamination scenarios (stale gate verdict for one repo while another repo is genuinely progressing), and the run-loop-service changes are honestly scoped to "single owner for now" with a clear note that fan-out is a later slice. CI is green across all checks.

## Issues

None blocking.

## Suggestions

- `src/main/run-engine/__tests__/driver.test.ts:677` — `db.prepare("UPDATE runs SET state = 'running' WHERE id = 'run-1'")` is missing the trailing `.run()`, so the statement is prepared but never executed. It's harmless today (the assertions rely on `run_owners.state`, which is set correctly via the `.run()` call a few lines above), but it's dead/misleading code that should either call `.run()` or be deleted.
- Worth a follow-up note (non-blocking, already acknowledged in the PR): `loopDeps.advance` and `answerRunPermission` in `run-loop-service.ts` both pick `listOwners(run.id)[0]` — the alphabetically-first repo — as "the" owner to drive. That's fine as an explicit interim step toward full fan-out, but flag it if the next slice doesn't land soon, since it means only one owner of a multi-owner run ever actually advances through the loop today.

PR #310, opened 2026-09-15
feat(run-engine): the loop drives every owner of a run (multi-owner, slice 3)

William-Long-II:
Refs #722. **Slice 3 of driving multiple repos per run** (CO-722 Phase 1). The loop no longer refuses a multi-owner run — it drives **each repo's owner on its own track** in one pass, and schedules the run off its **fastest** owner so a stuck one cannot starve the healthy ones. This removes the `owners.length > 1` skip that has stood since the engine's first slice.

## run-loop.ts
- **The skip is gone.** `passOne` iterates `listOwners` and dispatches on **each owner's** state: `running` → look at that owner's gate (`activeGate(runId, repo)`); a reconcile state → reconcile that owner's PR (its **own** phase decides how checks and reviews read, since two repos can be at different phases at once); anything else is the inbox's or done. A pass that drove nobody (the only movable owner is waiting on a person) is **not** a backoff.
- **Scheduling is decoupled from the rollup.** `schedulingState` picks the shortest-interval movable owner, so a repo at gate 2 (60s) isn't paced by a sibling at review (5m), and a run whose rollup is a person-state (`inconclusive`, `waitingPermission` — all null-interval) is **still due** while any owner can move. This is the crux: scheduling off the rollup would let one inconclusive owner zero out a healthy owner's cadence and starve it.
- `LoopDeps.advance` / `activeGate` are per owner; `run-loop-service` wires the loop's advance to that owner's executor.

## Verified
- Two owners driven in one pass (one at a gate, one reconciling), with no "drives one repo" skip.
- **The starvation regression**: a run whose rollup is `inconclusive` is still `due`, and its `running` owner still moves, while the stuck owner waits for the inbox — proving scheduling reads owner state, not the rollup.
- `schedulingState` unit tests; every existing loop test updated to the per-owner fake.
- **1412 tests green**, `tsc` clean on both configs, full `bun run build` green.

## What is deliberately NOT here (flagged)
A run still does **not start on its own** — nothing dispatches `prepared`, so an armed run sits in `preparing` (this is **pre-existing and true for single-repo too**, not introduced here). Starting a run — provision once (run-level), then fan out a gate 2 per owner — is the next piece, and it forces a design choice about how run-level provisioning meets per-owner state. I'll raise that separately.

## Next
4. Permission channel per owner in the inbox UI + `merge_order` display. Then the START/fan-out piece to make runs self-start end to end.


brannon-bowden (changes requested):
This is a well-structured refactor of the run loop to drive every owner of a multi-owner run independently, and the new tests (`schedulingState`, the two multi-owner `passOne` scenarios) exercise the core new logic directly, but CI is currently red and needs to be resolved before merge.

## Issues
- CI is failing: **SonarQube Code Analysis** and **quality-gate** are both red. Per this repo's gating policy this PR cannot be approved until those checks pass — please pull the Sonar findings and address whatever they flag (likely complexity/duplication introduced by the `lookAtOwner`/`reconcileOwner` split, or the new `schedulingState` reduce) before merge.

## Suggestions (non-blocking)
- `schedulingState` picks the run's tick cadence from the *fastest* movable owner (`run-loop.ts:94-118`), but `passOne` then drives **every** movable owner on each tick regardless of that owner's own cadence (`run-loop.ts:230-260`). For a run with one owner at gate-2 (60s) and another `waitingReview` (5m), the slower owner will now be reconciled every 60s instead of every 5m, since the whole run becomes due at the faster owner's interval. That may be intentional (simpler to always drive what's due at the run level), but if the intent was for each owner to respect its own cadence, this is worth a second look — could you confirm whether per-owner throttling exists elsewhere (e.g. inside `reconcile`) or whether this is an accepted trade-off?
- Test coverage for multi-owner `passOne` covers "one owner blocked, other advances" for the `look`/gate path, but there's no equivalent test for two RECONCILES-state owners where one owner's `reconcile` call returns `problems` — worth confirming the other owner's reconcile/advance still proceeds independently in that case, matching the `ok = (await ...) && ok` aggregation pattern.

William-Long-II:
Fixed on `6eb1b39`:

- **new_violations → 0**: S6959 — `schedulingState`'s reduce now has an initial value (seeded with the first movable state; it was already guarded non-empty).
- **Cadence trade-off** (your Q): confirmed **accepted** and documented in `passOne`. The run ticks at its fastest owner's interval and drives every movable owner each tick, so a slow owner (`waitingReview`, 5m) is reconciled at the faster cadence — one `gh` read, bounded by the repos in an initiative. There's no per-owner throttle today; that's a later refinement, not a correctness fix.
- **Suggested test added**: two `waitingChecks` owners where one's reconcile returns problems — the other still reconciles, and the run's pass counts a failure (the `ok = (await ...) && ok` aggregation doesn't short-circuit).

Suite green, tsc clean, build green.

brannon-bowden (approved):
Approve — this update resolves both open items from the previous review and CI is now green across the board (SonarQube, quality-gate, and both test platforms passing).

- The CI blocker is resolved: SonarQube Code Analysis and quality-gate are both reported passing now.
- The per-owner-cadence question is addressed with an explicit design-rationale comment on `passOne` (`run-loop.ts:247-253`), explaining that driving every movable owner at the run's fastest cadence is an accepted trade-off (bounded by repo count, one `gh` read per reconcile) rather than an oversight. That's a reasonable resolution — no code change was required, just making the intent explicit.
- The missing multi-owner `RECONCILES` test is now present (`run-loop.test.ts:320-343`): it drives two owners through `reconcile`, has one return `problems`, and asserts both owners are still visited (`seen` contains both `o/a` and `o/b`), the problem surfaces in the report, and the run's failure count reflects it. This directly confirms the `ok = (await ...) && ok` aggregation doesn't short-circuit and skip the second owner.
- The `schedulingState` reduce now takes an explicit initial value (`movable[0]`) instead of relying on the no-initial-value form defaulting to the first element (`run-loop.ts:113-118`). This is behavior-preserving (the no-arg reduce already used `movable[0]` as the seed) — just a clarity/defensiveness change, not a functional fix, but harmless.

## Suggestions (non-blocking)
- In the new test, the owner objects are constructed with `repo: 'repo-a'` / `'repo-b'`, but the `reconcile` fake branches on `t.repo === 'o/a'`, matching the org/repo parsed from `prUrl` rather than the label passed into `owner(...)`. It reads correctly once you trace the PR URL, but a future maintainer skimming the test could mistake this for a typo/mismatch. A short inline comment noting that `t.repo` is derived from `prUrl` (org/repo), distinct from the owner's `repo` label, would save the next reader a double-take.

PR #311, opened 2026-09-15
feat(run-engine): answer each owner's permission in the inbox (multi-owner, slice 4)

William-Long-II:
Refs #722. **Slice 4 of driving multiple repos per run** (CO-722 Phase 1). Closes the **permission-channel correctness gap** the design review flagged back in slice 2: a run can have several gates in flight, and two owners at gate 4 both run the `verifier`, so a person's allow/deny had to name **which repo** it belonged to — otherwise answering one owner could unblock another.

- **`listRunPermissions`** reads **every** owner's channel (each via its own `activeGate(runId, repo)`) and tags each pending request with its `repo`, so the inbox can show two blocked owners at once.
- **`answerRunPermission`** takes a `repo`: it writes into that owner's channel and advances **that owner's** track (checking the *owner's* state, not the run's rollup). Answering one owner never touches another's gate — the channel key already carries the repo, so the two resolve to different directories.
- IPC / preload / renderer thread the repo through; each request in the inbox shows which repo it is from, and Allow/Deny carry that repo.

## Verified
- The inbox shows two owners' requests, and answering the second carries `repo-b`; two owners at the same gate resolve to **different channel dirs**; existing answer tests updated to the repo-carrying signature.
- **1889 tests green** (main + renderer), `tsc` clean on both configs, full `bun run build` green, and the **built app launches clean** (`[RunLoop] started`, no error — the answer IPC signature changed, so this confirms wiring).

## Where multi-owner stands after this
Slices 1–4 make the engine **arm, drive, and answer** a cross-repo run per owner end to end: per-owner state + rollup (1), per-owner driver + gate/channel (2), the loop fans out over every owner (3), and now per-owner permission answering (4).

## Deliberately deferred (I'll raise separately)
- **`merge_order` display** — parse `seams.yaml`'s order and show it (display-only, non-correctness).
- **Runs don't self-start yet** — nothing dispatches `prepared` (pre-existing, single-repo too); the START/fan-out step (provision once, open a gate 2 per owner) is the piece that makes an armed run actually begin, and it needs a small design decision I'll bring you.


brannon-bowden (changes requested):
This PR correctly threads a `repo` parameter through the IPC/preload/renderer stack so each owner's permission gate can be answered independently, but the core service-layer fix lacks direct test coverage, so I'm requesting changes before merge.

## Issues

- **Missing test for the actual bug being fixed** (`src/main/run-engine/run-loop-service.ts:257-280`, `answerRunPermission`): the change from `listOwners(runId)[0]` to `listOwners(runId).find(o => o.repo === repo)`, and from gating on `run.state === 'waitingPermission'` to `runsRepo.ownerState(runId, repo) === 'waitingPermission'`, is the entire point of this slice — it's what stops answering owner A's permission from advancing owner B's track. Neither new test in this diff exercises that function directly: `permission-inbox.test.ts` only checks that `channelDirForGate` produces distinct directories, and `RunInbox.test.tsx` mocks `answer` entirely, so it only proves the UI *passes* the right `repo` string, not that `run-loop-service.ts` *uses* it correctly to pick the right owner/gate and call `advance` on the right track. Please add a unit test around `answerRunPermission` (and ideally `listRunPermissions`) with two owners simultaneously in `waitingPermission`, asserting that answering one does not call `advance` for the other.

## Suggestions (non-blocking)

- `decide`'s busy-lock in `RunInbox.tsx` still keys off `toolUseId` alone (`setBusy(toolUseId)` / `disabled={busy !== null}` globally disables all buttons while any single request is being answered), rather than per-`repo`. With two owners now able to be blocked concurrently, answering one owner's request will disable the other owner's buttons too until the first completes. Given `toolUseId`s are independently generated per conversation and collisions are effectively impossible, this isn't a correctness bug, but scoping busy state to `repo:toolUseId` would let a user act on both owners' queues without waiting on each other.

CI: only "GitGuardian Security Checks" is reported and it's passing; no other checks listed.

William-Long-II:
Fixed on `bb2f68a`:

- **The blocking finding** — added a direct service-layer test (`run-loop-service.test.ts`) over a real in-memory DB with the disk channel and the driver faked. Two owners both in `waitingPermission`, each with its own gate-4 in flight; asserts answering `repo-a` advances **only** repo-a's track (plus: a not-there request advances nobody; an owner not waiting on permission isn't advanced even when the reply is written). `listRunPermissions` covered too (each request tagged with its repo).
- **The busy-lock suggestion** — RunInbox's busy state is now per request (a Set keyed `repo:toolUseId`), so answering one owner no longer disables the other's buttons. Test: a hung answer on repo-a leaves repo-b's buttons usable.

1894 green (main + renderer), tsc clean, build green.

brannon-bowden (approved):
This PR now has direct test coverage for the exact bug it fixes, and the busy-lock scoping I'd previously flagged as a nice-to-have is also done — no blocking issues remain.

The new `run-loop-service.test.ts` sets up two owners (`repo-a`, `repo-b`) both parked in `waitingPermission` with their own gate 4, and asserts:
- answering `repo-a` calls `advance` for `repo-a` only, never `repo-b` (`src/main/run-engine/__tests__/run-loop-service.test.ts:69-75`) — this is the property the whole slice exists to guarantee, and it's now exercised against the real service function with a real in-memory DB, not just mocked at the UI layer.
- a decision that fails to write (`wrote = false`) advances nobody (`:77-81`).
- an owner that isn't in `waitingPermission` is not advanced even if the write succeeds (`:83-87`), covering the state-gating half of the fix.

This closes out the gap from the previous review.

The `RunInbox.tsx` busy-lock also moved from a single global `toolUseId` string to a `Set<string>` keyed by `${repo}:${toolUseId}` (`src/renderer/components/RunInbox.tsx:84-86`, `108-131`), so answering one owner's request no longer disables the other owner's buttons — previously a non-blocking suggestion, now implemented and covered by the new "leaves the other owner's buttons usable" test.

## Suggestions (non-blocking)

- The mocked `pendingRequests` in the new service test returns the same fixed request for any truthy `gate` argument, regardless of which owner's gate is passed in. That's fine for testing the `advance`-routing property, but it means this test can't catch a regression where the wrong owner's *gate* is looked up before `advance` is even called (channel/gate resolution is still only covered by the separate `permission-inbox.test.ts`). Not necessary to change, just noting the boundary of what this test proves.

CI: all 7 reported checks are passing (SonarQube, quality-gate, GitGuardian, postinstall on both platforms, tests on both platforms).

PR #312, opened 2026-09-15
feat(run-engine): the loop starts an armed run (multi-owner, START)

William-Long-II:
Refs #722. **The loop now starts an armed run** — the last piece that makes a run (single- or multi-repo) actually begin. Until now nothing dispatched the `prepared` event, so an armed run sat in `preparing` forever; the loop only drove runs already `running`/reconciling. This closes that foundational gap.

## What it does
- **A `preparing` run is due.** `baseIntervalFor('preparing')` returns a cadence, so the loop picks an armed run up and **provisions it once** (over the whole initiative) by advancing its **first owner** with `prepared`. The state machine couples `provisioned → running + gate 2` for that owner — provision is **not** run per repo.
- **The remaining owners fan out.** Each is a null-state owner of the now-running run, brought onto its own track by a thin driver op **`startOwnerGate`** (set `running`, open its gate 2). It reuses the driver's row-before-spawn ordering, is **idempotent** (an owner already on a track is skipped), and **does not touch the pure state machine** — it's orchestration, not a new state rule.
- **Scheduling keeps a run due while any owner is unstarted**: `schedulingState` maps a null owner to the `preparing` cadence, so the fan-out completes even after the started owners settle.

## The design choice (why this shape)
The fork was how run-level provisioning meets per-owner state. Provisioning is genuinely the **run's** (one `provision.sh` over the whole initiative dir), while gates are the **owner's**. Rather than fork the state machine, the first owner rides the provision (the machine already couples them) and the rest are fanned out explicitly — keeping `transitions.ts` untouched, which has been this whole feature's invariant.

## Known limitation (noted, follow-up)
Provisioning runs **synchronously inside the tick**, so a long install blocks the loop from other runs until it returns. Fine while a tester runs one initiative at a time; backgrounding provision is a separate improvement, not a correctness issue.

## Verified
- A `preparing` run provisions via its first owner's `prepared` (once, not per repo); a null-state owner of a running run is started at gate 2; `startOwnerGate` seeds `running` + opens gate 2 and is idempotent; `schedulingState` keeps a run with an unstarted owner due; the reconcile-loop cadence tests updated (preparing is now polled, provisioning is not).
- **1424 tests green**, `tsc` clean on both configs, full `bun run build` green, and the **built app launches clean** (`[RunLoop] started`, no error).

## Where this leaves the feature
With this, the engine **arms → starts → provisions → drives every owner → answers permissions → to approval**, cross-repo, entirely from the app. The only remaining item is the small, display-only `merge_order` surfacing.


brannon-bowden (approved):
This PR wires the run loop to automatically provision an armed run and bring additional owners onto their track, and it's backed by solid, well-targeted tests (idempotency of `startOwnerGate`, scheduling of null-state owners, the `preparing` polling cadence). No blocking issues found based on what's shown in the diff.

## Issues

None that block merge. All CI checks are passing.

## Suggestions (non-blocking)

- `run-loop.ts`: the `preparing` branch always calls `deps.advance(run, owners[0], { kind: 'prepared' })` — i.e. `owners[0]` is the owner that "rides" the provision. This is deliberate per the tests, but it does mean correctness depends on the owners array being returned in a stable, deterministic order (e.g. insertion order) across ticks — if a failed provision attempt were retried against a different "first" owner due to reordering, you could end up with inconsistent partial state. Worth double-checking that the owners repository query guarantees stable ordering (e.g. `ORDER BY` a stable key), since that isn't visible in this diff.
- `driver.ts` (`startOwnerGateOnce`): the owner is marked `running` via `recordOwnerTransition` before `execute` runs the `spawnGate` action. If the spawn fails, the owner row is left `running` with no live gate. This mirrors the existing row-before-spawn ordering used elsewhere (per the doc comment), and the "look at a running gate" reconcile path presumably catches this, so it's not a new risk — just flagging for confirmation that the existing recovery path indeed covers a gate that never actually launched.

Issue #313, opened 2026-09-15
Relay host publishes 47393 on 0.0.0.0, which the TRUST_PROXY=true rate limiting assumes it does not


William-Long-II:
Split out of #254, which fixed the relay docs but deliberately left the host alone. #254 can close without this; the divergence it found should not close with it.

## What diverges

`relay/README.md` documents — and its `docker run` recipe specifies — `-p 127.0.0.1:47393:8080`. During the #248 deploy, `docker inspect` on the live container reported the publish as `0.0.0.0:47393`.

## Why it is not just cosmetic

Production `.env` sets `TRUST_PROXY=true`. With that on, `clientIp` buckets rate-limit callers by the **rightmost** `X-Forwarded-For` entry (`relay/src/rate-limit.ts`), on the assumption that exactly one trusted proxy — Caddy — sits in front and wrote it. That assumption is what the loopback binding buys.

If the port is reachable on a routable address, a caller who connects directly writes that rightmost entry themselves and gets a fresh bucket per request, evading every limit — including the ones guarding link-code guessing.

## What is not established

- **Reachability.** A direct probe of 47393 on both relay hosts timed out rather than being refused. That is consistent with a host firewall or security group blocking the port, which would mean the Docker binding is wide but nothing external can use it. It is equally consistent with filtering that does not apply to some source. Unconfirmed either way.
- **Current state.** The `docker inspect` reading is from the #248 deploy, not from today. Nobody with SSH access has re-checked since.

So this is "a documented invariant the host may not be honouring", not a confirmed live exploit. Confirming it is the first step, not the fix.

## Suggested resolution

1. `docker inspect bodhi-relay --format '{{json .HostConfig.PortBindings}}'` on both `bodhilander.bodhilabs.dev` and `bodhilander.sytanek.tech` to establish the real binding.
2. If it is `0.0.0.0`, recreate the container on `127.0.0.1:47393` per the README recipe. #254 notes this is the better end state: the loopback rationale is sound and the binding should honour it. Caddy already proxies to `127.0.0.1:47393`, so nothing user-facing changes.
3. Independently, confirm the host firewall denies 47393 from off-host, so the binding is not the only thing standing between the limiter and the open internet.

Alternative, if loopback turns out to be infeasible: turn `TRUST_PROXY` off and accept bucketing by socket peer, or amend the docs to describe what actually runs and drop the safety claim. Leaving `TRUST_PROXY=true` on a routable binding is the one combination to avoid.

[...]

PR #314, opened 2026-09-15
254 — relay /health reports the deployed commit; deploy docs match the live host

### LINKED ISSUE #254: Relay deploy docs describe a host that does not match, and nothing records what is deployed
Found while deploying the relay for #248. Neither of these broke anything, but both cost real time and would mislead the next person.

**`relay/README.md` does not match the live host.** It documents the checkout at `/root/bodhi-relay`; it is actually `/home/bodhilabs/bodhi-relay`. It publishes the container on `127.0.0.1:47393`, with a paragraph explaining that binding to loopback keeps the plain-HTTP relay off the network — the live container publishes `0.0.0.0:47393`. Following the documented `docker run` verbatim would have silently changed the host binding, which is why that deploy was done from the container's own `docker inspect` instead.

Either the docs should describe what is running, or the host should be moved to match the docs. The second is the better end state — the loopback rationale in the README is sound and the current binding does not honour it — but that is a change to make deliberately, not as a side effect of a routine deploy.

**Nothing on the host says what is deployed.** `~/bodhi-relay` is an rsync target, not a checkout, so there is no ref, no tag, no build stamp. The only way to tell a stale relay from a current one was to `grep -c handoff src/http.ts` and compare against the repo. `/health` returns `version: 0.1.0`, which is the package version and has not moved in months, so it cannot answer the question either.

Stamping the deployed commit into the image as a build arg and surfacing it on `/health` would turn that investigation into one curl.

William-Long-II:
Closes #254

Initiative 254. `/health` now reports the commit baked into the image, and the relay deploy docs describe the host that is actually running.

## What changed

**The build stamp.** `docker-compose.yml` passes `RELAY_COMMIT` as a build arg; the `Dockerfile` takes it as `ARG RELAY_COMMIT` and re-exports it under a *different* name, `ENV RELAY_BUILD_COMMIT`. The rename is the point: `--env-file` is applied over the image's ENV at run time, so an image whose env shared the build arg's name could have a stray `.env` value silently outrank the real stamp. `loadConfig` reads `RELAY_BUILD_COMMIT` into `config.commit`, and `GET /health` returns it alongside `version`. The same value goes on the `relay listening` startup line, so the stamp is readable from logs as well as over HTTP.

The `ARG`/`ENV` pair sits last in the `Dockerfile` so that stamping a new sha reuses every layer above it.

**An unstamped production image says so at boot.** Nothing else enforces the ARG/ENV wiring, and the failure mode is that `/health` answers `null` forever — discovered on the day someone needs to know what is deployed, which is too late. `loadConfig` pushes a warning when `isProduction` and no commit is present.

**The docs.** `relay/README.md` documented a checkout at `/root/bodhi-relay`; it is `/home/bodhilabs/bodhi-relay`. It asserted the container publishes on `127.0.0.1:47393`; the live container publishes `0.0.0.0:47393`. The README no longer asserts either binding — it records the divergence as observed but unconfirmed from this machine, because the port could not be probed from here (see below).

## The host-side half

Issue #254 raises two possible ends for the binding mismatch: correct the docs, or move the host to match them. The issue itself says the host move is "a change to make deliberately, not as a side effect of a routine deploy." It is an ops action on a live host, not a branch change, so it is not in this PR — it is tracked in **#313** (open), which `relay/README.md` also references so it does not vanish when #254 closes.

## Why the absent case is spelled out rather than written with `||`

`loadConfig` reads the stamp as:

```ts
const stamp = env.RELAY_BUILD_COMMIT?.trim() ?? '';
const commit = stamp === '' ? null : stamp;
```

An earlier revision wrote this as `?.trim() || null`, which Sonar flagged (S6606, prefer `??` over `||`). The rule's literal suggestion — `?.trim() ?? null` — is the one rewrite that must not be made here: an unset `--build-arg` reaches the process as `''`, so the empty string is precisely the value that has to collapse to `null`, and `??` passes it through. `/health` would answer `"commit": ""`, which is the unstamped-image case this work exists to make visible, reported as though it were a stamp.

That is not an argument, it is a tested property. Replacing the two lines above with `?.trim() ?? null` and re-running the spec fails on the empty case, `Received: ""` against `expect(...).toBeNull()` at `config.test.ts` L63. The shape above satisfies the rule by removing the `||` without touching behaviour, and the Quality Gate passes on it.

## Verified

All evidence below was produced at `7994130`, which is this PR's head commit.

`scripts/verify.sh` at harness 2.22.1 (commit `bcd6491`), run on `feat/254-Bodhilander` at `7994130` against base `origin/development` (`7d5cb5d`):

```
  PASS         tests
  SKIP         typecheck
  SKIP         lint
  SKIP         seams
```

Scoped specs, run directly — `--isolate` is this repo's configured test command, and the bare form segfaults on `mock.module` leaking between files:

```
$ bun test --isolate src/config.test.ts src/index.test.ts src/relay.test.ts
 32 pass
 0 fail
 100 expect() calls
Ran 32 tests across 3 files. [91.00ms]
```

Type-check, run by hand because `verify.sh` reports SKIP for this repo — `npx tsc` resolves to a placeholder package here, so the real local binary was used (typescript 5.9.3 in `relay/node_modules`):

```
$ ./node_modules/.bin/tsc --noEmit
(exit 0)
```

On a real booted relay, not by inspection. Stamped, with `NODE_ENV=production RELAY_BUILD_COMMIT=7994130`:

```
$ curl -s localhost:47399/health
{"ok":true,"version":"0.1.0","commit":"7994130","uptime":0.5459863}

{"level":"info","msg":"relay listening","port":47399,"env":"production","commit":"7994130"}
```

Unstamped, same binary, `RELAY_BUILD_COMMIT` unset in production — the compensating control for the one hop that cannot be exercised here:

```
$ curl -s localhost:47398/health
{"ok":true,"version":"0.1.0","commit":null,"uptime":0.547402}

{"level":"warn","msg":"RELAY_BUILD_COMMIT is not set — /health cannot identify this deployment. Build the image with --build-arg RELAY_COMMIT=<sha>."}
```

**Sonar.** Not runnable locally (`SONAR_TOKEN` and `SONAR_HOST_URL` are unset on this machine), so the authoritative instrument is the server's: `SonarQube Code Analysis` and `quality-gate` both pass on this PR at head `7994130`, with 0 New Issues.

## NOT verified

- **Docker's own `ARG RELAY_COMMIT` → `ENV RELAY_BUILD_COMMIT` substitution.** No Docker daemon runs on this machine (`docker info` exits 1), so no image was built. Everything either side of that one hop is proven on a booted process, as above, and the name chain was confirmed statically across all five places it appears — compose build arg, Dockerfile ARG/ENV, config read, `/health`, startup log. The production boot warning is the compensating control, and it is observed firing above.
- **The live hosts' actual port binding and `TRUST_PROXY` value.** No SSH from here: `bodhilander.bodhilabs.dev` port 22 timed out, `bodhilander.sytanek.tech` refused publickey. A direct probe of `:47393` on both hosts exits 28 (timeout), not 7 (refused) — the port is filtered, which neither confirms nor refutes the `0.0.0.0` publish. Confirming it needs `docker inspect` on the hosts; #313 tracks that.
- **Lint.** `verify.sh` reports SKIP: no lint runner is configured in the registry for this repo, so there is nothing to run rather than a run that was missed.
- **Whether an unauthenticated `/health` should publish the exact source commit.** A deliberate trade, not a defect. This repo is PUBLIC, so the sha is independently resolvable by anyone and the endpoint leaks nothing the repo does not already give away; one-curl triage from outside is the stated goal of #254, and gating it rebuilds the cost the issue exists to remove. A human may ratify or reverse this.

## Design / Testing documents

None exist for initiative 254. It was bootstrapped as a single-repo task by `scripts/init-task.sh`, which writes `team.yaml` and `seams.yaml` but no Design Document or Testing Plan, and no `file` agent ran. There is therefore no Testing Plan to link and none whose Sign-off Criteria could be updated with the evidence above — that evidence lives in this PR body and in the gate receipts under `initiatives/254/gates/`.

`seams.yaml` declares `seams: []` — the claim that nothing in this repo depends on a contract owned elsewhere. That holds here: the change edits this repo's own build plumbing and its own `/health` payload, and reads no other repo's seed data, route shape or payload keys.

Order: Bodhilander


William-Long-II:
## Sonar: 1 new issue — blocking, and the obvious fix is wrong

Quality Gate failed on **1 New Issues (is greater than 0)**. The full list, from the check-run annotations:

> `relay/src/config.ts` **L145** — Prefer using nullish coalescing operator (`??`) instead of a logical or (`||`), as it is a safer operator.

```ts
const commit = env.RELAY_BUILD_COMMIT?.trim() || null;
```

**Do not swap `||` for `??` here.** The two are not interchangeable at this line, and `??` is the less safe operator for this case. An unset `--build-arg` reaches the process as the empty string, so the value this code must collapse to `null` is `''`, which `??` passes straight through. That would make `/health` report `"commit": ""` instead of `null`, and it breaks two assertions already on this branch (`config.test.ts` L63–64):

```ts
expect(loadConfig({ ...BASE, RELAY_BUILD_COMMIT: '' }).config.commit).toBeNull();
expect(loadConfig({ ...BASE, RELAY_BUILD_COMMIT: '   ' }).config.commit).toBeNull();
```

Those are the point of the change — the empty-string case *is* the unstamped production image this branch exists to make visible.

So the finding needs a resolution that keeps the semantics, not a rule-silencing swap. Making the absent case explicit satisfies the rule without touching behaviour:

```ts
const trimmed = env.RELAY_BUILD_COMMIT?.trim();
const commit = trimmed === undefined || trimmed === '' ? null : trimmed;
```

The alternative is a human marking the issue won't-fix in SonarQube with this reasoning attached. Either is fine; deciding between them is the repo owner's call, and this is application code so gate 4 is not making the edit.

Tests are green on both platforms (`test (ubuntu-latest)`, `test (windows-latest)`); GitGuardian passes. Sonar is the only thing red. This PR stays **draft** until it is green.

brannon-bowden (approved):
Approve — this is a well-scoped, thoroughly tested change that adds a build-time commit stamp to `/health` and the startup log, and brings the deploy docs in line with the actual host.

**Blocking issues**

None found.

**Non-blocking issues**

- `relay/README.md` (the "Production deployment" section): the docs disclose that the live host currently publishes the relay on `0.0.0.0:47393` instead of `127.0.0.1`, which — combined with `TRUST_PROXY=true` — lets any direct caller forge the rightmost `X-Forwarded-For` entry and bypass rate limiting, including on link codes. This is accurately flagged as pre-existing and tracked in #313 rather than something this PR should fix, but given it's a live bypassable rate limiter, it's worth confirming #313 is prioritized rather than left open-ended.
- The new commit-stamping behavior is covered thoroughly at the unit level (`config.ts`, `http.ts`, startup log) but there's no end-to-end test that actually runs `docker build --build-arg RELAY_COMMIT=<sha>` and checks the resulting image's `/health` output — understandable given the scope, but worth keeping in mind if the Dockerfile's `ARG`/`ENV` wiring ever regresses silently.

**Tests**

Coverage is solid: `config.test.ts` covers the stamped/unstamped/whitespace/dirty-suffix cases and the `RELAY_COMMIT` vs `RELAY_BUILD_COMMIT` precedence trap; `relay.test.ts` covers the production warning and the `/health` JSON shape (including that `commit` key is present-but-null rather than omitted); `index.test.ts` adds a check that the startup log line carries the commit. No gaps that need to block merge.

PR #315, opened 2026-09-16
feat(run-engine): surface the initiative's merge order

William-Long-II:
Refs #722. Display-only: a cross-repo initiative's `seams.yaml` declares the order its PRs should merge in, and nothing in the app showed it. Now `arm-run` parses `merge_order` (block + flow forms), `ignition` stores each owner's index and returns the ordered repos, and the arm result (RunArm/RunPrepare) shows "Merge order: a → b → c" while the inbox lists a run's repos in that order (unordered last). The engine does not gate on it — a person still merges. Tests: parser (block/flow/absent/stop-at-key), arm threading, inbox ordering. 1904 green, tsc clean both configs, build green.


src/main/repositories/runs.ts
```diff
@@ -409,6 +409,19 @@ export function recordOwnerPullRequest(
...
+/**
+ * Record an owner's place in the initiative's merge order (CO-722).
+ *
+ * Display only: the engine does not gate on it. A person merges the approved
+ * PRs in this order, so the inbox can show them in it. Null when the initiative
+ * declared no order (a single-repo run, or a manifest without one).
+ */
+export function recordOwnerMergeOrder(runId: string, repo: string, order: number | null): void {
+  getDatabase()
+    .prepare('UPDATE run_owners SET merge_order = ? WHERE run_id = ? AND repo = ?')
```
brannon-bowden: No migration is visible in this diff for a `merge_order` column on `run_owners`. Please confirm the column already exists (or add the migration here) — otherwise this UPDATE and the new ORDER BY in listInbox will fail at runtime.

src/main/run-engine/arm-run.ts
```diff
@@ -45,6 +45,33 @@ export function harnessFromTeamYaml(text: string): string | null {
...
+  const lines = text.split(/\r?\n/);
+  for (let i = 0; i < lines.length; i += 1) {
+    const flow = /^merge_order:\s*\[(.*)\]\s*$/.exec(lines[i]);
+    if (flow) {
+      return flow[1].split(',').map((s) => s.trim()).filter(Boolean);
+    }
+    if (/^merge_order:\s*$/.test(lines[i])) {
+      const repos: string[] = [];
+      for (let j = i + 1; j < lines.length; j += 1) {
+        const item = /^\s*-\s*(\S+)\s*$/.exec(lines[j]);
```
brannon-bowden: This regex will retain quotes on `- "repo-a"` style entries and won't handle repo names with spaces. Fine for now since it's display-only, but worth a comment noting the assumption.

brannon-bowden (changes requested):
This PR wires the initiative's declared merge order through arming, storage, and the inbox display in a well-tested, narrowly-scoped way, but CI is currently red and needs to go green before this merges.

## Issues

- **CI is failing**: `test (windows-latest)` is red. Given this PR's tests hard-code POSIX-style paths through `path.join` (e.g. `path.join('C:/init/BDH-239', 'team.yaml')` in `src/main/run-engine/__tests__/arm-run.test.ts`), it's worth checking whether the failure is a Windows path-separator mismatch introduced by these changes before merging. (`postinstall produces a working native module (windows-2022)` is still pending — not gating, but worth watching.)
- `recordOwnerMergeOrder` (`src/main/repositories/runs.ts:421`) updates `run_owners SET merge_order = ...`, and `listInbox` now sorts by that same column. This diff doesn't include a schema migration for `run_owners.merge_order`, so it's worth confirming one already exists (or is included elsewhere) — if the column isn't there, both the write and the inbox query will fail at runtime.
- No test exercises `recordOwnerMergeOrder`'s DB write or the new `listInbox` ORDER BY (nulls-last, then `merge_order`, then `repo`, grouped by `run_id`). The added tests cover `mergeOrderFromSeams` parsing and `armInitiative` wiring well, but the persistence/ordering side of this feature is untested.

## Suggestions (non-blocking)

- `mergeOrderFromSeams`'s block-form item regex (`src/main/run-engine/arm-run.ts:65`, `/^\s*-\s*(\S+)\s*$/`) will keep surrounding quotes on entries like `- "repo-a"` and won't match repo names containing spaces. Given this is display-only and repo names are typically simple slugs, this is low risk, but a short doc note on the assumption (or a slightly more permissive pattern) would help future maintainers.

William-Long-II:
Fixed on `fdce6fc`:

- **The Windows CI failure was not this change** — it's a flaky timing test, `permission-broker > nobody answering is an explicit refusal`: it asserts a ~1s broker wait finishes in `<5000ms`, but `node <broker>` spawn overhead on a loaded Windows runner hit 5114ms. The claim is "answers rather than hanging forever," so I loosened the bound to 15s (a real hang blows the runner timeout, not this assertion). Path separators weren't involved — the arm-run tests use `path.join` on both sides.
- **Added the persistence/ordering test** — `recordOwnerMergeOrder` writes the column and `listInbox` returns a run's repos in merge order, unordered last.
- **Migration**: `run_owners.merge_order` (+ guarded ALTER + backfill) shipped in the schema slice #308 and is on `development`; this PR only reads/writes the existing column.
- Noted the `mergeOrderFromSeams` bare-token assumption in a doc comment.

Suite green, tsc clean.

brannon-bowden (approved):
This round resolves all three previously-raised issues and CI is now green across the board, so this is ready to merge.

- CI is passing on all 7 checks, including both `test (windows-latest)` and `test (ubuntu-latest)`, resolving the prior red-CI blocker.
- `src/main/repositories/__tests__/runs.test.ts` now adds a test that seeds owners out of order (including one with a `null` merge order), calls `recordOwnerMergeOrder`, and asserts `listInbox()` returns repos ordered correctly with unordered entries last — this directly covers the persistence/ordering gap flagged previously. Its passing on both OS runners also gives confidence the `run_owners.merge_order` column exists and is being written/read correctly, addressing the earlier migration concern.
- `arm-run.ts` gained a doc comment explaining the block-form regex's quote/space limitation, addressing the earlier documentation suggestion.

## Suggestions (non-blocking)

- `permission-broker.test.ts:103` widens the hang-detection bound from 5s to 15s to absorb Windows CI spawn overhead. The accompanying comment explains the rationale well; just worth keeping an eye on whether this loosens the test enough to miss a real regression in broker latency, since 15s is a fairly generous ceiling for a ~1s expected wait.

PR #316, opened 2026-09-16
feat(run-engine): drive runs in concurrent lanes (multi-run, CO-722)

William-Long-II:
Refs #722. **Testers will run several initiatives at once — and today they can't.** The loop drove due runs in a **sequential await-each** loop under a global "one tick at a time" guard, and two operations block *inside* `advance()` for minutes — a **print gate (3/4)** and **provision** (both `await` the child to `close`). So one run's long operation froze every other run.

## The fix: per-run concurrent lanes
The loop now drives each due run in its own async lane:
- A **`running` set** tracks runs whose pass is in flight. Each tick starts a lane per due run **not already running**, and awaits only the lanes *it* started — so a sibling's minutes-long gate no longer holds the tick up. A run already in a lane is skipped, not re-entered (that would race its own rows).
- The global **`ticking` guard is gone** — overlapping 15s fires are exactly what let the other runs keep moving while one is blocked; the per-run `running` set is now what prevents driving a run twice at once.
- **`MAX_CONCURRENT_RUNS` (6)** caps how many lanes start at once, so a burst of freshly-armed initiatives doesn't spawn dozens of `claude`/provision children; the rest are due again next tick.

## Why this shape
No harness change, no gate-mode change. A single run's own gate is as slow as ever — it just no longer blocks the others, which is the multi-run requirement. The engine already tolerated this: the driver serializes per run id (`inFlight` keyed by run id — "runs do not block each other") and `advance` re-reads state from the DB, so concurrent advances of *different* runs are safe.

Two alternatives were rejected: a plain `Promise.all` fan-out (insufficient — the awaited tick + `ticking` guard still stalls the next tick behind the slowest run), and converting gates 3/4 to background+receipt (carries an external dependency on the harness writing receipts for those roles under `--bg`, and reworks `continueSequence` — too risky for beta, and not needed for concurrency).

## Verified
- **The regression**: run A blocked in its pass (a never-resolving gate) while run B is still driven in the same tick's lanes.
- The per-run guard: a run in flight is not re-driven by the next tick.
- The cap: only `MAX_CONCURRENT_RUNS` lanes start in one tick.
- Existing single-run tests unchanged. **1431 green**, `tsc` clean on both configs, full `bun run build` green, app launches clean.

## Out of scope (noted)
Making a *single* run's own gates non-blocking (background gates 3/4 + provision) is the larger, harness-dependent change — not needed for concurrent multi-run. **In-app cross-repo bootstrap** (driving the `arch` gate) is the other large item, separate.


src/main/run-engine/run-loop.ts
```diff
@@ -339,29 +351,48 @@ export function createRunLoop(deps: LoopDeps): RunLoop {
...
+      const lane = (async () => {
+        let ok = false;
+        try {
+          ok = await passOne(run, report);
+        } catch (err) {
+          // One run's failure is not another's, and not the loop's. Reported,
+          // counted against this run's cadence, and the other lanes go on.
+          report.problems.push({ runId: run.id, problem: err instanceof Error ? err.message : String(err) });
+        }
+        passed(run.id, ok, report);
```
brannon-bowden: passed(run.id, ok, report) is called outside the try/catch around passOne. Its promise chain is only ever awaited via Promise.allSettled with the result discarded, so a throw here is now silently swallowed instead of surfacing via the tick-level .catch log as it did before this change.

brannon-bowden (approved):
Approve — the move from a single global tick guard to per-run "lanes" is a sound fix for the freeze bug (one run's slow gate blocking every other run), and the cap plus the two new tests do a good job of exercising both the fix and the concurrency limit. All CI checks are passing. One error-handling regression worth a follow-up, plus a couple of minor notes.

## Issues

- **Major** — `src/main/run-engine/run-loop.ts:378`: `passed(run.id, ok, report)` runs outside the `try/catch` that wraps `passOne`, same as before this change. Previously, an uncaught throw there propagated out of `tick()` and was caught/logged by the `.catch` in `start()` ("the run loop tick threw…"). Now each run's work happens inside a lane promise that's only ever consumed via `Promise.allSettled(startedThisTick)` with the result discarded (`run-loop.ts:384`), so a throw from `passed()` is silently swallowed — no `report.problems` entry, no log line, nothing. Worth bringing `passed(...)` inside the same try/catch as `passOne`, or otherwise inspecting the `allSettled` results, so this failure mode doesn't go dark.
- **Minor** — `src/main/run-engine/run-loop.ts:366`: the `running`/`MAX_CONCURRENT_RUNS` cap is global, so if `MAX_CONCURRENT_RUNS` runs are simultaneously stuck in a slow gate, every other due run is starved until one of those frees a slot — a smaller-scale version of the exact freeze this PR fixes, just at cardinality 6 instead of 1. The doc comment above `MAX_CONCURRENT_RUNS` frames overflow as "wait for next tick," which reads as a burst-throttle rather than a starvation risk under sustained blocking — worth confirming that's the intended tradeoff.
- **Nit** — `src/main/run-engine/__tests__/run-loop.test.ts:458`: in the new "a run blocked in its pass does not stop another run…" test, `releaseA?.()` is called after the assertions but the test never awaits run A's lane (or the underlying `loop.tick()`) actually finishing, so that async work resolves after the test has already returned.

## Suggestions (non-blocking)

- The concurrency-safety reasoning (synchronous scan-and-dispatch per tick, `running` set shared across overlapping ticks) is solid and well documented in the comments — no changes needed there.

PR #319, opened 2026-09-16
feat: schema + model for in-app cross-repo bootstrap (CO-722)

William-Long-II:
Refs #722. **Slice 1 of 5** for driving gates 0/1 in-app, so a tester can kick off a **cross-repo** initiative entirely in the app instead of bootstrapping it in the harness terminal first. Inert on its own — it lands the run-level schema and model helpers; nothing reads them yet.

## Why a new shape at all
A cross-repo run can't be armed the way a single-repo one is: its owners don't exist until the `arch` agent has authored the seam manifest and `spawn.py` has cut a worktree per repo. So a `kind: 'multi'` run is **created with no owners** and driven through a pre-owner `bootstrap_state` (`scoping → architecting → awaitingManifest → spawning → done`) **outside** the pure per-owner machine in `transitions.ts`, which stays untouched. After spawn the run is byte-for-byte what `armRun` produces (`state: 'preparing'`, owner rows with a null gate state) and the existing provision → gates 2/3/4 path takes over.

## What's in this slice
- **Schema**: `runs` gains `kind` (`NOT NULL DEFAULT 'single'`), `bootstrap_state`, `scope_repos` — in `RUN_TABLES_SQL` (fresh installs) and as guarded, idempotent `addColumn` migrations that backfill every existing run as `single`. No `run_owners`/`run_gates` changes (the arch gate reuses `gate=1, repo='(scope)'` in the existing shape, a later slice).
- **Model**: `RunRow`/`RawRun`/`toRun`/`CreateRunInput`/`createRun` carry the fields; a malformed `scope_repos` reads as unknown rather than crashing the run view.
- **Helpers**: `setBootstrapState` / `setRunState` — thin run-level writes the bootstrap driver will use, so it need not borrow `recordTransition`'s blocked-state overloads.
- **`bootstrap.ts`** (new): `SCOPE_REPO` sentinel + `RunKind`/`BootstrapState` types, kept out of `transitions.ts` so the pure machine never learns of the bootstrap.

## Tests
- `run-schema.test.ts`: fresh schema has the columns; a fresh run is `single`; the migration adds the columns on a pre-bootstrap DB and backfills every in-flight run as `single`.
- `runs.test.ts`: single-by-default; a multi run round-trips `kind`/`bootstrapState`/`scopeRepos`; `setBootstrapState` advances and clears at handoff; `setRunState` sets + clears the reason; malformed `scope_repos` → null.
- 991 pass / 0 fail across `repositories` + `run-engine` + `main`; `tsc` clean on both configs; `bun run build` green.


brannon-bowden (approved):
This PR is clean, well-scoped scaffolding — schema migration plus repository plumbing for the CO-722 cross-repo bootstrap — with good test coverage for both the fresh-schema and backfill paths, and I don't see anything that blocks merge. Approving.

## Issues
None blocking. All 7 CI checks are passing.

## Suggestions (non-blocking)
- `src/main/repositories/runs.ts:100` (`parseScopeRepos`) — good defensive handling of malformed JSON, but the "valid JSON, wrong shape" branch (e.g. `'{"a":1}'` or `'[1,2]'`) isn't exercised by a test; only the truly malformed `'{not json'` case is covered. Worth a quick additional case since this function is the guard against a corrupt column taking down `getRun`.
- `src/main/repositories/runs.ts:190` (`setRunState`) — this writes `state`/`blocked_reason` directly, bypassing `recordTransition`/`rollupState` used by the per-owner machine. The docstring explains this is intentional for bootstrap-driven pre-owner moves, which is reasonable given owners don't exist yet, but it's worth confirming with the author that nothing downstream assumes all state writes go through the validated transition path — not something visible in this diff.
- Neither `kind` nor `bootstrap_state` has a `CHECK` constraint in the schema (`src/main/run-tables-sql.ts:30-36`), so enforcement of the `RunKind`/`BootstrapState` enums is TypeScript-only; a raw SQL write (or a future migration bug) could insert an out-of-band value that `toRun` would then blindly cast. Low risk given the existing codebase style for similar columns, but worth a thought if this table sees direct SQL elsewhere.

PR #320, opened 2026-09-16
feat: mechanical scope + cross-repo run creation, loop-driven (CO-722)

William-Long-II:
Refs #722. **Slice 2 of 5** for driving the cross-repo bootstrap in-app. **Stacked on #319** (schema+model) — GitHub retargets this to `development` when #319 merges; review here sees only Slice 2's diff.

A tester can now start a cross-repo initiative from the app: pick the repos, name the issue, and a `kind: 'multi'` run is created that the **loop** bootstraps — durably, so a closed app resumes mid-flight. This slice lands **scope (gate 0)** and the routing; the `arch` gate is stubbed, so a multi run scopes then holds at `architecting` (Slice 3 fills it).

## Gate 0 is mechanical (no scribe LLM)
The tester *picks* the repos, so scope needs no LLM. `scope-initiative.ts` writes an `owners.yaml` from the picks (a brief per repo, none of the four keys `spawn.py` stamps) and hands it to the harness's own `file_scope.py`, which authors `team.yaml`. Idempotent on resume: an existing `team.yaml` reads as already-scoped rather than tripping file_scope's overwrite refusal.

## Loop-driven & durable
`prepareCrossRepoRun` runs nothing eagerly — it validates config + picks and writes **one** run row in `bootstrap_state: 'scoping'`, returning the id. The run shows in the Runs view at once; the loop does the rest.

- **`cross-repo-prepare.ts`** — `planCrossRepoRun` (pure, DB-free) collects every fixable config gap, refuses a bad issue name / empty pick, derives the tracking key the harness way, and returns the exact multi `CreateRunInput`.
- **`bootstrap-driver.ts`** — `driveBootstrap` runs one durable step per pass and advances `bootstrap_state`, entirely **outside** `transitions.ts`. `scoping` → `architecting`; a scope failure parks the run **inconclusive**; later states are benign holds.
- **`run-loop.ts`** — `passOne` routes a multi run with a live `bootstrap_state` to `driveBootstrap` **before** the no-owner skip and the provision step, so a run with no owners never advances `prepared` on a non-existent owner. Single runs are byte-identical to before.
- **`RunPrepare.tsx`** — a single/cross-repo toggle; cross-repo collects a repo multi-select, creates the run **without arming**, and points the tester at the Runs view.
- IPC `db:runs:prepareCrossRepo` + preload + `RunCrossRepoPrepareResult`.

## Tests
Planning refusals + the produced row; `owners.yaml` content + `file_scope` argv + resume + failure→reason; sub-driver scoping advance / inconclusive / holds; `passOne` routes multi → sub-driver and **never** provisions (plus single unaffected, `done` falls through); RunPrepare cross-repo mode (picks → create, no arm; button gating; refusal shown). **1332 pass**, `tsc` clean on both configs, `bun run build` green.


brannon-bowden (approved):
Approve. This is a well-scoped, well-documented slice of the cross-repo bootstrap feature — the pure/impure split (`cross-repo-prepare.ts`, `scope-initiative.ts`, `bootstrap-driver.ts` vs. the IO-injecting wiring in `run-loop-service.ts`) is clean, the "collect every refusal before failing" pattern in `planCrossRepoRun` is good UX, and the new logic (plan validation, scope resume-on-crash, loop routing for `kind: 'multi'` runs) is backed by solid unit tests, including edge cases like duplicate repo picks, whitespace-only picks, and a `bootstrapState: 'done'` run falling through to the owner path.

I don't see anything that must change before merge.

**Issues**
- `src/main/run-engine/bootstrap-driver.ts:77` — On a scope refusal, only `scoped.reason` (the first line of stderr/stdout) is passed to `recordInconclusive`, and `scoped.log` (the full joined stdout+stderr, built at `src/main/run-engine/scope-initiative.ts:77`) is discarded entirely — it isn't logged via `deps.log` either. For a `file_scope.py` failure with a multi-line traceback, only the first line survives anywhere, which will make real failures harder to diagnose in production. Worth threading `scoped.log` into `deps.log(...)` at minimum. (minor)

- `src/main/run-engine/scope-initiative.ts:92-97` — `ownersYaml` interpolates repo names directly as unquoted YAML mapping keys. The comment notes repo names are "registry slugs validated by the picker and re-checked by `file_scope`," which is true for the intended UI flow, but `prepareCrossRepoRun`'s IPC handler (`src/main/index.ts`) accepts `repos: string[]` from the renderer with no server-side allow-list check before that string reaches `ownersYaml`. A repo value containing `:` or a newline would produce malformed/injected YAML written to disk before `file_scope.py` ever runs its registry check. Low risk given the trust boundary and downstream validation, but an explicit charset check alongside the existing `ISSUE_ID` regex in `cross-repo-prepare.ts` would close the gap cheaply. (nit)
- Not verifiable from this diff, but worth confirming with the author: does the run loop guard against a second tick calling `driveBootstrap` for the same run while a prior pass's `file_scope.py` invocation (up to the 15-minute timeout wired in `run-loop-service.ts`) is still in flight? If ticks can overlap per-run, `scopeInitiative` could be re-entered mid-write. This is a question about loop-scheduling code not shown in the diff, not a defect in what's shown here.

brannon-bowden (approved):
This is a well-structured, thoroughly-tested addition (pure planning/validation logic separated from IO, extensive unit coverage for the new bootstrap sub-driver, arch gate, and scope step). CI is fully green. Nothing here needs to change before merge; a couple of small items are worth a look.

## Issues

- **Minor** `src/main/run-engine/cross-repo-prepare.ts:56` — `issueId` is validated against a strict `ISSUE_ID` regex before being used, but the picked `repos` are only trimmed/deduped/filtered for emptiness — no format check. The UI currently only offers checkboxes populated from `listHarnessRepos()`, but the `db:runs:prepareCrossRepo` IPC handler and `planCrossRepoRun` accept any strings for `repos`, which then flow verbatim into a hand-built YAML string in `ownersYaml()` (`scope-initiative.ts:97`). A repo string containing a colon/newline could corrupt the generated `owners.yaml` structure. The comment notes `file_scope.py` "re-checks" repo names against the registry, which likely catches this downstream, but validating the shape at the same trust boundary as `issueId` (main process) would be more consistent and avoid relying solely on the harness script to reject malformed input.

## Suggestions (non-blocking)

- Worth confirming (not blocking, since the relevant scheduling code isn't in this diff): `runArchGate` awaits `deps.launch(...)`, which can run for up to `gateTimeoutMs`. If the run loop's "due" selection can re-select the same run for another tick while a prior `driveBootstrap` call for it is still in flight, a second `startGate`/launch could fire concurrently for the same `(runId, SCOPE_REPO, gate=1)`. If the existing owner-gate path already guards against this (e.g. via gate/run state), the same guard presumably covers this new path too — just flagging since the guard isn't visible here.
- `src/renderer/components/RunPrepare.tsx:31` (approx, the `Phase` type) — `{ kind: 'working'; step: ... | 'bootstrapping' }` and the separate `{ kind: 'bootstrapping'; runId: string }` phase share the name "bootstrapping" for two different states (in-flight vs. completed-and-watching). Slightly confusing for future readers; a distinct name for one of them (e.g. `step: 'starting'`) would read more clearly.
- `src/renderer/components/RunPrepare.tsx` mode toggle — uses `role="radiogroup"` on the container with `aria-pressed` toggle buttons inside. `radiogroup` expects `radio`-role children with `aria-checked`; the current combination is a mismatched ARIA pattern. Cosmetic only.
- The cross-repo form doesn't reset `issueId`/selected repos after a successful "Start bootstrap" (mirrors existing single-repo behavior), so a stray double-click could create two `multi` runs for the same issue. Not a regression from this PR, just worth considering while in the area.

PR #321, opened 2026-09-16
feat: drive the arch gate + manifest park for cross-repo runs (CO-722)

William-Long-II:
Refs #722. **Slice 3 of 5** — the one part of the cross-repo bootstrap that needs an LLM. **Stacked on #320**; GitHub retargets to `development` as the stack merges. Review here sees only Slice 3's diff.

The seam manifest is a set of producer/consumer contracts read off both sides' **live code** — a person can't pick it the way they pick scope. So a multi run in `architecting` now drives `arch` as a real print gate and parks for approval; a scope/arch failure parks it **inconclusive**.

## `bootstrap-arch.ts` — `runArchGate`
Launches `arch` through the **same** machinery every reading gate uses (`launchGate`, the permission broker, the verdict schema), with two differences from an owner gate:
- **no worktree yet** → it runs in the workspace root where the repo clones live (`cwd = bodhiRoot`);
- **no owner** → its `run_gates` row and permission channel are keyed on the **scope sentinel** (`(scope)`), so the channel key is byte-identical to what `channelDirForGate` derives.

The verdict is the **file + the harness verifier**, not the model's word: a pass with no `seams.yaml`, or one `verify_seams.py` rejects, is inconclusive — never a park. Idempotent on resume (an existing `seams.yaml` re-verifies rather than re-running the gate). arch is gate 1, which `Gate = 2\|3\|4` intentionally excludes; `launchGate` uses the number only as a label, so **one documented bridge cast** is the whole cost of reuse — `transitions.ts` stays untouched.

## Driver + service
- **`bootstrap-driver.ts`** — `architecting` drives arch, then advances to `awaitingManifest` + `runs.state 'waitingHumanGate'` (parks for a person; the loop stops polling until approval), or records inconclusive.
- **`run-loop-service.ts`** — `archDeps` wires the launcher/broker/verifier; **`listRunPermissions` is bootstrap-aware**: while a run is `architecting` it reads the arch channel on the scope sentinel — the only path a person has to a print gate that's blocking its own lane. `answerRunPermission` already routes a sentinel reply to that channel and drives no owner (arch isn't per-owner-driven).

`publish_manifest` is intentionally **not** run: it posts the manifest to the tracking issue for the harness's own `gh search` tooling; the app reads `seams.yaml` from disk, so the driven flow doesn't need it.

## Tests
arch happy path (print mode, `cwd=workspace`, sentinel channel, verify runs) + undriveable / fail / pass-without-file / verify-fail all → inconclusive + resume; driver `architecting` park vs inconclusive; service surfaces **and** answers the scope-sentinel permission. **646 run-engine pass**, `tsc` clean, `bun run build` green.

## Still ahead
Slice 4 (manifest-approval IPC + `materializeOwners` + `spawn.py` + handoff) and Slice 5 (polish), then a live 2-repo shakeout before the next beta.


brannon-bowden (approved):
Approve. This is a well-scoped, well-tested slice: `runArchGate` cleanly separates "model said pass" from "the file exists and the harness verifier accepts it," the resume path is idempotent, and the new `bootstrap-driver`/`run-loop-service` wiring is exercised by tests covering the happy path, undriveable/fail/no-manifest/rejected-manifest outcomes, and the no-owner permission-channel plumbing. All CI checks are green.

## Issues

- **Retry behavior on inconclusive arch is unclear** (`src/main/run-engine/bootstrap-driver.ts:93`, confirmed by the test at `bootstrap-driver.test.ts` showing `rec.bootstrap` and `rec.runState` stay empty on `inconclusive`): when arch is inconclusive, `bootstrapState`/`runState` are left unchanged, so the run remains eligible for the next `driveBootstrap` poll. Since `runArchGate` only skips re-launching when `seams.yaml` is already on disk, any inconclusive outcome that didn't produce a file (undriveable, fail verdict, pass-without-file) will cause the *next* poll to spawn a brand-new `arch` LLM session from scratch, with no backoff or attempt cap visible in this diff. Given that gate launches carry real cost (`costUsd` on `GateOutcome`), it's worth confirming there's throttling/backoff elsewhere in the loop scheduler that isn't shown here — if not, a failing arch run could retry unbounded on every tick.
- **Missing test coverage for two defensive branches** in `bootstrap-arch.ts`: the "gate row could not be opened" path (`bootstrap-arch.ts:90`) and the "unexpected background launch" path (`bootstrap-arch.ts:123-126`) are both reachable code with their own return values but aren't covered by `bootstrap-arch.test.ts`, unlike every other branch in the file.

## Suggestions (non-blocking)

- `archDeps.run` in `run-loop-service.ts` hardcodes a 15-minute timeout for `verify_seams.py` independent of `config.gateTimeoutMs`. That's probably fine (verification should be much faster than the LLM gate itself), but a one-line comment noting why it's decoupled from the configured gate timeout would save a future reader from wondering if it's an oversight.

PR #322, opened 2026-09-16
feat: manifest approval + spawn + handoff for cross-repo runs (CO-722)

William-Long-II:
Refs #722. **Slice 4 of 5** — the piece that makes the in-app cross-repo bootstrap **end-to-end drivable**. **Stacked on #321**; retargets to `development` as the stack merges.

A person approves the seam manifest in the Runs inbox; the loop then cuts the worktrees, writes the owner rows, and **hands off** to the per-owner machine, which provisions and drives gates 2/3/4 exactly as it does for a harness-bootstrapped initiative.

## The handoff (why the pure machine stays untouched)
- **`ignition.ts` — `materializeOwners`**: `armRun`'s post-check body **minus `createRun`** (the run already exists). Reads the now-populated `team.yaml` via `initiative.py`, resolves each repo's owner exactly as arming does, mirrors the owners block. A `>1-candidate` repo is the **same refusal** arming gives. Shares `resolveOwners`/`upsertOwner`/`recordOwnerMergeOrder` with arming so the two can't drift (`resolveOwners`'s param narrowed so both callers fit; `armRun` body unchanged).
- **`bootstrap-spawn.ts` — `runSpawn`**: runs `spawn.py` with `BODHI_ROOT` (as single-repo prepare does), reads the merge order from `seams.yaml`, materializes owners. Spawn failure / `>1-candidate` → one-line reason.
- **`bootstrap-driver.ts` — `spawning`**: runs `runSpawn`, then the handoff — clears `bootstrap_state` to null and drops to `preparing`, so the next pass sees owners + preparing and **the existing per-owner machine takes over**. A spawn refusal parks inconclusive.

## Approval
- **`run-loop-service.ts`** — `readRunManifest` / `approveRunManifest` / `rejectRunManifest`. Approve records `humanApprovedGate` (audit), then flips to `spawning` + `preparing` **without** going through `transition()` — whose `waitingHumanGate` exit spawns gate 2, which has no owner pre-spawn — guarded on `awaitingManifest` (a double-click is a no-op). Reject parks inconclusive.
- **`RunInbox.tsx`** — a `ManifestApproval` panel on a `waitingHumanGate` row: merge order + the **whole** `seams.yaml` (approving is approving what each repo builds to) + Approve/Reject. IPC `db:runs:bootstrap:{manifest,approve,reject}`.

## Tests
`materializeOwners` (writes owners + order / `>1-candidate` refusal / unreadable); `runSpawn` (BODHI_ROOT + merge order threaded / spawn fail / refusal / absent seams); driver `spawning` handoff vs inconclusive; service approve (spawning+preparing+event, guarded) / reject (inconclusive+reason); `ManifestApproval` panel. **1357 pass**, `tsc` clean on both configs, `bun run build` green.

## After this
The cross-repo bootstrap is functionally complete (scope → arch → approve → spawn → drive). Remaining: **Slice 5** (Runs-view bootstrap labels + out-of-scope UI, cosmetic) and a **live 2-repo shakeout** before the next beta — during arch, confirm a tool prompt surfaces in the Runs view and can be answered while the print gate blocks its lane, and that a mid-arch restart re-launches cleanly.


brannon-bowden (approved):
**Approving this PR.** The manifest approval → spawn → handoff flow is implemented cleanly, the state-machine changes to `bootstrap-driver.ts` are well-contained, and the new logic (`bootstrap-spawn.ts`, `materializeOwners`, the approve/reject IPC surface, and the `ManifestApproval` UI) is backed by solid unit and component tests covering both the happy path and refusal/edge cases (blank reject reason, missing `seams.yaml`, multi-candidate owner refusal). CI is fully green across all 7 checks.

## Suggestions (non-blocking)

- `src/main/run-engine/ignition.ts:255` (`materializeOwners`) builds the `initiative.py` path with template-literal concatenation (`${request.harnessPath}/scripts/lib/initiative.py`) instead of `path.join`, unlike the sibling `bootstrap-spawn.ts:33` which uses `path.join(run.harnessPath, 'scripts', 'lib', 'spawn.py')`. Worth aligning for consistency and to avoid odd separators if `harnessPath` ever has a trailing slash/backslash.
- `src/renderer/components/RunInbox.tsx` (`ManifestApproval`'s `useEffect`) has no cancellation/unmount guard around the `loadManifest(runId).then(setManifest)` call, so a slow response after the row leaves the DOM will attempt to set state on an unmounted component. Low real-world impact given `row.id` is stable across polls, but a common source of console warnings.
- `decide()` in the same component doesn't check the boolean returned by `approve`/`reject` before calling `onDecided()` — if a call is guarded off server-side (e.g., a lost race where `bootstrapState` already moved on), the UI silently treats it as success with no error surfaced to the user.
- Worth confirming (not necessarily in this PR) that `spawn.py` is safe to re-invoke if `spawnStep` were retried after a crash between the script succeeding and `setBootstrapState(run.id, null)` committing — the diff doesn't show `spawn.py`'s idempotency, so this is a question for the author rather than a defect.

PR #323, opened 2026-09-16
feat: manifest approval + spawn + handoff for cross-repo runs (CO-722)

William-Long-II:
Refs #722. **Slice 4 (final code slice)** of the in-app cross-repo bootstrap — the piece that makes it **end-to-end drivable**. Targets `development` directly (the earlier stacked PR #322 auto-merged into a since-deleted feature branch; this re-lands the identical, reviewed content cleanly on top of slices 1–3, now in `development`).

A person approves the seam manifest in the Runs inbox; the loop then cuts the worktrees, writes the owner rows, and **hands off** to the per-owner machine, which provisions and drives gates 2/3/4 exactly as for a harness-bootstrapped initiative.

- **`ignition.ts` — `materializeOwners`**: `armRun`'s post-check body minus `createRun` (the run already exists). Reads the now-populated `team.yaml`, resolves owners as arming does, mirrors the block. A `>1-candidate` repo is the same refusal arming gives. Shares `resolveOwners`/`upsertOwner`/`recordOwnerMergeOrder` with arming.
- **`bootstrap-spawn.ts` — `runSpawn`**: `spawn.py` with `BODHI_ROOT`, reads merge order from `seams.yaml`, materializes owners; failures → one-line reason.
- **`bootstrap-driver.ts` — `spawning`**: runs `runSpawn`, then the handoff — clears `bootstrap_state` and drops to `preparing`, so the per-owner machine takes over. A refusal parks inconclusive.
- **`run-loop-service.ts`**: `readRunManifest`/`approveRunManifest`/`rejectRunManifest`. Approve records `humanApprovedGate`, flips to `spawning`+`preparing` (guarded on `awaitingManifest`), **without** `transition()`. Reject parks inconclusive.
- **`RunInbox.tsx`**: a `ManifestApproval` panel on `waitingHumanGate` — merge order + the whole `seams.yaml` + Approve/Reject.

## Tests
`materializeOwners`, `runSpawn`, driver `spawning` handoff/inconclusive, service approve/reject, `ManifestApproval`. tsc clean on both configs, `bun run build` green, 78 targeted tests pass (full suite green pre-recovery at 1357).

## Note
This completes the feature (scope → arch → approve → spawn → drive). Next: **Slice 5** (cosmetic Runs-view labels) and a **live 2-repo shakeout** before beta.N+1.


brannon-bowden (approved):
This PR adds the manifest-approval and spawn/handoff steps for cross-repo runs (CO-722), and the changes are coherent, well-scoped, and backed by solid test coverage across the state machine, spawn logic, and the new UI panel — I'm approving it.

## Suggestions (non-blocking)

- `ManifestApproval`'s `useEffect` (`src/renderer/components/RunInbox.tsx:214-217`) fires `load(runId).then(setManifest)` with no cancellation/mount guard. If the component ever re-renders with a changed `runId` (rather than remounting), a stale response could overwrite newer state, and calling `setManifest` after unmount will warn. Likely harmless today if each row is keyed by `row.id` and thus remounts, but worth a quick look.
- `decide()` (`src/renderer/components/RunInbox.tsx:198-212`) has no `catch` around the `approve`/`reject` IPC calls — if either rejects, `busy` still resets via `finally`, but nothing surfaces to the user that the action failed. Consider showing an inline error on failure.
- `readRunManifest` (`src/main/run-engine/run-loop-service.ts`) is new non-trivial-ish logic (missing run, missing seams.yaml, present manifest) but doesn't have a direct unit test in `run-loop-service.test.ts` the way `approveRunManifest`/`rejectRunManifest` do — worth adding for parity.

## Question

- `bootstrap-spawn.ts` and `run-loop-service.ts` both import `mergeOrderFromSeams` from `./arm-run`, which isn't part of this diff. Can you confirm that export already exists there (vs. needing to be added), since I can't see `arm-run.ts` in this diff to verify it?

PR #324, opened 2026-09-16
feat: an active-runs list so a bootstrapping run is visible (CO-722)

William-Long-II:
Refs #722. **Slice 5** — the last planned piece of the in-app cross-repo bootstrap. Targets `development` directly.

## The gap this closes
A cross-repo run spends **minutes** in `state='preparing'` while it scopes and drives `arch`. The inbox only shows runs that need a person, so `preparing` is excluded — meaning until now the run was **invisible in the app exactly while the most was happening** ("I armed it… where did it go?"). This adds the *"what is the engine doing"* surface beside the *"what needs me"* inbox.

## What's in it
- **`repositories/runs.ts`** — `listActive()`: every in-flight run (state ∉ approved/done/failed) with `kind` + `bootstrap_state` + `repos`, newest-active first. Extracted `reposByRun` and pointed `listInbox` at it so both lists answer "which repos, in what order" identically.
- IPC `db:runs:active` + preload `getActiveRuns` + `electron.d.ts`; `RunActiveRow` in `shared/types` (loose strings like `RunInboxRow`, so the renderer needn't import the main-side unions).
- **`RunList.tsx`** — renders in-flight runs with a phase label: a multi run reads by its bootstrap phase (**"Scoping…", "Architecting seams…", "Awaiting manifest approval", "Spawning worktrees…"**), everything else by its state. Filters out the states the inbox owns (`waitingPermission`/`waitingHumanGate`/`inconclusive`) so a run shows in one place, not both; renders nothing when nothing's in flight (the inbox carries the empty message). Mounted in the Runs view between arm and the inbox.

## Tests
`listActive` (kind/phase, exclusion of finished runs, repos in merge order); `phaseFor` mapping; `RunList` renders the phase, hides inbox-owned rows, stays empty before load. **710 pass**, `tsc` clean on both configs, `bun run build` green.

This completes the feature set (scope → arch → approve → spawn → drive, now visible end to end). Next up is Windows code signing so betas actually install, then a live 2-repo shakeout.


brannon-bowden (approved):
Approve — this is a clean, well-tested addition. The `reposByRun` extraction is a sensible refactor that keeps the inbox and the new active list answering "which repos, in what order" identically, the new IPC channel follows the existing read-only pattern, and the SQL is parameterized correctly. Repository tests cover kind/bootstrap defaults, exclusion of finished runs, and repo merge order; component tests cover phase labeling and the inbox/active-list split. All CI checks are green.

- `src/renderer/components/RunList.tsx:31` — `STATE_LABEL.approved` is dead: `listActive()`'s SQL (`src/main/repositories/runs.ts`) explicitly excludes `state = 'approved'`, so this branch can never be reached. Harmless, but worth pruning or leaving a note if `approved` is meant to appear here later.
- Repository tests don't cover the `'approved'`/`'failed'` exclusion or the `ORDER BY updated_at DESC` sort — only `'done'` exclusion is exercised (`src/main/repositories/__tests__/runs.test.ts:166`). Not a blocker given the existing coverage, but worth adding since the sort order is part of the documented "most-recently-active first" behavior.
- `RunList` polls independently of `RunInbox` (`src/renderer/components/RunList.tsx:59`), so both components issue their own IPC/DB round trip every 5s. Fine at current scale, but if a shared poller is ever introduced for the inbox, this would be a natural place to consolidate.

PR #325, opened 2026-09-16
ci: sign the Windows build with Azure Trusted Signing (optional, gated)

William-Long-II:
Fixes the Smart App Control block you hit installing beta.15. Unsigned Windows builds are **hard-blocked** by Smart App Control (no "run anyway" button), so a beta can't be installed on a fresh Windows 11 machine.

Wires **Azure Trusted Signing** (your chosen path): cloud-based (no hardware token), CI-native, and Microsoft-run so it carries good SAC/SmartScreen standing.

## Safe by construction
Signing is **gated** — it never fails a release:
- `release.yml`'s Windows step signs **only when** the Azure credentials + identifiers are configured; otherwise it logs a warning and builds **UNSIGNED**, exactly as today.
- `electron-builder.yml` declares **no** signing — it's injected at release time via `-c.win.azureSignOptions.*` overrides + the standard `AZURE_*` env — so a local `bun run pack` / dev build needs no Azure setup.
- Merging this cuts **no release**: `check-version` gates on a `-beta.` version whose tag doesn't exist yet, and `v3.5.1-beta.15` is already tagged.

## What you set up (in `docs/windows-signing.md`)
The parts that need your accounts/credentials:
- **Secrets:** `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET` (a service principal with the *Trusted Signing Certificate Profile Signer* role).
- **Variables:** `AZURE_CODE_SIGNING_ENDPOINT`, `AZURE_CODE_SIGNING_ACCOUNT`, `AZURE_CODE_SIGNING_PROFILE`, `AZURE_PUBLISHER_NAME`.
- Prereq: a Trusted Signing account + **org identity verification** (auto if your domain is 3+ yrs old, else a few days of doc review) + a Public Trust certificate profile.

The doc has the click-path and a **dry-run** recipe (Actions → Run workflow → `dry_run: true`) that publishes a draft release so you can download the `.exe`, check its Digital Signatures tab, and confirm SAC no longer blocks it.

## Files
- `.github/workflows/release.yml` — gated signing step.
- `electron-builder.yml` — note explaining signing is CI-injected.
- `docs/windows-signing.md` — full setup + verification guide.


.github/workflows/release.yml
```diff
@@ -150,9 +150,40 @@ jobs:
...
+          AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+          AZ_ENDPOINT: ${{ vars.AZURE_CODE_SIGNING_ENDPOINT }}
+          AZ_ACCOUNT: ${{ vars.AZURE_CODE_SIGNING_ACCOUNT }}
+          AZ_PROFILE: ${{ vars.AZURE_CODE_SIGNING_PROFILE }}
+          AZ_PUBLISHER: ${{ vars.AZURE_PUBLISHER_NAME }}
+        run: |
+          # `bun run build` already ran in the "Build application" step, so this
+          # invokes electron-builder directly (as the mac step does).
+          if [[ -n "$AZURE_CLIENT_ID" && -n "$AZ_ACCOUNT" && -n "$AZ_PROFILE" && -n "$AZ_ENDPOINT" ]]; then
```
brannon-bowden: Gate omits AZURE_TENANT_ID and AZURE_CLIENT_SECRET. If either is unset/mistyped while AZURE_CLIENT_ID + the three variables are present, this takes the signing branch and electron-builder will fail Azure auth, breaking the release instead of falling back to unsigned. Add both to the condition.

brannon-bowden (changes requested):
Request changes: the design (opt-in signing, unsigned fallback, secrets vs. non-secret variables) is sound and the documentation is thorough, but the gate that decides "is signing configured" doesn't check all the credentials electron-builder actually needs, which can turn a partial/typo'd secret configuration into a hard release failure — the exact outcome this PR says it's designed to avoid.

## Blocking issues

- **`.github/workflows/release.yml:171`** — The gate `if [[ -n "$AZURE_CLIENT_ID" && -n "$AZ_ACCOUNT" && -n "$AZ_PROFILE" && -n "$AZ_ENDPOINT" ]]` never checks `AZURE_TENANT_ID` or `AZURE_CLIENT_SECRET`. Both are read by electron-builder's Azure Trusted Signing support per the added `env:` block and per `docs/windows-signing.md`'s own description of how signing works. If a repo has `AZURE_CLIENT_ID` plus the three variables set but `AZURE_TENANT_ID` or `AZURE_CLIENT_SECRET` is missing or mistyped (an easy mistake — GitHub silently resolves an unset/mistyped secret name to an empty string rather than erroring), the gate passes, the signed branch is taken, and electron-builder will fail authenticating to Azure — failing the whole Windows build/release. That directly contradicts the stated goal in the added comment ("this step never fails a release for a repo that has not set signing up yet"). Fix: add `-n "$AZURE_TENANT_ID" && -n "$AZURE_CLIENT_SECRET"` to the condition (or check all five env vars explicitly).

## Non-blocking issues

- **`.github/workflows/release.yml:153-185`** — The previous step ran `bun run dist:win`; this now calls `bunx electron-builder --win` directly, on the stated assumption that `dist:win` was equivalent to "build + electron-builder --win" and that the build half is now covered by the earlier "Build application" step. That script isn't in the diff, so it's worth the author double-checking `dist:win` didn't carry extra flags (e.g. an explicit `--publish` mode, a custom config path) that are now silently dropped — not something that's verifiable from this diff alone.
- Consider logging which specific credential(s) are missing in the unsigned-fallback warning (or once the gate above is fixed, in a secondary check) rather than a generic "not configured" message — makes it faster to diagnose a partially-configured repo from the Actions log alone.

## Tests

No automated coverage for the branch logic itself, which is reasonable for a shell conditional in a workflow file. The doc's manual `dry_run` verification path is a reasonable substitute, but given the blocking issue above, it's worth explicitly dry-running with an intentionally partial credential set (e.g. `AZURE_CLIENT_ID` set, `AZURE_TENANT_ID` unset) once the gate is fixed, to confirm the fallback still triggers correctly in that case.

William-Long-II:
Thanks — good catch on the gate. Fixed in e69ea06: the gate now requires **all six** values signing needs (the 3 credentials `AZURE_TENANT_ID`/`AZURE_CLIENT_ID`/`AZURE_CLIENT_SECRET` + the 3 identifiers endpoint/account/profile; `publisherName` stays optional). A partial/typo'd config now falls back to UNSIGNED, and the warning names the missing value(s).

On the non-blocking points:
- **`dist:win` equivalence:** confirmed — `dist:win` is exactly `bun run build && electron-builder --win`, no extra flags (no `--publish`, no `--config`). The build half is covered by the earlier "Build application" step, and publishing is done by the separate `release` job (softprops/action-gh-release), not electron-builder — same as the mac step which already runs `bunx electron-builder --mac`. So nothing is silently dropped.
- **Log which credential is missing:** done in the same commit — the unsigned-fallback warning now lists the missing names.

brannon-bowden (approved):
Approve — this update resolves the previously-blocking gate issue.

The gate at `.github/workflows/release.yml:180-193` now checks all six values electron-builder's Azure Trusted Signing path actually needs (`AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZ_ENDPOINT`, `AZ_ACCOUNT`, `AZ_PROFILE`), not the partial set from before. A repo with, say, `AZURE_CLIENT_ID` set but `AZURE_TENANT_ID` unset/mistyped will now correctly fall through to the unsigned path instead of taking the signed path and failing Azure authentication mid-release. That was the blocking concern in the prior review, and it's fixed.

One subtlety worth confirming was checked: each `[[ -z "$VAR" ]] && MISSING+=(VAR)` line, when the test is false (variable is set), returns a non-zero status as a non-final member of an `&&` list — under `bash -e` (GitHub Actions' default for `shell: bash` steps) that specific case is exempted from triggering `errexit`, so this won't unexpectedly abort the step. No issue there.

## Blocking issues

None.

## Non-blocking issues

- **`.github/workflows/release.yml:180-186`** — Minor readability nit: the six `[[ -z ... ]] && MISSING+=(...)` lines repeat the same pattern with a local-var-name-to-secret-name mapping baked in twice (e.g., `AZ_ENDPOINT` → `AZURE_CODE_SIGNING_ENDPOINT`). Not a defect, but a small associative-array or loop could reduce duplication if this list grows again.

## Tests

As before, this is a shell conditional in a workflow file, so automated test coverage isn't practical here. Given the fix, it'd be worth a manual `dry_run` verification with an intentionally partial credential set (e.g., `AZURE_CLIENT_ID` set but `AZURE_TENANT_ID` unset) to confirm the fallback path and the new missing-credential message both fire as expected — not a merge blocker, just good validation before relying on this in a real release.

